PR #1, opened 2024-02-24
SYCL building system

fengyuan14:
Main parts of the PR include,

1. Find SYCL compiler tool kit in cmake. (cmake/Module/FindSYCLToolkit.cmake)
2. Setup cmake variables for SYCL compilation flags, includes and runtime library. (cmake/public/xpu.cmake, cmake/Dependencies.cmake)
3. Implement two cmake custom target building helpers, sycl_add_library and sycl_add_executable. (cmake/Module/FindSYCL.cmake, cmake/Module/FindSYCL/run_sycl.cmake)
4. Provide unit test cases to verify SYCL building system, since there is no SYCL kernel added in PyTorch. (test/cpp/sycl/CMakeLists.txt)

SYCL program building is similar as CUDA and HIP. SYCL program requires custom compilation and additional device object linkage steps. We implement sycl_add_library and sycl_add_excutable to introduce SYCL compiler,

1. Introduce SYCL compiler to compile .cpp, in which we have SYCL kernels. As to "host only" .cpp, we use CXX compiler for compilation. So far, we only enable GCC compiler in SYCL separate compilation. Next step we will enable CLANG compiler to align with PyTorch building requirement.
2. Introduce SYCL compiler to link device object and produce a host relocatable object which could be linked by host linker in subsequence.


fengyuan14:
Can build standalone SYCL test case. See `test/sycl/`.

fengyuan14:
Move to ghstack.

PR #16, opened 2024-03-06
view, as_strided: add xpu operator

xytintel:


test/python/tensor_shape.py
```diff
@@ -0,0 +1,10 @@
+import torch
```
fengyuan14, line 1: Please use pytest backbone, then we can run on preci easily.

xytintel:
pr change

PR #18, opened 2024-03-07
resize, full_like, _copy_from_and_resize: add xpu operator

xytintel:


test/python/tensor_resize.py
```diff
@@ -0,0 +1,14 @@
+import torch
```
fengyuan14, line 1: Port to pytest.

xytintel:
pr change

PR #59, opened 2024-03-21
Add unfold

etaf:
The view op "aten::unfold" should not be fallback to CPU. So add the xpu implement first.


src/aten/TensorShape.cpp
fengyuan14, line 1: Please add example cases.

fengyuan14 (approved):

PR #77, opened 2024-03-26
Raise error when a View op fallback to CPU

etaf:
Raise error when a View op fallback to CPU


etaf:
@fengyuan14 we need to open this flag to make sure we did not fallback a View Op to CPU. We want this feature to exclude unreasonable fallbacks when we debug e2e benchmark failure. Please have a review.

EikanWang (approved):

fengyuan14 (approved):

PR #87, opened 2024-03-27
[CI] enhance e2e test workflow

chuanqi129:


.github/scripts/inductor_xpu_test.sh
```diff
@@ -36,6 +36,14 @@ if [[ $MODE == "training" ]]; then
     Mode_extra="--training "
 fi
 
+Real_DT=$DT
+if [[ $DT == amp_bf16 ]]; then
+    Real_DT=amp
+elif [[ $DT == "amp_fp16" ]]; then
+    Real_DT=amp
+    export INDUCTOR_AMP_DT=float16
+fi
```
EikanWang, line 45: Should be `fp16` and `bf16` rather than amp?

EikanWang (approved):

.github/scripts/inductor_xpu_test.sh
chuanqi129: This is a general test launch script, the datatype will pass to this script. Take bfloat16 test as an example https://github.com/intel/torch-xpu-ops/pull/87/files#diff-97cd5c8211b5b26755e1fd4e03a03b567774f243a142ed9e830f108a94362a84R68

PR #89, opened 2024-03-27
Fix abs complex to float issue

xytintel:
Note [Complex abs and angle] Complex inputs to abs and angle return float results by default. abs and angle, in both NumPy and C++, returns a float result when given a complex input. This makes sense mathematically since the absolute value and angle of a complex number has no imaginary part.

### DIFF
--- a/src/aten/UnaryOps.cpp
+++ b/src/aten/UnaryOps.cpp
@@ -1,31 +1,118 @@
 #include <ATen/ScalarOps.h>
 #include <ATen/XPUNativeFunctions.h>
 #include <ATen/core/Tensor.h>
+#include <ATen/native/Resize.h>
 #include <ATen/native/TensorIterator.h>
-#include <torch/library.h>
-
 #include <aten/sycl/UnaryKernels.h>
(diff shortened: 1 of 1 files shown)


src/aten/UnaryOps.cpp
```diff
@@ -1,31 +1,118 @@
...
 #include <ATen/native/TensorIterator.h>
-#include <torch/library.h>
-
 #include <aten/sycl/UnaryKernels.h>
 #include <aten/sycl/UnaryLogKernels.h>
+#include <torch/library.h>
 
 namespace at {
 
+template <typename Stub>
```
fengyuan14, line 12: Can we implement a lambda to align the API as `stub`. The lambda is a wrapper to route to our `abs_kernel`. Like,
```
auto xpu_abs_stub = [](DeviceType dev, TensorIteratorBase iter) { abs_kernel(iter) };
```

fengyuan14: Dismiss. `unary_op_impl_with_complex_to_float_out` is static function in PyTorch.

fengyuan14 (approved):

PR #93, opened 2024-03-29
Enable test_ops::TestCommon

fengyuan14:
PYTORCH_TEST_WITH_SLOW=1 PYTORCH_ENABLE_XPU_FALLBACK=1 pytest -vs test_ops.py


test/xpu/test_xpu.py
```diff
@@ -0,0 +1,26 @@
+# Owner(s): ["module: intel"]
```
EikanWang, line 1: We do not need to provide such things. Actually, I think we just need to patch the `instantiate_device_type_tests` before importing the test cases.

```python
import torch
torch.testing._internal.common_device_type.onlyCUDA = onlyXPU
torch.testing._internal.common_device_type.onlyNativeDeviceTypes = onlyXPU

def instantiate_device_type_tests_dummy(*args, **kwargs):
    pass

instantiate_device_type_tests_fn = torch.testing._internal.common_device_type.instantiate_device_type_tests
torch.testing._internal.common_device_type.instantiate_device_type_tests = instantiate_device_type_tests_dummy

import test_ops

class TestCommonXPU(TestCase):
    pass

instantiate_device_type_tests_copy(TestCommonXPU, test_ops.TestCommon)
instantiate_device_type_tests_fn(TestCommonXPU, globals(), only_for="xpu")
```

EikanWang (changes requested):

test/xpu/run_test.py
```diff
@@ -0,0 +1,369 @@
...
+)
+
+
+skip_options = " -k 'not " + skip_list[0]
+for skip_case in skip_list[1:]:
+    skip_option = " and not " + skip_case
+    skip_options += skip_option
+skip_options += "'"
+
+test_command = "PYTORCH_ENABLE_XPU_FALLBACK=1 PYTORCH_TEST_WITH_SLOW=1 pytest -v _test_ops.py"
```
chuanqi129, line 366: Will we have other `_test_xxx.py` under this folder? if have, do we need to define multiple skip lists for those test suites respectively? And we will have multiple test commands in here, right?

fengyuan14: Ideally, I believe no, since each case gets unique test name.

But so far, `pytest -v` looks not working for the ported cases. We need use `pytest -v _test_xxx.py`. I believe it should work  with some change in patching scheme.

fengyuan14: We can optimize validation script in another task thread.

EikanWang (approved):
LGTM

chuanqi129 (approved):

PR #99, opened 2024-04-01
Dispatch explicitly to bypass ATen redispatch in CPU fallback routine.

fengyuan14:


fengyuan14:
@EikanWang 

After rebuilding PyTorch with full ENV vars, and fixing the issue, TestCommon reaches,
== 347 failed, 19838 passed, 1839 skipped, 322 xfailed in 3741.02s (1:02:21) ===

EikanWang (approved):

PR #141, opened 2024-04-12
Rebasing SYCL IndexKernel template and add aten::index_select

fengyuan14:
1. Add IndexKernel template. It could be shared among index_select, index_add, index_put and index_cpu.
2. Add ate::index_select
3. Use IndexKernel in index_add instead of CUDA ported kernels, since we have XPU specific implementation.


src/aten/sycl/Indexing.cpp
```diff
@@ -7,17 +7,160 @@
...
+template <typename ValType>
+class IndexSelectScalarFunctor {
+ public:
+  void operator()(
+      ValType* dst,
+      ValType* src,
+      int64_t dst_off,
+      int64_t src_off,
+      int64_t idx,
+      ValType alpha) const {
```
EikanWang, line 33: Why are the `idx` and `alpha` not used?

src/aten/sycl/Indexing.cpp
```diff
@@ -7,17 +7,160 @@
...
+}
+
+void index_select_kernel(
+    const Tensor& src,
+    int64_t dim,
+    const Tensor& indices,
+    const Tensor& dst) {
+  at::assert_no_internal_overlap(dst);
+  at::assert_no_overlap(dst, src);
+  at::assert_no_overlap(dst, indices);
```
EikanWang, line 76: Do these operation require XPU specific implementation and register for XPU?

src/aten/sycl/Indexing.cpp
```diff
@@ -7,17 +7,160 @@
...
+
+  dim = at::maybe_wrap_dim(dim, src.dim());
```
EikanWang, line 78: ditto

src/aten/sycl/Indexing.h
```diff
@@ -1,17 +1,358 @@
...
-    const Tensor& self,
-    int64_t dim,
-    const Tensor& index,
-    const Tensor& source,
-    const Scalar& alpha,
-    const Tensor& result);
+template <
+    class IdxConfig,
+    bool TrivialOffCal = false,
+    bool known_problem_inner = false>
```
EikanWang, line 351:
```suggestion
    bool TrivialOffCal = false,
    bool KnownProblemInner = false>
```

src/aten/sycl/Indexing.h
```diff
@@ -1,17 +1,358 @@
...
+        dinfo_(dinfo),
+        iinfo_(iinfo),
+        alpha_(alpha),
+        index_num_(index_num),
+        indexing_dimension_size_(indexing_dimension_size),
+        indexing_dst_(indexing_dst),
+        problem_inner_(problem_inner),
+        func_(func) {}
+
+  template <class TarInfo>
```
EikanWang, line 62: What's the meaning of `Tar`?

EikanWang (commented):
Please address my comments. Meanwhile, which test case could cover the changes?

src/aten/sycl/Indexing.cpp
fengyuan14: IndexKernel is an IPEX-designed kernel, since before stock CUDA implementation is not good enough. The general kernel template is shared with `aten::index_select`, `aten::index_add`, `aten::index_copy`, and `aten::index_fill` so far in IPEX. The scalar functor is abstracted to following the broadest semantics. So here, for `aten::index_select`, `idx` and `alpha` are not necessary.

src/aten/sycl/Indexing.cpp
fengyuan14: No, they don't. The utils check the overlap of tensors by checking tensor storage on host.

src/aten/sycl/Indexing.cpp
fengyuan14: Not IPEX specific neither.

src/aten/sycl/Indexing.h
fengyuan14: Since it is a general kernel template shared among multiple indexing operators, some operator indexes along src tensor and some operators indexes along dst tensor. To give an abstract name, using `target` (`Tar`), which means the target tensor to index.

fengyuan14:
> Please address my comments. Meanwhile, which test case could cover the changes?

Yes, 

PYTORCH_ENABLE_XPU_FALLBACK=1 PYTORCH_TEST_WITH_SLOW=1 pytest -k 'index_select' -v _test_ops.py
image

PYTORCH_ENABLE_XPU_FALLBACK=1 PYTORCH_TEST_WITH_SLOW=1 pytest -k 'index_add' -v _test_ops.py
image

src/aten/sycl/Indexing.h
EikanWang: Regarding these non-common abbreviations, I recommend keeping the full word for better code understanding. I would prefer to ensure the code self is the comments.

EikanWang (approved):

PR #172, opened 2024-04-26
[Fallback] Add a list of default fallback op without PYTORCH_ENABLE_XPU_FALLBACK.

etaf:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #172


src/aten/XPUFallback.template
```diff
@@ -80,4 +80,352 @@ TORCH_LIBRARY_IMPL(aten, XPU, m) {
         torch::CppFunction::makeFromBoxedFunction<&xpu_fallback>());
   }
 }
+
+/*
+ * Register fallback to CPU for ops specified in fallback_list.txt
```
fengyuan14, line 85: What is `fallback_list.txt`?

etaf: old txt, updated

src/aten/XPUFallback.template
```diff
@@ -80,4 +80,352 @@ TORCH_LIBRARY_IMPL(aten, XPU, m) {
         torch::CppFunction::makeFromBoxedFunction<&xpu_fallback>());
   }
 }
+
+/* 
+ * These ops are not supported via XPU backend currently, and we fallback to run on CPU.
```
fengyuan14, line 85:
```suggestion
 * Declare CPU fallback implementation of these operators as last resort for XPU backend. Explicit registration of XPU implementation can override CPU fallback implementation here.
```

PR #186, opened 2024-05-07
Workflow will be go on even if PR applied got failed

mengfei25:
Workflow will be go on even if PR applied got failed


mengfei25:
abandon

PR #205, opened 2024-05-13
Add UT for test_autograd_fallback.py

daisyden:
Enabled test_reducations.py and test_autograd_fallback.py to run with XPU patch. Updated XPU patch for instantiate_parametrized_tests().
We did not enable skiplist in order to report failures in log.

### DIFF
--- a/test/xpu/test_autograd_fallback.py
+++ b/test/xpu/test_autograd_fallback.py
@@ -1,0 +1,18 @@
+# Owner(s): ["module: intel"]
+
+from torch.testing._internal.common_utils import run_tests, instantiate_parametrized_tests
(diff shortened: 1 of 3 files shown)


test/xpu/test_autograd_fallback.py
fengyuan14, line 1: Can you give a separate PR for the case?

fengyuan14: Since we need fine grain fixing for cases and merge it not to block for a long time.

fengyuan14:
LGTM
image

fengyuan14 (approved):

Issue #223, opened 2024-05-15
Issues report of unit tests for test_autocast_xpu.py


PenghuiCheng:
### 🚀 The feature, motivation and pitch

Log:
```
test/xpu/test_autocast_xpu.py::TestAutocastGPU::test_cache_disabled FAILED [ 20%]
  File "/home/gta/penghuic/pytorch_stock/third_party/torch-xpu-ops/test/xpu/test_autocast_xpu.py", line 90, in test_cache_disabled
    torch._C._set_cached_tensors_enabled(False)
AttributeError: module 'torch._C' has no attribute '_set_cached_tensors_enabled'
```
command:
```python
PYTORCH_ENABLE_XPU_FALLBACK=1 PYTORCH_TEST_WITH_SLOW=1 pytest -v test/xpu/test_autocast_xpu.py
```

guangyey:
I see, checking...

guangyey:
Currently, _set_cached_tensors_enabled is tailored to CUDAGraph. So I don't think you need this API in our UT at this moment.
We also have no reason to refactor them unless there is a real case that requires it.

PR #224, opened 2024-05-15
Enable test_autocast

PenghuiCheng:
Add autocast test cases.
issue report:
new featrue:
https://github.com/intel/torch-xpu-ops/issues/223


fengyuan14:
@EikanWang  It is device agnostic case? Do we need? I assume XPU implementation of ATen won't break autocast architecture. Just like we don't need XPU runtime cases in the repo.
What's your opinion?

test/xpu/run_test_with_skip.py
```diff
@@ -803,5 +803,14 @@ def launch_test(test_case, skip_list=None, exe_list=None):
 )
 res += launch_test("test_transformers_xpu.py", skip_list)
 
+
+# test_autocast
+skip_list = (
+    # Frontend API support
+    # https://github.com/intel/torch-xpu-ops/issues/223
```
fengyuan14, line 810:
```suggestion
    Unsupported XPU runtime functionality,  '_set_cached_tensors_enabled'
    # https://github.com/intel/torch-xpu-ops/issues/223
```

fengyuan14 (approved):

fengyuan14:
> @EikanWang It is device agnostic case? Do we need? I assume XPU implementation of ATen won't break autocast architecture. Just like we don't need XPU runtime cases in the repo. What's your opinion?

Reviewed. The case tests `with autocast(XPU)`. It is hard-coded for XPU test.

PR #226, opened 2024-05-15
Skip unsupported dtypes for fine-grain tests.

xytintel:
Skip unsupported dtypes for fine-grain tests.


test/xpu/fin_grain/test_ops_xpu.py
```diff
@@ -149,10 +150,12 @@ class TestCommon(TestCase):
     #@ops(_xpu_computation_ops_no_numpy_ref, dtypes=any_common_cpu_xpu_all)
     @ops(_xpu_computation_ops, dtypes=cpu_xpu_all)
     def test_compare_cpu(self, device, dtype, op):
-        self.proxy = Namespace.TestCommonProxy()
-
-        test_common_test_fn = get_wrapped_fn(Namespace.TestCommonProxy.test_compare_cpu)
-        test_common_test_fn(self.proxy, device, dtype, op)
+        if dtype in op.supported_dtypes(device):
```
fengyuan14, line 153: We have not supported supported_dtypes for XPU device. What is the return of the API? dtypes are claimed by CPU device I remember?

fengyuan14: Resolved. It's op dtype. May not align CUDA. But it is reasonable to the operator tested. Will have further enhancement in test infrastructure to support XPU specific op dtype.

test/xpu/fin_grain/test_ops_xpu.py
```diff
@@ -149,10 +150,12 @@ class TestCommon(TestCase):
...
+            self.proxy = Namespace.TestCommonProxy()
+            test_common_test_fn = get_wrapped_fn(Namespace.TestCommonProxy.test_compare_cpu)
+            test_common_test_fn(self.proxy, device, dtype, op)
+        else:
+            pytest.skip("dtype not supported")
```
fengyuan14, line 158: We don't need the warning? Can we remove it?

EikanWang: It is a skip message. We can keep it but the message should be more informative.

```python
    pytest.skip(f"{op.name} has not supported {dtype} yet for {device}")
```

EikanWang (approved):

test/xpu/fin_grain/test_ops_xpu.py
xytintel: sure

fengyuan14: @EikanWang There are 3 dtypes list in test infrastructure,
1. All dtypes PyTorch has,
2. All dtypes PyTorch defines for the current op,
3. All dtypes PyTorch defines for the current op has and supported by specified device.

Here `supported_dtype` means #2. So a dtype going to `else` branch might be a dtype PyTorch doesn't define for the op. Like, complex might not be supported in `quantize` according to PyTorch definition, but we warn it with `qunatize has not supported complex yet for XPU`.

fengyuan14: CUDA won't warn it, when the dtype is not claimed by the op. So I suggest removing it.

xytintel: From a script semantics perspective, it's the scope of a specific device (which is 3). Do we need to upstream?

EikanWang: @fengyuan14 , the `supported_dtype` should be 3th bullet.

this PR has checked the supported data types for a particular operation on a particular device platform. The supported data types may be different for different devices. For example, the CPU supports a data type while XPU does not support it. The message could prompt us why the case is not applicable to XPU and help us to better understand if we need to do additional things to support it.

By the way, the message does not make things worse.

fengyuan14: Yes, you are right. It should be 3th, but now we haven't enhanced test infrastructure to enable `dtypeIfXPU`. And now we are working with 2nd. Anyway the `else` branch of both 3th and 2nd might involve dtype not claimed by the PyTorch operator. Agree, just remainder ourselves. LGTM

PR #243, opened 2024-05-16
support test_ops.py TestCompositeCompliance xpu test

daisyden:
The issues detected including: 235~242, 233, 232, 229.
To reproduce:
timeout 10000 python run_test_with_skip.py 2>&1|tee log


fengyuan14:
Enable the case in fine gran infrastructure. https://github.com/intel/torch-xpu-ops/pull/282

PR #250, opened 2024-05-16
Enable H2D/D2H asynchronized copy when XPU HostCachingAlloctor works

fengyuan14:


src/aten/Copy.cpp
```diff
@@ -239,11 +244,42 @@ void _copy_xpu(TensorIterator& iter, bool non_blocking) {
...
+        q.memcpy(dst, stage_mem, nbytes);
+        at::xpu::CachingHostAllocator_recordEvent(
+            const_cast<void*>(stage_mem),
+            stage_mem_dptr.get_context(),
+            at::xpu::getCurrentXPUStream());
+      }
+    } else {
+      q.memcpy(dst, src, nbytes);
+      if (at::detail::getXPUHooks().isPinnedPtr(dst)) {
+        at::xpu::CachingHostAllocator_recordEvent(
```
gujinghui, line 277: @guangyey   when will we abstract and use the unified API for all device types?

guangyey: After the device allocator...

src/aten/Copy.cpp
```diff
@@ -225,10 +227,13 @@ void _copy_xpu(TensorIterator& iter, bool non_blocking) {
 
   // Copy between CPU and GPU
   c10::xpu::OptionalXPUGuard device_guard;
+  enum { _H2D_, _D2H_ } copy_kind;
```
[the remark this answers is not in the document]
fengyuan14: CUDA backend is using a CUDA runtime definition for copy direction (`cudaMemcpyKind`) and pass it to CUDA runtime copy API (`cudaMemcpyAsync`).

For us, q.memcpy is transparent for copy direction. We use the enum to distinguish code-paths to handle different Application logic. Since it is used in a local scope, following common coding style in PyTorch, leave `_` for the variables to tell developers they are not used as interface.

gujinghui (approved):

PR #251, opened 2024-05-16
enhance on demand test

mengfei25:
support multiple configs for on demand launch


.github/actions/inductor-xpu-e2e-test/action.yml
```diff
@@ -96,35 +80,54 @@ runs:
...
+                bash inductor_xpu_test.sh ${suite} ${dt} ${mode} ${scenario} xpu 0 static 8 0 &
+                bash inductor_xpu_test.sh ${suite} ${dt} ${mode} ${scenario} xpu 1 static 8 1 &
+                bash inductor_xpu_test.sh ${suite} ${dt} ${mode} ${scenario} xpu 2 static 8 2 &
+                bash inductor_xpu_test.sh ${suite} ${dt} ${mode} ${scenario} xpu 3 static 8 3 &
+                bash inductor_xpu_test.sh ${suite} ${dt} ${mode} ${scenario} xpu 4 static 8 4 &
+                bash inductor_xpu_test.sh ${suite} ${dt} ${mode} ${scenario} xpu 5 static 8 5 &
+                bash inductor_xpu_test.sh ${suite} ${dt} ${mode} ${scenario} xpu 6 static 8 6 &
+                bash inductor_xpu_test.sh ${suite} ${dt} ${mode} ${scenario} xpu 7 static 8 7 &
+                wait
+                LOG_DIR="inductor_log/${suite}/${dt}"
```
chuanqi129, line 101: Still suggest to move out the log post-process and analysis as a standalone step. In that step, we can summary all configs summary in one file and show it. It will be better if we create a standalone log process scripts, especially for future performance test

.github/actions/inductor-xpu-e2e-test/action.yml
```diff
@@ -96,35 +80,54 @@ runs:
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: Inductor-XPU-E2E-Data-${{ github.event.pull_request.number || github.sha }}
+        path: ${{ github.workspace }}/inductor_log
+
     - name: Upload Inductor XPU E2E Data (${{ inputs.suite }} ${{ inputs.dt }} ${{ inputs.mode }} ${{ inputs.scenario }})
+      if: ${{ inputs.expected_pass_num }}
       uses: actions/upload-artifact@v4
       with:
         name: Inductor-XPU-E2E-Data-${{ inputs.suite }}-${{ inputs.dt }}-${{ inputs.mode }}-${{ inputs.scenario }}-${{ github.event.pull_request.number || github.sha }}
```
chuanqi129, line 133: Comment for the last step. We need consider to create a new file to record the expected pass number for each config, especially for the multi-configs scenario. In the last result check step, if the result check is needed (eg. `inputs.need_check` ), we can read the expected value for each config in the parameters combinations. How do you think?

.github/workflows/inductor_xpu_e2e_nightly.yml
```diff
@@ -13,41 +13,25 @@ on:
...
         required: true
-        type: choice
+        type: string
         default: 'huggingface'
-        options:
-          - huggingface
-          - timm_models
-          - torchbench
-        description: Dynamo benchmarks test suite, huggingface / timm_models / torchbench
+        description: Dynamo benchmarks test suite. huggingface,timm_models,torchbench
```
chuanqi129, line 18: Add "support multiple suites split with ," in the description, similar change for other parameters

.github/actions/inductor-xpu-e2e-test/action.yml
mengfei25: Is it better to check failed number, which should be more easy because it can be 0?

.github/actions/inductor-xpu-e2e-test/action.yml
```diff
@@ -96,35 +80,54 @@ runs:
...
+                echo "num_failed: $num_failed" | tee -a ${LOG_DIR}/${suite}_${dt}_${mode}_${scenario}_e2e_summary.log
+                echo "pass_rate: $pass_rate" | tee -a ${LOG_DIR}/${suite}_${dt}_${mode}_${scenario}_e2e_summary.log
+              done
+            done
+          done
+        done
         cd ${{ github.workspace }} && cp -r ../pytorch/inductor_log .
+
+    - name: Upload Inductor XPU E2E Data
+      if: ${{ ! inputs.expected_pass_num }}
```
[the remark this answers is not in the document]
mengfei25: Looks like we don't need split several artifact for on demand launch. One is enough I think.

.github/actions/inductor-xpu-e2e-test/action.yml
chuanqi129: But sometimes failed model info didn't written into the CSV due to some reason, so still suggest to add such config. How do you think?

It can be a csv file, e.g.
|suite|dt|mode|scenario|pass_num|
|--|--|--|--|--|
|huggingface|float32|training|accuracy|46|

.github/actions/inductor-xpu-e2e-test/action.yml
```diff
@@ -96,47 +80,130 @@ runs:
...
-        bash inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dt }} ${{ inputs.mode }} ${{ inputs.scenario }} xpu 3 static 8 3 & \
-        bash inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dt }} ${{ inputs.mode }} ${{ inputs.scenario }} xpu 4 static 8 4 & \
-        bash inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dt }} ${{ inputs.mode }} ${{ inputs.scenario }} xpu 5 static 8 5 & \
-        bash inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dt }} ${{ inputs.mode }} ${{ inputs.scenario }} xpu 6 static 8 6 & \
-        bash inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dt }} ${{ inputs.mode }} ${{ inputs.scenario }} xpu 7 static 8 7 & wait
-        LOG_NAME=inductor_${{ inputs.suite }}_${{ inputs.dt }}_${{ inputs.mode }}_xpu_${{ inputs.scenario }}_all.log
-        for log in $(ls inductor_log/${{ inputs.suite }}/${{ inputs.dt }}/*card*.log); do cat $log >> inductor_log/${{ inputs.suite }}/${{ inputs.dt }}/${LOG_NAME}; done
-    - name: Test Results Overview (${{ inputs.suite }} ${{ inputs.dt }} ${{ inputs.mode }} ${{ inputs.scenario }})
+        # check param
```
chuanqi129, line 83: Suggest to split the param check as a standalone step, and write the valid value into $GITHUB_ENV and use them in below steps directly. 
```
echo "SUITES=huggingface,timms_model" >> ${GITHUB_ENV} 
echo "DTS=float32" >> ${GITHUB_ENV} 
...
```

chuanqi129 (approved):
LGTM, still have 2 minor suggestion, you can decide do them in this PR or a new standalone PR. BTW, have we launch a on-demand test based on this PR branch?

mengfei25:
Yes, ever launched one https://github.com/intel/torch-xpu-ops/actions/runs/9107345605/job/25036072421

PR #252, opened 2024-05-16
Add unit test case for comprarison_utils ops

PenghuiCheng:
Add unit test case for comprarison_utils ops!
Tests passed!


daisyden (approved):
LGTM

PR #255, opened 2024-05-16
add foreach op fallback

etaf:
Add not implemented foreach aten ops used in `pytorch/test/inductor/test_compiled_optimizers.py` to default fallback list.


fengyuan14:
The operators are not claimed by torch-xpu-opx. Should not add them into explicit CPU fallback list. We are discussing on how to enable them for inductor unit tests.

PR #270, opened 2024-05-20
Register groupnorm fallback

retonym:
Missing groupnorm forward fallback registration, causing nan in TIMM benchmark


etaf:
Doese this op appeares only in forward? I used to collecte fallback ops in the training mode but missed this op.

retonym:
> Doese this op appeares only in inference? I used to collecte fallback ops in the training mode and curious that I missed this op.

I find the this op in poolformer_m36 inference

retonym:
@etaf @EikanWang Could we merge this pr

retonym:
@EikanWang  could we merge this PR first to avoid TIMM accuracy issue in poolformer_m36 inference, before we implement such kernels.

etaf (approved):

PR #274, opened 2024-05-20
add test_segment_reductions_xpu.py

daisyden:
All passed


fengyuan14:
Close the PR, since the operator is not supported yet.

PR #278, opened 2024-05-20
add test_segment_reductions_xpu.py

daisyden:
all passed


fengyuan14:
The case covers aten::segment_reduce, which has not been supported in XPU backend. Close the PR for now.

PR #279, opened 2024-05-20
Expose TORCH_XPU_API to Pytorch

etaf:
Expose TORCH_XPU_API to Pytorch

### DIFF
--- a/src/ATen/xpu/EmptyTensor.cpp
+++ b/src/ATen/xpu/EmptyTensor.cpp
@@ -1,6 +1,7 @@
 #define TORCH_ASSERT_NO_OPERATORS
 #include <ATen/Context.h>
 #include <ATen/EmptyTensor.h>
+#include <ATen/xpu/EmptyTensor.h>
 #include <c10/core/Allocator.h>
 #include <c10/core/DeviceGuard.h>
 
--- a/src/ATen/xpu/EmptyTensor.h
+++ b/src/ATen/xpu/EmptyTensor.h
@@ -3,14 +3,13 @@
 
 namespace at::detail {
 
-// XXX: add TORCH_XPU_API
-TensorBase empty_xpu(
+TORCH_XPU_API TensorBase empty_xpu(
     IntArrayRef size,
     ScalarType dtype,
     c10::optional<Device> device_opt,
     c10::optional<c10::MemoryFormat> memory_format_opt);
 
-TensorBase empty_xpu(
+TORCH_XPU_API TensorBase empty_xpu(
     IntArrayRef size,
     c10::optional<ScalarType> dtype_opt,
     c10::optional<Layout> layout_opt,
@@ -18,15 +17,16 @@ TensorBase empty_xpu(
     c10::optional<bool> pin_memory_opt,
     c10::optional<c10::MemoryFormat> memory_format_opt);
 
-TensorBase empty_xpu(IntArrayRef size, const TensorOptions& options);
+TORCH_XPU_API TensorBase
+empty_xpu(IntArrayRef size, const TensorOptions& options);
 
-TensorBase empty_strided_xpu(
+TORCH_XPU_API TensorBase empty_strided_xpu(
     IntArrayRef size,
     IntArrayRef stride,
     ScalarType dtype,
     c10::optional<Device> device_opt);
 
-TensorBase empty_strided_xpu(
+TORCH_XPU_API TensorBase empty_strided_xpu(
     IntArrayRef size,
     IntArrayRef stride,
     c10::optional<ScalarType> dtype_opt,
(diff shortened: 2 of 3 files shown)


etaf:
This PR is relied upon stock Pytorch PR: https://github.com/pytorch/pytorch/pull/126678

EikanWang:
@etaf , is it critical to catch up PT 2.4?

etaf:
> @etaf , is it critical to catch up PT 2.4?

No, it's a performance optimization.

etaf:
Since this Pytorch 2.4 release branch has been cut, I think we can start land this PR and the https://github.com/pytorch/pytorch/pull/126678 that depend on it.

src/aten/EmptyTensor.h
```diff
@@ -4,37 +4,38 @@
 namespace at::detail {
 
 // XXX: add TORCH_XPU_API
```
EikanWang, line 6: The comment should be invalid now.

src/aten/EmptyTensor.cpp
```diff
@@ -1,6 +1,7 @@
 #define TORCH_ASSERT_NO_OPERATORS
 #include <ATen/Context.h>
 #include <ATen/EmptyTensor.h>
+#include <ATen/xpu/EmptyTensor.h>
```
EikanWang, line 4: Since we have included the xpu/EmptyTensor, is ATen/EmptyTensor still required?

EikanWang (commented):
LGTM.

src/aten/EmptyTensor.cpp
etaf: `ATen/xpu/EmptyTensor.h` is the exported interface of `EmptyTensor.cpp`
and `ATen/EmptyTensor.h` is for `at::detail::empty_generic` which is used in this  `EmptyTensor.cpp` file.

src/aten/EmptyTensor.h
etaf: removed

src/aten/EmptyTensor.cpp
EikanWang: Then my question will be why `ATen/xpu/EmptyTensor.h` includes `ATen/EmptyTensor.h`?

etaf: Sorry, but `ATen/xpu/EmptyTensor.h` does not include `ATen/EmptyTensor.h`.

etaf:
@fengyuan14  can you help review this PR?

src/aten/EmptyTensor.cpp
```diff
@@ -1,4 +1,5 @@
 #define TORCH_ASSERT_NO_OPERATORS
+#include "EmptyTensor.h"
```
fengyuan14, line 2: Why add it?

etaf: It's the interface of xpu empty_strided.

fengyuan14: Before adding it, we can pass compilation. Why add it?

fengyuan14: Ok, you mean the decorator declared in header need to be included during cpp compilation to let compiler know the attribute.

fengyuan14: Can we use `aten/EmptyTensor.h` here?

etaf: if you mean `#include<aten/EmptyTensor.h>` , yes.

fengyuan14:
You can expose the var to PyTorch cmake,
https://github.com/intel/torch-xpu-ops/blob/a7c3bd2a7f7b17ac692bc7faa3067f3b2a005383/src/CMakeLists.txt#L7
And add it into cmake var of Caffe2 or ATen include dirs.

etaf:
> You can expose the var to PyTorch cmake,
> 
> https://github.com/intel/torch-xpu-ops/blob/a7c3bd2a7f7b17ac692bc7faa3067f3b2a005383/src/CMakeLists.txt#L7
> 
> 
> And append it into cmake var of Caffe2 or ATen include dirs.

Wouldn't that make the two PRs dependent on each other?

fengyuan14:
> > You can expose the var to PyTorch cmake,
> > https://github.com/intel/torch-xpu-ops/blob/a7c3bd2a7f7b17ac692bc7faa3067f3b2a005383/src/CMakeLists.txt#L7
> > 
> > And append it into cmake var of Caffe2 or ATen include dirs.
> 
> Wouldn't that make the two PRs dependent on each other?

ATen_XPU_INCLUDE_DIR is the committed formal and stable API of torch-xpu-ops CMAKE. It is not implementation dependent.

fengyuan14:
Can we do like this ? What do you think of it?
image

fengyuan14:
Looks I have exposed it by CMAKE  target torch-xpu-ops, like this,
image

etaf:
The original design is to align the include path with CPU/CUDA, eg. `ATen/xpu/EmptyTensor.h`, can I still keep this path with your design ?

fengyuan14 (approved):

fengyuan14 (approved):

PR #307, opened 2024-05-23
Fix div ut failures caused by the ineffective type promotion

xytintel:
Fix div ut failures caused by the ineffective type promotion.
1. Align div with mul implementation by using opmath type.
2. Fix div trunc failure when result of division between two same floating point values is not 1.f sometimes.
3. Remove unnecessary kernels.


src/comm/XPUMathCompat.h
```diff
@@ -29,4 +29,23 @@ __MATH_FUNCTIONS_DECL__ double rsqrt(double x) {
...
 
+template <typename T>
+inline auto div(const T& a, const T& b) {
+  return a / b;
+}
+
+template <>
+inline auto div<c10::Half>(const c10::Half& a, const c10::Half& b)
+    __ubsan_ignore_float_divide_by_zero__ {
+  volatile float res = static_cast<float>(a) / static_cast<float>(b);
```
fengyuan14, line 40: Please add comments here for readability. Let us not make latter maintainer confused.

fengyuan14, line 40:
```suggestion
  // Suppress compiler optimization to get data type promotion.
  volatile float res = static_cast<float>(a) / static_cast<float>(b);
```

src/comm/XPUMathCompat.h
```diff
@@ -29,4 +29,23 @@ __MATH_FUNCTIONS_DECL__ double rsqrt(double x) {
...
+  return res;
+}
+
+template <>
+inline auto div<c10::BFloat16>(const c10::BFloat16& a, const c10::BFloat16& b)
+    __ubsan_ignore_float_divide_by_zero__ {
+  volatile float res = static_cast<float>(a) / static_cast<float>(b);
```
fengyuan14, line 47:
```suggestion
  // Suppress compiler optimization to get data type promotion.
  volatile float res = static_cast<float>(a) / static_cast<float>(b);
```

src/comm/XPUMathCompat.h
```diff
@@ -29,4 +29,30 @@ __MATH_FUNCTIONS_DECL__ double rsqrt(double x) {
...
+inline auto div<c10::BFloat16>(const c10::BFloat16& a, const c10::BFloat16& b)
+    __ubsan_ignore_float_divide_by_zero__ {
+  // Suppress compiler optimization to get data type promotion.
+  volatile float res = static_cast<float>(a) / static_cast<float>(b);
+  return res;
+}
+
+template <typename T>
+inline T div_trunc(const T& a, const T& b) {
+  return ((a == b) && std::isfinite(a)) ? (T)1 : (T)std::trunc(a / b);
```
fengyuan14, line 55: Do we need data type promotion here?

fengyuan14 (approved):

Issue #262, opened 2024-05-17
Provide a way to debug explicit CPU fallback


dvrogozh:
@fengyuan14 - The commit https://github.com/intel/torch-xpu-ops/commit/5bf9e0cc768f7a3b13d829118683275f324399f1 muted debug logs of "explicit" CPU fallbacks. This complicated debug for 3d party contributors trying to evaluate XPU backend capabilities - now I am forced to revert noted commit to understand which operations are not currently implemented by XPU. Please:

1. Explain what "explicit CPU fallback" means - this seems to be internal to xpu team classification which is unclear and confusing
2. Extend `PYTORCH_DEBUG_XPU_FALLBACK=1` to track any CPU fallback happening in XPU backend. Note: I am fine if "explicit" fallback will be muted by default, but I really need a way to be able to track it.

```
commit 5bf9e0cc768f7a3b13d829118683275f324399f1 (origin/meng_max_2d)
Author: Feng Yuan <feng1.yuan@intel.com>
Date:   Mon Apr 29 13:05:51 2024 +0800

    Register operator's implementation lazily. (#177)

    1. Avoid dangling operator's implementation (m.impl(torchvision::nms) is
    ahead of `import torchvision` sometime)
    2. Mute debug log of explicit CPU fallback.
    3. Add torchvision.roi_align/_roi_align_backward example case
```

CC: @jgong5 @mingfeima @XiaobingSuper @ashokei @jingxu10 @gujinghui @EikanWang @fengyuan14 @guangyey

dvrogozh:
Also filed https://github.com/pytorch/pytorch/issues/126488 for visibility at pytorch level.

fengyuan14:
Thanks for your feedback. Replied in https://github.com/pytorch/pytorch/issues/126488.

PR #318, opened 2024-05-24
Always warn on cpu fallback

dvrogozh:
Fixes: #262, https://github.com/pytorch/pytorch/issues/126488

5bf9e0c ("Register operator's implementation lazily") disabled warnings printout on explicit CPU fallback. I believe that users and customers will benefit from these warnings in all the cases. Note that "explicit fallback" seems to be some internal intel classification term for supported/unsupported operations unlikely known to others. Thus, non-intel users will likely care for cpu fallback in general regardless of its type. This PR adds warning back for all CPU fallback cases. We did discuss in https://github.com/pytorch/pytorch/issues/126488 that maybe printout in Release build might not be needed - I am thinking otherwise and added printout for all the build modes. Let's discuss this in the PR review.

CC: @gujinghui @EikanWang @fengyuan14 @guangyey @jgong5


EikanWang (approved):
LGTM

EikanWang:
@dvrogozh , do you have the privilege to add a reviewer?

dvrogozh:
> @dvrogozh , do you have the privilege to add a reviewer?

No, unfortunately I do not.

dvrogozh:
@EikanWang @fengyuan14. Some of `pull / preci-ut ` tests fail. Are these some known failures since I can hardly attribute them to the change in this PR?

dvrogozh:
@fengyuan14 : tests are still failing. Should this be retriggered again?

fengyuan14:
> @fengyuan14 : tests are still failing. Should this be retriggered again?
Two new errors never encounter. Let me try retrigger 
```
FAILED test_autograd_xpu.py::TestAutogradDeviceTypeXPU::test_copy_r_to_c_xpu - AssertionError: False is not true : 
FAILED test_autograd_xpu.py::TestAutogradDeviceTypeXPU::test_to_r_to_c_xpu - AssertionError: False is not true : 
```

fengyuan14:
I checked the failure again. The failures should be caused by the change in the PR.
image
The case says there should be no warning, but we warn in CPU fallback code path.

dvrogozh:
@fengyuan14 : Ok, so there are some tests in pytorch which don't allow warnings. However, the tests which fail actually run few basic ops - in a way someone might consider that these ops should not fail and should run on appropriate device. I.e. actually someone might consider that these tests are actually broken and incorrectly report status.

There are few ways to proceed with this PR:
1. Actually fix the tests. This means identify and implement missing xpu ops.
2. Silence warning conditional to something like release build or environment variable disabled by default.

I need your opinions on this. From my side, I will check tomorrow which ops are not implemented. As of now I checked only 1 test:
* `TestAutogradDeviceTypeXPU.test_copy_r_to_c_xpu` - need `aten::all.all_out`

fengyuan14:
> @fengyuan14 : Ok, so there are some tests in pytorch which don't allow warnings. However, the tests which fail actually run few basic ops - in a way someone might consider that these ops should not fail and should run on appropriate device. I.e. actually someone might consider that these tests are actually broken and incorrectly report status.
> 
> There are few ways to proceed with this PR:
> 
> 1. Actually fix the tests. This means identify and implement missing xpu ops.
> 2. Silence warning conditional to something like release build or environment variable disabled by default.
> 
> I need your opinions on this. From my side, I will check tomorrow which ops are not implemented. As of now I checked only 1 test:
> 
> * `TestAutogradDeviceTypeXPU.test_copy_r_to_c_xpu` - need `aten::all.all_out`

I prefer the first option. Looks there won't be too many efforts to take. And `aten::all.all_out` is in our development plan. We won't take extra efforts. We can prioritize the op.

fengyuan14:
https://github.com/intel/torch-xpu-ops/pull/368

dvrogozh:
> From my side, I will check tomorrow which ops are not implemented.

As far as I can tell from the log, we have 2 tests failing, both requiring same op:
* `TestAutogradDeviceTypeXPU.test_copy_r_to_c_xpu` - need `aten::all.all_out`
* `TestAutogradDeviceTypeXPU.test_to_r_to_c_xpu` - need `aten::all.all_out`

PR #331, opened 2024-05-27
Enable AOT targets for MTL

fengyuan14:


fengyuan14:
AOT targets work,
image

majing921201 (approved):

PR #335, opened 2024-05-27
add test_multihead_attention

daisyden:
Besides known oneDNN issue, only one issue at #342


daisyden:
merged with https://github.com/intel/torch-xpu-ops/pull/407

PR #341, opened 2024-05-28
Evaluate fine gran cases and add comments

huaiyuzh:
update fine grain cases skip reason

### DIFF
--- a/test/xpu/fin_grain/run_fine_grain.py
+++ b/test/xpu/fin_grain/run_fine_grain.py
@@ -2,30 +2,41 @@ import os
 import sys
 
 skip_list = (
-    # Known issue
-    "test_compare_cpu__refs_rsub_xpu_bfloat16", # CPU result is not golden reference
-    "test_compare_cpu_add_xpu_bfloat16",# CPU result is not golden reference
-    "test_compare_cpu_cumsum_xpu_bfloat16",# Caused by different accumulation order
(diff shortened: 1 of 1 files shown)


test/xpu/fin_grain/run_fine_grain.py
```diff
@@ -3,28 +3,28 @@
 
 skip_list = (
     # Known issue
-    "test_compare_cpu__refs_rsub_xpu_bfloat16",
-    "test_compare_cpu_add_xpu_bfloat16",
-    "test_compare_cpu_cumsum_xpu_bfloat16",
-    "test_compare_cpu_cumsum_xpu_float16",
+    "test_compare_cpu__refs_rsub_xpu_bfloat16", # CPU result is not golden reference
```
fengyuan14, line 6:
```
Calculation error between XPU implementation and CPU implementation,
1. Compiler optimization causes failing to promote data type to higher precision.
2. Accumulate error is amplified by some operations in some extreme cases. (std::exp(extreme_large_num))
3. Accumulate error is amplified by a large number of accumalate operations.
4. Accumulate error is different on different implementations due to different accumulation order.
	a. Different kernel implementations.
	b. Different std functions. (std::log, std::tanh, std::exp)
5. The result of division between two same float values is not 1.
```

fengyuan14 (approved):

fengyuan14:
Random failure by an extreme small error.
image

PR #344, opened 2024-05-28
Add environment var of XPU Archs for overriding default AOT build tar…

fengyuan14:
…gets


fengyuan14:
Not operator specific error. Will rerun cases.
image

majing921201 (approved):

Issue #197, opened 2024-05-09
Complement operator variants when implementing a required operator


fengyuan14:
### 🚀 The feature, motivation and pitch

As to staging goal of PyTorch 2.5, we collect 484 operators which are required working with XPU backend. Part of them are required XPU specific implementation.
When we give XPU implementation for an ATen operator, we need register all variants of the operator, like xxx.out, xxx.Tensor, xxx.Scalar, xxx_ and so on.
Following the rule,
1. We won't take additional efforts to be back for lack of registration in future and complement them. Adding variants at the moment is cheap.
2. When we align with CUDA registration, in-tree would be seamless.

- [x] random_.to
- [x] clamp.Tensor_out
- [x] clamp_min.Tensor_out
- [x] clamp_max.Tensor_out
- [x] fmod.Scalar // remove
- [x] fmod_.Scalar // remove
- [x] index_add_
- [x] index_add
- [x] remainder.Scalar_out // remove
- [x] remainder.Scalar // remove
- [x] remainder_.Scalar // remove
- [x] rsub.Scalar // remove
- [x] rsub.Scalar_out // remove
- [x] rsub.Tensor_out // remove
- [x] sub.Scalar // remove
- [x] sub_.Scalar // remove
- [x] sub.Scalar_out // remove
- [x] sum.out // remove
- [x] sum.dim_IntList // remove

chunhuanMeng:
we can close this issue

chuanqi129:
@chunhuanMeng please help to double check if we can close this issue or not

chunhuanMeng:
> @chunhuanMeng please help to double check if we can close this issue or not
We can close this issue

PR #366, opened 2024-06-04
Remove ATen variants not required to register

chunhuanMeng:
remove some variants which were not registered in cuda too.

https://github.com/intel/torch-xpu-ops/issues/197


yaml/xpu_functions.yaml
```diff
@@ -16,9 +16,6 @@ supported:
   - sub.Tensor
   - sub_.Tensor
   - sub.out
-  - sub.Scalar
```
fengyuan14, line 19: Should remove corresponding implementation in cpp.

yaml/xpu_functions.yaml
```diff
@@ -228,7 +215,7 @@ supported:
   - min
   - min.unary_out
   - min.dim_min
-  - sum.out
+  - sum
```
fengyuan14, line 218: You are saying it? It should not be registered? It's dispatched by  `CompositeExplicitAutograd` 
image

PR #392, opened 2024-06-11
Enable correctly rounded divide and sqrt

xytintel:
Add -cl-fp32-correctly-rounded-divide-sqrt flag.


cmake/BuildFlags.cmake
```diff
@@ -65,6 +66,7 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   set(SYCL_DEVICE_LINK_FLAGS ${SYCL_DEVICE_LINK_FLAGS} ${SYCL_TARGETS_OPTION})
 
   set(SYCL_OFFLINE_COMPILER_CG_OPTIONS ${SYCL_OFFLINE_COMPILER_CG_OPTIONS} "-options \"-cl-intel-enable-auto-large-GRF-mode\"")
+  set(SYCL_OFFLINE_COMPILER_CG_OPTIONS ${SYCL_OFFLINE_COMPILER_CG_OPTIONS} "-options \"-cl-fp32-correctly-rounded-divide-sqrt\"")
```
dvrogozh, line 69: > `-cl-fp32-correctly-rounded-divide-sqrt`

Marvelous option name. As well as the behavior it suggests by its name.

fengyuan14 (approved):

cmake/BuildFlags.cmake
fengyuan14: Correctly rounded operations are defined in IEEE754. We are talking with SYCL compiler team to give a formal option exposed by SYCL compiler directly not by offline compiler (ocloc). The option will be similar as conventional name used in GCC, CLANG, and NVCC.

PR #394, opened 2024-06-12
python from 3.8 to 3.10 for CI and Nightly

mengfei25:
python from 3.8 to 3.10


.github/workflows/inductor_xpu_e2e_nightly.yml
```diff
@@ -48,12 +53,11 @@ jobs:
...
-          which conda
-          if ! conda env list | grep -q "^e2e_ci "; then
-            conda create -n e2e_ci python=3.8 cmake ninja -y
-          fi
+          which conda && conda clean -ay
+          conda remove --all -y -n e2e_ci || rm -rf $(dirname ${CONDA_EXE})/../envs/e2e_ci
+          conda create -n e2e_ci python=${{ inputs.python }} cmake ninja -y
           source activate e2e_ci
-          conda install intel::mkl-static intel::mkl-include -y
+          conda install -c intel mkl-static mkl-include -y
```
chuanqi129, line 60: why we need this change?

.github/workflows/pull.yml
```diff
@@ -93,6 +94,7 @@ jobs:
             fi
           done
           # Run Pytorch XPU python UT
+          pip install -r .ci/docker/requirements-ci.txt
```
chuanqi129, line 97: we can mv this line before all UT tests

.github/workflows/inductor_xpu_e2e_nightly.yml
mengfei25: Ever use intel::mkl-static in CPU inductor test, it may be failed and don't know why.

PR #395, opened 2024-06-12
Add aten::_foreach_lerp and its variants

fengyuan14:
Including aten::_foreach_lerp/_/.List/.Scalar


dvrogozh:
Thanks. That's one of the ops needed in https://github.com/pytorch/pytorch/issues/127931.

xytintel (approved):

xytintel:
Pls add fine grain test case

fengyuan14:
> Pls add fine grain test case

Fine gran cases base on test_ops. test_ops will test all operators in op_db. But foreach operators are not covered by test_ops, since foreach operators have their own operator database, not op_db.

PR #420, opened 2024-06-18
fix bug of pre ci

yuchengliu1:


test/xpu/run_test_with_skip.py
```diff
@@ -992,6 +992,7 @@ def launch_test(test_case, skip_list=None, exe_list=None):
 # test_sort_and_select
 skip_list = (
     # The following isin case fails on CPU fallback, as it could be backend-specific.
+    "test_isin_xpu_float16", # RuntimeError: "isin_default_cpu" not implemented for 'Half'
```
fengyuan14, line 995: Please check if the skip is needed, since we have got correct dtypes for XPU test.

yuchengliu1: This case is still fail. It has a cuda/xpu bias decorator like `@dtypesIfCUDA(*all_types_and(torch.half))`. It does not get dtype from opdb.

fengyuan14 (approved):

PR #428, opened 2024-06-19
Add aten::_nnz/_values for SparseXPU and SparsCsrXPU dispatch key

fengyuan14:


majing921201 (approved):

fengyuan14:
No dedicated cases cover the operators. The operators are covered by cases of sparse related calculation operators.

PR #429, opened 2024-06-19
Add complex half for sigmoid kernel

yucai-intel:


fengyuan14:
Wait for related case ported. And need switch the case on.

fengyuan14 (approved):

PR #439, opened 2024-06-20
Add aten::std/var/_mean and thier variants

fengyuan14:


test/xpu/xpu_test_utils.py
```diff
@@ -95,6 +95,8 @@
     "sum",
     "amin",
     "amax",
+    "std_mean",
+    "var_mean",
```
majing921201, line 99: Can we also add "std", "var" ?

fengyuan14: Complemented.

majing921201 (approved):

PR #459, opened 2024-06-25
Optimize nightly test

mengfei25:
1. check test status in the end
2. always upload artifacts


.github/actions/inductor-xpu-e2e-test/action.yml
```diff
@@ -158,11 +154,6 @@ runs:
           cat $var >> inductor_log/summary_accuracy.csv
```
chuanqi129, line 154: Why we need this csv file? And if the on-demand test both for accuracy and performance, all csv file will be concat together, for example https://github.com/intel/torch-xpu-ops/actions/runs/9659586596

mengfei25: Because want to keep one file for all tests instead of to check one-by-one when finished. I will split the performance and accuracy.

chuanqi129: Seems we don't need this file anymore? I don't find it has been used in any place

chuanqi129 (approved):

PR #465, opened 2024-06-25
Update test_binary_ufuncs_xpu.py UT

PenghuiCheng:
Update test_binary_ufuncs_xpu.py unit test cases.


test/xpu/run_test_with_skip.py
```diff
@@ -950,11 +950,9 @@ def launch_test(test_case, skip_list=None, exe_list=None):
 
 # test_binary_ufuncs
 skip_list = (
-    "jiterator", # Jiterator is only supported by CUDA
-    "cuda", # Skip cuda hard-coded case
+    "test_cuda_tensor_pow_scalar_tensor_xpu",
```
fengyuan14, line 953: What's the reason of skip?

PenghuiCheng: the reason is the result dtype of pow in numpy is float64, and dtype of torch.pow is float32:
AssertionError: The values for attribute 'dtype' do not match: torch.float32 != torch.float64.
Will debug it later.

daisyden (approved):
LGTM

PR #516, opened 2024-07-01
Add aten::is_set_to

fengyuan14:


fengyuan14:
The case was enabled.
`test_torch_xpu::test_is_set_to`

majing921201 (approved):

PR #531, opened 2024-07-03
Support XPU ABI=0 build

guangyey:
# Motivation
Support XPU ABI neutral build.
Starting from compiler 2025.0, `libsycl.so` is ABI-neutral lib. Before this, the `libsycl-preview.so` is ABI-neutral.

refer to https://github.com/pytorch/pytorch/pull/130110

### DIFF
--- a/cmake/Modules/FindSYCL.cmake
+++ b/cmake/Modules/FindSYCL.cmake
@@ -407,7 +407,6 @@ macro(SYCL_LINK_DEVICE_OBJECTS output_file sycl_target)
       OUTPUT ${output_file}
       DEPENDS ${object_files}
       COMMAND ${SYCL_EXECUTABLE}
-      -fsycl
       ${SYCL_device_link_flags}
       -fsycl-link ${object_files}
       -Xs "\"${SYCL_OFFLINE_COMPILER_FLAGS}\""
@@ -471,7 +470,7 @@ macro(SYCL_ADD_LIBRARY sycl_target)
   target_link_libraries(
     ${sycl_target}
     ${SYCL_LINK_LIBRARIES_KEYWORD}
-    ${SYCL_LIBRARIES})
+    ${SYCL_LIBRARY})
 
   set_target_properties(${sycl_target}
     PROPERTIES
@@ -530,7 +529,7 @@ macro(SYCL_ADD_EXECUTABLE sycl_target)
   target_link_libraries(
     ${sycl_target}
     ${SYCL_LINK_LIBRARIES_KEYWORD}
-    ${SYCL_LIBRARIES})
+    ${SYCL_LIBRARY})
 
   set_target_properties(${sycl_target}
     PROPERTIES
--- a/cmake/Modules/FindSYCLToolkit.cmake
+++ b/cmake/Modules/FindSYCLToolkit.cmake
@@ -25,15 +25,19 @@ This will define the following variables:
 
 #]=======================================================================]
 
-set(SYCLTOOLKIT_FOUND False)
-include(${CMAKE_ROOT}/Modules/FindPackageHandleStandardArgs.cmake)
+include(${TORCH_ROOT}/cmake/Modules/FindSYCLToolkit.cmake)
 
-set(SYCL_ROOT "")
-if(DEFINED ENV{SYCL_ROOT})
-  set(SYCL_ROOT $ENV{SYCL_ROOT})
-elseif(DEFINED ENV{CMPLR_ROOT})
-  set(SYCL_ROOT $ENV{CMPLR_ROOT})
+if(NOT SYCL_FOUND)
+  set(SYCLTOOLKIT_FOUND FALSE)
+  return()
 endif()
+
+if(SYCLTOOLKIT_FOUND)
+  return()
+endif()
+set(SYCLTOOLKIT_FOUND TRUE)
+
+include(${CMAKE_ROOT}/Modules/FindPackageHandleStandardArgs.cmake)
 
 if(WIN32)
   set(SYCL_EXECUTABLE_NAME icx)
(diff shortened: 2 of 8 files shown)


src/ATen/native/xpu/sycl/AmpKernels.cpp
```diff
@@ -102,9 +102,9 @@ void amp_foreach_non_finite_check_and_unscale_kernel(
 }
 
 struct AmpUpdateScaleKernelFunctor {
-  void operator()(sycl::item<1> item) const {
+  void operator()(sycl::nd_item<1> item) const {
     // There is only single item/task scheduled.
-    if (item.get_linear_id() != 0)
+    if (item.get_global_linear_id() != 0)
       return;
```
guangyey, line 108: otherwise, raise error
```
/home/guangyey/intel/oneapi/compiler/2024.1/bin/compiler/../../include/sycl/handler.hpp:1360:9: error: static assertion failed due to requirement 'std::is_convertible_v<sycl::nd_item<1>, sycl::item<1, true>>': Kernel argument of a sycl::parallel_for with sycl::nd_range must be either sycl::nd_item or be convertible from sycl::nd_item
 1360 |         std::is_convertible_v<sycl::nd_item<Dims>, LambdaArgType>,
```

cmake/Modules/FindSYCLToolkit.cmake
```diff
@@ -202,6 +202,11 @@ set(SYCL_FLAGS "")
 set(SYCL_LINK_FLAGS "")
 list(APPEND SYCL_FLAGS "-fsycl")
 list(APPEND SYCL_LINK_FLAGS "-fsycl")
+# Workaround about XPU ABI neutral build
+if(LINUX)
```
min-jean-cho, line 206: Let's do the same for Windows.

cmake/Modules/FindSYCL.cmake
```diff
@@ -427,7 +427,6 @@ macro(SYCL_LINK_DEVICE_OBJECTS output_file sycl_target)
       OUTPUT ${output_file}
       DEPENDS ${object_files}
       COMMAND ${SYCL_EXECUTABLE}
-      -fsycl
```
min-jean-cho, line 430: Why `-fsycl` needs to be removed? Should `-fpreview-breaking-changes -fsycl` link against `sycl-preview`?

cmake/Modules/FindSYCLToolkit.cmake
guangyey: Seems like Windows doesn't have [dual abi](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html) issue.

cmake/Modules/FindSYCL.cmake
guangyey: `-fsycl` is already contained in `SYCL_device_link_flags`.
Yes, the goal is to link against `sycl-preview`.

cmake/Modules/FindSYCLToolkit.cmake
```diff
@@ -96,7 +96,7 @@ find_file(
 
 find_library(
   SYCL_LIBRARY
-  NAMES sycl
+  NAMES sycl-preview
```
min-jean-cho, line 99: Port Windows path as in [here](https://github.com/pytorch/pytorch/blob/main/cmake/Modules/FindSYCLToolkit.cmake#L58-L74) to link against `sycl7.lib` instead of `sycl-preview`.

min-jean-cho (approved):
LGTM, otherwise.

cmake/Modules/FindSYCLToolkit.cmake
guangyey: Update.
refer to https://github.com/pytorch/pytorch/pull/130110/files#diff-65bed5904d10dccd55edec2d52c25c4ac49db608da9c650ada99cb409675e168R58-R74

cmake/Modules/FindSYCL.cmake
EikanWang: Regarding these SYCL libs, I'd like to reuse PyTorch variable rather than hard coding it.

cmake/Modules/FindSYCLToolkit.cmake
EikanWang: ditto

src/ATen/native/xpu/sycl/AmpKernels.cpp
fengyuan14: The pairs of kernel launch usage,
1. sycl::queue::parallel_for(sycl::range) + sycl::item
2. sycl::queue::parallel_for(sycl::nd_range) + sycl::nd_item

So after the fixing, we follow 2.

cmake/Modules/FindSYCLToolkit.cmake
fengyuan14, line 1: Do we need the file any more. Can we reuse all in `${TORCH_ROOT}/cmake/Modules/FindSYCLToolkit.cmake`?

fengyuan14: I see. Some specific logic is inevitable in torch-xpu-ops.

src/ATen/native/xpu/sycl/AmpKernels.cpp
fengyuan14: Interesting thing is there is no compilation error before?

cmake/Modules/FindSYCLToolkit.cmake
guangyey: yes

cmake/Modules/FindSYCLToolkit.cmake
guangyey: updated

src/ATen/native/xpu/sycl/AmpKernels.cpp
guangyey: compilation error only happens when abi=0

dvrogozh (changes requested):
Same comment as in https://github.com/pytorch/pytorch/pull/130110#pullrequestreview-2194153211. `ABI=0`, `ABI neutral`, usage of "preview" requires detailed explanation in PR description and commit message.

PR #532, opened 2024-07-03
Add aten::log_sigmoid and its variants

yucai-intel:
- [x]  log_sigmoid.out
- [x] log_sigmoid
- [x] log_sigmoid_forward.output
- [x] log_sigmoid_forward
- [x] log_sigmoid_backward.grad_input
- [x] log_sigmoid_backward


yaml/xpu_functions.yaml
```diff
@@ -361,6 +361,12 @@ supported:
   - hardsigmoid_
   - hardsigmoid_backward.grad_input
   - hardsigmoid_backward
+  - log_sigmoid.out
+  - log_sigmoid
```
majing921201, line 365: no need to register log_sigmoid.out/log_sigmoid, It's not structure delegation, CUDA doesn't register too

yucai-intel: Removed.

fengyuan14 (approved):

PR #542, opened 2024-07-06
Using kernel specific max work group size instead of device max work …

fengyuan14:
…group size.

Max work group size of kernel is not a static and device related only property. It now in SYCL depends on driver/compiler implementation. Device max work group means the probable max work group size allowd by the device. But actual max work group size depends on driver/compiler implementation, like compilaton optimization. Using kernel specific max work group size could get actual max work group allowed correctly. For example, on Xe, if compiler chooses SIMD16 and large GRF (32 HW threads per SS), the actual max work group size will be 512 (16 * 32), not 1024 queried by device::info::max_work_group_size.


majing921201:
I have a question about **syclMaxWorkItemsPerEU**, **syclMaxWorkItemsPerTile** API, which defined by us. We use them to compute actual work group size, when compiler doesn't give us correct number before. Now they seems duplicate, right ?

fengyuan14:
> I have a question about `syclMaxWorkItemsPerEU` API, which defined by us. We use them to compute actual work group size per EU, when compiler doesn't give us correct number before. Now they seems duplicate, right ?

Yes, we can make `syclMaxWorkItemsPerEU` and `syclMaxWorkItemsPerTile` actual. Since we asked compiler for simd width of the kernel.

majing921201 (approved):

majing921201:
This PR LGTM
But I

PR #545, opened 2024-07-09
[PyTorch uplift breaks XPU] Follow utils usage of latest PyTorch definition

fengyuan14:


xytintel (approved):

dvrogozh (approved):
yeah, was about to submit the same. The problem is however that this won't fix pytorch build immediately since pytorch hardcodes some prev. revision of the project: https://github.com/pytorch/pytorch/blob/main/third_party/xpu.txt.

@fengyuan14 : have you considered to track torch-xpu-ops `main` instead of hardcoding specific revision (only in pytorch main, of course)?

fengyuan14:
> yeah, was about to submit the same. The problem is however that this won't fix pytorch build immediately since pytorch hardcodes some prev. revision of the project: https://github.com/pytorch/pytorch/blob/main/third_party/xpu.txt.
> 
> @fengyuan14 : have you considered to track torch-xpu-ops `main` instead of hardcoding specific revision (only in pytorch main, of course)?

We discussed with Meta about this. Meta people suggested us using commit pin. Although they didn't tell the reason, I thought we did same thing as other third party components. Third party components always have a commit pin.
So regarding the break issue of building, we will give them feedbacks and expect these breaks would not happen frequently.

EikanWang (approved):

PR #552, opened 2024-07-09
build: fix -Wunused-function warning

dvrogozh:
PyTorch enforced unused-function and unused-variable compilation options recently. It leads to torch-xpu-ops building failure with -Werror.
Fixing: https://github.com/pytorch/pytorch/pull/130084
```
build/aten/src/ATen/xpu/RegisterXPU.cpp:5452:13: warning: ‘void at::xpu_force_fallback(const c10::OperatorHandle&, torch::jit::Stack*)’ defined but not used
```

cc: @fengyuan14

### DIFF
--- a/src/ATen/native/xpu/XPUFallback.template
+++ b/src/ATen/native/xpu/XPUFallback.template
@@ -109,10 +109,6 @@ static void xpu_lazy_registration_or_error_fallback(
   }
 }
 
-static void xpu_force_fallback(
-    const c10::OperatorHandle& op,
-    torch::jit::Stack* stack) {}
-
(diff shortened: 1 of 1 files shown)


fengyuan14 (approved):

fengyuan14:
@dvrogozh  Fail due to case regression. PyTorch should changes the case logic. I cannot push a commit to skip the case, since it's your forked branch. I will refile a PR for building issue and skipping case.

fengyuan14:
We need fix it ASAP to resolve the break issues in PyTorch preci. So switch to https://github.com/intel/torch-xpu-ops/pull/554
Thanks for your PR. @dvrogozh

PR #568, opened 2024-07-11
Add aten::adaptive_max_pool2d/backward and their variants

fengyuan14:


src/ATen/native/xpu/AdaptiveMaxPooling2d.cpp
```diff
@@ -0,0 +1,214 @@
...
+      grad_output, input, indices, grad_input);
+  return grad_input;
+}
+
+Tensor& XPUNativeFunctions::adaptive_max_pool2d_backward_out(
+    const Tensor& grad_output,
+    const Tensor& input,
+    const Tensor& indices,
+    Tensor& grad_input) {
+  TensorArg grad_input_arg{grad_input, "grad_input", 1};
```
xytintel, line 195: where is "alertNotDeterministic"

xytintel (approved):

src/ATen/native/xpu/AdaptiveMaxPooling2d.cpp
fengyuan14: It's the logic of kernel. So I put it in kernel level.

PR #569, opened 2024-07-11
Add aten::histogram and variant

majing921201:
CPU/MPS only ops


majing921201:
two cases failed:
SampleInput(input=tensor([-6.7103,  5.7013,  0.7920,  2.8814, -4.1030], device='xpu:0'), args=(tensor([-6.7103,  5.7013,  0.7920], device='xpu:0'),), kwargs={'weight': None, 'density': False}, broadcasts_input=False, name='')

SampleInput(input=6.362441917740258, args=(tensor([ 6.3624, -3.1650,  7.6229,  4.4501], device='xpu:0',
       dtype=torch.float64),), kwargs={'weight': None, 'density': False}, broadcasts_input=False, name='')

cases are not reasonable, for out-of-order bin.
1, pytorch doc requires increasing bin
image
2, CPU implemented kernel with std::upper_bound, which also requires increasing sequence

src/ATen/native/xpu/Histogram.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+std::tuple<Tensor&, Tensor&> XPUNativeFunctions::histogram_out(
+    const Tensor& self,
+    int64_t bin_ct,
+    std::optional<c10::ArrayRef<double>> range,
+    const std::optional<Tensor>& weight,
+    bool density,
+    Tensor& hist,
+    Tensor& bin_edges) {
+  globalContext().alertNotDeterministic("histogram_bin_count_xpu");
```
fengyuan14, line 171: Move it into kernel level as it depends kernel implementation.

test/xpu/extended/run_test_with_skip.py
```diff
@@ -162,6 +162,10 @@
     "test_compare_cpu_std_mean_xpu_bfloat16",
     "test_compare_cpu_sub_xpu_float16",
     "test_compare_cpu_var_mean_xpu_bfloat16",
+
+    # test case doesn't make sense, will file an issue to track it.
```
fengyuan14, line 166:
```suggestion
    # test case doesn't make sense, will file an issue to track it.
    # https://github.com/pytorch/pytorch/issues/130916
```

test/xpu/run_test_with_skip.py
```diff
@@ -782,6 +782,8 @@ def launch_test(test_case, skip_list=None, exe_list=None):
     
     # torch.complex32 - "sinh_cpu" not implemented for 'ComplexHalf'
     "test_dtypes_cosh_xpu",
+    # CUDA doesn't support operator
+    "test_dtypes_histogram_xpu",
```
fengyuan14, line 786: What's the error log of the case? Maybe we have
correct comments?

fengyuan14 (approved):

PR #581, opened 2024-07-15
Split binary of PyTorch XPU operators into multiple libraries

fengyuan14:
Including 1. libtorch_xpu.so (Host only, operator level) 2. libtorch-xpu-ops-sycl-ker-partx.so (Device code) libtorch-xpu-ops-sycl-ker-partx.so at most includes 50 SYCL kernel related source files, will be about 1GB.

### DIFF
--- a/src/ATen/native/xpu/sycl/MultinomialKernel.cpp
+++ b/src/ATen/native/xpu/sycl/MultinomialKernel.cpp
@@ -450,8 +450,11 @@ void multinomial_kernel(
             std::min(maxThreads, requiredSubGroups * SubGroupSize);
         int requiredShared = requiredThreads * sizeof(accscalar_t);
         if (n_sample == 1 && maxShared >= requiredShared) {
-          Tensor sampled =
-              at::detail::empty_xpu({numDist, n_sample}, self_v.options());
+          // Break per source build. Build every kernel source file into a so.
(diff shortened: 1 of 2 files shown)


fengyuan14:
image

EikanWang:
In general, the split solution looks good to me. But we should categorize the kernel implementations, like binary, unary, reduction, and so on. 

In addition, please help file an issue. We need to understand how the SYCL compiler organizes the object file. Is there any spec to describe the format of SYCL object files? I'm trying to understand the root cause and check if we should submit requirements to the compiler team.

fengyuan14:
> In general, the split solution looks good to me. But we should categorize the kernel implementations, like binary, unary, reduction, and so on.
> 
> In addition, please help file an issue. We need to understand how the SYCL compiler organizes the object file. Is there any spec to describe the format of SYCL object files? I'm trying to understand the root cause and check if we should submit requirements to the compiler team.

1. May I know what's the intention to categorize lib as what kernels are included. So far, in my understanding, they are opaque to users?

2. Most likely, there is nothing specific in last linkage step to reorganize structure of device code in data segment, since we are using gcc to do final linkage. It might be the default behavior of gcc. I will file issue to track it, if we have a better solution.

EikanWang (approved):

PR #606, opened 2024-07-17
Add aten::polar and its variants

yucai-intel:


fengyuan14:
Looks we have better dtype coverage than expected?
image

yucai-intel:
The data types are consistent with those supported by CUDA.

fengyuan14:
Please skip the case, and add error log as comments

> The data types are consistent with those supported by CUDA. Screenshot 2024-07-22 152455

The error log is saying unexpected data type is supported in Polar backward. Do yo know which operator is invoked in backward? Can you have an investigation, if the backward op supports bfloat16, does not aligned with CUDA, you can skip the case and add error log and list the backward operator as comments.

yucai-intel:
Polar's backward is calculated using complex(), which does not support bfloat16, so the test of this datatype is skipped.

fengyuan14:
> o

If we align with CUDA, CUDA should fail as well?

yucai-intel:
Cuda fails with same error:
RuntimeError: Expected both inputs to be Half, Float or Double tensors but got BFloat16 and BFloat16.

test/xpu/extended/run_test_with_skip.py
```diff
@@ -154,6 +154,10 @@
     # Greatest relative difference: 0.00396728515625 at index (610,) (up to 0.001 allowed)
     "test_compare_cpu_hypot_xpu_bfloat16",
 
+    # RuntimeError: Expected both inputs to be Half, Float or Double tensors but got BFloat16 and BFloat16.
+    # Polar's backward is calculated using complex(), which does not support bfloat16. Cuda fails with same error.
```
fengyuan14, line 158:
```suggestion
    # Polar's backward is calculated using complex(), which does not support bfloat16. CUDA fails with same error.
```

src/ATen/native/xpu/TensorFactories.cpp
```diff
@@ -151,6 +151,29 @@ Tensor& XPUNativeFunctions::complex_out(
...
+  return result;
+}
+
+Tensor XPUNativeFunctions::polar(const Tensor& abs, const Tensor& angle) {
+  complex_check_floating(abs, angle);
+  c10::TensorOptions options = abs.options();
+  options = options.dtype(toComplexType(abs.scalar_type()));
+  Tensor result = at::empty(0, options);
+  return at::polar_out(result, abs, angle);
+}
```
fengyuan14, line 175:
```suggestion
```

yaml/xpu_functions.yaml
```diff
@@ -268,6 +268,8 @@ supported:
   - eye.m_out
   - _efficientzerotensor
   - complex.out
+  - polar.out
+  - polar
```
fengyuan14, line 272:
```suggestion
```

fengyuan14 (approved):

PR #609, opened 2024-07-17
add hook for _check_arg_device2 and freeze_rng_state to support xpu

daisyden:
Added hook to support xpu backend, so that test_multihead_self_attn_two_masks_fast_path_mock_xpu and test_Dropout1d_xpu and test_Dropout3d_xpu can pass.


fengyuan14:
Can we give a simple summary as title, and an elaboration for description?

fengyuan14 (approved):

Issue #253, opened 2024-05-16
[Evaluated] OneDNN issue in TestMathBits of test_ops.py


daisyden:
### 🐛 Describe the bug
There are THREE kinds of issues:

1. TestMathBitsXPU , totally 200 cases got RuntimeError: Double and complex datatype matmul is not supported in oneDNN
ONEDNN_VERBOSE=2 PYTORCH_ENABLE_XPU_FALLBACK=1 PYTORCH_TEST_WITH_SLOW=1 pytest -v test_ops_xpu.py -k 'test_conj_view_addmm_xpu_complex64'
 
    "test_conj_view___rmatmul___xpu_complex64",
    "test_conj_view__refs_linalg_svd_xpu_complex64",
    "test_conj_view_addmm_decomposed_xpu_complex64",
    "test_conj_view_addmm_xpu_complex64",
    "test_conj_view_addmv_xpu_complex64",
    "test_conj_view_addr_xpu_complex64",
    "test_conj_view_baddbmm_xpu_complex64",
    "test_conj_view_bmm_xpu_complex64",
    "test_conj_view_cholesky_inverse_xpu_complex64",
    "test_conj_view_cholesky_solve_xpu_complex64",
    "test_conj_view_cholesky_xpu_complex64",
    "test_conj_view_corrcoef_xpu_complex64",
    "test_conj_view_cov_xpu_complex64",
    "test_conj_view_einsum_xpu_complex64",
    "test_conj_view_geqrf_xpu_complex64",
    "test_conj_view_inner_xpu_complex64",
    "test_conj_view_linalg_cholesky_ex_xpu_complex64",
    "test_conj_view_linalg_cholesky_xpu_complex64",
    "test_conj_view_linalg_cond_xpu_complex64",
    "test_conj_view_linalg_det_singular_xpu_complex64",
    "test_conj_view_linalg_det_xpu_complex64",
    "test_conj_view_linalg_eig_xpu_complex64",
    "test_conj_view_linalg_eigh_xpu_complex64",
    "test_conj_view_linalg_eigvals_xpu_complex64",
    "test_conj_view_linalg_eigvalsh_xpu_complex64",
    "test_conj_view_linalg_householder_product_xpu_complex64",
    "test_conj_view_linalg_inv_ex_xpu_complex64",
    "test_conj_view_linalg_inv_xpu_complex64",
    "test_conj_view_linalg_ldl_factor_ex_xpu_complex64",
[...]

PR #617, opened 2024-07-18
Add skip cases

majing921201:
RuntimeError: Double and complex datatype matmul is not supported in oneDNN


majing921201:
updated https://github.com/intel/torch-xpu-ops/issues/253

fengyuan14 (approved):

PR #619, opened 2024-07-19
Disable nondeterministic implementation for `max_pool2d_backward`

Stonepia:
We disable the nondeterministic implementation of `max_pool2d_backward` by default, so that to avoid potential accuracy issues.

The background is that we encountered eager two-run-differ problems with `alexnet` (without enabling deterministic). However, CUDA could run without enabling it. To align with the CUDA's behavior, we decided to always go to the deterministic path in this kernel.

 This PR adds a flag `XPU_ALLOW_UNDETERMINISTIC`. In the future, we may enable this path for performance consideration.


fengyuan14 (approved):

fengyuan14:
Please fix compilation error.

fengyuan14 (approved):

PR #624, opened 2024-07-22
Reduce register pressure for radix sort kernel

xytintel:
With both key and value data type being 64 bytes, there will be occasional computation errors on MTL machines. This issue can be resolved by reducing register pressure.

### DIFF
--- a/src/ATen/native/xpu/sycl/SortingKernels.h
+++ b/src/ATen/native/xpu/sycl/SortingKernels.h
@@ -419,50 +419,54 @@ void segmented_sort_pairs_(
     value_t* values_out,
     int num_segments,
     int num_elements) {
-  if (num_elements > 4096) {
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -419,50 +419,54 @@ void segmented_sort_pairs_(
     value_t* values_out,
     int num_segments,
     int num_elements) {
-  if (num_elements > 4096) {
+  constexpr int scaling_coef = sizeof(key_t) * sizeof(value_t) >= 64
+      ? 2
+      : 1; // Attempt to reduce register pressure. For some reason, register
```
fengyuan14, line 424: Please change the comment, since we have no idea of the root cause. The cause might not be memory spill. Just to say, the result will be incorrect when using too many local variables (registers). Please also paste an issue here to track the root cause of it.

fengyuan14 (approved):

PR #635, opened 2024-07-23
SYCL: using SYCL group algorithm API instead of old style for sub group shift utilities

guangyey:
Change 1:
SYCL is shifting to SYCL group algorithm API to unify sub-group, work-group APIs, e.g. reduce_over_group, group_barrier, and shift_group here. And old style (use separate member function of each structure) is being deprecated.

| Old | SYCL-2020|
| ------------- | ------------- |
| sg.shuffle_down(x, 1) | sycl::shift_group_left(sg, x, 1) |
| sg.shuffle_up(x, 1) | sycl::shift_group_right(sg, x, 1) |
| sg.shuffle(x, id) | sycl::select_from_group(sg, x, id) |
| sg.shuffle_xor(x, mask) | sycl::permute_group_by_xor(sg, x, mask) |

FYI: Please **don't** use sg.shuffle members anymore since they are deprecated.

Change 2:
`sycl::shfit_group_xxx` is more restricted than `sg.shuffle`, requiring the object shifted is C++ trivially copyable object. We implemented private `pair` instead of `std::pair` in the commit.
FYI: CUDA is using `thrust::pair` in kernel.


EikanWang (approved):

fengyuan14 (approved):

fengyuan14:
What's the bundle required?
image

guangyey:
> What's the bundle required? image

THe reason is that `std::pair` is not supported in `sycl::shift_group_left`. @xytintel will be going to implement a customed `pair` structure, a countpart `thrust::pair`, supported in `sycl::shift_group_left` and use the customed `pair` in our SYCL kernel code.

fengyuan14:
```
/opt/intel/oneapi/pytorch-gpu-dev-0.5/include/sycl/group_algorithm.hpp:541:1: note: candidate template ignored: requirement 'std::is_trivially_copyable_v<std::pair<unsigned char, unsigned char>> || detail::is_vec<std::pair<unsigned char, unsigned char>>::value' was not satisfied [with Group = sub_group, T = std::pair<unsigned char, unsigned char>]
```
Cannot totally understand the failure. As SYCL spec, std::pair is supposed to C++ trivially copyable.
image

fengyuan14:
Verify on c++17, std::pair is not a C++ trivially copyable structure.
```
(dev) fengyuan@fy-9900:~/workspace/test$ cat test.cpp
#include <iostream>
#include <utility>
#include <type_traits>

int main() {
  std::cout << std::is_trivially_copyable<std::pair<unsigned char, unsigned char>>::value << std::endl;
  return 0;
}
(dev) fengyuan@fy-9900:~/workspace/test$ g++ -std=c++17 test.cpp  -o test
(dev) fengyuan@fy-9900:~/workspace/test$ ./test
0
```

fengyuan14:
> ```
> /opt/intel/oneapi/pytorch-gpu-dev-0.5/include/sycl/group_algorithm.hpp:541:1: note: candidate template ignored: requirement 'std::is_trivially_copyable_v<std::pair<unsigned char, unsigned char>> || detail::is_vec<std::pair<unsigned char, unsigned char>>::value' was not satisfied [with Group = sub_group, T = std::pair<unsigned char, unsigned char>]
> ```
> 
> Cannot totally understand the failure. As SYCL spec, std::pair is supposed to C++ trivially copyable. image

Spec is saying the std structures listed are a plus list to support device copyable in SYCL.

fengyuan14:
Verified,
```
(dev) fengyuan@fy-9900:~/workspace/test$ cat test.cpp
#include <iostream>
#include <utility>
#include <type_traits>
#include <tuple>
#include <sycl/sycl.hpp>

int main() {
  std::cout << std::is_trivially_copyable<std::pair<unsigned char, unsigned char>>::value << std::endl;
  std::cout << sycl::is_device_copyable<std::pair<unsigned char, unsigned char>>::value << std::endl;
  return 0;
}
(dev) fengyuan@fy-9900:~/workspace/test$ icpx -fsycl -std=c++17 test.cpp  -o test
(dev) fengyuan@fy-9900:~/workspace/test$ ./test
0
1
```

jbrodman:
Hi - I think several cases where you're migrating to shifts or permutes could also be greatly simplified by using SYCL 2020 reduce_over_group at either sub-group or work-group scope.

jbrodman:
Additionally, any calls to nd_item::barrier should move to sycl::group_barrier.

fengyuan14:
> Additionally, any calls to nd_item::barrier should move to sycl::group_barrier.

@jbrodman Thanks for the remainder.

Yes. It's on our plan. We reviewing all APIs which should move to group algorithm API in IPEX two years ago. But at that time, these APIs, like group_barrier or reduce_over_group, get some performance issues. So we didn't adopt them. For example, memory fence implicated in `group_barrier` is not performant than `sycl::nd_item::barrier`.

Exactly we should follow SYCL spec to move to new APIs and cowork with implementation team to optimize performance.

fengyuan14:
@xytintel Ready to preci?

xytintel:
> @xytintel Ready to preci?

I think yes

src/comm/XPUPair.h
```diff
@@ -0,0 +1,248 @@
+#pragma once
+
+#include <ATen/Context.h>
+
+namespace at::xpu {
+
+#pragma once
```
guangyey, line 7: Please remove this line(duplicated).

xytintel: Done

src/comm/XPUPair.h
```diff
@@ -0,0 +1,246 @@
...
+    class T1 = null_type,
+    class T2 = null_type,
+    class T3 = null_type,
+    class T4 = null_type,
+    class T5 = null_type,
+    class T6 = null_type,
+    class T7 = null_type,
+    class T8 = null_type,
+    class T9 = null_type>
+class tuple;
```
guangyey, line 47: where is `class tuple` used?

src/comm/XPUPair.h
```diff
@@ -0,0 +1,246 @@
...
+struct tuple_size<T const volatile> : public tuple_size<T> {};
+
+template <class T>
+struct tuple_size {
+  static const int value = 1 + tuple_size<typename T::tail_type>::value;
+};
+
+// specializations for tuple_size
+template <>
+struct tuple_size<tuple<>> {
```
guangyey, line 190: I am not sure if `tuple` and `tuple_size` are valid structs as they are defined as forward declarations.

jbrodman:
We just pushed a fix into IGC that should solve the barrier performance issue. It would be a huge benefit for the SYCL compiler and runtime to dogfood things like the group algorithms so we can make sure they're performing properly.

src/comm/XPUPair.h
```diff
@@ -0,0 +1,246 @@
...
+    class T8 = null_type,
+    class T9 = null_type>
+class tuple;
+
+template <typename T1, typename T2>
+inline void swap(T1& a, T2& b) {
+  T1 temp = a;
+  a = b;
+  b = temp;
+}
```
rolandschulz, line 54: Why is this reimplementing swap and not using `std::swap`? This doesn't work correctly if `T1` or `T2` have their custom swap specialization.

src/comm/XPUPair.h
```diff
@@ -0,0 +1,246 @@
...
+  template <typename U1, typename U2>
+  inline pair(const pair<U1, U2>& p) : first(p.first), second(p.second) {}
+
+  template <typename U1, typename U2>
+  pair(const std::pair<U1, U2>& p) : first(p.first), second(p.second) {}
+
+  inline void swap(pair& p) {
+    swap(first, p.first);
+    swap(second, p, second);
+  }
```
rolandschulz, line 78: Should use `std::swap` and typo "p, second".

```suggestion
  inline void swap(pair& p) {
    using std::swap;
    swap(first, p.first);
    swap(second, p.second);
  }
```

src/comm/XPUPair.h
```diff
@@ -0,0 +1,246 @@
...
+  inline typename tuple_element<1, Pair>::type& operator()(Pair& p) const {
+    return p.second;
+  }
+};
+
+} // namespace detail
+
+template <unsigned int N, typename T1, typename T2>
+inline typename tuple_element<N, pair<T1, T2>>::type& get(pair<T1, T2>& p) {
+  return detail::pair_get<N, pair<T1, T2>>()(p);
```
rolandschulz, line 237: unnecessary complicated.

```suggestion
  if constexpr(N==0) return p.first; else return p.second.;
```

xytintel: Agree with your proposal. I will remove the unnecessary code.

fengyuan14:
> We just pushed a fix into IGC that should solve the barrier performance issue. It would be a huge benefit for the SYCL compiler and runtime to dogfood things like the group algorithms so we can make sure they're performing properly.

We will change the APIs gradually. Thanks.

fengyuan14:
@rolandschulz Any more comments?

src/comm/XPUPair.h
```diff
@@ -0,0 +1,88 @@
...
+  inline pair(const T1& x, const T2& y) : first(x), second(y) {}
+
+  template <typename U1, typename U2>
+  inline pair(const pair<U1, U2>& p) : first(p.first), second(p.second) {}
+
+  template <typename U1, typename U2>
+  pair(const std::pair<U1, U2>& p) : first(p.first), second(p.second) {}
+
+  inline void swap(pair& p) {
+    std::swap(first, p.first);
```
rolandschulz, line 27: This still doesn't work with custom swap (e.g. if you have `pair<pair<...>...>`). Correct use needs to enable ADL. For details see e.g. https://stackoverflow.com/questions/28130671/how-does-using-stdswap-enable-argument-dependent-lookup-adl.

Nit: This member function seems unnecessary. The implementation could go directly into the free-function.

rolandschulz (commented):
no other comments

src/comm/XPUPair.h
fengyuan14: @xytintel Please address.

xytintel: Removed

src/comm/XPUPair.h
```diff
@@ -0,0 +1,83 @@
...
+}
+
+template <typename T1, typename T2>
+bool operator>=(const pair<T1, T2>& x, const pair<T1, T2>& y) {
+  return !(x < y);
+}
+
+template <typename T1, typename T2>
+inline void swap(pair<T1, T2>& x, pair<T1, T2>& y) {
+  return x.swap(y);
```
rolandschulz, line 59: you removed it but didn't move the implementation.

xytintel: Done

rolandschulz (approved):

PR #637, opened 2024-07-23
Enable weekly test

mengfei25:
1. enable weekly test contains 3 suites full e2e and full ut
2. always() to not cancelled()


.github/workflows/nightly_ondemand.yml
```diff
@@ -216,7 +216,7 @@ jobs:
           hf_token: ${{ secrets.HUGGING_FACE_HUB_TOKEN }}
       - name: Summarize archieve files
         id: summary
-        if: always()
+        if: ${{ ! cancelled() }}
```
chuanqi129, line 219: Why we need this change?

.github/workflows/weekly.yml
```diff
@@ -0,0 +1,277 @@
...
+        with:
+          suite: torchbench
+          env_prepare: true
+          dt: float32,bfloat16,float16,amp_bf16,amp_fp16
+          mode: inference,training
+          scenario: accuracy,performance
+          hf_token: ${{ secrets.HUGGING_FACE_HUB_TOKEN }}
+      # Timm_models
+      - name: Weekly Timm_models Full Test
+        uses: ./.github/actions/inductor-xpu-e2e-test
```
chuanqi129, line 174: Let's merge this file into nightly one, we only need add those 3 steps in nightly and make the if condition as `if: github.event.schedule ==  '0 16 * * 5'`

.github/workflows/weekly.yml
```diff
@@ -0,0 +1,277 @@
...
+            exit 0
+          fi
+          # Test Type
+          if [ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ];then
+            test_type="On-demand"
+            test_issue_id=426
+            cc_comment="CC @${GITHUB_TRIGGERING_ACTOR}"
+          else
+            test_type="Weekly"
+            test_issue_id=432
```
chuanqi129, line 258: Do we need a new issue for weekly?

.github/workflows/nightly_ondemand.yml
mengfei25: if set always, it will be always run even if cancelled.

.github/workflows/weekly.yml
mengfei25: will do

.github/workflows/weekly.yml
mengfei25: I think not, just test scope is different, and we can cancel the nightly in weekend, is it okay?

chuanqi129 (approved):

mengfei25:
This PR is not related with preci-ut

PR #652, opened 2024-07-25
Add aten::masked_scatter_

xytintel:
Add aten::masked_scatter_.


xytintel:
image
image

src/ATen/native/xpu/Indexing.cpp
```diff
@@ -44,4 +43,35 @@ Tensor XPUNativeFunctions::index_select(
...
+
+  c10::MaybeOwned<Tensor> b_mask =
+      expand_inplace(self, mask, "masked_scatter_");
+
+  if (self.numel() == 0) {
+    return self;
+  }
+
+  auto maskPrefixSum = at::empty(self.sizes(), mask.options().dtype(kLong));
+  native::xpu::launch_masked_scatter_kernel(
```
fengyuan14, line 71: To align our style, there should be no launch as prefix.

src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h
```diff
@@ -23,10 +22,10 @@ struct KSScanKernelFunctor : public __SYCL_KER_CONFIG_CONVENTION__ {
     // initialize local_input
     auto cur_init = init_;
     if (scan_type == 1) {
-      local_scan_[local_id] = first_[local_id];
+      local_scan_[local_id] = c10::load(&first_[local_id]);
```
fengyuan14, line 25: When we should use `c10::load` and when we could use subscript.

src/ATen/native/xpu/Indexing.cpp
xytintel: Fixed

fengyuan14 (approved):

PR #655, opened 2024-07-26
Use safe API to access Tensor data pointer

Stonepia:
The un-templated data_ptr() does not check about the type check and storage_initialized() check. Thus, it may introduce potential bugs.

By using the template data_ptr, this could throw an error like below:

```
torch._dynamo.exc.TorchRuntimeError: Failed running call_function torchvision.roi_align(*(FakeTensor(..., device='xpu:0', size=(1, 1024, 50, 75), dtype=torch.bfloat16), FakeTensor(..., device='xpu:0', size=(1000, 5), dtype=torch.bfloat16), 0.0625, 14, 14, 0, True), **{}):
The tensor has a non-zero number of elements, but its data is not allocated yet. Caffe2 uses a lazy allocation, so you will need to call mutable_data() or raw_mutable_data() to actually allocate memory.
```

### DIFF
--- a/src/ATen/native/xpu/NMS.cpp
+++ b/src/ATen/native/xpu/NMS.cpp
@@ -55,7 +55,7 @@ Tensor nms(const Tensor& dets, const Tensor& scores, double iou_threshold_) {
 
   at::Tensor keep =
       at::empty({dets_num}, dets.options().dtype(at::kLong).device(at::kCPU));
-  int64_t* keep_out = (int64_t*)keep.data_ptr();
+  int64_t* keep_out = keep.mutable_data_ptr<int64_t>();
 
   int num_to_keep = 0;
   for (int i = 0; i < dets_num; i++) {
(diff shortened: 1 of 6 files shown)


Stonepia:
Note that currently, there is no bug found in stock pytorch about this mentioned unsafe data ptr (but we found on in IPEX).

This PR doesn't address the `static_cast` data_ptr. These occurrences do not use the template data_ptr. Similarly, the data_ptr can't be set as `const_data_ptr`.

```C++
static_cast<TLMetaForAddressScalar<scalar_vals_t, depth>*>(
          addressStorage.data_ptr());

// Could not be changed to int64_t
static_cast<TLMetaForAddressScalar<scalar_vals_t, depth>*>(
          addressStorage.mutable_data_ptr<int64_t>());

// Set as mutable_data_ptr. Could not use const_data_ptr
static_cast<TLMetaForAddressScalar<scalar_vals_t, depth>*>(
          addressStorage.mutable_data_ptr());
``` 

To fix this, we may need to further change these static_cast usage (to align with PyTorch implementation).

EikanWang (approved):

PR #668, opened 2024-07-31
adaptive_avgpool2d_backward: Propagate memory format correctly

retonym:
In inductor usage, generated Triton kernel assumes specific memory layout after compile. The fallback operator in the graph need propagate memory format correctly.

### DIFF
--- a/src/ATen/native/xpu/AdaptiveAveragePooling2d.cpp
+++ b/src/ATen/native/xpu/AdaptiveAveragePooling2d.cpp
@@ -69,23 +69,23 @@ Tensor sum_backward(
 
 Tensor mean_backward(
     const Tensor& grad,
+    const Tensor& input,
     c10::SymIntArrayRef shape,
     OptionalIntArrayRef opt_dim,
     c10::SymInt numel,
-    bool keepdim,
-    const Tensor& input) {
+    bool keepdim) {
   bool is_all_reduce = !opt_dim.has_value() || opt_dim.value().empty();
   auto n =
       is_all_reduce ? std::move(numel) : _safe_size(shape, opt_dim.value());
 
-  Tensor grad_input_ =
+  Tensor grad_input =
       sum_backward(grad, shape, opt_dim, keepdim) / std::move(n);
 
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/AdaptiveAveragePooling2d.cpp
```diff
@@ -72,11 +72,20 @@ Tensor mean_backward(
...
   bool is_all_reduce = !opt_dim.has_value() || opt_dim.value().empty();
   auto n =
       is_all_reduce ? std::move(numel) : _safe_size(shape, opt_dim.value());
-  return sum_backward(grad, shape, opt_dim, keepdim) / std::move(n);
+
+  Tensor grad_input_ =
+      sum_backward(grad, shape, opt_dim, keepdim) / std::move(n);
+
+  if (input.suggest_memory_format() == at::MemoryFormat::ChannelsLast) {
+    grad_input_ = grad_input_.contiguous(input.suggest_memory_format());
```
fengyuan14, line 85: I could not understand the acc failure. The fixing here adjusts the memory format, there is no change in content of `grad_input_`. So when the next node in the autograd graph could handle plain format, the model should work.

src/ATen/native/xpu/AdaptiveAveragePooling2d.cpp
```diff
@@ -72,11 +72,20 @@ Tensor mean_backward(
     c10::SymIntArrayRef shape,
     OptionalIntArrayRef opt_dim,
     c10::SymInt numel,
-    bool keepdim) {
+    bool keepdim,
+    const Tensor& input) {
```
fengyuan14, line 76: Please put the `input` as the second parameter of the function.

src/ATen/native/xpu/AdaptiveAveragePooling2d.cpp
fengyuan14: Synced offline. There is an assumption in Inductor usage with Triton kernel. Inductor generates hardcoded memory access for specific memory layout. For example, Inductor enables CL optimization for a model, the Triton kernel will be generated for CL only. The side-effect is fallback operators should propagate memory layout correctly.

PR #679, opened 2024-08-01
reduction: Add an extra operation for data type casting as SYCL reduction kernel doesn't support dynamic cast so far.

yuchengliu1:
As pytorch comment https://github.com/pytorch/pytorch/blob/8ea4c72eb275d9ca97917dfd9386c93f16f4b49f/aten/src/ATen/native/ReduceOps.cpp#L1569.
Reduce backbone do not support dynamic input cast. Using `make_reduction_from_out_ty` instead of `make_reduction`.
fix for case on mtl:
    "test_python_ref__refs_all_xpu_complex128",
    "test_python_ref_executor__refs_all_executor_aten_xpu_complex128",
    "test_python_ref_torch_fallback__refs_all_xpu_complex128",


src/ATen/native/xpu/ReduceOps.cpp
```diff
@@ -320,7 +320,8 @@ inline TensorIterator get_allany_iter(
     const Tensor& result,
     OptionalIntArrayRef dims,
     bool keepdim) {
-  return meta::make_reduction(self, result, dims, keepdim, self.scalar_type());
+  return meta::make_reduction_from_out_ty(
```
fengyuan14, line 323: It looks a device agnostic change. Why we pass on PVC before the change?

fengyuan14 (approved):

yuchengliu1:
need to futher inspect why these case pass in PVC before

PR #680, opened 2024-08-02
changes to allow winUT

poliset1:
The following changes were made to compile and run on windows
1. resolve std namespace conflicts by changing it to ::std
2. __restrict__ causes an error on windows. 

This is still a draft as there is a link error that I am not able to resolve


src/ATen/native/xpu/sycl/BatchNormKernels.cpp
```diff
@@ -1008,9 +1008,9 @@ struct BatchNormCollectStatisticsChannelsLastKernelFunctor
   }
 
   BatchNormCollectStatisticsChannelsLastKernelFunctor(
-      const scalar_t* __restrict__ input,
-      accscalar_t* __restrict__ out_mean,
-      accscalar_t* __restrict__ out_invstd,
+      const scalar_t* input,
+      accscalar_t*  out_mean,
+      accscalar_t*  out_invstd,
```
fengyuan14, line 1013: The changes impact Linux path. You can declare a macro to use here, and distinguish Win/Linux in the macro.

src/ATen/native/xpu/sycl/LossNLL2dKernels.cpp
```diff
@@ -246,6 +246,7 @@ void nll_loss2d_forward_kernel(
   output.zero_();
   total_weight.zero_();
 
+
```
fengyuan14, line 249: Please remove these unnecessary changes.

fengyuan14:
Could not see the changes of `::std`. Is the PR done?

src/ATen/native/xpu/sycl/BatchNormKernels.cpp
min-jean-cho: Refer to here https://github.com/intel/torch-xpu-ops/blob/1d70431c072db889d9a47ea4956049fe340a426d/src/ATen/native/xpu/sycl/TensorTransformationsKernels.cpp#L9-L13

PR #690, opened 2024-08-05
MKL installation from conda to pypi

mengfei25:


mengfei25:
UT failed is the known random issue, not related with this change

Issue #256, opened 2024-05-16
[Evaluated] skips in test_module


fengyuan14:
### 🐛 Describe the bug

- [ ] test_cpu_gpu_parity_nn_CrossEntropyLoss_xpu_float16 // Should be open when nll_loss2d is enabled.
- [ ] oneDNN failures // Please check `run_test_with_skip`
- [ ] Lack of aten::_thnn_fused_gru_cell // Need evaluate whether operator is needed.
- [ ] CPU fallback failures // To evaluate when there is an XPU implementation.

### Versions

main

PenghuiCheng:
✔ test_cpu_gpu_parity_nn_CrossEntropyLoss_xpu_float16 // Should be open when nll_loss2d is enabled.-----removed from skip list
 ✔oneDNN failures // Please check run_test_with_skip----checked: https://github.com/intel/torch-xpu-ops/pull/691
 ✔Lack of aten::_thnn_fused_gru_cell // Need evaluate whether operator is needed.----it is only align with cuda.
 ✔CPU fallback failures // To evaluate when there is an XPU implementation.

PR #691, opened 2024-08-05
unittest: Remove skips from skip list

PenghuiCheng:
Remove unnecessary cases from skip list after evaluating https://github.com/intel/torch-xpu-ops/issues/256


PenghuiCheng:
Issue: https://github.com/intel/torch-xpu-ops/issues/256

fengyuan14 (approved):

Issue #611, opened 2024-07-18
MTL specific issues


daisyden:
Aug 2 updated status
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/daisyden/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/daisyden/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">

<div ccp_infra_version=3 ccp_infra_timestamp=1722592942444 ccp_infra_user_hash=3987611503
ccp_infra_copy_id="" data-ccp-timestamp=1722592942444>

test_reductions_xpu.py::TestReductionsXPU::test_median_nan_values_xpu_float64 | Nan | Penghui | P1 | passed
-- | -- | -- | -- | --
test_reductions_xpu.py::TestReductionsXPU::test_median_real_values_xpu_float64 | Nan | Penghui | P1 | passed
test_reductions_xpu.py::TestReductionsXPU::test_quantile_xpu_float64 | Nan | Penghui | P1 | passed
test_nn_xpu.py::TestNNDeviceTypeXPU::test_grid_sample_large_xpu | Nan | Jing | P1 | JIRA:https://jira.devtools.intel.com/browse/GSD-9622
TestCommonXPU.test_compare_cpu_index_put_xpu_bool | Inf | Daisy | P1 | Passed
TestCommonXPU.test_compare_cpu_index_put_xpu_uint8 | Inf | Daisy | P1 | Passed
TestCommonXPU.test_python_ref__refs_all_xpu_complex128 | Inf | Yucheng | P1 |  
TestCommonXPU::test_compare_cpu_all_xpu_complex128 | Acc | Yucheng | P1 | WIP
test_reductions_xpu.py::TestReductionsXPU::test_min_xpu_bool | Inf | Penghui->Yutao | P1 | WIP
TestCommonXPU.test_python_ref__refs_all_xpu_complex128 | Inf | Yucheng | P1 |  
[...]

PR #692, opened 2024-08-05
mtl: reduction: Bypass group reduction on SLM by sparing workload to other SGs when SIMD=8

xytintel:
Fixing: https://github.com/intel/torch-xpu-ops/issues/611 (test_min_xpu_bool)
Tracking: https://github.com/intel/torch-xpu-ops/issues/698
Another change: Aligning memory access utility `c10::load` with PyTorch usage, `c10::load` makes bool data with value compatible.

### DIFF
--- a/src/ATen/native/xpu/sycl/Reduce.h
+++ b/src/ATen/native/xpu/sycl/Reduce.h
@@ -245,6 +245,7 @@ struct ReduceConfig {
     // reduction of small shape input only requires some shift operations
     // in side of SG. It is functional WA. We got case failures on some
     // platforms supporting SIMD8.
+    // https://github.com/intel/torch-xpu-ops/issues/698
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/Reduce.h
```diff
@@ -241,19 +241,19 @@ struct ReduceConfig {
   template <typename T, class KernelClass>
   void set_group_dimension(int64_t dim0, int64_t dim1) {
     auto max_wg_sz = syclMaxWorkGroupSize<KernelClass>();
-    auto max_sg_sz = syclMaxSubGroupSize();
+    auto min_sg_sz = syclMinSubGroupSize();
```
fengyuan14, line 244: We should use `compile_sub_group_size`?

fengyuan14 (approved):

PR #693, opened 2024-08-05
arc: Comment failure of test_variable_sequence_xpu_float16 with justification after evaluation

hjhee:
[Certain step](https://github.com/pytorch/pytorch/blob/28b0ad4f4616497e9e297d3970992aad1d3cba44/aten/src/ATen/native/RNN.cpp#L740) in current LSTM implementation returns incorrect result consistently. It is believed that the root cause should be in Op `addmm` using oneDNN matmul implementation.


fengyuan14 (approved):

fengyuan14:
Arc specific. We are discussing to separate skip lists for different platforms.

PR #702, opened 2024-08-07
copy_: Align meta process of XPU implementation with PyTorch in-tree implementation

majing921201:
Add handle of ZeroTensor.

### DIFF
--- a/src/ATen/native/xpu/Copy.cpp
+++ b/src/ATen/native/xpu/Copy.cpp
@@ -293,25 +293,21 @@ void _copy_xpu(TensorIterator& iter, bool non_blocking) {
   }
 }
 
-Tensor& _copy_xpu(Tensor& self, const Tensor& src, bool non_blocking) {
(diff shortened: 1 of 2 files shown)


src/ATen/native/xpu/Copy.cpp
```diff
@@ -294,13 +294,45 @@ void _copy_xpu(TensorIterator& iter, bool non_blocking) {
...
+        "output with shape ",
+        self.sym_sizes(),
+        " doesn't match the broadcast shape ",
+        shape);
+    return self;
+  }
+
+  if (src.is_meta()) {
+    TORCH_CHECK_NOT_IMPLEMENTED(false, "Cannot copy out of meta tensor; no data!")
+  }
```
fengyuan14, line 314: We don't need it. Since, Meta dispatch key shares the code with CPU/CUDA. Here is XPU specific code path.

fengyuan14 (approved):

PR #709, opened 2024-08-07
_foreach_norm: Align with PyTorch operator semantics on allocation scheme of return tensors

chunhuanMeng:
PyTorch requires separate copies returned  in foreach_norm. The existing XPU implementation follows an out-of-date allocation scheme, to share storage among returned tensor. In latest PyTorch unit test, the behavior is not allowed.
related case:
- test_dispatch_meta_outplace__foreach_norm_xpu_bfloat16
- test_dispatch_meta_outplace__foreach_norm_xpu_float
- test_dispatch_symbolic_meta_outplace__foreach_norm_xpu_bfloat16
- test_dispatch_symbolic_meta_outplace__foreach_norm_xpu_float
- test_dispatch_symbolic_meta_outplace_all_strides__foreach_norm_xpu_float32
- test_meta_outplace__foreach_norm_xpu_bfloat16
- test_meta_outplace__foreach_norm_xpu_float


fengyuan14:
The error is exposed by -WError in preci. Please fix it,
image

src/ATen/native/xpu/sycl/ForeachReduceKernels.cpp
```diff
@@ -230,12 +234,18 @@ std::vector<Tensor> foreach_norm_kernel(
                     max_chunks_per_tensor);
 
                 // sum final val for all chunks
+                void* tensor_list_addresses[ntensors];
```
fengyuan14, line 237: Does the var need `const`?

src/ATen/native/xpu/sycl/ForeachReduceKernels.cpp
```diff
@@ -268,12 +278,18 @@ std::vector<Tensor> foreach_norm_kernel(
                     output_per_tensor.mutable_data_ptr<out_opmath_t>(),
                     max_chunks_per_tensor);
 
+                const void* tensor_list_addresses[ntensors];
+                for (int i = 0; i < ntensors; i++) {
+                  tensor_list_addresses[i] =
+                      ret_per_tensor[i].mutable_data_ptr<out_t>();
+                }
+
```
fengyuan14, line 286: Are these two code pieces all same? We can have only one before the `if-else`?

fengyuan14 (approved):

fengyuan14:
image

src/ATen/native/xpu/sycl/ForeachReduceKernels.cpp
```diff
@@ -194,8 +194,15 @@ std::vector<Tensor> foreach_norm_kernel(
...
-  auto res_option = options.dtype(output_dtype);
-  auto ret_per_tensor = at::empty({ntensors}, res_option);
+  std::vector<at::Tensor> ret_per_tensor;
+  ret_per_tensor.reserve(ntensors);
+  const auto res_option = options.dtype(output_dtype);
+  for (int i = 0; i < ntensors; i++) {
+    ret_per_tensor.push_back(at::empty({}, res_option));
+  }
+  sycl::queue q{sycl::property::queue::in_order()};
+  void** tensor_list_addresses = sycl::malloc_shared<void*>((ntensors), q);
```
fengyuan14, line 204: Do not use raw runtime malloc, but helpers provided by PyTorch XPU backend. The usage here, should be,
1. Allocating ping memory
2. Initing meta on ping memory
3. Allocating device memory
4. memcpy from ping to dev

fengyuan14 (approved):

PR #734, opened 2024-08-08
Port test_stable_sort_against_numpy

Stonepia:
CUDA has skipped the index_fill call in this UT by hard coding. We need to do the same. Otherwise, there will be `idx` tensor on CPU while `self` tensor on XPU, which would result in page fault.


Stonepia:
Need to be merged after https://github.com/intel/torch-xpu-ops/pull/735

fengyuan14:
Please resolve the error.
image

fengyuan14 (approved):

PR #735, opened 2024-08-08
index_fill: Add device check

Stonepia:
We followed PyTorch code gen, but which doesn't include common device check in index_fill. Add device check in operator source.


fengyuan14:
The PR depends on https://github.com/intel/torch-xpu-ops/pull/734

fengyuan14 (approved):

PR #736, opened 2024-08-09
reduce: Align device of temp buffer with input tensor

daisyden:
to fix torch.all() Runtime Error on xpu:1


fengyuan14:
Not a defect.

PR #739, opened 2024-08-09
 Add aten::histc and its variants

yucai-intel:


src/ATen/native/xpu/SummaryOps.cpp
```diff
@@ -21,4 +22,30 @@ Tensor XPUNativeFunctions::bincount(
...
 
+Tensor XPUNativeFunctions::histc(
+    const Tensor& self,
+    int64_t nbins,
+    const Scalar& min,
+    const Scalar& max) {
+  if (self.scalar_type() == ScalarType::Half) {
+    AT_ERROR("HalfTensor is not supported");
+  }
+  // See Note [Writing Nondeterministic Operations]
```
majing921201, line 33: let's move deterministic check to kernel level, which is related to kernel algorithm

xytintel (approved):

PR #740, opened 2024-08-09
Separate skip cases list by platform name

PenghuiCheng:
Separate skip cases list by platform name: "Arc", "MTL", "PVC".


daisyden (approved):
LGTM

mengfei25:
@daisyden  There is not client related regular tests now, you don't need change the workflow

PR #775, opened 2024-08-19
Add aten::_embedding_bag_dense_backward and aten::_embedding_bag_per_sample_weights_backward

hjhee:
- _embedding_bag_dense_backward
  - _embedding_bag_per_sample_weights_backward


src/ATen/native/xpu/sycl/EmbeddingBag.cpp
```diff
@@ -351,6 +360,217 @@ void embedding_bag_max_template(
...
+    int64_t weight_stride1,
+    const index_t* indices, // contiguous
+    const index_t* offset2bag, // contiguous
+    int64_t num_samples,
+    int64_t embedding_features,
+    scalar_t* output,
+    index_t padding_idx) {
+  using accscalar_t = at::acc_type<scalar_t, true>;
+
+  int64_t max_group_size = 64;
```
majing921201, line 546: Can we use general work group size here, instead of hardcode

xytintel (approved):

PR #786, opened 2024-08-20
[Windows] Fix for MSVC compilation errors

ratnampa:
The following changes will overcome compilation errors for windows build due to MSVC. It is in continuation of the work from this PR: https://github.com/intel/torch-xpu-ops/pull/680

1. Explicit static cast to double is required for cmath func isfinite in MSVC.
2. Creating Restrict macro for windows/linux compatibility.
3. Explicit list of value by reference for lambda to capture, implicit capture failing with MSVC might be scope issue.
4. Changing Math.h to MathExtensions.h as it is confusing the compiler with another Math.h.
5. Adding permissive flag for MSVC build to overcome ambiguous symbol error


min-jean-cho (approved):

fengyuan14:
Merged to https://github.com/intel/torch-xpu-ops/pull/838. Thanks.

PR #798, opened 2024-08-21
Fix pow operator complex32 UT failure on Windows MTL/LNL

Kanya-Mo:
Fix for test_binary_ufuncs_xpu.py::TestBinaryUfuncsXPU::test_pow_xpu_float16
The previous way of casting into opmath_t leads to accuracy failure for complex half dtype on windows. This change uses memory::LoadWithCast instead which fixes the problem.

### DIFF
--- a/src/ATen/native/xpu/sycl/PowKernels.cpp
+++ b/src/ATen/native/xpu/sycl/PowKernels.cpp
@@ -130,7 +130,8 @@ void pow_tensor_tensor_kernel(TensorIteratorBase& iter) {
       pow_chalf_tensor_scalar_impl(iter, exp);
     } else {
       TORCH_INTERNAL_ASSERT(!iter.is_cpu_scalar(1) && !iter.is_cpu_scalar(2));
-      auto f = PowTensorTensorCastFunctor<scalar_t>();
+      using opmath_t = at::opmath_type<scalar_t>;
+      auto f = PowTensorTensorFunctor<opmath_t>();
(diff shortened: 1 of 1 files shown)


xytintel:
This change will cause complex pow to take the dynamic cast path, which will hurt performance.

Kanya-Mo:
> This change will cause complex pow to take the dynamic cast path, which will hurt performance.
Thanks for pointing out. The failure of test_pow_xpu_float16 on complex input happens only on windows, so I updated and limited the change to windows only. Now this should not have performance regression at currently working cases as windows platforms (MTL, LNL and A770) are failing the case.

min-jean-cho (approved):

src/ATen/native/xpu/sycl/PowKernels.cpp
```diff
@@ -36,13 +36,23 @@ static inline c10::complex<T> pow_(c10::complex<T> base, c10::complex<T> exp) {
 
 } // namespace impl
 
+#ifdef _MSC_VER
```
fengyuan14, line 39: Please file an issue to record the ut failure https://github.com/intel/torch-xpu-ops/issues. And comment with the issue link and with some words to explain it is a Window specific issue.

Kanya-Mo: Opened https://github.com/intel/torch-xpu-ops/issues/842

PR #799, opened 2024-08-22
Improved some of the UT codes and cleaned the skip list.

PenghuiCheng:
Improved some of the UT codes and cleaned the skip list.


test/xpu/test_modules_xpu.py
```diff
@@ -90,7 +90,7 @@ def fn_to_gradcheck(*flat_input_and_params):
 @wrapSwapTensorsTest()
 def _test_to(self, device, dtype, module_info, training, swap, set_grad):
     module_cls = module_info.module_cls
-    devices = ['xpu']
+    devices = ['cpu', 'xpu']
```
fengyuan14, line 93: What's the meaning of adding `cpu` to `devices`?

test/xpu/nn/test_packed_sequence_xpu.py
```diff
@@ -15,12 +15,15 @@
...
+            *args, **kwargs
+        )
+        if ex.is_xpu:
             return self.to(*args, **kwargs)
-        return self.to(*args, device='xpu', **kwargs)
+        kwargs["device"] = "xpu"
+        return self.to(*args, **kwargs)
 
-    rnn_utils.PackedSequence.xpu = types.MethodType(rnn_utils.PackedSequence, myxpu)
+    rnn_utils.PackedSequence.xpu = myxpu
```
fengyuan14, line 26: What's the difference? Looks logics are same?

PenghuiCheng: types.MethodType is for object, but we want to change a class function member. so we use rnn_utils.PackedSequence.xpu = myxpu.

test/xpu/test_modules_xpu.py
PenghuiCheng: the case is to compare cpu and xpu op output, we aligned with cuda. If didn't add "cpu" , the case will error.

daisyden: It is ok to align with cuda but why the test will fail if no cpu in the devices list?

PenghuiCheng: because the order of testing will affect the results

PR #805, opened 2024-08-22
Adopt std::log & std::exp in softmax kernel to avoid unnecessary fp64 instructions

xytintel:
Using `std::exp` and `std::log` to replace `::exp` and `::log` respectively.


fengyuan14:
Please file a PR for release/2.5.

xytintel:
Move to https://github.com/intel/torch-xpu-ops/pull/812

PR #825, opened 2024-08-27
[DO NOT MERGE][Windows] Fix Split DLL Win with fix cyclic dep

ratnampa:
**Only use for windows build, linux part will be added later**


fengyuan14:
https://github.com/intel/torch-xpu-ops/pull/838

PR #828, opened 2024-08-27
[Windows] Add a skip list for ARC display card on windows platform.

cfgfung:
It is found that the SYCL compiler on Windows will remove extra FP64-related functions. This will trigger Windows-specific UT failures. This is a list of functions that will trigger unexpected failures. Adding this can avoid unnecessary false alarms at this moment. The compiler problem has been reported to related teams.


daisyden (approved):
LGTM

min-jean-cho (approved):

PR #831, opened 2024-08-27
Implementation of complex64 division without FP64 instructions

xytintel:
In compiler header files, FP64 calculations might be introduced.


xytintel:
image

PR #836, opened 2024-08-28
Resolve the FP64 issue in the gamma operators on ARC

xytintel:
Resolve the FP64 issue in the gamma operators on ARC.


xytintel:
image

test/xpu/skip_list_arc.py
```diff
@@ -8,4 +8,7 @@
     "test_tensor_creation_ops_xpu.py": (
         "test_float_to_int_conversion_finite_xpu_int64",
     ),
+    "test_ops_xpu.py": (
+        "test_compare_cpu_digamma_xpu_float32",
+    ),
```
fengyuan14, line 13: Do we need the skip?

xytintel: No, we don't need

fengyuan14 (approved):

PR #838, opened 2024-08-28
Windows: Enable building for latest torch-xpu-ops

fengyuan14:
In latest torch-xpu-ops, we have split libtorch_xpu.so into multiple libraries to avoid a binary with large size. The change of building system leads to two issues on Windows. The PR is to resolve them. 1) Cyclic dependence, 2) Symbols visibility.


src/ATen/native/xpu/sycl/AbsKernel.h
```diff
@@ -4,6 +4,6 @@
 
 namespace at::native::xpu {
 
-void abs_kernel(TensorIteratorBase& iter);
+TORCH_XPU_API void abs_kernel(TensorIteratorBase& iter);
```
EikanWang, line 7: We expose the symbol because of the current solution. In the future, it should be not necessary if we combine the libraries as a single one as long as device image compression is ready. So I suggest replacing `TORCH_XPU_API` with another name like `TORCH_XPU_API_MSVC`. In the future, we can easily remove `TORCH_XPU_API_MSVC` by searching and replacing.

src/BuildOnWindows.cmake
```diff
@@ -0,0 +1,101 @@
...
+# libtorch_xpu_ops.a dlopens libtorch_xpu_ops_aten.so (Break cycle)
+# libtorch_xpu_ops_aten.so links to libtorch_xpu_ops_sycl_unary_binary_kernels.so and libtorch_xpu_ops_sycl_kernels.so
+# libtorch_xpu_ops_sycl_unary_binary_kernels.so and libtorch_xpu_ops_sycl_kernels.so links to libtorch_xpu.so
+add_library(
+  torch_xpu_ops
+  STATIC
+  ${ATen_XPU_CPP_SRCS}
+  "bridge.cpp")
+
+target_compile_options(torch_xpu_ops PRIVATE -DPATH_TO_TORCH_XPU_OPS_ATEN_LIB=\"torch_xpu_ops_aten.dll\")
```
EikanWang, line 17: `torch_xpu_ops_aten` should be a cmake variable.

src/BuildOnWindows.cmake
```diff
@@ -0,0 +1,101 @@
...
+else()
+  # Split SYCL kernels into 2 libraries as categories 1) Unary+Binary 2) Others.
+  set(ATen_XPU_SYCL_UNARY_BINARY_SRCS)
+  set(ATen_XPU_SYCL_NON_UNARY_BINARY_SRCS)
+
+  foreach(sycl_src ${ATen_XPU_SYCL_SRCS})
+    string(REGEX MATCH "Binary" IS_BINARY ${sycl_src})
+    string(REGEX MATCH "Unary" IS_UNARY ${sycl_src})
+    string(REGEX MATCH "Pow" IS_POW ${sycl_src})
+    string(REGEX MATCH "Copy" IS_COPY ${sycl_src})
```
EikanWang, line 50: Please add informative comments here to elaborate on why we need to take `Pow` and `Copy` as a special case.

src/ATen/native/xpu/RangeFactories.cpp
```diff
@@ -29,7 +29,7 @@ Tensor& XPUNativeFunctions::arange_out(
 
         TORCH_CHECK(xstep > 0 || xstep < 0, "step must be nonzero");
         TORCH_CHECK(
-            std::isfinite(xstart) && std::isfinite(xend),
+            std::isfinite(static_cast<double>(xstart)) && std::isfinite(static_cast<double>(xend)),
```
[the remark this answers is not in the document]
fengyuan14: Verified. Not part of windows build. But alignment with intree's code piece.

src/ATen/native/xpu/sycl/AbsKernel.h
fengyuan14: I see your concerns. I suggest keeping it,
1. As for replacement in future, the changes all are included in `XXXKernel.h`. It is easy to find out them.
2. As for correct logic of these macros, basically TORCH_XPU_API is designed and includes the compatibility on Windows on Linux. Add a new one let people confused, like some people would say why could `TORCH_XPU_API` not do this, or is there new concept involved in the new macro.

src/BuildOnWindows.cmake
fengyuan14: Will fix

src/BuildOnWindows.cmake
fengyuan14: Will fix

EikanWang (approved):

ratnampa (approved):

EikanWang (approved):

PR #844, opened 2024-08-29
Arc: Add cases skip list for extended cases

PenghuiCheng:
Add cases skip list for ARC device in extended folder.


daisyden (approved):
LGTM

PR #847, opened 2024-08-29
Fix pow operator complex32 UT failure on Windows MTL/LNL

Kanya-Mo:
(Windows only) Fix for test_binary_ufuncs_xpu.py::TestBinaryUfuncsXPU::test_pow_xpu_float16 

The previous way of casting into opmath_t leads to accuracy failure for complex half dtype on windows. This change uses memory::LoadWithCast instead which fixes the problem.

(cherry picked from commit 5955b5cd4e8ee93a7607c1cd8cc0a9acda3c891b)

### DIFF
--- a/src/ATen/native/xpu/sycl/PowKernels.cpp
+++ b/src/ATen/native/xpu/sycl/PowKernels.cpp
@@ -37,6 +37,7 @@ static inline c10::complex<T> pow_(c10::complex<T> base, c10::complex<T> exp) {
 } // namespace impl
 
 #ifdef _MSC_VER
+// Divergence for MSVC due to accuracy issue. https://github.com/intel/torch-xpu-ops/issues/842.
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/PowKernels.cpp
```diff
@@ -36,13 +36,23 @@ static inline c10::complex<T> pow_(c10::complex<T> base, c10::complex<T> exp) {
 
 } // namespace impl
 
+#ifdef _MSC_VER
```
fengyuan14, line 39:
```suggestion
#ifdef _MSC_VER
// Divergence for MSVC due to accuracy issue. https://github.com/intel/torch-xpu-ops/issues/842.
```

fengyuan14 (approved):

PR #857, opened 2024-08-31
add rpath of sycl libraries location in python env to avoid manually activating sycl environment

jingxu10:
as title

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -83,6 +83,11 @@ foreach(lib ${TORCH_XPU_OPS_LIBRARIES})
   target_include_directories(${lib} PUBLIC ${SYCL_INCLUDE_DIR})
 
   target_link_libraries(${lib} PUBLIC ${SYCL_LIBRARY})
+  if(LINUX)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -83,6 +83,11 @@ foreach(lib ${TORCH_XPU_OPS_LIBRARIES})
   target_include_directories(${lib} PUBLIC ${SYCL_INCLUDE_DIR})
 
   target_link_libraries(${lib} PUBLIC ${SYCL_LIBRARY})
+  if(LINUX)
+    target_link_options(${lib} PRIVATE "-Wl,-rpath,$ORIGIN")
+    target_link_options(${lib} PRIVATE "-Wl,-rpath,$ORIGIN/../../../../")
+    target_link_options(${lib} PRIVATE "-Wl,--disable-new-dtags")
+  endif()
```
fengyuan14, line 90: Do you hope PyTorch XPU could work at runtime without sourcing oneAPI bundle?

jingxu10: yes, when pip packages are released.
The PR in PyTorch is https://github.com/pytorch/pytorch/pull/134912

PR #862, opened 2024-09-03
remove fp64 and complex128 test cases for ARC device in extended folder

PenghuiCheng:
remove fp64 and complex128 test cases for ARC device in extended folder


daisyden (approved):
LGTM

PR #866, opened 2024-09-04
Windows: Give default AOT targets to Windows build.

fengyuan14:
Default AOT target list on Windows prefers client GPUs, including ARC, LNL and MTLs.


min-jean-cho (approved):
May I know the final Windows binary size with the four AOT targets?

PR #867, opened 2024-09-04
Windows: Give default AOT targets to Windows build.

fengyuan14:
Default AOT target list on Windows prefers client GPUs, including ARC, LNL and MTLs.


min-jean-cho (approved):

EikanWang (approved):

cmake/BuildFlags.cmake
```diff
@@ -116,7 +116,11 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "MSVC"
   set(SYCL_OFFLINE_COMPILER_CG_OPTIONS "${SYCL_OFFLINE_COMPILER_CG_OPTIONS} -cl-fp32-correctly-rounded-divide-sqrt")
   set(SYCL_OFFLINE_COMPILER_CG_OPTIONS "-options '${SYCL_OFFLINE_COMPILER_CG_OPTIONS}'")
 
-  set(AOT_TARGETS "pvc,xe-lpg,ats-m150")
+  if(WIN32)
+    set(AOT_TARGETS "ats-m150,lnl-m,mtl-u,mtl-h")
+  else()
+    set(AOT_TARGETS "pvc,xe-lpg,ats-m150")
```
EikanWang, line 122: Since it kind of depends on the driver implementation, should we replace xe-lpg with mlt-u and mtl-h? It is not critical for PT 2.5

PR #871, opened 2024-09-05
Add aten::tril_indices, aten::triu_indices

hjhee:
- tril_indices
  - tril_indices.out
  - triu_indices
  - triu_indices.out


yaml/xpu_functions.yaml
```diff
@@ -722,3 +722,7 @@ supported:
   - _weight_norm_interface
   - _weight_norm_interface_backward
   - range.out
+  - tril_indices
+  - tril_indices.out
```
xytintel, line 726: We don't need to register `tril_indices.out`

yaml/xpu_functions.yaml
```diff
@@ -722,3 +722,7 @@ supported:
...
+  - triu_indices
+  - triu_indices.out
```
xytintel, line 728: We don't need to register `triu_indices.out`

PR #872, opened 2024-09-05
fix UT test_unary_op_out_casting_xpu_int64_float32

huaiyuzh:
fix UT test_type_promotion.py::TestTypePromotionXPU::test_unary_op_out_casting_xpu_int64_float32.


src/ATen/native/xpu/UnaryOps.cpp
```diff
@@ -1028,11 +1028,12 @@ Tensor& XPUNativeFunctions::ceil_(Tensor& self) {
 }
 
 Tensor& XPUNativeFunctions::ceil_out(const Tensor& self, Tensor& out) {
+  auto iter = ceil_meta(self, out);
```
guangyey, line 1031: I am very curious about what happens if we don't reorder these two lines of code. And why this reorder can affect the result?

huaiyuzh: The test case expects to catch a runtime error such as "RuntimeError: Found dtype Float but expected Long", which is raised in ceil_meta. According to the original code logic, if the input type is int, ceil_meta will not be called.

guangyey: Make sense. Thanks very much~

guangyey (approved):

fengyuan14 (approved):

PR #883, opened 2024-09-09
Add one more AT_DISPATCH for layer norm gamma scalar type

Liangliang-Ma:
Original kernel only do AT_DISPATCH on input dtype, while weight/bias (gamma/beta) may have different data types. It will break when trying to get `gamma.data_ptr<weight_t>()`, for its wrong scalar type.

Same issue break comfirmed on cuda.

Fixing this by adding one more AT_DISPATCH.

### DIFF
--- a/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
@@ -593,6 +593,29 @@ void _layer_norm_backward_kernel(
       dY, X, mean_data, var_data, dgamma, dbeta, config_w);
 }
 
+template <typename x_t>
+void bridge_layer_norm_kernel(
+    const Tensor& X,
+    const Tensor& gamma,
+    const Tensor& beta,
+    int64_t M,
+    int64_t N,
+    double eps,
+    Tensor& Y,
+    Tensor& mean,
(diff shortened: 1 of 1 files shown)


xytintel:
Checked the logic of CUDA. They use `is_mixed_type` to differentiate between paths. Of course, they only assume that `gamma` is float.

xytintel:
@Liangliang-Ma What scenarios need this fix? Since CUDA doesn’t specifically address this issue, is it necessary? @fengyuan14 @EikanWang

Liangliang-Ma:
> @Liangliang-Ma What scenarios need this fix? Since CUDA doesn’t specifically address this issue, is it necessary? @fengyuan14 @EikanWang

Met this in a ipex rebase 2.5 workload jira, which is a bart model doing question answering task on float16 precision(no AMP). It turned out to be easy to reproduce in single UT, also failing with lastest stock pytorch whl on CUDA.

```
import torch
from torch.nn import LayerNorm

dtype1 = torch.float16
inputs = torch.rand([1, 1024, 4096], dtype=dtype1, requires_grad=True).cuda()
ln = LayerNorm(4096, eps=1e-05, elementwise_affine=True).cuda()
outputs = ln(inputs)
```

PR #886, opened 2024-09-09
Windows: Build: Split kernel lib into smaller libs

fengyuan14:
Fixing loading kernel dll error.

### DIFF
--- a/src/BuildOnWindows.cmake
+++ b/src/BuildOnWindows.cmake
@@ -40,8 +40,12 @@ if(BUILD_SEPARATE_OPS)
   endforeach()
 else()
   # Split SYCL kernels into 2 libraries as categories 1) Unary+Binary 2) Others.
-  set(ATen_XPU_SYCL_UNARY_BINARY_SRCS)
-  set(ATen_XPU_SYCL_NON_UNARY_BINARY_SRCS)
+  set(ATen_XPU_SYCL_BINARY_SRCS)
+  set(ATen_XPU_SYCL_UNARY_SRCS)
+  set(ATen_XPU_SYCL_REDUCE_SRCS)
+  set(ATen_XPU_SYCL_ACTIVATION_SRCS)
+  set(ATen_XPU_SYCL_FOREACH_SRCS)
+  set(ATen_XPU_SYCL_OTHERS_SRCS)
 
   foreach(sycl_src ${ATen_XPU_SYCL_SRCS})
     string(REGEX MATCH "Binary" IS_BINARY ${sycl_src})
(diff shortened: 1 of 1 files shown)


src/BuildOnWindows.cmake
```diff
@@ -40,8 +40,12 @@ if(BUILD_SEPARATE_OPS)
...
 else()
   # Split SYCL kernels into 2 libraries as categories 1) Unary+Binary 2) Others.
-  set(ATen_XPU_SYCL_UNARY_BINARY_SRCS)
-  set(ATen_XPU_SYCL_NON_UNARY_BINARY_SRCS)
+  set(ATen_XPU_SYCL_BINARY_SRCS)
+  set(ATen_XPU_SYCL_UNARY_SRCS)
+  set(ATen_XPU_SYCL_REDUCE_SRCS)
+  set(ATen_XPU_SYCL_ACTIVATION_SRCS)
+  set(ATen_XPU_SYCL_FOREACH_SRCS)
+  set(ATen_XPU_SYCL_OTHERS_SRCS)
```
airMeng, line 48: @JianpingChen066 per our experience, separate kernels libraries according to AOT arch will be better than operators, for the binary size

fengyuan14: If splitting bin according to AOT arch, do you have any change on host? I cannot image how to call MTL add_kernel without changes on host.

airMeng: ok for stock pytorch, we don't have a device management so no dispatch available

src/BuildOnWindows.cmake
```diff
@@ -53,31 +57,101 @@ else()
...
+      list(APPEND ATen_XPU_SYCL_ACTIVATION_SRCS ${sycl_src})
     else()
-      list(APPEND ATen_XPU_SYCL_UNARY_BINARY_SRCS ${sycl_src})
+      list(APPEND ATen_XPU_SYCL_OTHERS_SRCS ${sycl_src})
     endif()
   endforeach()
 
-  set(sycl_unary_binary_lib torch_xpu_ops_sycl_unary_binary_kernels)
+  # Binary kernel lib
+  set(sycl_binary_lib torch_xpu_ops_sycl_binary_kernels)
```
airMeng, line 80: You can try the interface library to make here explicit
https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries

JianpingChen066:
for xetla kernels, generally we have a function for arch dispatch according to the current arch type. ideally, it should make kernels that for the same compiler options to be together to reduce the code size. 
Not sure about the sycl/onednn side, how they handling the AOT build and manage different executable for different arch ?
and what about this PR purpose, is that for code reduce or something else ?

src/BuildOnWindows.cmake
fengyuan14: Yes. But I am not sure to uploading device management from runtime to fwk is reasonable so far. We expose the complexity.

src/BuildOnWindows.cmake
fengyuan14: Could you elaborate the suggestion? Interface library won't involve building, but a cmake target to hold some properties?

fengyuan14:
> for xetla kernels, generally we have a function for arch dispatch according to the current arch type. ideally, it should make kernels that for the same compiler options to be together to reduce the code size. Not sure about the sycl/onednn side, how they handling the AOT build and manage different executable for different arch ? and what about this PR purpose, is that for code reduce or something else ?

The PR is to reduce bin size. But according to your comments, I think it is not proper to introduce arch management (arch dispatch) to PyTorch so far. PyTorch so far only covers arch agnostic kernels, and leave complexity of arch dispatch to runtime/driver.

src/BuildOnWindows.cmake
airMeng: you can set an interface library like ```ATen_XPU_SYCL_SRCS``` and link all separated libraries into the interface library. then you only need to link the interface library to ```torch_xpu_ops_aten```
next time if you have more new libraries, you only need to link the new library to the interface, don't need to change the interface linking.

airMeng: here xetla_kernels is an interface library https://github.com/intel/intel-extension-for-pytorch/blob/8d4670e4c131a476c6aabb686f33688ca9949243/csrc/gpu/aten/operators/xetla/kernels/CMakeLists.txt#L117
where it is used https://github.com/intel/intel-extension-for-pytorch/blob/8d4670e4c131a476c6aabb686f33688ca9949243/csrc/gpu/CMakeLists.txt#L249

fengyuan14: Look good to me to make the logic simple. BTW, I need additional logic like not only,
`target_link_library(torch_xpu_ops_aten, torch_xpu_ops_kernel_itf)`, But also,
`target_link_library(torch_xpu_ops_kernel1, torch_xpu)`
`target_link_library(torch_xpu_ops_kernel2, torch_xpu)`
`target_link_library(torch_xpu_ops_kernel3, torch_xpu)`
So for part 1, I can use interface library, for part 2, I have to config link library separately.

airMeng: interface support bidirectional linking
https://github.com/intel/intel-extension-for-pytorch/blob/8d4670e4c131a476c6aabb686f33688ca9949243/csrc/gpu/CMakeLists.txt#L118

fengyuan14: LGTM. But so far the current impl looks good to me, as only the code piece uses these targets. If these targets need to be used in many places, interface target will be the choice. Thanks.

EikanWang (approved):

EikanWang:
Thanks @airMeng and @JianpingChen066 . In general, we serve PyTorch 2.5 release now. The validation team is testing the codebase. We will not change the code according to your comments. In addition, it is a temporary solution because of the missing of kernel compression. For PT 2.6, these codes will be removed and only a single binary.

PR #915, opened 2024-09-13
unittest: update decomp test 

yuchengliu1:
Re-test test_decomp, and it has some new failures. 
This PR fixed some new failures by hooking test function.
The rest failures has not fixed it. The cases with unsupported OPs are commented at #584. The case with supported OPs are in #918


test/xpu/test_decomp_xpu.py
```diff
@@ -2,7 +2,7 @@
 
 import torch
 from torch.testing._internal.common_device_type import instantiate_device_type_tests
-from torch.testing._internal.common_utils import run_tests
+from torch.testing._internal.common_utils import run_tests, skipIfCrossRef
```
fengyuan14, line 5: Could you comment the usage of the decorator in PR description?

yuchengliu1: ok

PR #933, opened 2024-09-25
batch_normalization: Introduce vectorization optimization in the batch norm elementwise kernel.

xytintel:
Due to performance issues with the low-precision data type implementation of group stride loops on PVC ([jira: PYTORCHDGQ-5162](https://jira.devtools.intel.com/browse/PYTORCHDGQ-5162)), partial vectorization optimization is used.


src/ATen/native/xpu/sycl/BatchNormKernels.cpp
```diff
@@ -1263,6 +1263,147 @@ struct BatchNormTransformInputKernelFunctor {
...
+                gamma * (i[feature] - mean) * invstd + beta);
+          }
+        } else {
+          using vec_t = memory::aligned_vector<input_scalar_t, VEC_SIZE>;
+          vec_t vec;
+#pragma unroll
+          for (int vt = 0; vt < VEC_SIZE; ++vt) {
+            index_t feature = feature_vec_begin + vt;
+            vec[vt] = static_cast<input_scalar_t>(
+                gamma * (i[feature] - mean) * invstd + beta);
```
fengyuan14, line 1323: Why not using vectorization for inputs? I see `vt` is contiguous across the iteration.

src/ATen/native/xpu/sycl/BatchNormKernels.cpp
```diff
@@ -1319,14 +1460,27 @@ void batch_norm_elemt_template(
...
+      memory::can_vectorize_up_to<input_scalar_t>(output_ptr) >= 4 &&
+      sizeof(input_scalar_t) < sizeof(float)) {
+    auto kfn = BatchNormTransformInputVectorizedKernelFunctor<
+        4,
+        input_scalar_t,
+        stat_scalar_t,
+        stat_accscalar_t,
+        true,
+        index_t>(input, output, mean, invstd, weight, bias, dummy_epsilon);
+    sycl_kernel_submit(global_range, local_range, queue, kfn);
```
fengyuan14, line 1474: Why we still keep the `global_range` unchanged, even the workloads on each work items increase?

src/ATen/native/xpu/sycl/BatchNormKernels.cpp
fengyuan14: If we know the `vt` size, we can calculate offset with linear id 0, 3, 7, 11? if the `vt` size is 4?

EikanWang:
@xytintel , may I know if this PR is still available?

xytintel:
> @xytintel , may I know if this PR is still available?

Yes, In the process of merging it.

xytintel:
Group stride kernel: 12.493ms
Vectorized kernel: 10.024ms
```python
import torch
a = torch.rand(2, 256, 128, 128).xpu().bfloat16()
bn = torch.nn.BatchNorm2d(256).xpu().bfloat16()
    activities=[
)
with prof_xpu:
    for i in range(120):
        output = bn(a)
print(prof_xpu.key_averages().table(sort_by="self_xpu_time_total", row_limit=100000))
print(output.dtype)
```

src/ATen/native/xpu/sycl/BatchNormKernels.cpp
xytintel: Apply a coalescing manner for the input access would be quite complex, let we discuss it later.

EikanWang (approved):

PR #953, opened 2024-10-08
Add aten::soft_margin_loss and its variants

yucai-intel:
soft_margin_loss
soft_margin_loss_backward

### DIFF
--- a/src/ATen/native/xpu/Loss.cpp
+++ b/src/ATen/native/xpu/Loss.cpp
@@ -3,12 +3,15 @@
 #include <ATen/native/BinaryOps.h>
 #include <ATen/native/DispatchStub.h>
 #include <ATen/native/PointwiseOps.h>
-#include <comm/xpu_aten.h>
-
 #include <ATen/native/xpu/sycl/BinaryMiscOpsKernels.h>
 #include <ATen/native/xpu/sycl/LossKernels.h>
 #include <ATen/native/xpu/sycl/PointwiseOpsKernels.h>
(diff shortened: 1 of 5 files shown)


src/ATen/native/xpu/Loss.cpp
```diff
@@ -73,5 +76,85 @@ Tensor& binary_cross_entropy_backward_out_xpu(
...
+    int64_t reduction) {
+  if (reduction == at::Reduction::Mean) {
+    return unreduced.mean();
+  } else if (reduction == at::Reduction::Sum) {
+    return unreduced.sum();
+  }
+  return unreduced;
+}
+
+Tensor soft_margin_loss_xpu(
```
xytintel, line 90: I don't see any reference to `soft_margin_loss_xpu`. Do we still need to register it?

src/ATen/native/xpu/Loss.cpp
```diff
@@ -73,5 +76,85 @@ Tensor& binary_cross_entropy_backward_out_xpu(
...
+                  .add_output(grad_input)
+                  .add_input(self)
+                  .add_input(target)
+                  .add_input(grad_output)
+                  .build();
+  xpu::soft_margin_backward_kernel(iter, norm);
+  return grad_input;
+}
+
+Tensor soft_margin_loss_backward_xpu(
```
xytintel, line 148: ditto.

xytintel:
Close this PR because CUDA is not registered separately; it is a composite op.

Issue #426, opened 2024-06-18
[DONT CLOSE] On-demand test result Tracker


chuanqi129:
This issue just for track on-demand test result and inform the result to scheduler.

PR #958, opened 2024-10-10
[CI] Fix no output of KERNEL_VERSION for on demand test

Stonepia:
Fixes the On-demand Test has no Kernel version output.


Stonepia:
This is to fix that the on demand test, the column Kernel is empty:
image

Stonepia:
Fixed result in https://github.com/intel/torch-xpu-ops/issues/426#issuecomment-2404570285 .

chuanqi129 (approved):

PR #966, opened 2024-10-15
Fix foreach binary op scalarlist macro

min-jean-cho:
Probably was a typo. Fix is needed prior to adding any `FOREACH_BINARY_OP_SCALARLIST` ops. Otherwise, will raise an undefined symbol error like below:
 
```
   from torch._C import *  # noqa: F403
ImportError: libtorch_xpu.so: undefined symbol: _ZN2at6native46foreach_tensor_clamp_max_scalarlist_kernel_xpuEN3c108ArrayRefINS_6TensorEEENS2_INS1_6ScalarEEE
```


min-jean-cho:
Closing as merged in https://github.com/intel/torch-xpu-ops/pull/974.

PR #968, opened 2024-10-16
Add 4 Bessel Special Operators

Kanya-Mo:
- [x] special_bessel_j0.out
- [x] special_bessel_j0
- [x] special_bessel_j1.out
- [x] special_bessel_j1
- [x] special_bessel_y0.out
- [x] special_bessel_y0
- [x] special_bessel_y1.out
- [x] special_bessel_y1


test/xpu/xpu_test_utils.py
```diff
@@ -252,6 +252,10 @@
     "square",
     "heaviside",
     "argsort",
+    "special_bessel_j0",
+    "special_bessel_j1",
```
majing921201, line 256: Those operators are not in common_methods_invocations, don't need to add here. please remove.

xytintel:
`aten/src/ATen/native/Math.h` can be reused

src/ATen/native/xpu/sycl/MathExtensions.h
```diff
@@ -361,4 +361,240 @@ static inline C10_HOST_DEVICE scalar_t calc_i1e(scalar_t _x) {
   return (_x < scalar_t{0.0}) ? -out : out;
 }
 
+template <typename scalar_t>
+static inline C10_HOST_DEVICE scalar_t bessel_j1_forward(scalar_t x) {
```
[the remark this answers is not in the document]
majing921201: > Why no `bessel_j0_forward` and `bessel_y0_forward` here?

They are in `torch/testing/_internal/opinfo/definitions/special.py`, Our datatype extended test only convers ops in `common_methods_invocations.py`
I think PyTorch test will cover

Kanya-Mo:
> `aten/src/ATen/native/Math.h` can be reused for `bessel_j1_forward` and `bessel_y1_forward`.

Hi @xytintel, the implementation of `bessel_j1_forward` and `bessel_y1_forward` in aten/src/ATen/native/Math.h include recursive call for some input which is not supported by sycl (compilation will fail with a message 'SYCL kernel cannot call a recursive function"), so I also upstreamed their ipex implementation. On the other hand, torch's `bessel_j0_forward` and `bessel_y0_forward` are fine and that's why I didn't include them in MathExtension.h.

test/xpu/xpu_test_utils.py
Kanya-Mo: Thanks for pointing out. Will remove.

xytintel (approved):

PR #770, opened 2024-08-16
Add aten::mode

hjhee:
- mode
- mode.values

PR #984, opened 2024-10-18
Modernize C10_UNUSED

xytintel:
[[[maybe_unused]]](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) is part of C++17 standard.
Depends on: https://github.com/pytorch/pytorch/pull/138102


xytintel:
move to https://github.com/intel/torch-xpu-ops/pull/770

PR #985, opened 2024-10-18
Add aten::_foreach_copy_

cfgfung:
Add the implementation opeartor _foreach_copy_
Tested on PVC.


src/ATen/native/xpu/ForeachOpList.cpp
```diff
@@ -145,5 +147,24 @@ void foreach_tensor_lerp_ternary_xpu_(
   }
 }
 
+void foreach_tensor_copy_list_kernel_xpu_(
+    at::TensorList self,
+    at::TensorList src,
+    bool non_blocking) {
+      at::native::check_foreach_api_restrictions(self, src);
+      if (!at::native::can_use_fast_route(
```
majing921201, line 155: no need at::native, here is the same namespace

majing921201, line 155: The check condition seems not aligned CUDA, can you help check ?
https://github.com/pytorch/pytorch/blob/d8f22a114142f40f78e1145e45d7eaf823b516cb/aten/src/ATen/native/cuda/ForeachBinaryOpList.cu#L418

cfgfung: The extra checks by CUDA is caused by this feature enhancement - https://github.com/pytorch/pytorch/issues/115171
They wish to improve the GPU bandwidth utilization when src and dst are of different types.

Shall we stick to the current design as the copy kernel does not support that functionality?

cfgfung: Removed that from the code.

majing921201: > The extra checks by CUDA is caused by this feature enhancement - [pytorch/pytorch#115171](https://github.com/pytorch/pytorch/issues/115171) They wish to improve the GPU bandwidth utilization when src and dst are of different types.
> 
> Shall we stick to the current design as the copy kernel does not support that functionality?

if only for performance improvement, We could hold on the change.

majing921201 (approved):

PR #990, opened 2024-10-21
Add basic sparse operators

xiaowangintel:
Supports same sparse ops which are needed for dlrm


src/ATen/native/sparse/xpu/sycl/SparseTensorMathKernel.cpp
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  void operator()(T* out, T* in1, T* in2) const {
+    *out = *in1 + val * *in2;
+  }
+
+  T val;
+};
+
+TORCH_XPU_API void add_out_dense_sparse_kernel(
```
fengyuan14, line 172: Move the decorator to the related header.

xiaowangintel: done

xytintel:
Move to https://github.com/intel/torch-xpu-ops/pull/1030

Issue #1027, opened 2024-10-29
Fail on 2025 compiler


ZzEeKkAa:
### 🐛 Describe the bug

While building pytorch using DLE with 2025.0.0 compiler:

```
cd /home/jovyan/Projects/pytorch/build/caffe2/aten_xpu/src/CMakeFiles/torch_xpu_ops_sycl_kernels.dir/ATen/native/xpu/sycl && /home/jovyan/Projects/triton/.venv/lib/python3.9/site-packages/cmake/data/bin/cmake -E make_directory /home/jovyan/Projects/pytorch/build/caffe2/aten_xpu/src/CMakeFiles/torch_xpu_ops_sycl_kernels.dir/ATen/native/xpu/sycl/. && /home/jovyan/Projects/triton/.venv/lib/python3.9/site-packages/cmake/data/bin/cmake -D verbose:BOOL=OFF -D generated_file:STRING=/home/jovyan/Projects/pytorch/build/caffe2/aten_xpu/src/CMakeFiles/torch_xpu_ops_sycl_kernels.dir/ATen/native/xpu/sycl/./torch_xpu_ops_sycl_kernels_generated_TensorModeKernel.cpp.o -P /home/jovyan/Projects/pytorch/build/caffe2/aten_xpu/src/CMakeFiles/torch_xpu_ops_sycl_kernels.dir/ATen/native/xpu/sycl/torch_xpu_ops_sycl_kernels_generated_TensorModeKernel.cpp.o.Release.cmake
FAILED: caffe2/aten_xpu/src/CMakeFiles/torch_xpu_ops_sycl_kernels.dir/ATen/native/xpu/sycl/torch_xpu_ops_sycl_kernels_generated_TensorModeKernel.cpp.o /home/jovyan/Projects/pytorch/build/caffe2/aten_xpu/src/CMakeFiles/torch_xpu_ops_sycl_kernels.dir/ATen/native/xpu/sycl/torch_xpu_ops_sycl_kernels_generated_TensorModeKernel.cpp.o
[...]

PR #1017, opened 2024-10-24
SYCL: Fixing breaking issue of sort related kernel level API due to Intel SYCL compiler uplift

chunhuanMeng:

PR #1032, opened 2024-10-30
Update default_sorter to work with 2025 compiler

ZzEeKkAa:
Fixes: #1027

### DIFF
--- a/src/ATen/native/xpu/sycl/TensorModeKernel.cpp
+++ b/src/ATen/native/xpu/sycl/TensorModeKernel.cpp
@@ -792,11 +792,21 @@ void mode_kernel_impl(
     auto group_size = problem_size;
 
     // scratch memory size needed by built-in sort
+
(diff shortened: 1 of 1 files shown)


ZzEeKkAa:
Part of the code is compiled with clang++, so `__INTEL_LLVM_COMPILER` can't be used there, so I don't know how to address it

fengyuan14:
Hi, @ZzEeKkAa. We have got a pr about it https://github.com/intel/torch-xpu-ops/pull/1017. We build torch-xpu-ops with separate compilation solution, compile host with gcc and compile device code with clang++. And the macro is not a source-code-defined macro in SYCL headers but a clang++ builtin. Please refer the pr pasted, we have a pre-process in cmake process to expose the flag.

ZzEeKkAa:
PR #1017 does exactly what we need. Thank you!

Issue #1022, opened 2024-10-28
sort bool type has issue


guizili0:
### 🐛 Describe the bug

```
import torch
import random

torch.manual_seed(0)
random.seed(0)

xpu_device = torch.device("xpu")
cpu_device = torch.device("cpu")

#tensor_dtype = torch.int64
tensor_dtype = torch.bool
value_range = 2
a = torch.randint(value_range, (4099,)).to(dtype=tensor_dtype).to(xpu_device)

print(a)
print(a.shape)

for dim in reversed(range(a.dim())):
    #output, output_indices = torch.mode(a, dim)
    #print(output)
    sorted, indices = torch.sort(a)
    print(sorted)
    sorted, indices = a.sort()
    print(sorted)
    sorted, indices = a.sort(stable=True)
    print(sorted)
    sorted_cpu, indices = torch.sort(a.to(cpu_device))
    res = torch.equal(sorted.to(cpu_device), sorted_cpu)
    print(res)

```

### Versions

sort result is wrong:

tensor([False,  True,  True,  ..., False,  True,  True], device='xpu:0')
torch.Size([4099])
tensor([False,  True,  True,  ..., False, False, False], device='xpu:0')
tensor([False,  True,  True,  ..., False, False, False], device='xpu:0')
tensor([False,  True,  True,  ..., False, False, False], device='xpu:0')
False

PR #1037, opened 2024-10-31
Complete the logic for the AB buffer in segmented radix sort.

xytintel:
Resolve https://github.com/intel/torch-xpu-ops/issues/1022

### DIFF
--- a/src/ATen/native/xpu/sycl/SortingKernels.h
+++ b/src/ATen/native/xpu/sycl/SortingKernels.h
@@ -317,7 +317,7 @@ void segmented_radix_sort_pairs_downsweep_kernel(
 // ======================= large sort =======================
 
 template <typename scalar_t>
-struct InternalCopyFunctor {
+struct ABBufferCopyFunctor {
   scalar_t operator()(scalar_t x) const {
     return x;
   }
@@ -364,8 +364,6 @@ void segmented_radix_sort_pairs_kernel(
   value_t* values_in_ = const_cast<value_t*>(values_in);
   value_t* values_out_ = values_temp;
 
-  bool result_in_temp = false;
-
   while (true) {
     segmented_radix_sort_pairs_upsweep_kernel<
         key_t,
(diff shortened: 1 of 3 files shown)


src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -397,10 +407,32 @@ void segmented_radix_sort_pairs_kernel(
       std::swap(keys_in_, keys_out_);
       std::swap(values_in_, values_out_);
     }
+    result_in_temp = !result_in_temp;
```
fengyuan14, line 410: There is no other logic except for `result_in_temp = false` before it? Using `!result_in_temp` makes confusion.
```
result_in_temp = true;
```

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -315,6 +316,13 @@ void segmented_radix_sort_pairs_downsweep_kernel(
 
 // ======================= large sort =======================
 
+template <typename scalar_t>
+struct InternalCopyFunctor {
```
fengyuan14, line 320: What's the `Internal` mean? Internal for sort? Or naming with `ABBufferCopyFunctor`.

src/ATen/native/xpu/sycl/SortingKernels.h
xytintel: This is the basic operation of the AB Buffer, where the inverse is taken at each step.

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -397,10 +407,32 @@ void segmented_radix_sort_pairs_kernel(
...
       break;
   }
+
+  if (result_in_temp) {
+    auto input_calc = TrivialOffsetCalculator<2>();
+    at::detail::Array<char*, 2> data;
+    if (keys_out) {
+      data[0] = (char*)keys_out;
+      data[1] = (char*)keys_temp;
+      auto fn = InternalCopyFunctor<key_t>();
```
fengyuan14, line 422: `key_t == char` ?

fengyuan14, line 422: I wonder if we can reduce kernel size, when we know only `bool` needs the kernel?

fengyuan14: Or we will have kernels for all dtpye instantiated by `key_t`, part of which are not used actually?

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -397,10 +407,32 @@ void segmented_radix_sort_pairs_kernel(
...
+      auto vec_size = memory::can_vectorize_up_to<decltype(fn)>(data);
+      launch_vectorized_kernel(
+          num_segments * num_elements, fn, data, input_calc, vec_size);
+    }
+    if (values_out) {
+      data[0] = (char*)values_out;
+      data[1] = (char*)values_temp;
+      auto fn = InternalCopyFunctor<value_t>();
```
fengyuan14, line 430: Ditto.

fengyuan14 (approved):

PR #1044, opened 2024-11-04
Calling pytest from python code since skip list too long

PenghuiCheng:
This PR fixed two issues:
1. The skip list is too long to run pytest in shell in windows OS. So Calling pytest from python code can fix this issue.
2. Skipped some new cases which not currently implemented for the XPU device.


daisyden:
Please also apply the long command fix to the run_test_xxx.py files under test/xpu, no only to extended UT.

daisyden:
Why we need to skip test_non_standard_bool_values_index_put_xpu_bool ?

PenghuiCheng:
> Why we need to skip test_non_standard_bool_values_index_put_xpu_bool ?

It is failed in LNL machine. I will split it to another PR.

PR #1045, opened 2024-11-04
Add basic nested tensor operators

min-jean-cho:
Add basic nested tensor operators.


min-jean-cho:
Verified locally that the following two PyTorch UTs for `_nested_from_padded` PASS. 
`test_nested_tensor_from_padded`: https://github.com/pytorch/pytorch/blob/main/test/test_nestedtensor.py#L2961-L2975
`test_nested_tensor_from_padded_fused`: https://github.com/pytorch/pytorch/blob/main/test/test_nestedtensor.py#L2977-L2991

Hi @daisyden, could you please help port these UTs to torch-xpu-ops? Thanks!

src/ATen/native/xpu/sycl/NestedTensorTransformerFunctionsKernels.h
```diff
@@ -0,0 +1,253 @@
...
+      batch_size);
+  int64_t max_wg_size = syclMaxWorkGroupSize(kfn);
+  sycl::range<2> global_range{(size_t)batch_size * max_wg_size, GRID_DIM_Y};
+  sycl::range<2> local_range{(size_t)max_wg_size, 1};
+
+  sycl_kernel_submit(global_range, local_range, queue, kfn);
+}
+
+template <typename T>
+void remove_padding_kernelLauncher(
```
fengyuan14, line 211: Please fix the mixing naming. In PyTorch, usually, we have `launch_xxx_kernel`.

src/ATen/native/xpu/NestedTensorTransformerFunctions.cpp
```diff
@@ -0,0 +1,109 @@
...
+            offsets_ptr,
+            input_size_ptr,
+            output_size_ptr,
+            padded_contiguous.dim() - 1,
+            padded_contiguous.sizes()[0]);
+      }
+    } else {
+      AT_ERROR("Only support fp32/fp16 for padded input");
+    }
+    return at::detail::make_tensor<at::native::NestedTensorImpl>(
```
fengyuan14, line 102: I remember we have not upstreamed `DispatchKey::NestedTensorXPU`. It should not work.

fengyuan14:
I think we can leave Nested Tensor related operators for now,
1. I asked workload team people, there is no usage of Nested Tensor operators in real popular workloads.
2. We need a full plan to support Nested Tensor, like we need upstream the dispatch key first.
3. Why does IPEX have the operators. It's an internal requirement to align op coverage with HPU.
According to the time line, it is a bit tight for us to catch up 2.6, if we need upstream the dispatch key. We can treat it as an entire feature to support Nested Tensor in 2.7.

src/ATen/native/xpu/sycl/NestedTensorTransformerFunctionsKernels.h
min-jean-cho: Thanks, updated.

src/ATen/native/xpu/NestedTensorTransformerFunctions.cpp
min-jean-cho: @fengyuan14 , I created https://github.com/intel/torch-xpu-ops/issues/1141 to keep track of PRs related to `NestedTensor` support for xpu.

min-jean-cho:
Depends on https://github.com/intel/torch-xpu-ops/pull/1140, https://github.com/intel/torch-xpu-ops/pull/1142.

xytintel (approved):

PR #1047, opened 2024-11-05
Add aten::record_stream

Kanya-Mo:
- [x] record_stream


Kanya-Mo:
This op has no unit tests for xpu in stock pytorch at the moment, essentially due to lack of xpu counterpart of torch.cuda._sleep(). In CUDA unit tests of this op, torch.cuda._sleep() is used to delay one stream suffciently long. So, during local testing I replace it with an expensive kernel and translate two tests from [test_cuda.py](https://github.com/pytorch/pytorch/blob/a7479fa2828ad55a056a60a629dff6b7a0cb6b98/test/test_cuda.py#L703) to effectively test this op (and it passed).  

I will work on finding a good way to add unit tests for this op.

zhangxiaoli73:
Please be aware - this OP is widely used in distributed FSDP.

guangyey (approved):

PR #1057, opened 2024-11-06
[PT2.6][Windows] Fix Windows Codegen

ratnampa:
Running codegen on windows is running into below error recently,

`FileNotFoundError: [Errno 2] No such file or directory: 'C:/Users/sdp/ratnam-work/pytorch/third_party/torch-xpu-ops/yaml/templates\\RegisterDispatchDefinitions.ini' `

This is caused due to this symlink command in windows not running a shell or cmd as an administrator https://github.com/intel/torch-xpu-ops/blob/b1582e14f9e27bcbc0666ff636389d2be61783e4/cmake/Codegen.cmake#L61

Also, when symlink is not performed due to access issues, cmake throws such error: 
`You do not have sufficient privilege to perform this operation.`

**Resolution**
Change symlink to copy templates folder and delete it once codegen is complete.


cmake/Codegen.cmake
```diff
@@ -56,9 +56,9 @@ function(GEN_XPU file_yaml)
 
   # Codegen prepare process
   if(WIN32)
-    string(REPLACE "/" "\\" LinkPATH "${CODEGEN_TEMPLATE}templates")
-    string(REPLACE "/" "\\" TargetPATH "${CMAKE_SOURCE_DIR}/aten/src/ATen/templates")
-    execute_process(COMMAND cmd /c mklink /D ${LinkPATH} ${TargetPATH})
+    string(REPLACE "/" "\\" DestPATH "${CODEGEN_TEMPLATE}templates")
+    string(REPLACE "/" "\\" SrcPATH "${CMAKE_SOURCE_DIR}/aten/src/ATen/templates")
+    execute_process(COMMAND cmd /c xcopy ${SrcPATH} ${DestPATH} /E /H /C /I /Y > nul)
```
EikanWang, line 61: @ratnampa , do you know why `mklink` did not impact the Windows build in the past?

EikanWang (commented):
Pls. check the build failures.

cmake/Codegen.cmake
ZhiweiYan-96: > If you're not in the Administrators group, you can still create a symbolic link if: UAC is on and You have the "Create Symbolic Links" privilege

Maybe an explanation for this. Seems the symbolic links right can be set independently.

ratnampa: @ZhiweiYan-96 when there is an admin privilege issue to create symlink, there are two errors, first the obvious visible one is `FileNotFoundError: [Errno 2] No such file or directory: 'C:/Users/sdp/ratnam-work/pytorch/third_party/torch-xpu-ops/yaml/templates\\RegisterDispatchDefinitions.ini' ` and the other one is the message saying `You do not have sufficient privilege to perform this operation.` in the cmake configuration info outputted at the start of the build, which is very hard to find for any user. Then they will need to perform some extra steps for the same. There is another option I found is to activate developer mode. 

Now the copying might be a concern if the templates folder is large in size, currently it is only 70KB. 

Another option is to create a junction using (/J) which will be similar, but it needs the files to be on the same volume (not andissue for us).

ZhiweiYan-96: >> Now the copying might be a concern if the templates folder is large in size
@ratnampa  Thanks for your info, the template file size should be steady and small, as all files in this directory are just come code templates. I think we the copy is affordable for us.

ZhiweiYan-96 (approved):

EikanWang (approved):

PR #1058, opened 2024-11-06
Add tests for aten::record_stream

Kanya-Mo:
Currently there's no available UT in pytorch to test record_stream. These two tests are adapted from corresponding [cuda tests](https://github.com/pytorch/pytorch/blob/a7479fa2828ad55a056a60a629dff6b7a0cb6b98/test/test_cuda.py#L703). The only difference is I use an actual expensive kernel in place of `torch.cuda._sleep` to create delay in one stream. The add kernel here would create sufficient delay based on max memory bandwidth among current supported gpus.


xytintel:
@Kanya-Mo Pls move `src/ATen/native/xpu/sycl/RecordStream.cpp` to `src/ATen/native/xpu/RecordStream.cpp`

Kanya-Mo:
> @Kanya-Mo Pls move `src/ATen/native/xpu/sycl/RecordStream.cpp` to `src/ATen/native/xpu/RecordStream.cpp`

Moved

xytintel (approved):

PR #1087, opened 2024-11-15
CTCLoss: Fix the hang issue caused by barrier divergence

xytintel:
Resolve https://github.com/pytorch/pytorch/issues/140781

### DIFF
--- a/src/ATen/native/xpu/sycl/LossCTCKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LossCTCKernels.cpp
@@ -111,7 +111,6 @@ struct CTCLossLogAlphaKernelFunctor {
         have_three = false;
       }
       for (int64_t t = 1; t < max_input_length_; t++) {
-        item.barrier(sycl_local_fence);
         if ((t < input_length) && (s < 2 * target_length + 1)) {
           // only for valid t, s. This is equation (6) and (7), la1, la2, la3
           // are the three summands, lamax is the maximum for the logsumexp
@@ -161,7 +160,6 @@ struct CTCLossLogAlphaKernelFunctor {
         }
       }
     }
-    item.barrier(sycl_local_fence);
 
     // compute the loss (eq (8))
     if (tid_x == 0) {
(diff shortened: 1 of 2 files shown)


src/ATen/native/xpu/sycl/LossCTCKernels.cpp
```diff
@@ -111,7 +111,6 @@ struct CTCLossLogAlphaKernelFunctor {
         have_three = false;
       }
       for (int64_t t = 1; t < max_input_length_; t++) {
-        item.barrier(sycl_local_fence);
```
EikanWang, line 114: Currently, it depends on workloads to expose such kind of issues. My question is why we could not find issues earlier.

EikanWang (approved):

dvrogozh:
With this PR issue reported in https://github.com/pytorch/pytorch/issues/140781 is gone. The HF tests for hubert model pass as follows for me: `156 passed, 257 skipped, 3 warnings in 52.34s`.

src/ATen/native/xpu/sycl/LossCTCKernels.cpp
dvrogozh: To catch such issues you typically need to do one of the 2 things, better both:
1. Each or almost each change to the code should be accompanied with dedicated test. If you reuse existing test, we need to review that it tests all corner cases. With the issue we spotted we could check that early exit conditions are actually being tried out.
2. Run real life tests, preferably from real life 3rd party framework or library. Huggingface Transformers gives you excellent way to do this.

1st item is a better guarantee that issues won't be missed. 2nd item is a lesser guarantee, but gives certainty that at least some real life cases will work. In both cases issues might be missed, but utilizing both we reduce the probability that something will get missed.

dvrogozh (approved):
I trust you decision that barrier really is not needed. Other than that change works to fix the issue I noticed. Consider to extend test coverage to cover the missed case.

xytintel:
> I trust you decision that barrier really is not needed. Other than that change works to fix the issue I noticed. Consider to extend test coverage to cover the missed case.

We still need barriers. The reason for the hang is that some threads exit prematurely, preventing the counter from resetting to zero. We are now planning to use `named barrier` to solve this problem.

dvrogozh:
I verified updated version. It works to address reported issue.

PR #1098, opened 2024-11-19
Update skip list for sparse related test cases

xytintel:
As we supported `SparseXPU` backend for basic operators, some UTs can be enabled.


daisyden (approved):
LGTM

PR #1099, opened 2024-11-19
Update skip list in `test_torch_xpu`

xytintel:
As we enhanced XPU operator coverage, some UTs in `test_torch_xpu.py` can be enabled.


daisyden (approved):
LGTM

PR #1115, opened 2024-11-22
Support lazy initialization for empty_xpu

guangyey:
# Motivation
Fix https://github.com/pytorch/pytorch/issues/140877
Some PyTorch C++ users could call empty op directly. In this situation, lazy initialization would not have been triggered. So we have to add `lazyInitDevice` here, which also aligns with CUDA [convention](https://github.com/pytorch/pytorch/blob/150ffb6e07f3802f8d0d3e843486e77a872803cf/aten/src/ATen/cuda/EmptyTensor.cpp#L13).


xytintel (approved):

guangyey:
XPU build pass and lint is green. Unrelated failures.

PR #1116, opened 2024-11-22
Fix the low performance issue in `GroupNorm` kernels

xytintel:
Primarily adopt vectorization approaches for `GroupNormForward` kernel.


xytintel:
image
image
image

xytintel:
We use the following code for performance test:
```python
import torch
a = torch.rand(2, 256, 128, 128).xpu().bfloat16()
# bn = torch.nn.BatchNorm2d(256).xpu().bfloat16()
bn = torch.nn.GroupNorm(num_channels=256, num_groups=4).xpu().bfloat16()
    activities=[
)
with prof_xpu:
    for i in range(120):
        output = bn(a)
print(prof_xpu.key_averages().table(sort_by="self_xpu_time_total", row_limit=100000))
print(output.dtype)
```
The kernel time improves from 130.589ms to 103.744ms

yucai-intel (approved):

PR #1146, opened 2024-12-06
Skipped cases in extended folder

PenghuiCheng:
1. Skipped cases since div floor and rounding accuracy tolerance is large on bfloat16, float16 and float64 dtype.
2. Skipped cases for 'Kernel is incompatible with all devices' error.


PenghuiCheng:
merge it to https://github.com/intel/torch-xpu-ops/pull/1161

PR #1148, opened 2024-12-09
Building System: Enable device code compression (Lin/Win)

fengyuan14:
This is the first stage to align binary distribution with CUDA. That is only `libtorch_cuda.so` at all. The first stage helps on combining all kernel libraries into one. After the commit, we will have two libraries `libtorch_xpu.so` and `libtorch_xpu_ops_sycl_kernel.so`, and the commit also helps us make the bin size on par as CUDA. In the next stage, we will
1. Combine all XPU libraries into one, `libtorch_xpu.so`.
2. Revert lin/win divergence.


cmake/Modules/FindSYCL.cmake
```diff
@@ -215,10 +215,17 @@ macro(SYCL_WRAP_SRCS sycl_target generated_files)
...
+  else()
+    set(library_name ${sycl_target})
+  endif()
+  
   set(SYCL_include_dirs "${SYCL_INCLUDE_DIR}")
-  list(APPEND SYCL_include_dirs "$<TARGET_PROPERTY:${sycl_target},INCLUDE_DIRECTORIES>")
+  list(APPEND SYCL_include_dirs "$<TARGET_PROPERTY:${library_name},INCLUDE_DIRECTORIES>")
 
-  set(SYCL_compile_definitions "$<TARGET_PROPERTY:${sycl_target},COMPILE_DEFINITIONS>")
+  set(SYCL_compile_definitions "$<TARGET_PROPERTY:${library_name},COMPILE_DEFINITIONS>")
```
fengyuan14, line 228: The logic of customizing libname should be out of the file. `FindSYCL.cmake` is a file to  provide general helpers.

cmake/Modules/FindSYCL.cmake
```diff
@@ -434,6 +441,16 @@ macro(SYCL_ADD_LIBRARY sycl_target)
...
+  
+  # For one library, in which case the sycl_target should be shorter to fit the 
+  # character limit of Windows batch files, and the library_name is set to be 
+  # more recognizable.
+  set(library_name)
+  if(${sycl_target} STREQUAL "xpu_ops_kernels")
+    set(library_name torch_xpu_ops_sycl_kernels)
+  else()
+    set(library_name ${sycl_target})
+  endif()
```
fengyuan14, line 453: Ditto.

chunhuanMeng (approved):

cmake/BuildFlags.cmake
```diff
@@ -115,16 +115,17 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "MSVC"
...
   set(SYCL_OFFLINE_COMPILER_CG_OPTIONS "${SYCL_OFFLINE_COMPILER_CG_OPTIONS} -cl-intel-enable-auto-large-GRF-mode")
   set(SYCL_OFFLINE_COMPILER_CG_OPTIONS "${SYCL_OFFLINE_COMPILER_CG_OPTIONS} -cl-fp32-correctly-rounded-divide-sqrt")
   set(SYCL_OFFLINE_COMPILER_CG_OPTIONS "-options '${SYCL_OFFLINE_COMPILER_CG_OPTIONS}'")
 
   if(WIN32)
-    set(AOT_TARGETS "ats-m150,mtl-u,mtl-h,xe2-lpg,xe2-hpg")
+    set(AOT_TARGETS "dg2-g10,mtl-u,mtl-h,xe2-lpg,xe2-hpg")
   else()
-    set(AOT_TARGETS "pvc,xe-lpg,ats-m150")
+    set(AOT_TARGETS "pvc,xe-lpg,dg2-g10,xe2-lpg,xe2-hpg")
```
chuanqi129, line 128: As we discussed before, let's remove BMG in default linux aot target. Because we have LTS driver for linux server gpu, it will break the build on LTS driver.

PR #1155, opened 2024-12-10
Windows: Fixing random runtime error when import torch (#1043)

Stonepia:
PyTorch has self logic of loading dll on Windows. The logic in torch-xpu-ops:bridge.cpp gets conflict against PyTorch logic.

Thus when the user trying to `import torch` within the env that has pypi installed whl, it will throw error.
 
Cherry-picked from https://github.com/intel/torch-xpu-ops/pull/1043


Stonepia:
Please do not merge for now, this is a draft pr for fixing.

tye1:
why the ci failed?

Stonepia:
Because the ci pulls the latest pytorch, and it is not compatible with the branch that ipex uses. I think that should be ignored.

fengyuan14 (approved):

fengyuan14:
> Because the ci pulls the latest pytorch, and it is not compatible with the branch that ipex uses. I think that should be ignored.

Could you ask Chuanqi how to specify pytorch branch for torch-xpu-ops release branch PRECI?

Stonepia:
After discuss, since this PR is already been tested from the IPEX side, we will directly merge this.

PR #1179, opened 2024-12-18
[CI] Add ccl and mpi env source for XCCL backend related PR test

chuanqi129:


.github/scripts/env.sh
```diff
@@ -4,6 +4,8 @@ if [ "$1" != "nightly_wheel" ];then
     source /opt/intel/oneapi/compiler/latest/env/vars.sh
     source /opt/intel/oneapi/umf/latest/env/vars.sh
     source /opt/intel/oneapi/pti/latest/env/vars.sh
+    source /opt/intel/oneapi/ccl/latest/env/vars.sh
+    source /opt/intel/oneapi/mpi/latest/env/vars.sh
```
zhangxiaoli73, line 8: For building, you don't need it. In runtime, I think you will use `pip install` instead of `source`.

zhangxiaoli73:
@chuanqi129 Could you help check the CI failure?

PR #1180, opened 2024-12-19
Add at::_safe_softmax op

jianyizh:
fuse safe softmax https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/attention.cpp#L656


test/xpu/nn/test_safe_softmax.py
```diff
@@ -0,0 +1,44 @@
+import torch
+from torch.testing._internal.common_utils import TestCase
+
+cpu_device = torch.device("cpu")
+xpu_device = torch.device("xpu")
+
+
+class TestSafeSoftMax(TestCase):
```
xytintel, line 8: If it is not present in the stock PyTorch tests, please move it to `test/regression/`

jianyizh: done

xytintel (approved):

Issue #1166, opened 2024-12-13
Preci report pass even when there are some UT failures. 


daisyden:
### 🐛 Describe the bug

Please check this test, you can  find some UT failures in log, but the result of preci is passed. 
![image](https://github.com/user-attachments/assets/3c59578b-ec89-41d8-8370-2150afea5e9c)
https://github.com/intel/torch-xpu-ops/actions/runs/12270141920/job/34234894443

### Versions

latest version

Stonepia:
This should already done.

PR #1181, opened 2024-12-19
[CI] UT logs post op and artifact

RUIJIEZHONG66166:
For fixing https://github.com/intel/torch-xpu-ops/issues/1166 and track the UT results for every PRs

### DIFF
--- a/.github/workflows/_linux_ut.yml
+++ b/.github/workflows/_linux_ut.yml
@@ -142,17 +142,19 @@ jobs:
         if: contains(inputs.ut, 'op_regression') || github.event_name == 'schedule'
         run: |
           cd ${{ github.workspace }}
+          mkdir -p ut_log/op_regression
           xpu-smi discovery
           source .github/scripts/env.sh ${{ inputs.pytorch }}
           source activate xpu_op_${ZE_AFFINITY_MASK}
           cd ${{ github.workspace }}
           cd test/regressions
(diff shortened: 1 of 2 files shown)


.github/workflows/_linux_ut.yml
```diff
@@ -142,17 +142,19 @@ jobs:
...
           cd ${{ github.workspace }}
+          mkdir -p ut_log/op_regression
           xpu-smi discovery
           source .github/scripts/env.sh ${{ inputs.pytorch }}
           source activate xpu_op_${ZE_AFFINITY_MASK}
           cd ${{ github.workspace }}
           cd test/regressions
           pip install pytest
-          timeout 8000 pytest -v
+          timeout 8000 pytest -v 2>&1 | tee ${{ github.workspace }}/ut_log/op_regression/op_regression_test.log
```
Stonepia, line 152: this `2>&1` will still redirect all logs from `stderr` to `stdout`, so the log will still be way too long for this part.

Personally, I would recommend we only show stdout to the user, and cat `stderr` when the test failed.

RUIJIEZHONG66166: Have separated the error information as a single log, and keep the whole output at the same time.

chuanqi129 (approved):

Stonepia (approved):

PR #1184, opened 2024-12-19
ci: print annotations for key package versions in transformers test

dvrogozh:
Annotations are available on a summary page of executed workflow.

### DIFF
--- a/.github/workflows/_linux_transformers.yml
+++ b/.github/workflows/_linux_transformers.yml
@@ -50,6 +50,7 @@ jobs:
       DisableScratchPages: ${{ inputs.driver == 'rolling' && '1' || '0' }}
       python: ${{ inputs.python != '' && inputs.python || '3.10' }}
       pytorch: ${{ inputs.pytorch != '' && inputs.pytorch || 'nightly' }}
+      transformers: ${{ inputs.transformers != '' && inputs.transformers || 'v4.47.0' }}
       TRANSFORMERS_TEST_DEVICE_SPEC: 'spec.py'
     steps:
       - name: Checkout torch-xpu-ops
@@ -60,7 +61,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: huggingface/transformers
-          ref: ${{ inputs.transformers != '' && inputs.transformers || 'v4.47.0' }}
+          ref: ${{ env.transformers }}
           path: transformers
       - name: Prepare OS environment
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/_linux_transformers.yml
```diff
@@ -106,12 +107,54 @@ jobs:
...
+          # printing annotations for the key packages
+          echo "### Annotations" >> $GITHUB_STEP_SUMMARY
+          echo "| | |" >> $GITHUB_STEP_SUMMARY
+          echo "| --- | --- |" >> $GITHUB_STEP_SUMMARY
+          echo "| jobs.$GITHUB_JOB.versions.os | $(source /etc/os-release && echo $VERSION_ID) |" >> $GITHUB_STEP_SUMMARY
+          echo "| jobs.$GITHUB_JOB.versions.linux-kernel | $(uname -r) |" >> $GITHUB_STEP_SUMMARY
+          echo "| jobs.$GITHUB_JOB.versions.python | $(python --version | cut -f2 -d' ') |" >> $GITHUB_STEP_SUMMARY
+          packages=" \
+            level-zero \
+            libigc1 \
```
dvrogozh, line 125: Note that this accounts to both cases: LTS and rolling drivers. They have some components available in the packages with the different names, like libigc1 vs. libigc2 and same for level-zero. This will result in printing empty versions for not installed packages.

RUIJIEZHONG66166:
LGTM. @chuanqi129 Could you please double check?

.github/workflows/_linux_transformers.yml
```diff
@@ -106,12 +107,54 @@ jobs:
         id: installed
         run: |
           source activate huggingface_transformers_test
-          echo "TORCH_BRANCH_ID=$(python -c 'import torch; print(torch.__version__)')" |tee -a "${GITHUB_OUTPUT}" >> "${GITHUB_ENV}"
-          echo "TORCH_COMMIT_ID=$(python -c 'import torch; print(torch.version.git_version)')" |tee -a "${GITHUB_OUTPUT}" >> "${GITHUB_ENV}"
           echo "pip installed packages:"
           pip list | tee ${{ github.workspace }}/transformers/tests_log/pip_list.txt
+          echo "lspci gpu devices:"
+          lspci -d ::0380 | tee ${{ github.workspace }}/transformers/tests_log/lspci_0380.txt
```
chuanqi129, line 113: Can we use `xpu-smi discovery` to get the gpu devices information? Because the runner procession with `ZE_AFFINITY_MASK` by default, the `lspci` always show all devices information.

.github/workflows/_linux_transformers.yml
```diff
@@ -106,12 +107,54 @@ jobs:
...
+          echo "| jobs.$GITHUB_JOB.drm.render_nodes_vendor_ids | $(echo $var | sed "s/ /,/g") |" >> $GITHUB_STEP_SUMMARY
+          var="[$(cat /sys/class/drm/render*/device/device)]"
+          echo "| jobs.$GITHUB_JOB.drm.render_nodes_device_ids | $(echo $var | sed "s/ /,/g") |" >> $GITHUB_STEP_SUMMARY
+          var=$(python -c "import torch; print(torch.version.xpu)")
+          echo "| jobs.$GITHUB_JOB.torch.version.xpu | $var |" >> $GITHUB_STEP_SUMMARY
+          var=$(python -c "import torch; print(torch.xpu.device_count())")
+          echo "| jobs.$GITHUB_JOB.torch.xpu.device_count | $var |" >> $GITHUB_STEP_SUMMARY
+          # printing annotations with key environment variables
+          echo "| jobs.$GITHUB_JOB.env.ZE_AFFINITY_MASK | $ZE_AFFINITY_MASK |" >> $GITHUB_STEP_SUMMARY
+          echo "| jobs.$GITHUB_JOB.env.NEOReadDebugKeys | $NEOReadDebugKeys |" >> $GITHUB_STEP_SUMMARY
```
chuanqi129, line 157: how do you think if we convert this step as a standalone scripts instead of long step in the workflow?

chuanqi129 (approved):
Overall LGTM, add 2 minor comments, FYI

.github/workflows/_linux_transformers.yml
dvrogozh: Good idea. But let's add running xpu-smi in a stand-alone step. I remember it gives lengthy output by itself and for us it has special interest. Good to separate from this step. I will submit another PR on this since I need to experiment with xpu-smi a little for the better output.

.github/workflows/_linux_transformers.yml
dvrogozh: It's not that long at the moment to strongly require converting to the script per my opinion.... Eventually complicated stuff, or often reused stuff, or lengthy stuff needs conversion to the script. At the same time I personally am trying to avoid calling scripts from github actions since this impacts workflow transparency and ultimately complicates workflow. So, right balance needs to be looked for here.

For this particular step, I actually wonder maybe we need to convert it to the stand-alone action rather than script? In this way it can be reused by other workflows in the repo in their steps and bring unified format for reporting this information.

Overall, let's merge in this PR as is and discuss these proposals.

dvrogozh:
PR did not modify the scope of failing ci test (pull/preci-ut/Torch-XPU-UT-Tests). Can be merged.

PR #1186, opened 2024-12-20
[n, k//2] int4pack to [n, k//8] w/o transpose

ZhiweiYan-96:
# Motivation
1. Original should be [n, k//2] int4x2, target should be [n, k//8] int4x8, no need to transpose, so the indexing logic is changed
2. **[Need to be decided]** We should not consider big endian/small endian here right? I think int4 should be packed in easiest way, aka, just right shift per bits.


src/ATen/native/xpu/sycl/WeightInt4PackKernel.cpp
```diff
@@ -9,19 +9,22 @@ namespace at::native::xpu {
 struct WeightToInt4PackKernelFunctor {
   void operator()(sycl::item<1> item) const {
     auto idx = item.get_linear_id();
-    int out_y = idx / N_;
-    int out_x = idx % N_;
-    int in_y = out_x;
-    int in_x = out_y * 4;
     int K_div_2 = K_ / 2;
+    int K_div_8 = K_ / 8;
```
airMeng, line 13: although might never happened, but could you handle ```K_``` not divided by 8?

ZhiweiYan-96: good point, let me add an assert

ZhiweiYan-96: added https://github.com/intel/torch-xpu-ops/pull/1186/files#diff-8988cf89a354f8324e695e52356e1f1d45e70a0a31db3db12a9023da8e2fe7e2R20

liangan1 (approved):
LGTM

PR #1188, opened 2024-12-20
Event free: Make the event free APIs compatible with different compil…

fengyuan14:
…er versions

In addition,
1. The PR includes compatibility changes of PRECI for PyTorch 2.5.
2. The PR mutes some warning errors raised from SYCL headers.


cmake/Modules/FindSYCLToolkit.cmake
```diff
@@ -35,6 +35,7 @@ endif()
 if(SYCLTOOLKIT_FOUND)
   return()
 endif()
+
```
fengyuan14, line 38:
```suggestion
```

gujinghui (approved):

PR #1206, opened 2024-12-24
ci/transformers: dump xpu-smi output in the log

dvrogozh:


.github/workflows/_linux_transformers.yml
```diff
@@ -112,6 +112,8 @@ jobs:
           lspci -d ::0380 | tee ${{ github.workspace }}/transformers/tests_log/lspci_0380.txt
           echo "GPU render nodes:"
           cat /sys/class/drm/render*/device/device | tee ${{ github.workspace }}/transformers/tests_log/device_IDs.txt
+          echo "xpu-xmi output:"
```
chuanqi129, line 115:
```suggestion
          echo "xpu-smi output:"
```

dvrogozh: fixed

chuanqi129 (approved):

PR #1218, opened 2024-12-26
Fix huggingface import cached_download issue

mengfei25:
Downgrading huggingface-hub to 0.25.0


chuanqi129:
Please describe that if we don't downgrade the version, what will be happened?

mengfei25:
Will use pytorch pinned torchbench synced with CUDA instead of downgrading huggingface-hug
https://github.com/intel/torch-xpu-ops/pull/1226

PR #1225, opened 2024-12-27
Skip cases on MTL device for windows

PenghuiCheng:
Skip cases on MTL device for windows


daisyden (approved):
LGTM， the triage of these failures is TBD.

PR #1243, opened 2025-01-03
Building System: Only build one shared library for the user

chunhuanMeng:
This is the second phase of aligning binary distribution with CUDA. Currently, there is only libtorch_xpu.so on both windows and linux, but there are still some differences between win and lin, which we will fix after the sycl compiler is fixed.


src/BuildOnWindows.cmake
```diff
@@ -160,96 +131,96 @@ elseif(BUILD_SPLIT_KERNEL_LIB)
...
-  target_link_libraries(${sycl_poly_lib} PUBLIC torch_xpu)
-  list(APPEND TORCH_XPU_OPS_LIBRARIES ${sycl_poly_lib})
+  install(TARGETS ${sycl_lib} DESTINATION "${TORCH_INSTALL_LIB_DIR}")
 
-  # Decouple with PyTorch cmake definition.
-  install(TARGETS ${sycl_poly_lib} DESTINATION "${TORCH_INSTALL_LIB_DIR}")
+  list(APPEND TORCH_XPU_OPS_LIBRARIES torch_xpu_ops)
+  list(APPEND TORCH_XPU_OPS_LIBRARIES torch_xpu_ops_aten)
+else()
+  add_library(
```
fengyuan14, line 152: Please add comments why we need split objects into two separate archive libraries on Windows.

src/BuildOnLinux.cmake
```diff
@@ -32,6 +33,20 @@ if(BUILD_SEPARATE_OPS)
     install(TARGETS ${sycl_lib} DESTINATION "${TORCH_INSTALL_LIB_DIR}")
   endforeach()
 elseif(BUILD_SPLIT_KERNEL_LIB OR __INTEL_LLVM_COMPILER LESS 20250001 OR ICX_DATE LESS 20241211)
```
fengyuan14, line 35: Please give an informative comment, like `Working with the compilers which don't support device code compression, we have to split kernels into multiple libraries to meet the bin size limitation.`

src/BuildOnWindows.cmake
```diff
@@ -36,7 +38,27 @@ if(BUILD_SEPARATE_OPS)
     # Decouple with PyTorch cmake definition.
     install(TARGETS ${sycl_lib} DESTINATION "${TORCH_INSTALL_LIB_DIR}")
   endforeach()
+  list(APPEND TORCH_XPU_OPS_LIBRARIES torch_xpu_ops)
+  list(APPEND TORCH_XPU_OPS_LIBRARIES torch_xpu_ops_aten)
 elseif(BUILD_SPLIT_KERNEL_LIB OR __INTEL_LLVM_COMPILER LESS 20250001 OR ICX_DATE LESS 20241211)
```
fengyuan14, line 43: Ditto.

src/BuildOnWindows.cmake
```diff
@@ -36,7 +38,27 @@ if(BUILD_SEPARATE_OPS)
...
+  add_library(
+    torch_xpu_ops
+    STATIC
+    ${ATen_XPU_CPP_SRCS}
+    ${ATen_XPU_MKL_SRCS})
```
fengyuan14, line 48: Is it right here? MKL sources are needed by `ATen_XPU_NATIVE_CPP_SRCS`? Operators involving MKL implementation are in `ATen_XPU_NATIVE_CPP_SRCS`.

src/BuildOnWindows.cmake
```diff
@@ -230,27 +252,87 @@ elseif(BUILD_SPLIT_KERNEL_LIB OR __INTEL_LLVM_COMPILER LESS 20250001 OR ICX_DATE
...
+    string(REGEX MATCH "Binary" IS_BINARY ${sycl_src})
+    string(REGEX MATCH "Unary" IS_UNARY ${sycl_src})
+    string(REGEX MATCH "Pow" IS_POW ${sycl_src})
+    string(REGEX MATCH "Copy" IS_COPY ${sycl_src})
+    string(REGEX MATCH "Reduce" IS_REDUCE ${sycl_src})
+    string(REGEX MATCH "Activation" IS_ACTIVATION ${sycl_src})
+    string(REGEX MATCH "Foreach" IS_FOREACH ${sycl_src})
+
+    if(NOT IS_FOREACH STREQUAL "")
+      list(APPEND ATen_XPU_SYCL_UNARY_BINARY_SRCS ${sycl_src})
```
fengyuan14, line 286: Please optimize the naming, `ATen_XPU_SYCL_UNARY_BINARY_SRCS`, as the list contains sources not only unary binary related, but also foreach and reduce related.

src/BuildOnWindows.cmake
CuiYifeng: You're right. `${ATen_XPU_MKL_SRCS}` should not be here.

fengyuan14 (approved):

xytintel:
Double checked

xytintel (approved):

PR #1257, opened 2025-01-07
ci: add test for Huggingface Accelerate

dvrogozh:


.github/workflows/_linux_accelerate.yml
```diff
@@ -0,0 +1,156 @@
...
+        required: true
+        type: string
+        default: 'linux.idc.xpu'
+        description: Runner label
+      nightly_whl:
+        required: false
+        type: string
+        default: ''
+        description: Pytorch nightly wheel version
+      accelerate:
```
chuanqi129, line 33: Do we need add a parameter for transformers version?

.github/workflows/_linux_accelerate.yml
```diff
@@ -0,0 +1,156 @@
...
+        run: |
+          which conda && conda clean -ay
+          conda remove --all -y -n $CONDA_ENV_NAME || rm -rf $(dirname ${CONDA_EXE})/../envs/$CONDA_ENV_NAME
+          conda create -y -n $CONDA_ENV_NAME python=${{ env.python }}
+          source activate $CONDA_ENV_NAME
+          pip install junitparser
+      - name: Prepare Stock XPU Pytorch
+        run: |
+          source activate $CONDA_ENV_NAME
+          if [ -z "${{ inputs.nightly_whl }}" ]; then
```
chuanqi129, line 76: Do we plan this test only target for torch cpu nightly wheel? latest nightly or a specific nightly version?

.github/workflows/_linux_accelerate.yml
```diff
@@ -0,0 +1,156 @@
...
+      NEOReadDebugKeys: 0
+      DisableScratchPages: 0
+      accelerate: ${{ inputs.accelerate != '' && inputs.accelerate || 'v1.2.1' }}
+      python: ${{ inputs.python != '' && inputs.python || '3.10' }}
+      pytorch: ${{ inputs.pytorch != '' && inputs.pytorch || 'nightly' }}
+      PYTORCH_DEBUG_XPU_FALLBACK: 1
+      ZE_AFFINITY_MASK: 0
+      PARSE_JUNIT: ${{ github.workspace }}/torch-xpu-ops/.github/scripts/parse-junitxml.py
+    steps:
+      - name: Checkout torch-xpu-ops
```
chuanqi129, line 56: Seems we don't need checkout torch-xpu-ops

.github/workflows/_linux_accelerate.yml
```diff
@@ -0,0 +1,156 @@
...
+          source activate $CONDA_ENV_NAME
+          # These checks are to exit earlier if for any reason torch
+          # packages were reinstalled back to CUDA versions (not expected).
+          pip show torch | grep Version | grep xpu
+          pip show torchaudio | grep Version | grep xpu
+          pip show torchvision | grep Version | grep xpu
+          python -c 'import torch; exit(not torch.xpu.is_available())'
+      - name: Run -k backbone tests
+        env:
+          TEST_CASE: 'tests_backbone'
```
chuanqi129, line 112: Will it reuse transformers test case? Or it has itself tests?

chuanqi129 (commented):
I'd prefer to clean the transformers tests of the first step and then expand  to the ecosystem libraries

.github/workflows/_linux_accelerate.yml
dvrogozh: We need it. Workflow uses:
* .github/scripts/parse-junitxml.py
* .github/actions/print-environment/action.yml

from torch-xpu-ops repo. However, I forgot to add .py script to the dependencies of this workflow and I also I forgot to drop `spec.py` which is Transformers specific - I will fix that.

dvrogozh: > However, I forgot to add .py script to the dependencies of this workflow and I also I forgot to drop spec.py which is Transformers specific - I will fix that.

Done.

.github/workflows/_linux_accelerate.yml
dvrogozh: It has its own tests. I forgot to change this copying from Transformers workflow. Thanks for the catch. Fixed.

.github/workflows/_linux_accelerate.yml
dvrogozh: That's a good question. These 2 projects have circle dependencies. My understanding is:
1. Transformers depend on Accelerate library for basic functionality
2. Accelerate library either does not depend on Transformers library or at least does not depend for basic features
3. Accelerate uses Transformers in tests
4. Accelerate uses Transformers in some utilities. Mostly for tests, but without a deep dive I am not 100% sure that there is no usage in Accelerate for some library features.

When we install Accelerate for testing we also get Transformers installed. As far as I see it will install latest Transformers version.

Overall, I think it will be wise to pin Transformers version for better test stability. I will add a change today.

.github/workflows/_linux_accelerate.yml
dvrogozh: I suggest to target it for latest nightly triggered per schedule daily. @chuanqi129 : are you ok if I will drop input parameter to set nightly version?

.github/workflows/_linux_accelerate.yml
dvrogozh: Added pin for Transformers version.

.github/workflows/_linux_accelerate.yml
dvrogozh: I dropped `inputs.nightly_whl` for now and just install latest nightly. Note that this parameter in any case would need a change since it controls only `torch` version and does not allow to set version for `torchvision` and `torchaudio`. So, if needed, we can add these options later on.

.github/workflows/_linux_accelerate.yml
```diff
@@ -0,0 +1,151 @@
...
+        default: 'v4.47.1'
+        description: Transformers version
+
+permissions: read-all
+
+jobs:
+  Torch-XPU-Accelerate-Tests:
+    runs-on: ${{ inputs.runner != '' && inputs.runner || 'linux.idc.xpu' }}
+    env:
+      CONDA_ENV_NAME: 'huggingface_accelerate_test'
```
dvrogozh, line 45: Add ZE_AFFINITY_MASK to the environment name to make it unique.

dvrogozh: done

chuanqi129 (approved):

dvrogozh:
pull ci failures are unrelated with the new test introduced in the PR. Can merge.

PR #1282, opened 2025-01-13
ci: update Transformers to v4.51.3

dvrogozh:
Changes:
* Benchmarking scripts are pruned from Transformers by v4.49.0 due to deprecation. So we don't need to test them anymore.
* Some cuda specific tests were generalized to cover non-cuda devices which uncovered some issues.
* Some new tests were added which fail for both cuda and xpu.
* Few regressions due to changes on Transformers side
* Time required to run the test on a single system increased from around 2h to 4h
* Test split by matrix to be able to test in smaller rerunnable chunks
* Test chunks are run sequentially due to issue with parallel HF downloads

Signed-off-by: Dmitry Rogozhkin <dmitry.v.rogozhkin@intel.com>

### DIFF
--- a/.github/actions/print-environment/action.yml
+++ b/.github/actions/print-environment/action.yml
@@ -16,6 +16,11 @@ inputs:
     type: string
     default: ''
     description: "Space separated list of PyPi packages to evaluate"
+  to:
+    required: false
+    type: string
+    default: ''
+    description: "File to print environment to"
 
 runs:
   using: composite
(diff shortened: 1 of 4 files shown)


dvrogozh:
Transformers v4.48.0 has a regression in tests:
```
# TRANSFORMERS_TEST_DEVICE_SPEC=spec.py python3 -m pytest tests/models/marian/test_modeling_marian.py -k backbone

E   ModuleNotFoundError: No module named 'transformers.models.marian.convert_marian_to_pytorch'
```

I don't see this issue in main. It might be reasonable to wait for v4.49.0 or bring this issue to HF if it will persist.

dvrogozh:
> Transformers v4.48.0 has a regression in tests:

Same issue observed on v4.49.0. I reported it to Transformers project and will exclude the test manually for now:
* https://github.com/huggingface/transformers/issues/36267

dvrogozh:
> @dvrogozh do we still need this PR? There are some code conflict

We need it. That's my attempt to update tested Transformers version. This unfortunately stepped into an issue of increased test time due to which we started to catch infrastructure errors (like fetching some content from HF) without a way to efficiently rerun some tests. I have an idea how to handle that, but this needs some time which slips away. I will work on that in the coming weeks.

chuanqi129 (approved):
LGTM

PR #1289, opened 2025-01-15
Make the code pass lint checking with "--skip CLANGTIDY,CLANGFORMAT --all-files"

daisyden:
This PR is to clean up lint issue on torch-xpu-ops and enable lint in CI. We will reuse upstream pytorch lint checkers. PyTorch lint checker .github/scripts/lintrunner.sh has two kinds of check, one is with CLANGDITY and GLANGFORMAT skipped, another is with them enabled. The later also scan the code created by torchgen in build folder. 
This PR refined the .lintrunner.toml introduced in https://github.com/intel/torch-xpu-ops/pull/1267,  and cleaned all the issues detected on static code. With this PR, the code can pass
```
lintrunner --force-color --tee-json=lint.json --skip CLANGTIDY,CLANGFORMAT --all-files
```

I will create different PRs to enable the checker in CI and enable CLANGDITY and GLANGFORMAT.


.github/workflows/_linux_transformers.yml
```diff
@@ -45,6 +45,7 @@ permissions: read-all
 
 jobs:
   Torch-XPU-Transformers-Tests:
+    if: github.repository_owner == 'pytorch'
```
dvrogozh, line 48: Why are you adding this here? This workflow is intended to be run in torch-xpu-ops repository. With this change  workflow won't run at all which is wrong.

dvrogozh (dismissed):
Please, update:
1. PR title - "Daisyden/lint3" does not give any understanding on the goal of the PR
2. PR description clearly stating what are intended changes of this PR

Until above won't be done I strongly suggest that this PR should be moved to drafts or labelled as draft or named with DO-NOT-MERGE.

I also see that PR modifies workflow execute conditions. What are you trying to achieve here and why? This should be put into the PR description.... Are you trying to avoid workflows executed in the user forks? or you plan to execute some workflows from github.com/pytorch/pytorch repo instead of from here? I honestly can't guess the intent with the lack of description. And some of the changes, for example to `_linux_transformers.yml` seem clearly wrong to me.

.github/workflows/_linux_transformers.yml
daisyden: Good catch! It is added by lint NO_WORKFLOWS_ON_FORK check auto correction, maybe we need to skip this checker. Let me roll back this change.

daisyden: Description is added. Thanks.

daisyden:
@dvrogozh  I did update according to your comments, please help to review the PR.

PenghuiCheng (approved):

PR #1287, opened 2025-01-15
Fix keyword restrict to fix Windows Build

CuiYifeng:
Replace `__restrict__` with macro `RESTRICT` to maintain cross platform compatibility.

PR #1297, opened 2025-01-17
[CI] Enable long paths on Windows

RUIJIEZHONG66166:
Enable long paths on Windows CI

### DIFF
--- a/.github/workflows/_windows_ut.yml
+++ b/.github/workflows/_windows_ut.yml
@@ -72,6 +72,9 @@ jobs:
           call conda install -y libuv
           call conda install -y rust 
           git config --system core.longpaths true
+          git config --global core.symlinks true
+          git config --global core.fsmonitor false
+          powershell -Command "Set-ItemProperty -Path "HKLM:\\SYSTEM\CurrentControlSet\Control\FileSystem" -Name "LongPathsEnabled" -Value 1"
           git status 
           git show -s 
           git submodule sync && git submodule update --init --recursive
--- a/.github/workflows/_windows_ut.yml
+++ b/.github/workflows/_windows_ut.yml
@@ -109,7 +109,7 @@ jobs:
           set MAX_JOBS=4
           python setup.py bdist_wheel > build_torch_wheel_log.log
           echo "[INFO] begin to install torch whls"
-          for /r C:\runner\actions-runner\_work\torch-xpu-ops\pytorch\dist %%i in (torch*.whl) do (
+          for /r C:\actions-runner\_work\torch-xpu-ops\pytorch\dist %%i in (torch*.whl) do (
             set TORCH_WHL=%%i
           )
           echo "[INFO] the torch version is %TORCH_WHL%"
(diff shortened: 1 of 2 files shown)


.github/workflows/_windows_ut.yml
```diff
@@ -72,6 +72,9 @@ jobs:
           call conda install -y libuv
           call conda install -y rust 
           git config --system core.longpaths true
+          git config --global core.symlinks true
+          git config --global core.fsmonitor false
+          powershell -Command "Set-ItemProperty -Path "HKLM:\\SYSTEM\CurrentControlSet\Control\FileSystem" -Name "LongPathsEnabled" -Value 1"
```
dvrogozh, line 77: Long path issues in XXI century :). Yeah, that's known problem. However, setting long paths as you are doing above might not fully help. It did not help for me in one of the projects I was doing before. The full solution which ultimately resolved the issue was twofold with the setting you applied being one of the fixes. There is a chance that you will still see issues with MSVC toolchain and to resolve that it is required to patch the toolchain executables with long paths support.

I hope you won't need that, but if you will still see the issues, below is a second part of the solution which helped us on that other project. Below is a snapshot of commit message and a relevant script (I can't just post a link since it was inhouse project):
```
We've stepped into Windows long files limitation which is that by default Windows
does not allow file and path name lenght to be more than 260. We step into this
threefold: 1) with git clone, 2) with meson subproject archives fetch and 3) with MSVC
toolchain. First 2 issues are getting fixed by Windows setting to enable long files.
Toolchain issue requires a workaround to apply manifest to toolchain executables to
enable long files support. Since it's not good to modify system binaries, this commit
copies toolchain executables and patches them with required manifest.
```
The script:
```
$ErrorActionPreference = "Stop"

function PatchToolset  {
    mkdir _bin >$null
    Set-Location _bin

    $cvtres = $(Get-Command cvtres -TotalCount 1).Source
    $link = $(Get-Command link -TotalCount 1).Source
    $rc = $(Get-Command rc -TotalCount 1).Source

    # Windows does not allow symlinks for generic users by default
    # so we will just copy some DLLs which link, rc, etc. tools require.
    cp "$(Split-Path $link -Parent)\mspdb140.dll" .\
    cp "$(Split-Path $link -Parent)\mspdbcore.dll" .\
    cp "$(Split-Path $link -Parent)\tbbmalloc.dll" .\

    cp $cvtres .\
    cp $rc .\
    cp $link .\

    # Apply longPathAware manifest to copied executables
    mt -nologo -manifest "$PSScriptRoot\longPathAware.manifest" -outputresource:"rc.exe;#1"
    if( -not $? ) { exit 1 }
    mt -nologo -manifest "$PSScriptRoot\longPathAware.manifest" -outputresource:"cvtres.exe;#1"
    if( -not $? ) { exit 1 }
    mt -nologo -manifest "$PSScriptRoot\longPathAware.manifest" -outputresource:"link.exe;#1"
    if( -not $? ) { exit 1 }

    # Export patched toolset location to github environment
    $env:Path = "$(Get-Location);$env:Path"
    Write-Output "Path=$env:Path" >> $env:GITHUB_ENV

    # Print out search queries for the tools we patched. First line
    # should point to .\_bin, second line - to original executable.
    Write-Output "Where cvtres?"
    $(Get-Command cvtres -TotalCount 2).Source
    Write-Output "Where rc?"
    $(Get-Command rc -TotalCount 2).Source
    Write-Output "Where link?"
    $(Get-Command link -TotalCount 2).Source
}

PatchToolset
```
With the `longPathAware.manifest` being:
```
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
<application  xmlns="urn:schemas-microsoft-com:asm.v3">
    <windowsSettings xmlns:ws2="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
        <ws2:longPathAware>true</ws2:longPathAware>
    </windowsSettings>
</application>
</assembly>
```

CuiYifeng:
@RUIJIEZHONG66166 Due to more than one issue with Windows Build, I have temporarily applied a patch from #1287.

CuiYifeng:
Temporarily apply patches from #1303 to test Windows Build.

CuiYifeng:
@RUIJIEZHONG66166 There's a new Windows Build error. Seems it's related to the file system.

Issue #1169, opened 2024-12-16
torch.nextafter has an incorrect result for bf16 on XPU


guangyey:
As titled, this is a regression. IPEX could give a correct result for bf16 on XPU.

```python
import torch

torch.manual_seed(0)

a = torch.randn(5, dtype=torch.bfloat16)
b = torch.randn(5, dtype=torch.bfloat16)
print(f"a: {a}")
print(f"b: {b}")
print(torch.nextafter(a, b))

x_a = a.to('xpu')
x_b = b.to('xpu')
print(torch.nextafter(x_a, x_b))
```
Output:
```
a: tensor([ 1.5391, -0.2930, -2.1719,  0.5703, -1.0859], dtype=torch.bfloat16)
b: tensor([-1.3984,  0.4043,  0.8398, -0.7188, -0.4043], dtype=torch.bfloat16)
tensor([ 1.5312, -0.2910, -2.1562,  0.5664, -1.0781], dtype=torch.bfloat16)
tensor([ 1.5391, -0.2930, -2.1719,  0.5703, -1.0859], device='xpu:0',
       dtype=torch.bfloat16)
```

### Versions

None

cc @fengyuan14 @xytintel

PR #1298, opened 2025-01-17
Improve bfloat16 accuracy of `nextafter` kernel

xytintel:
Resolve https://github.com/intel/torch-xpu-ops/issues/1169

### DIFF
--- a/src/ATen/native/xpu/sycl/StepKernels.cpp
+++ b/src/ATen/native/xpu/sycl/StepKernels.cpp
@@ -7,10 +7,66 @@
 
 namespace at::native::xpu {
 
+static inline c10::BFloat16 nextafteri(c10::BFloat16 from, c10::BFloat16 to) {
+  // Reference:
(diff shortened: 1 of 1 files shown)


fengyuan14 (approved):

guangyey:
It is weird. Could you help me understand why this could work? I don't see any difference between this partial specialization and its std definition in https://github.com/pytorch/pytorch/blob/6cb186e279bc179a6bb63f0226e24ab42a07b394/c10/util/BFloat16-math.h#L249

xytintel:
> It is weird. Could you help me understand why this could work? I don't see any difference between this partial specialization and its std definition in https://github.com/pytorch/pytorch/blob/6cb186e279bc179a6bb63f0226e24ab42a07b394/c10/util/BFloat16-math.h#L249

Just missed the header: https://github.com/intel/torch-xpu-ops/pull/1302

PR #1302, opened 2025-01-20
Improve bfloat16 accuracy of `nextafter` kernel (step2)

xytintel:
Resolve https://github.com/intel/torch-xpu-ops/issues/1169


guangyey (approved):
Make sense.

PR #1303, opened 2025-01-20
Reduce the character number of linkage command

CuiYifeng:


chuanqi129:
@CuiYifeng do we still need this PR?

PR #1306, opened 2025-01-20
Apply vectorization for `batch_norm` channels last kernel

xytintel:


xytintel:
```python
import torch
import torch.nn as nn

N, C, H, W = 4, 24, 160, 256
x = torch.randn(N, C, H, W).to(memory_format=torch.channels_last).bfloat16().xpu()
bn = nn.BatchNorm2d(C).xpu()

    activities=[
)
with prof_xpu:
    for i in range(120):
        output = bn(x)
print(prof_xpu.key_averages(group_by_input_shape=True).table(sort_by="self_xpu_time_total", row_limit=100000))
print(output.dtype)
```

xytintel:
Move to https://github.com/intel/torch-xpu-ops/pull/1317

PR #1308, opened 2025-01-21
Support compatibility with lower version compilers

chunhuanMeng:
Device code compression is a feature that is only available in compiler 2025.0.1 and above. It is not available in lower versions. Compatibility with lower versions is required.


src/CMakeLists.txt
```diff
@@ -12,6 +12,14 @@ add_subdirectory(ATen)
...
   add_subdirectory(xccl)
 endif()
+
+execute_process(
+  COMMAND icpx --version
+  OUTPUT_VARIABLE ICX_VERSION_OUTPUT
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+string(REGEX REPLACE ".*\\.([0-9]+)\\).*" "\\1" ICX_DATE ${ICX_VERSION_OUTPUT})
+
```
fengyuan14, line 22: Move to the top level CMakelists.txt. Move to cmake/SYCL.cmake

fengyuan14 (approved):

PR #1316, opened 2025-01-22
Fix workflows issue

mengfei25:


.github/workflows/nightly_ondemand.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       runner: pvc_e2e
 
   Linux-Nightly-Ondemand-UT-Tests:
-    if: github.event_name == 'schedule' || ${{ inputs.ut_suite }}
+    if: ${{ github.event_name == 'schedule' || inputs.ut != 'skipped' }}
```
chuanqi129, line 84: Does it means if we want to skip the ut test in on-demand test, we can't leave this parameter as space value?

mengfei25: Changed to use space

chuanqi129 (approved):

PR #1318, opened 2025-01-22
Set USE_XCCL default as OFF

Chao1Han:


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ list(APPEND CMAKE_MODULE_PATH ${TORCH_XPU_OPS_ROOT}/cmake/Modules)
 include(${TORCH_XPU_OPS_ROOT}/cmake/SYCL.cmake)
 include(${TORCH_XPU_OPS_ROOT}/cmake/BuildFlags.cmake)
 
-option(USE_XCCL "Build with XCCL support" ON)
+option(USE_XCCL "Build with XCCL support" OFF)
```
dvrogozh, line 41: why?

chuanqi129:
@Chao1Han Please add `export USE_XCCL=1` in https://github.com/intel/torch-xpu-ops/blob/main/.github/workflows/_linux_build.yml#L91, to make sure the torch-xpu-ops CICD build with XCCL

zhangxiaoli73:
@gujinghui Please be aware that we plan to disable XCCL by default (also a suggestion from Validation)  and enable when we are ready for XCCL binary release.

CMakeLists.txt
zhangxiaoli73: Currently, XCCL backend does not get ready in PyTorch release. So we disable it by default (you still can build XCCL with `USE_XCCL`=1) and will enable again in PyTorch 2.8 release by our expectation.

chuanqi129 (approved):

CMakeLists.txt
dvrogozh: That's quite unfortunate since a number of projects awaits distributed support fromXPU. What are outstanding issues preventing to enable XCCL by default?

PR #1339, opened 2025-02-06
Look for ocloc-<version> executable.

githubsgi:
Also, print more details when SYCL compile check fails.


CuiYifeng:
Hi @githubsgi, is this PR still valid?

PR #1353, opened 2025-02-10
Print summary results in page for E2E

mengfei25:


chuanqi129 (approved):
Todo: refine the detail summary as table also

PR #1363, opened 2025-02-14
enable non aot for pytorch compilation

jingxu10:
as title


guangyey (approved):
It means the only method to disable aot is setting by `TORCH_XPU_ARCH_LIST=''`
LGTM.

jingxu10:
> It means the only method to disable aot is setting by `TORCH_XPU_ARCH_LIST=''` LGTM.

yes. Thanks.

PR #1371, opened 2025-02-17
[CI]Split XPU distributed UT into another job

zxd1997066:


.github/workflows/pull.yml
```diff
@@ -233,6 +221,31 @@ jobs:
       ut: op_extended
       runner: linux.idc.xpu
   
+  preci-linux-build-distributed:
```
chuanqi129, line 224: remove this job, we don't need 2 builds

chuanqi129 (changes requested):
The split means, we add 2 jobs in `./.github/workflows/_linux_ut.yml` one for non-distributed ut test(already exist), another for distributed UT

chuanqi129:
Please add `if` condition for `ut_test` job also, if the ut scope only has distributed ut this job will not be triggered also

.github/workflows/_linux_ut.yml
```diff
@@ -49,6 +49,7 @@ permissions: read-all
 jobs:
   ut_test:
     runs-on: ${{ inputs.runner }} 
+    if: contains(inputs.ut, 'xpu_distributed') == false
```
chuanqi129, line 52:
```suggestion
    if: ${{ inputs.ut != 'xpu_distributed' }}
```

chuanqi129 (approved):

PR #1377, opened 2025-02-18
Update USE_XCCL back to PyTorch

guangyey:
# Motivation
We need to update `USE_XCCL` back to PyTorch to facilitate https://github.com/pytorch/pytorch/pull/147161

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,9 +40,35 @@ include(${TORCH_XPU_OPS_ROOT}/cmake/ONEMKL.cmake)
 include(${TORCH_XPU_OPS_ROOT}/cmake/BuildFlags.cmake)
 
 option(USE_XCCL "Build with XCCL support" OFF)
+option(USE_C10D_XCCL "Build with XCCL support for C10D" OFF)
 
-if(NOT WIN32 AND USE_XCCL)
+# -- [ Re-generate the macros file for https://github.com/pytorch/pytorch/pull/147161
+macro(update_caffe2_macros_file)
+  configure_file(
+    ${TORCH_ROOT}/caffe2/core/macros.h.in
+    ${CMAKE_BINARY_DIR}/caffe2/core/macros.h)
+endmacro()
+
+if(NOT DEFINED USE_XCCL)
+  # Propagate the option to PyTorch
+  caffe2_update_option(USE_XCCL OFF)
+  update_caffe2_macros_file()
+endif()
+
+if(USE_XCCL)
   include(${TORCH_XPU_OPS_ROOT}/cmake/XCCL.cmake)
+  if(NOT PYTORCH_FOUND_XCCL)
+    # Propagate the option to PyTorch
+    caffe2_update_option(USE_XCCL OFF)
+    update_caffe2_macros_file()
+  endif()
+endif()
+
+# Propagate the option to PyTorch
+if(USE_DISTRIBUTED AND USE_XCCL AND USE_C10D_XCCL)
+  caffe2_update_option(USE_C10D_XCCL ON)
+else()
+  caffe2_update_option(USE_C10D_XCCL OFF)
 endif()
 
 if(BUILD_TEST)
--- a/cmake/Modules/FindXCCL.cmake
+++ b/cmake/Modules/FindXCCL.cmake
@@ -6,6 +6,12 @@
 
 include(${CMAKE_ROOT}/Modules/FindPackageHandleStandardArgs.cmake)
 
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -41,8 +41,15 @@ include(${TORCH_XPU_OPS_ROOT}/cmake/BuildFlags.cmake)
 
 option(USE_XCCL "Build with XCCL support" OFF)
 
-if(NOT WIN32 AND USE_XCCL)
+if(USE_XCCL)
   include(${TORCH_XPU_OPS_ROOT}/cmake/XCCL.cmake)
+  if(NOT PYTORCH_FOUND_XCCL)
```
guangyey, line 46: refer to https://github.com/pytorch/pytorch/blob/d9cf1debf9831edac51aab7ca29d8a0584474cc5/cmake/Dependencies.cmake#L94

CMakeLists.txt
```diff
@@ -40,9 +40,20 @@ include(${TORCH_XPU_OPS_ROOT}/cmake/ONEMKL.cmake)
...
+    caffe2_update_option(USE_XCCL OFF)
+    # -- [ Re-generate the macros file for https://github.com/pytorch/pytorch/pull/147161
+    configure_file(
+      ${TORCH_XPU_OPS_ROOT}/../../caffe2/core/macros.h.in
+      ${CMAKE_BINARY_DIR}/caffe2/core/macros.h)
+  else()
+    # Propagate the option to PyTorch
+    caffe2_update_option(USE_C10D_XCCL ON)
+  endif()
 endif()
```
pkourdis, line 57: @guangyey can you also propagate the option to PyTorch when `USE_XCCL=OFF` also? I believe this will fix my failing unit test: 
https://github.com/pytorch/pytorch/actions/runs/13425292655/job/37507957344?pr=147161#step:14:9872
when your PR is merged.

guangyey: `USE_C10D_XCCL` is defined as `OFF` by default in cmake. So if `USE_XCCL=OFF`, `USE_C10_XCCL` will be `OFF`. It is unnecessary to update it, right?

pkourdis: I think if we don't update it we end up with `USE_XCCL=` in `torch.__config__.show()` because the user will (normally) not explicitly use `export USE_XCCL=0`.

guangyey: I see your point.

guangyey: updated in line 45

pkourdis: I think we need also to regenerate the macros file when `USE_XCCL` is not defined like you did in the other case?

CMakeLists.txt
```diff
@@ -40,9 +40,25 @@ include(${TORCH_XPU_OPS_ROOT}/cmake/ONEMKL.cmake)
...
 option(USE_XCCL "Build with XCCL support" OFF)
+set(USE_C10D_XCCL OFF CACHE BOOL "Build with XCCL support for C10D")
 
-if(NOT WIN32 AND USE_XCCL)
+if(NOT DEFINED USE_XCCL)
+  # Propagate the option to PyTorch
+  caffe2_update_option(USE_XCCL OFF)
+endif()
+
+if(USE_XCCL)
```
zhangxiaoli73, line 50: Should you have a linux check here?

CMakeLists.txt
```diff
@@ -40,9 +40,25 @@ include(${TORCH_XPU_OPS_ROOT}/cmake/ONEMKL.cmake)
 include(${TORCH_XPU_OPS_ROOT}/cmake/BuildFlags.cmake)
 
 option(USE_XCCL "Build with XCCL support" OFF)
+set(USE_C10D_XCCL OFF CACHE BOOL "Build with XCCL support for C10D")
 
-if(NOT WIN32 AND USE_XCCL)
+if(NOT DEFINED USE_XCCL)
+  # Propagate the option to PyTorch
+  caffe2_update_option(USE_XCCL OFF)
```
zhangxiaoli73, line 47: Please be aware that keeping `USE_XCCL` in `torch-xpu-ops` is a short-term implementation and we will upstream this option to stock PyTorch (should happen in PyTorch 2.8 release).

CMakeLists.txt
guangyey: Good catch!

CMakeLists.txt
guangyey: I move to the check to findxccl.cmake

EikanWang (approved):

PR #1379, opened 2025-02-19
Modify int4 kernel scale&zp layout

sunjiweiswift:
1 Modify the layout of scale and zeropt to row-major.
2 Add more groupsize


mingfeima (approved):
LGTM

PR #1393, opened 2025-02-21
Add float8 datatype to XPU OPs

yucai-intel:
kernels related:
_local_scalar_dense
foreach_tensor_copy
cat_xpu
where_xpu


src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
```diff
@@ -6,6 +6,65 @@
 
 #include <ATen/native/xpu/sycl/ForeachCopyKernels.h>
 
+#define AT_DISPATCH_SOURCE_TYPES(TYPE, NAME, ...)                                                \
+  AT_DISPATCH_SWITCH(                                                                            \
+      TYPE,                                                                                      \
+      NAME,                                                                                      \
+      AT_PRIVATE_CASE_TYPE_USING_HINT(                                                           \
```
dvrogozh, line 13: Why so strange indent below? If I find definition correctly it expands to the regular `case <id>: {}`. So each item can be placed just below the previous one.

Did you just copy from https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/ForeachBinaryOpList.cu#L288? I suggest to fix the indent here for sycl kernel.

https://github.com/pytorch/pytorch/blob/7ce4974e50fa68e7be3caafb35208505328dedc9/aten/src/ATen/Dispatch.h#L64

```
#define AT_PRIVATE_CASE_TYPE_USING_HINT(enum_type, HINT, ...)                 \
  case enum_type: {                                                           \
    AT_PRIVATE_CHECK_SELECTIVE_BUILD(enum_type);                              \
    using HINT [[maybe_unused]] = c10::impl::ScalarTypeToCPPTypeT<enum_type>; \
    return __VA_ARGS__();                                                     \
  }
```

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
```diff
@@ -17,23 +76,49 @@ struct Identity {
...
+                  UnaryOpFunctor<
+                      scalar_t,
+                      /* depth */ 2,
+                      /* r_args_depth */ 1,
+                      /* res_arg_index */ 1>(),
+                  Identity<opmath_t>());
+            });
       });
+
+  // AT_DISPATCH_ALL_TYPES_AND_COMPLEX_AND3(
```
dvrogozh, line 104: what's the reason to have this commented out? Either remove or add in-source comment explaining why this might be needed.

dvrogozh (changes requested):
> kernels related: _local_scalar_dense foreach_tensor_copy cat_xpu where_xpu

@yucai-intel : what this comment tries to say? Can you, please, update PR description highlighting which XPU ops will be available for FP8 datatype after merging this PR? and whether there are any ops which won't be available with FP8 and we should expect over PRs to bring support for them?

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
yucai-intel: Hi, since our code format is using clang-format, this kind of indentation will be generated.

yucai-intel: Hi, @dvrogozh thank you for your kind suggestion. We hope you know that we choose to list the kernel names instead of the corresponding ops because these names can help us immediately locate the datatype dispatch part when searching.

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
```diff
@@ -6,6 +6,65 @@
...
+                                                              at::ScalarType::                   \
+                                                                  Float8_e5m2,                   \
+                                                              src_t,                             \
+                                                              __VA_ARGS__)                       \
+                                                              AT_PRIVATE_CASE_TYPE_USING_HINT(   \
+                                                                  at::ScalarType::               \
+                                                                      Float8_e5m2fnuz,           \
+                                                                  src_t,                         \
+                                                                  __VA_ARGS__))
+
```
dvrogozh, line 67: clang-format is a good tool, but that's not right to sacrifice human readability for the tool which can't work in certain cases. And it's known that clang-format sometimes incorrectly formats some cases. Ours being one of them.

I think in this particular sense it makes sense to disable clang-format over this code block. This will render better readable code which is always an ultimate goal of using a tool such as clang-format.

```suggestion
// clang-format off
#define AT_DISPATCH_SOURCE_TYPES(TYPE, NAME, ...)                 \
  AT_DISPATCH_SWITCH(                                             \
      TYPE,                                                       \
      NAME,                                                       \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Byte, src_t, __VA_ARGS__)               \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Char, src_t, __VA_ARGS__)               \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Long, src_t, __VA_ARGS__)               \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Short, src_t, __VA_ARGS__)              \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Int, src_t, __VA_ARGS__)                \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Double, src_t, __VA_ARGS__)             \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Float, src_t, __VA_ARGS__)              \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::ComplexDouble, src_t, __VA_ARGS__)      \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::ComplexFloat, src_t, __VA_ARGS__)       \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Half, src_t, __VA_ARGS__)               \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::BFloat16, src_t, __VA_ARGS__)           \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Bool, src_t, __VA_ARGS__)               \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Float8_e4m3fn, src_t, __VA_ARGS__)      \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Float8_e4m3fnuz, src_t, __VA_ARGS__)    \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Float8_e5m2, src_t, __VA_ARGS__)        \
      AT_PRIVATE_CASE_TYPE_USING_HINT(                            \
          at::ScalarType::Float8_e5m2fnuz, src_t, __VA_ARGS__))
// clang-format on
```

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
dvrogozh: I closed this comment thread to be able to suggest a code change for the whole block. See that other comment.

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
yucai-intel: Hi, @dvrogozh , thank you for your detailed suggestions.
One point worth noting is that although clang-format's processing may not seem very readable, it is reasonable because it clearly prompts readers that these parallel functions are parameters rather than complete sentences. Maybe this is a bit confusing for you, who is an experienced developer, but for ordinary developers, such prompts are logical.

In addition, a kind reminder is that if clang-format has some inappropriate places, you can propose improvements to the overall code format review method instead of modifying individual cases. Because a single and personalized modification will cause trouble to other collaborators and reviewers, resulting in unnecessary duplication of work. After all, clear rules are more efficient than constraints.
Thank you again for your suggestions.

yucai-intel:
image

PR #1407, opened 2025-02-25
Change `batch_norm` vectorization policy

xytintel:
We don’t need to vectorize FP32 currently.


xytintel:
> What's the motivation? Why does FP32 not require vectorization?

The optimized batch normalization has an accuracy issue in CosmicTagger's long train convergence (FP32), and the cause is still unclear. We will disable it for now.

chunhuanMeng (approved):

fengyuan14:
> > What's the motivation? Why does FP32 not require vectorization?
> 
> The optimized batch normalization has an accuracy issue in CosmicTagger's long train convergence (FP32), and the cause is still unclear. We will disable it for now.

Are you sure of if the changes leaded to the accuracy regression?

PR #1416, opened 2025-02-27
Dispatch NestedTensorXPU LayerNormKernel to xpu::layer_norm_kernel

min-jean-cho:
per title.


PenghuiCheng (approved):

src/ATen/native/xpu/LayerNorm.cpp
```diff
@@ -177,6 +177,8 @@ ::std::tuple<at::Tensor, at::Tensor, at::Tensor> layer_norm_backward_xpu(
       grad_bias,
       grad_input_mask);
 }
+
+REGISTER_XPU_DISPATCH(LayerNormKernel, &xpu::layer_norm_kernel);
```
xytintel, line 181: Where is `LayerNormBackwardKernel`?

min-jean-cho: `xpu::layer_norm_backward_kernel` needs to be refactored to align with CUDA `LayerNormBackwardKernelImpl`.

PR #1425, opened 2025-03-03
Remove roi_align autocast test due to vision change

frost-intel:
Current XPU test of autocast for `roi_align` op in torchvision is failing. Recent changes in upstream torchvision change the behavior of this op for XPU when `requires_grad` is true. I have been unable to reproduce the test error locally, but will continue attempts to reproduce and fix. In the meantime, I suggest disabling the test in torch-xpu-ops for now.


daisyden (approved):
LGTM

PR #1421, opened 2025-02-28
Refine XCCL build option

guangyey:
# Motivation
Prior to this PR, building PyTorch XPU with CCL distribution support required setting `USE_XCCL=ON USE_C10D_XCCL=ON python setup.py install`.
With this PR, it only requires `USE_XCCL=ON python setup.py install`, simplifying the process.

PR #1430, opened 2025-03-05
[CI]Change USE_XCCL from 1 to ON

zxd1997066:


zxd1997066:
Hi @Chao1Han , we found that USE_XCCL only workable when setting "ON", setting "1" does not work, is it reasonable ?

zxd1997066:
https://github.com/intel/torch-xpu-ops/pull/1421 merged, it can build with USE_XCCL=1 now, close this pr

PR #1434, opened 2025-03-05
group start after comm init

Chao1Han:


src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -587,13 +600,15 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::send(
           at::xpu::XPUStream& stream,
```
zhangxiaoli73, line 600: should you change it to oneCCL stream?

Chao1Han: xpu stream use for device synchronize, I think it not need change to oneCCL stream

PR #1467, opened 2025-03-14
register xpu logger reducer

Chao1Han:
Distribute logger supported by torch/csrc/distributed/c10d/logger.cpp, and op compute time measure need device event support. Register XpuTimer like https://github.com/pytorch/pytorch/blob/main/torch/csrc/distributed/c10d/reducer_cuda.cpp.


src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -267,16 +269,54 @@ bool ProcessGroupXCCL::WorkXCCL::wait(std::chrono::milliseconds timeout) {
   return true;
 }
 
+static std::atomic<size_t> process_group_id = 0;
```
zhangxiaoli73, line 272: what's this for? Is it `group_name` in python frontend?

src/xccl/ProcessGroupXCCL.hpp
```diff
@@ -380,6 +387,15 @@ class TORCH_API ProcessGroupXCCL : public Backend {
 } // namespace c10d
 
 namespace {
+
+std::string getXcclVersion() {
+  auto xccl_version = ccl::get_library_version();
```
zhangxiaoli73, line 392: Please make sure if this API is supported in oneCCL new C API.

src/xccl/ProcessGroupXCCL.cpp
Chao1Han: Like a group number counter, use for track local group.

src/xccl/ProcessGroupXCCL.hpp
Chao1Han: Not support now, new C API have other version api.

Chao1Han:
unrelated fail
test_unary_ufuncs_xpu.py::TestUnaryUfuncsXPU::test_contig_vs_transposed_exp_xpu_complex128

PR #1476, opened 2025-03-17
Add benchmark for operator level performance track

xytintel:


xytintel:
Move to https://github.com/intel/torch-xpu-ops/pull/1484

PR #1490, opened 2025-03-20
Avoid invoking `-fsycl-targets=spir64_gen,spir64` when `aot` is set to `none`

jingxu10:
Since `-fsycl-targets=spir64_gen,spir64` has to work with `-device`, moving the `-fsycl-targets` to conditions as well.


chunhuanMeng (approved):

jingxu10:
The failed CI is not related to this PR.

PR #1501, opened 2025-03-24
Fix TorchDynamo Cache Issue

chunhuanMeng:
Resolved failures in `test_autocast` cases caused by TorchDynamo's compilation cache limit. When deterministic mode is enabled, Python-implemented ops are compiled using TorchDynamo instead of their C++ counterparts. Added `torch._dynamo.reset()` to clear the cache and allow recompilation.

Reference: [torchvision/ops/roi_align.py#L256](https://github.com/pytorch/vision/blob/8ea4772e97bc11b2cfee48a415e7df8cd17fa682/torchvision/ops/roi_align.py#L256)


frost-intel (approved):
LGTM.

PR #1516, opened 2025-03-27
softmax: enable half_to_float

weishi-deng:
To enable half_to_float flag in softmax.

### DIFF
--- a/src/ATen/native/xpu/sycl/SoftMaxKernels.cpp
+++ b/src/ATen/native/xpu/sycl/SoftMaxKernels.cpp
@@ -203,7 +203,8 @@ template <
     int INNER_LOOP,
     int vec_size,
     int SIMD,
-    typename scalar_t,
+    typename inscalar_t,
+    typename outscalar_t,
     typename accscalar_t,
     typename IndexType,
     bool LogSoftMax,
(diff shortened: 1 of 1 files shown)


xytintel:
Is there any test case?

weishi-deng:
> Is there any test case?

UT updated.

xytintel (approved):

test/regressions/test_softmax.py
```diff
@@ -0,0 +1,38 @@
...
+                x = torch.randn(shape[i]).to(input_type)
+                grad = torch.randn(shape[i]).to(output_type)
+                x_cpu = x.clone().requires_grad_()
+                y_cpu = F.softmax(x_cpu, dim, dtype=output_type)
+                y_cpu.backward(grad.clone())
+
+                x_xpu = x.clone().to(xpu_device).requires_grad_()
+                y_xpu = F.softmax(x_xpu, dim, dtype=output_type)
+                y_xpu.backward(grad.clone().to(xpu_device))
+                self.assertEqual(y_cpu, y_xpu.cpu())
```
guangyey, line 37: I have a question, without https://github.com/pytorch/pytorch/pull/150278, why could this UT pass.

weishi-deng: > I have a question, without [pytorch/pytorch#150278](https://github.com/pytorch/pytorch/pull/150278), why could this UT pass.

I think this ut doesn't work as pytorch do not dispatch to this path. It can pass my local test. So the next step is to update the torch-xpu-ops commit and test this ut in pytorch, I think.

guangyey: Make sense. Thanks.

PR #1522, opened 2025-03-31
Modify CI/Nightly XPU build

mengfei25:
1. remove nightly check in source build
2. update LKG torch only in nightly


.github/workflows/nightly_ondemand_rolling.yml
```diff
@@ -81,6 +81,7 @@ jobs:
       python: ${{ github.event_name == 'schedule' && '3.10' || inputs.python }}
       driver: rolling
       runner: pvc_rolling
+      update_lkg: lkg
```
chuanqi129, line 84: Use true or false?

chuanqi129 (approved):

PR #1539, opened 2025-04-02
Add retry mechanism for file open to handle `PermissionError`

chunhuanMeng:
This PR addresses `PermissionError` issues during codegen caused by multiple processes writing to the same file. It introduces a retry mechanism for file opening operations, ensuring the script waits and retries when encountering a PermissionError, thus improving the robustness of the build process.


xytintel:
Move to https://github.com/intel/torch-xpu-ops/pull/1541

PR #1541, opened 2025-04-02
Improve torch-xpu-ops codegen

guangyey:
# Motivation
Improve code to avoid `PermissionError` on Windows:
Previously, the script n `install_xpu_headers.py` opened the file twice - once for reading and once for writing:
```python
for file in files:
    with open(file) as fr:
        # modify fr
    with open(file, "w") as fw:
        # write fw
```
However, on Windows, this can lead to `PermissionError` due to file access conflicts. It looks like a bug related to the Windows system or Python. I am not sure. 
To avoid this issue, we now open the file only once using `"r+"` mode, allowing both reading and writing within the same context:
```python
for file in files:
    with open(file, "r+") as f:
        # modify f
        # rewrite f
```
Additionally, add dependencies for torch-xpu-ops codegen to ensure proper execution.

# Solution
Trying to prevent writing back an unchanged file. And open the file only once to allow both reading and writing.

# Additional Context
Windows CI build pass refer to https://github.com/pytorch/pytorch/actions/runs/14236812185/job/39897594304?pr=139971

### DIFF
--- a/cmake/Codegen.cmake
+++ b/cmake/Codegen.cmake
@@ -106,6 +106,8 @@ function(GEN_XPU file_yaml)
     ${CODEGEN_XPU_YAML_DIR}/native/${file_yaml}
     ${XPUFallback_TEMPLATE}
     ${TORCH_XPU_OPS_ROOT}/tools/codegen/install_xpu_headers.py
+    ${ops_generated_headers}
+    ${ops_generated_sources}
   )
 
   # Post codegen delete the copied templates folder only on Windows.
--- a/tools/codegen/install_xpu_headers.py
+++ b/tools/codegen/install_xpu_headers.py
@@ -97,18 +97,22 @@ def append_xpu_ops_headers(src_dir, dst_dir, common_headers, xpu_ops_headers):
                 re.findall(r"struct TORCH_XPU_API.*xpu.*?{.*?};\n", src_text, re.DOTALL)
             )
 
-        with open(dst) as fr:
-            dst_lines = fr.readlines()
+        if not xpu_declarations:
+            continue
+
+        with open(dst, "r+") as f:
+            dst_lines = f.readlines()
             dst_text = "".join(dst_lines)
-            for line in dst_lines:
+            for index, line in enumerate(dst_lines):
                 if re.match(r"^(TORCH_API.*;|struct TORCH_API.*)", line):
                     for xpu_declaration in xpu_declarations:
                         if not re.search(re.escape(xpu_declaration), dst_text):
-                            dst_lines.insert(dst_lines.index(line), xpu_declaration)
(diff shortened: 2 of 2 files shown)


cmake/Codegen.cmake
```diff
@@ -106,6 +106,8 @@ function(GEN_XPU file_yaml)
     ${CODEGEN_XPU_YAML_DIR}/native/${file_yaml}
     ${XPUFallback_TEMPLATE}
     ${TORCH_XPU_OPS_ROOT}/tools/codegen/install_xpu_headers.py
+    ${ops_generated_headers}
+    ${ops_generated_sources}
```
EikanWang, line 110: @guangyey , as we discussed, have you concluded why the cmake dependency does not solve the multiple process contention?

tools/codegen/install_xpu_headers.py
```diff
@@ -97,18 +97,22 @@ def append_xpu_ops_headers(src_dir, dst_dir, common_headers, xpu_ops_headers):
...
                 if re.match(r"^(TORCH_API.*;|struct TORCH_API.*)", line):
                     for xpu_declaration in xpu_declarations:
                         if not re.search(re.escape(xpu_declaration), dst_text):
-                            dst_lines.insert(dst_lines.index(line), xpu_declaration)
+                            dst_lines.insert(index, xpu_declaration)
                     break
 
-        with open(dst, "w") as fw:
-            fw.writelines(dst_lines)
+            f.seek(0)
```
EikanWang, line 113: `r+` means that the stream is positioned at the beginning of the file. `f.seek(0)` required?

cmake/Codegen.cmake
guangyey: I suspect the contention isn't coming from the ATen codegen itself. This part should behave as expected — the issue likely originates elsewhere.

EikanWang (approved):

EikanWang:
In general, I do not think the current solution is a perfect solution as we have not figured out the root cause. Hence, Pls. file a GitHub issue to track it. @guangyey , @chunhuanMeng

guangyey:
> In general, I do not think the current solution is a perfect solution as we have not figured out the root cause. Hence, Pls. file a GitHub issue to track it. @guangyey , @chunhuanMeng

I completely agree. This is a workaround to mitigate the issue for now, but it might occur again in the future. I don't have the bandwidth to investigate the root cause at the moment. @chunhuanMeng Could you help file an issue to track it and consider trying FileLock as a potential solution?

guangyey:
Let's land this PR to avoid blocking the Windows CI build.

chunhuanMeng:
https://github.com/intel/torch-xpu-ops/issues/1554

PR #1542, opened 2025-04-03
Remove abi=0 tests

mengfei25:


.github/workflows/_linux_build.yml
```diff
@@ -90,15 +85,11 @@ jobs:
           source .github/scripts/env.sh ${{ inputs.pytorch }}
           export USE_ONEMKL=1
           export USE_XCCL=1
+          export _GLIBCXX_USE_CXX11_ABI=1
```
chuanqi129, line 88: ABI=1 is by default, we can remove this line

chuanqi129 (commented):
Please also trigger an on-demand test to prove the change doesn't impact the workflow functionality

mengfei25:
Passed with nightly launch https://github.com/intel/torch-xpu-ops/actions/runs/14235749589

PR #1560, opened 2025-04-09
[Nightly] Modify UT tests scope check and single case timeout

mengfei25:
1. UT test scope
2. UT timeout for each single case


mengfei25:
Failures are not related with this change

chuanqi129 (approved):
Overall LGTM. Minor comment, can we define a unify parameter of timeout for all places use.   The advantage is we just modify one place in the future if we need change the timeout

PR #1579, opened 2025-04-16
[wip] merge oneccl v1&v2 api

Chao1Han:


src/xccl/xccl.cpp
```diff
@@ -0,0 +1,440 @@
+#include "xccl/xccl.h"
+
+namespace c10d {
+namespace xccl {
+
+void oneccl_v2_group_start() {
```
zhangxiaoli73: No need to mention V2 in API name.

src/xccl/xccl.cpp
```diff
@@ -0,0 +1,440 @@
...
+  if (isCCLV2EnabledCached()) {
+    onecclGroupStart();
+  } else {
+    ccl::group_start();
+  }
+}
+
+void oneccl_v2_group_end() {
```
zhangxiaoli73: No need to mention V2 in API name.

src/xccl/ProcessGroupXCCL.hpp
```diff
@@ -6,13 +6,14 @@
...
 
-#include <oneapi/ccl.hpp>
+// #include <oneapi/ccl.hpp>
 #include <exception>
 #include <future>
 #include <list>
 #include <mutex>
 #include <unordered_map>
 #include <vector>
+#include "xccl/xccl.h"
```
guangyey:
```suggestion
#include <xccl/xccl.h>
```

src/xccl/ProcessGroupXCCL.hpp
```diff
@@ -6,13 +6,14 @@
 #define CCL_ENABLE_ZE
 #define CCL_ENABLE_SYCL
 
-#include <oneapi/ccl.hpp>
+// #include <oneapi/ccl.hpp>
```
guangyey:
```suggestion
```
Remove it if it is not necessary any more.

src/xccl/xccl.cpp
```diff
@@ -0,0 +1,447 @@
+#include "xccl/xccl.h"
```
guangyey:
```suggestion
#include <xccl/xccl.h>
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,478 @@
...
+  }
+  return versionString;
+}();
+
+namespace c10d {
+
+struct XCCLStream {
+  at::xpu::XPUStream xpuStream;
+  ccl::stream cclStream;
+  void* syclQueue;
```
guangyey:
```suggestion
```
I think `syclQueue` is a redundant member.

guangyey: It always can be fetched from `xpuStream`

Chao1Han: You’re right — previously, converting an XPU stream to a CCL stream incurred significant overhead, so it was stored in a map. Since `sycl::queue` is essentially an alias for the XPU stream, I’ll make the necessary modification.

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+namespace c10d {
+
+struct XCCLStream {
+  at::xpu::XPUStream xpuStream;
+  ccl::stream cclStream;
+};
+
+struct xcclComm_t {
+  std::optional<ccl::communicator> cclComm;
+  onecclComm_t onecclComm;
```
guangyey:
```suggestion
  onecclComm_t onecclComm{nullptr};
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+
+  xcclComm_t() : onecclComm(nullptr) {}
```
guangyey:
```suggestion
  xcclComm_t() = default;
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+  xcclComm_t(ccl::communicator comm)
```
guangyey:
```suggestion
  explicit xcclComm_t(ccl::communicator comm)
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+      : cclComm(std::move(comm)), onecclComm(nullptr) {}
+  xcclComm_t(onecclComm_t comm) : onecclComm(comm) {}
```
guangyey:
```suggestion
  explicit xcclComm_t(onecclComm_t comm) : onecclComm(comm) {}
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+            return std::tolower(c);
+          });
+      return value == "on" || value == "yes" || value == "1";
+    }
+    return false;
+  }();
+  return cachedValue;
+}
+
+static std::string versionString = []() {
```
guangyey:
```suggestion
inline const std::string& getVersionString() {
static std::string versionString = []() {

...
return versionString;
```
let's init it lazily.

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+    {ReduceOp::MIN, ccl::reduction::min},
+    {ReduceOp::MAX, ccl::reduction::max},
+    {ReduceOp::SUM, ccl::reduction::sum},
+    {ReduceOp::PRODUCT, ccl::reduction::prod},
+#ifdef XCCL_HAS_AVG
+    {ReduceOp::AVG, ccl::reduction::avg},
+#endif // XCCL_HAS_AVG
+};
+
+const std::map<at::ScalarType, onecclDataType_t> xcclDatatypesV2 = {
```
guangyey:
```suggestion
inline const std::map<at::ScalarType, onecclDataType_t> xcclDatatypesV2 = {
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+    {at::kBFloat16, onecclDataType_t::onecclBfloat16},
+    {at::kBool, onecclDataType_t::onecclUint8},
+    // use for non-reduction op like allgather
+    {at::kFloat8_e5m2, onecclDataType_t::onecclUint8},
+    {at::kFloat8_e4m3fn, onecclDataType_t::onecclUint8},
+    {at::kFloat8_e4m3fnuz, onecclDataType_t::onecclUint8},
+    {at::kFloat8_e5m2fnuz, onecclDataType_t::onecclUint8},
+};
+
+const std::map<at::ScalarType, ccl::datatype> xcclDatatypesV1 = {
```
guangyey:
```suggestion
inline const std::map<at::ScalarType, ccl::datatype> xcclDatatypesV1 = {
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+    // use for non-reduction op like allgather
+    {at::kFloat8_e5m2, ccl::datatype::uint8},
+    {at::kFloat8_e4m3fn, ccl::datatype::uint8},
+    {at::kFloat8_e4m3fnuz, ccl::datatype::uint8},
+    {at::kFloat8_e5m2fnuz, ccl::datatype::uint8},
+};
+
+namespace {
+
+ccl::datatype getXcclDataTypeV1(
```
guangyey:
```suggestion
inline ccl::datatype getXcclDataTypeV1(
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+  TORCH_CHECK_WITH(
+      TypeError,
+      it != xcclDatatypesV1.end(),
+      "Input tensor data type is not supported for XCCL process group: ",
+      type);
+  return it->second;
+}
+
+// V2 specific function to avoid variant overhead
+onecclDataType_t getXcclDataTypeV2(
```
guangyey:
```suggestion
inline onecclDataType_t getXcclDataTypeV2(
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+  auto it = xcclDatatypesV2.find(type);
+  TORCH_CHECK_WITH(
+      TypeError,
+      it != xcclDatatypesV2.end(),
+      "Input tensor data type is not supported for XCCL process group: ",
+      type);
+  return it->second;
+}
+
+ccl::reduction getXcclReduceOpV1(const ReduceOp& reduceOp, at::Tensor& input) {
```
guangyey:
```suggestion
inline ccl::reduction getXcclReduceOpV1(const ReduceOp& reduceOp, at::Tensor& input) {
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+#endif
+    return xcclOpsV1.at(reduceOp);
+  } catch (const std::out_of_range&) {
+    C10_THROW_ERROR(
+        ValueError,
+        "Cannot use ReduceOp." + reduceOpToString(reduceOp) + " with XCCL");
+  }
+}
+
+onecclRedOp_t getXcclReduceOpV2(const ReduceOp& reduceOp, at::Tensor& input) {
```
guangyey:
```suggestion
inline onecclRedOp_t getXcclReduceOpV2(const ReduceOp& reduceOp, at::Tensor& input) {
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+              "] is setting up XCCL communicator and "
+              "retrieving xcclUniqueId from [0] via c10d key-value store by key '",
+              storeKey,
+              "'",
+              ". This may indicate a possible application crash on rank 0 or a network set up issue."));
+    }
+  }
+}
+
+std::shared_ptr<xcclComm_t> createXCCLCommHelper(
```
guangyey:
```suggestion
inline std::shared_ptr<xcclComm_t> createXCCLCommHelper(
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+    int rank,
+    int numRanks,
+    int rank_,
+    c10::Device& device,
```
guangyey, line 348: unused parameter?

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+    }
+    ccl::kvs::address_type main_addr;
+    std::copy_n(
+        ccl_kvs_addr.begin(), ccl::kvs::address_max_size, main_addr.begin());
+    kvs = ccl::create_kvs(main_addr);
+  }
+  return kvs;
+}
+
+void broadcastUniqueXCCLID(
```
guangyey:
```suggestion
inline void broadcastUniqueXCCLID(
```

src/xccl/xccl.h
```diff
@@ -0,0 +1,477 @@
...
+#endif
+    return xcclOpsV2.at(reduceOp);
+  } catch (const std::out_of_range&) {
+    C10_THROW_ERROR(
+        ValueError,
+        "Cannot use ReduceOp." + reduceOpToString(reduceOp) + " with XCCL");
+  }
+}
+
+std::mutex kvs_mutex;
```
guangyey: Here is definitely what you don't expect! The original code `kvs_mutex` is a member of `ProcessGroupXCCL`, it ensures that get_kvc is mutex within each instance. But now, the behavior changed.

src/xccl/xccl.h
Chao1Han: Not quite — the device is actually used in `onecclSetDevice(device.index())`. In most cases, `device.index` equals `deviceKey`, but they differ in single P2P scenarios.

src/xccl/xccl.h
Chao1Han: Good hint, let me restore original change.

src/xccl/xccl.h
```diff
@@ -0,0 +1,491 @@
...
+  }
+}
+
+} // namespace
+
+namespace xccl {
+
+void oneccl_group_start();
+void oneccl_group_end();
+void onecclAllReduce(
```
guangyey, line 406: Is it better to use `XCCLStream` as the input parameter, anyway it is OK to me.

cmake/Modules/FindXCCL.cmake
```diff
@@ -49,7 +49,14 @@ find_library(
   NO_DEFAULT_PATH
 )
 
-if((NOT XCCL_INCLUDE_DIR) OR (NOT XCCL_LIBRARY_DIR) OR (NOT XCCL_LIBRARY))
+find_file(
+  XCCL_LIBRARY_2_0
```
zhangxiaoli73, line 61: I think we don't need to find this .2 so, right?

src/xccl/xccl.h
```diff
@@ -0,0 +1,491 @@
...
+    case c10d::ReduceOp::AVG:
+      return "AVG";
+    case c10d::ReduceOp::PREMUL_SUM:
+      return "PREMUL_SUM";
+    default:
+      return "UNKNOWN";
+  }
+}
+
+inline bool isCCLV2EnabledCached() {
```
zhangxiaoli73, line 54: What do you mean for "cache"?

cmake/Modules/FindXCCL.cmake
Chao1Han: No, still standalone so file in 2021.17

src/xccl/xccl.h
```diff
@@ -0,0 +1,491 @@
+#pragma once
+
+#define CCL_ENABLE_ZE
```
zhangxiaoli73, line 3: why do you need ZE enabling?

src/xccl/xccl.h
Chao1Han: That means once this function is called, it becomes effectively static — it won’t perform the environment variable lookup every time it’s invoked.

src/xccl/xccl.h
```diff
@@ -0,0 +1,491 @@
...
+    }
+    return false;
+  }();
+  return cachedValue;
+}
+
+inline const std::string& getVersionString() {
+  static std::string versionString = []() {
+    bool useCCLV2 = isCCLV2EnabledCached();
+    std::string versionString;
```
zhangxiaoli73, line 73: I think this version means public release version like 2021.15. Then why do you have to care about V2 or not?

src/xccl/xccl.h
Chao1Han: You should recall this legacy issue. when compiling CCL-related code with **g++** instead of **icpx**, these two defines are required.

src/xccl/xccl.h
```diff
@@ -0,0 +1,491 @@
...
+      : cclComm(std::move(comm)), onecclComm(nullptr) {}
+  explicit xcclComm_t(onecclComm_t comm) : onecclComm(comm) {}
+};
+
+const std::map<c10d::ReduceOp, onecclRedOp_t> xcclOpsV2 = {
+    {ReduceOp::MIN, onecclRedOp_t::onecclMin},
+    {ReduceOp::MAX, onecclRedOp_t::onecclMax},
+    {ReduceOp::SUM, onecclRedOp_t::onecclSum},
+    {ReduceOp::PRODUCT, onecclRedOp_t::onecclProd},
+#ifdef XCCL_HAS_AVG
```
zhangxiaoli73: V2 always support "avg", no need to do check micro check.

src/xccl/xccl.h
Chao1Han: For easier future code migration, in any case, it behaves the same as **isCCLV2EnabledCached**, which is invoked only once.

src/xccl/xccl.h
Chao1Han: ok

src/xccl/xccl.h
```diff
@@ -0,0 +1,491 @@
...
+
+inline onecclRedOp_t getXcclReduceOpV2(
+    const ReduceOp& reduceOp,
+    at::Tensor& input) {
+  try {
+    if (input.scalar_type() == at::kBool) {
+      if (reduceOp == ReduceOp::SUM) {
+        return onecclRedOp_t::onecclMax;
+      }
+#ifdef XCCL_HAS_AVG
```
zhangxiaoli73: Let us remove those redundant micro check. Keep code clean for V2.

src/xccl/xccl.h
```diff
@@ -0,0 +1,491 @@
...
+        *store_,
+        xcclCommCounter_,
+        kvs_mutex,
+        singleP2POp,
+        deviceKey,
+        p2pRank);
+    auto comms = ccl::create_communicators(numRanks, devs_rank, ctx, xccl_kvs);
+    return std::make_shared<xcclComm_t>(std::move(comms[0]));
+  } else {
+    LOG(INFO) << "USE_CCL_V2=1";
```
zhangxiaoli73: Add more comments when using v2.

src/xccl/xccl.h
zhangxiaoli73: @Chao1Han Please refine input argument.

Chao1Han: done

src/xccl/xccl.h
Chao1Han: done

src/xccl/xccl.h
Chao1Han: done

PR #1580, opened 2025-04-16
[BE] Move torch xpu wheel build in manylinux_2_28 container both for CI/CD

mengfei25:
1. build xpu wheel in container manylinux_2_28
2. split build steps to standalone script
3. build wheel post op

### DIFF
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -33,7 +33,7 @@ cp -r ${GITHUB_WORKSPACE}/torch-xpu-ops third_party/torch-xpu-ops
 sed -i "s/checkout --quiet \${TORCH_XPU_OPS_COMMIT}/log -n 1/g" caffe2/CMakeLists.txt
 
 # oneAPI DLE
-source ${GITHUB_WORKSPACE}/torch-xpu-ops/.github/scripts/env.sh
+source third_party/torch-xpu-ops/.github/scripts/env.sh
 icpx --version
 
 # Pre Build
@@ -57,7 +57,7 @@ WERROR=1 python setup.py bdist_wheel
 # Post Build
 python -m pip install patchelf
 rm -rf ./tmp
-bash ${GITHUB_WORKSPACE}/torch-xpu-ops/.github/scripts/rpath.sh ${GITHUB_WORKSPACE}/pytorch/dist/torch*.whl
+bash third_party/torch-xpu-ops/.github/scripts/rpath.sh ${GITHUB_WORKSPACE}/pytorch/dist/torch*.whl
 python -m pip install --force-reinstall tmp/torch*.whl
 
 # Verify
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -11,7 +11,7 @@ TORCH_XPU_OPS_VERSION=${TORCH_XPU_OPS_VERSION:-"main"}
 for var; do
     eval "export $(echo ${var@Q} |sed "s/^'-*//g;s/=/='/")"
 done
-exit
+
 # Set pytorch
 rm -rf ${WORKSPACE}/pytorch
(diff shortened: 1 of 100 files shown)


mengfei25:
The failures are not related with this change

.github/workflows/_linux_build.yml
```diff
@@ -44,25 +44,37 @@ permissions:
 jobs:
   build:
     runs-on: ${{ inputs.runner }}
+    container:
+      image: 'pytorch/manylinux2_28-builder:xpu-main'
+      volumes:
+        - ${{ github.workspace }}:${{ github.workspace }}
+      env:
+        PATH: /opt/xpu-build/bin:/usr/share/Modules/bin:/opt/rh/gcc-toolset-11/root/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
chuanqi129, line 52: why we need this line?

.github/workflows/_linux_build.yml
```diff
@@ -44,25 +44,37 @@ permissions:
...
       - name: Prepare Stock Pytorch
         run: |
           pwd
-          which conda && conda clean -ay
-          conda remove --all -y -n xpu_build || \
-                rm -rf $(dirname ${CONDA_EXE})/../envs/xpu_build
-          conda create -n xpu_build python=${{ inputs.python }} cmake=3.28 ninja -y
-          source activate xpu_build
+          which python && python -V && pip list
+          pip install -U pip wheel setuptools
```
chuanqi129, line 77: Let's move those 2 lines into last step

chuanqi129 (commented):
Overall LGTM, will we have another PR to add rpath change?

.github/workflows/_linux_build.yml
```diff
@@ -44,25 +44,37 @@ permissions:
 jobs:
   build:
     runs-on: ${{ inputs.runner }}
+    container:
+      image: 'pytorch/manylinux2_28-builder:xpu-main'
+      volumes:
+        - ${{ github.workspace }}:${{ github.workspace }}
```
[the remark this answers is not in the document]
mengfei25: Yes, it will mount to /__w/torch-xpu-ops but not same as the host path, I mount it to avoid hard code path issue, of course we can remove it if need

.github/workflows/_linux_build.yml
mengfei25: in order to remove conda env

.github/workflows/_linux_build.yml
```diff
@@ -44,68 +44,61 @@ permissions:
...
           set -xe
-          source activate xpu_build
-          source .github/scripts/env.sh ${{ inputs.pytorch }}
-          export USE_ONEMKL=1
-          export USE_XCCL=1
-          if [[ ${{ inputs.driver }} == 'lts' ]]; then
-            export TORCH_XPU_ARCH_LIST='pvc'
-          fi
-          pip install mkl-static==2025.0.1 mkl-include==2025.0.1
+          build_status="$(
```
chuanqi129, line 82: It has a little bit over-design in here I think, how do we deal with build for PR? It seems will always use the latest main rather than the PR source branch/ merge code if we let the keep-torch-xpu-ops as false.

mengfei25: split the build steps to script, it will help the next bisect search for build, and for PR it will use the merged code in the build.sh at "# Set torch-xpu-ops", if set false, it will copy the checkout merged code.

.github/scripts/build.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+set -xe
+export GIT_PAGER=cat
+
```
chuanqi129, line 4: Please add usage in here

chuanqi129 (approved):

PR #1600, opened 2025-04-23
[Nightly/CI] Enable performance comparison

mengfei25:


.github/scripts/perf_comparison.py
```diff
@@ -0,0 +1,111 @@
...
+import re
+import os
+import fnmatch
+import argparse
+import pandas as pd
+from statistics import geometric_mean
+
+parser = argparse.ArgumentParser(description="Analysis", formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument("-xpu", default=None, help="XPU file")
+parser.add_argument("-refer", default=None, help="refer file")
```
chuanqi129, line 11: Same here for those parameters

.github/scripts/perf_comparison.py
```diff
@@ -0,0 +1,111 @@
+
+import re
+import os
+import fnmatch
+import argparse
+import pandas as pd
+from statistics import geometric_mean
+
```
chuanqi129, line 8: Better to add usage comments here

.github/scripts/calculate_best_perf.py
```diff
@@ -0,0 +1,93 @@
+
+import re
+import os
+import fnmatch
+import argparse
+import pandas as pd
+from datetime import date
+
```
chuanqi129, line 8: Better to add usage comments here

Issue #1163, opened 2024-12-12
torch._standard_gamma() has accuracy gap compared to scipy and torch.cpu


daisyden:
### 🐛 Describe the bug

test_distributions_xpu.py::TestDistributionsXPU::test_gamma_gpu_sample_xpu FAILED                                                                               [100%]

============================================================================== FAILURES ===============================================================================
___________________________________________________________ TestDistributionsXPU.test_gamma_gpu_sample_xpu ____________________________________________________________
Traceback (most recent call last):
  File "/home/gta/daisyden/rebase/n1_ci/third_party/torch-xpu-ops/test/xpu/test_distributions_xpu.py", line 119, in _test_gamma_gpu_sample
    self._check_sampler_sampler(
  File "/home/gta/daisyden/rebase/n1_ci/third_party/torch-xpu-ops/test/xpu/../../../../test/distributions/test_distributions.py", line 1289, in _check_sampler_sampler
    self.assertLess(-threshold, bias, message)
  File "/home/gta/miniforge3/envs/n2/lib/python3.10/unittest/case.py", line 1232, in assertLess
    self.fail(self._formatMessage(msg, standardMsg))
  File "/home/gta/miniforge3/envs/n2/lib/python3.10/unittest/case.py", line 675, in fail
    raise self.failureException(msg)
AssertionError: -0.06743384440339613 not less than -0.259 : Gamma(alpha=1.0, beta=0.1).sample() is biased:
[-0.259 -0.264 -0.243 -0.059  1.     0.951 -0.279 -0.274 -0.296 -0.277]

To execute this test, run the following from the base repo dir:
    python test/distributions/test_distributions.py TestDistributionsXPU.test_gamma_gpu_sample_xpu

### Versions

with 2024/12/09 nightly wheel, no such issue on preci.

daisyden:
Tried cpu _standard_gamma tensor, no such issue.

PenghuiCheng:
fixed：https://github.com/intel/torch-xpu-ops/pull/1161

daisyden:
Seems this issue exists or not on different driver versions, investigation is WIP.

PR #1608, opened 2025-04-24
Fix gamma accuracy gap compared to scipy

xytintel:
Resolve https://github.com/intel/torch-xpu-ops/issues/1163. This PR aims to resolve the gamma accuracy gap compared to scipy by adjusting two key parameters in the XPU distributions. **Motivation:** Misaligned key parameters can cause computation differences with CUDA, which may lead to failures in CUDA bias cases. We have decided to follow these parameters.


EikanWang:
@xytintel , I cannot quite understand why the changes can fix the issue https://github.com/intel/torch-xpu-ops/issues/1163. You need to provide an informative and detailed description.

EikanWang (changes requested):
Pls. provide an informative and detailed description.

EikanWang (approved):
LGTM. By the way, @xytintel , should we always comment on the magic number? For example, why do we choose 512 as `max_threads_per_block`? Is it always optimal for different GPU Archs?

xytintel:
> LGTM. By the way, @xytintel , should we always comment on the magic number? For example, why do we choose 512 as `max_threads_per_block`? Is it always optimal for different GPU Archs?
It is recommended to use the same parameters for the random generator as CUDA, because these parameters can affect the generated sequence.

PR #1619, opened 2025-04-28
Don't use c10:string_view

cyyever:
`c10::string_view` is an alias of `std::string_view` and we are planning to deprecate it.


guangyey (approved):
@cyyever Thanks.

guangyey:
lint is green and build pass.

PR #1622, opened 2025-04-28
[Nightly] Add op performance regression check

RUIJIEZHONG66166:
For op performance track, add the regression check funtion into the OP nightly performance test.


RUIJIEZHONG66166:
Ondemand Test: https://github.com/intel/torch-xpu-ops/actions/runs/14929100483
Regression cases and improvement cases are shown on the summary. The threshold can be set by parameter.

RUIJIEZHONG66166:
Save baseline artifact number into a github issue: https://github.com/intel/torch-xpu-ops/issues/1689.
Ondemand test action job is https://github.com/intel/torch-xpu-ops/actions/runs/15337311965
Previously artifact number is:
image
After ondemand test finished, it changed to the ondemand action job number:
image

.github/workflows/_linux_op_benchmark.yml
```diff
@@ -121,9 +120,53 @@ jobs:
           bash ${{ github.workspace }}/.github/scripts/microbench_summary.sh ${{ github.workspace }}/op_benchmark ${{ github.workspace }}/op_benchmark/forward_op_summary.csv
           # Summary backward op time
           bash ${{ github.workspace }}/.github/scripts/microbench_summary.sh ${{ github.workspace }}/op_benchmark ${{ github.workspace }}/op_benchmark/backward_op_summary.csv True
+      - name: Download OP Baseline
```
chuanqi129, line 123: Suggest to split below part as a standalone job and run with Github official runner

chuanqi129 (approved):
Overall, LGTM. Is the on-demand test demo https://github.com/intel/torch-xpu-ops/actions/runs/15337311965 still available? Do we need to rerun a new demo? 
Base on the report of current demo, give below comments:

- [ ] The regression and improvement table sort with ratio -> case name -> op name -> dt, if possible
- [ ] I saw there are some 99% drop cases in the regression table, are those real issue? Similar for improvement table, there are some >>100% improvement.
- [ ] Current table has too many column, can we concat some op config in one column or other better format
- [ ] I saw the tables are duplicate in the summary, please double check it

.github/workflows/_linux_op_benchmark.yml
RUIJIEZHONG66166: Got it. This on-demand test is still available.
Later, I will raise a enhancement PR to do these:
1. Concat the op config in one column with sort of ratio
2. The tables are not duplicate because it covered forward and backward(Shown in "op_name" column).  I will update the summary title to clarify it.
3. Split the check part as a standalone job and run with Github official runner

PR #1629, opened 2025-04-30
Replace c10::guts::apply with std::apply 

cyyever:
c10::guts::apply is an alias of std::apply


EikanWang (approved):

guangyey (approved):
Thanks.

guangyey:
lint is green and build pass.

PR #1633, opened 2025-05-06
[Nightly] To fix 'requests.get too many requests'

mengfei25:
To address this issue, consider the following approaches: 
1. Add auth to use larger request limitation
2. Use git to get diff instead of Github API

### DIFF
--- a/.github/scripts/apply_torch_pr.py
+++ b/.github/scripts/apply_torch_pr.py
@@ -5,6 +5,7 @@ import urllib
 import subprocess
 import sys
 import time
+import os
 
 
 parser = argparse.ArgumentParser()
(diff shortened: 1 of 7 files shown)


mengfei25:
UT failures are not related with this change

chuanqi129 (approved):

EikanWang:
@chuanqi129 , why does this PR cannot pass the CI?

chuanqi129:
> @chuanqi129 , why does this PR cannot pass the CI?

@EikanWang Because there are some new UT failures, and those failures are not related to this PR. At the same time, this PR is needed for others PR CI test to unblock git rate limit problem

PR #1638, opened 2025-05-07
skip gather cases to avoid core dump

daisyden:


xytintel:
Move to https://github.com/intel/torch-xpu-ops/pull/1643

PR #1642, opened 2025-05-07
Rename USE_ONEMKL as USE_ONEMKL_XPU and set it as default ON

CuiYifeng:
- USE_ONEMKL (for oneMKL XPU) may be confused with USE_MKL (for oneMKL CPU). New name USE_ONEMKL_XPU provides better readability.
- Set USE_ONEMKL_XPU as default ON.


CuiYifeng:
@chuanqi129 @fengyuan14 Please help review, thanks.

.github/scripts/build.sh
```diff
@@ -50,7 +50,7 @@ git submodule sync && git submodule update --init --recursive
 python -m pip install -r requirements.txt
 python -m pip install mkl-static mkl-include
 export USE_STATIC_MKL=1
-export USE_ONEMKL=1
+export USE_ONEMKL_XPU=1
```
chuanqi129, line 53: Is there any simple API can check whether the oneMKL is compiled and enabled? If have we can add it in the `verify` part of this script

CuiYifeng: These is no such API available. Checking if `libtorch_xpu.so => libmkl_sycl_*.so` is an alternative method.

.github/workflows/_windows_ut.yml
```diff
@@ -114,7 +114,7 @@ jobs:
           pip install cmake setuptools==72.1.0 clang-format
           pip install mkl-static==2025.0.1 mkl-include==2025.0.1
           set USE_KINETO=0
-          set USE_ONEMKL=1
+          set USE_ONEMKL_XPU=1
```
fengyuan14, line 117: Please remove it if it is default on by design.

CuiYifeng: Done.

chuanqi129 (approved):

PR #1671, opened 2025-05-15
Enhance Debug Build Configuration with Conditional Compiler Options

chunhuanMeng:
This pull request introduces a small but important update to the `cmake/BuildFlags.cmake` file, enhancing the build configuration for debugging. Specifically, it adds debug-specific compiler options when the build type is set to `Debug`.

### Debug build enhancements:
* Added a conditional block to include `-g` for generating debug information and `-Rno-debug-disables-optimization` to suppress warnings about debug information potentially disabling optimizations when `CMAKE_BUILD_TYPE` is set to `Debug`.


cmake/BuildFlags.cmake
```diff
@@ -84,6 +84,12 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "MSVC"
     set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -Wno-absolute-value)
     set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -no-ftz)
   endif()
+
+  if(CMAKE_BUILD_TYPE MATCHES Debug)
+    set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -g)
+    set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -Rno-debug-disables-optimization)
```
jianyizh, line 90: why we use `-Rno-debug-disables-optimization` on kernel option but `-O0` in host option? we should keep them the same. https://github.com/intel/torch-xpu-ops/blob/main/cmake/BuildFlags.cmake#L47

xytintel (approved):

PR #1677, opened 2025-05-16
[CI] Enable matrix test for UT

RUIJIEZHONG66166:
Enable matrix test method to accelerate the UT test.


.github/workflows/_linux_ut.yml
```diff
@@ -51,17 +51,158 @@ jobs:
...
         run: |
+          export CONDA_PKGS_DIRS="${{ runner.temp }}/conda_pkgs_${{ github.run_id }}"
+          mkdir -p "$CONDA_PKGS_DIRS"
           pwd
           which conda && conda clean -ay
-          conda remove --all -y -n xpu_op_${ZE_AFFINITY_MASK} || \
-                rm -rf $(dirname ${CONDA_EXE})/../envs/xpu_op_${ZE_AFFINITY_MASK}
-          conda create -n xpu_op_${ZE_AFFINITY_MASK} python=${{ inputs.python }} cmake ninja -y
-          source activate xpu_op_${ZE_AFFINITY_MASK}
+          conda remove --all -y -n $CONDA_ENV_NAME || \
```
dvrogozh, line 202: Command on a previous line, `conda clean -ay` is a big problem if you have a few runners running on the same system. This effectively means that you can damage environment you create/work in one runner from another and get all sorts of fancy random failures.

.github/workflows/_linux_ut.yml
```diff
@@ -51,17 +51,158 @@ jobs:
...
+            env: {}
+            additional_steps: |
+              mkdir -p $LOG_DIR/profile_test/issue_reproduce
     steps:
       - name: Checkout torch-xpu-ops
         uses: actions/checkout@v4
+      - name: Create unique workspace
+        run: |
+          # Create unique conda env for each UT test
+          echo "CONDA_ENV_NAME=xpu_op_${ZE_AFFINITY_MASK}_${{ matrix.test.name }}" >> $GITHUB_ENV
```
dvrogozh, line 195: That's not truly unique actually. Meaning that if there are 2 of these workloads running in parallel, for example for different PRs, these will start to fail. To address this you can reuse approach from #1282. It will work only with Linux though. For Windows you will need something else:
```suggestion
          random=$(head /dev/urandom | tr -dc A-Za-z0-9_ | head -c ${1:-5} | xargs)
          echo "CONDA_ENV_NAME=xpu_op_${ZE_AFFINITY_MASK}_${{ matrix.test.name }}_${random}" >> $GITHUB_ENV
```

RUIJIEZHONG66166: Got it. Thank you so much for your help~~

.github/workflows/_linux_ut.yml
RUIJIEZHONG66166: Got it. Thank you~~

.github/workflows/_linux_ut.yml
```diff
@@ -51,17 +51,153 @@ jobs:
       NEOReadDebugKeys: ${{ inputs.driver == 'rolling' && '1' || '0' }}
       DisableScratchPages: ${{ inputs.driver == 'rolling' && '1' || '0' }}
       ut_skip_issue: 1624
+    strategy:
+      fail-fast: false
+      matrix:
+        test:
+          - name: 'op_regression'
```
chuanqi129, line 63: Let's add runner as a config of the matrix

.github/workflows/_linux_ut.yml
```diff
@@ -51,17 +51,153 @@ jobs:
...
+            condition: ${{ contains(inputs.ut, 'op_regression') }}
+            directory: 'test/regressions'
+            command: 'pytest --timeout 600 -v --junit-xml=../../ut_log/op_regression.xml'
+            log_prefix: 'op_regression'
+            timeout: 8000
+            additional_steps: |
+              pip install pytest pytest-timeout
+          - name: 'op_regression_dev1'
```
chuanqi129, line 72: So that this part can dispatch to multi-cards runner type

.github/workflows/_linux_ut.yml
```diff
@@ -51,17 +51,153 @@ jobs:
...
+              cp op_ut_with_only.xml $GITHUB_WORKSPACE/ut_log
+            additional_steps: |
+              pip install pytest pytest-timeout
+          - name: 'torch_xpu'
+            condition: ${{ contains(inputs.ut, 'torch_xpu') }}
+            directory: '../pytorch'
+            command_script: |
+              export PYTORCH_TEST_WITH_SLOW=1
+              export PYTORCH_TESTING_DEVICE_ONLY_FOR="xpu"
+              for xpu_case in build/bin/*{xpu,sycl}*; do
```
chuanqi129: let's remove this part binary tests, because we split pytorch build as a standalone job, and we don't artifact other binaries  together.

.github/workflows/_linux_ut.yml
```diff
@@ -51,17 +51,153 @@ jobs:
...
+                if [[ "$xpu_case" != *"*"* && "$xpu_case" != *.so && "$xpu_case" != *.a ]]; then
+                  case_name=$(basename "$xpu_case")
+                  echo "Testing ${case_name} ..."
+                  "$xpu_case" --gtest_output=xml:"$TEST_REPORTS_DIR"/"$case_name".xml
+                fi
+              done
+              test_cmd="python test/run_test.py --include "
+              for test in $(ls test/inductor | grep test); do test_cmd="${test_cmd} inductor/$test"; done
```
chuanqi129, line 144: Please consider how to further parallel when the inductor ut enabled in this part. Can we leverage shard mechanism in stock pytorch directly?

.github/workflows/_linux_ut.yml
```diff
@@ -51,17 +51,153 @@ jobs:
...
+          echo "CONDA_ENV_NAME=xpu_op_${ZE_AFFINITY_MASK}_${{ matrix.test.name }}_${random}" >> $GITHUB_ENV
       - name: Prepare Stock Pytorch
         run: |
           pwd
-          which conda && conda clean -ay
-          conda remove --all -y -n xpu_op_${ZE_AFFINITY_MASK} || \
-                rm -rf $(dirname ${CONDA_EXE})/../envs/xpu_op_${ZE_AFFINITY_MASK}
-          conda create -n xpu_op_${ZE_AFFINITY_MASK} python=${{ inputs.python }} cmake ninja -y
-          source activate xpu_op_${ZE_AFFINITY_MASK}
+          which conda
```
chuanqi129, line 198: Consider to use setup-python https://github.com/actions/setup-python instead of conda
cc @mengfei25

.github/workflows/_linux_ut.yml
RUIJIEZHONG66166: Done

.github/workflows/_linux_ut.yml
RUIJIEZHONG66166: Done. Dispatched to pytorch-06.

.github/workflows/_linux_ut.yml
RUIJIEZHONG66166: Done

.github/workflows/_linux_ut.yml
```diff
@@ -48,25 +48,156 @@ permissions: read-all
...
+          - name: 'op_regression_dev1'
+            condition: ${{ contains(inputs.ut, 'op_regression_dev1') }}
+            directory: 'test/regressions'
+            command: 'pytest --timeout 600 -v test_operation_on_device_1.py --junit-xml=$GITHUB_WORKSPACE/ut_log/op_regression_dev1.xml'
+            log_prefix: 'op_regression_dev1'
+            timeout: 8000
+            additional_steps: |
+              unset ZE_AFFINITY_MASK
+              pip install pytest pytest-timeout
+            runner: 'pytorch-06'
```
chuanqi129: please use label, for example `pvc_e2e`

RUIJIEZHONG66166: Got it. Done

chuanqi129 (approved):

PR #1680, opened 2025-05-20
[CI] Enhance distributed test

mengfei25:
1. Return error once failed contains timeout via bash flags
2. Always set and reset ptrace_scope before and after tests


mengfei25:
Distributed test timeout is expected.

chuanqi129 (approved):

PR #1688, opened 2025-05-21
Optimize work_group reduce for GroupNorm Forward

yucai-pro:
Optimized the calculation of data on shared memory in GroupReduceWithoutBroadcast, so that the performance of GroupNorm Forward when shape=(1024, 128, 32, 32) is improved by 20+% (latency 2700+ms->~2000ms).
image


toyxu (approved):

xytintel:
> @xytintel , @yucai-intel , may I know which workload drives such an optimization?

1. Subgroup reduce for reducing computation
image

2. Use more fine-grained paths to allow small workloads to go directly into subgroup reduce.
image

PR #1696, opened 2025-05-23
Support half precision compute for histc kernel

xytintel:
We decided to keep the data type consistent with the CPU for the `histc` op.

### DIFF
--- a/src/ATen/native/xpu/sycl/SummaryOpsKernels.cpp
+++ b/src/ATen/native/xpu/sycl/SummaryOpsKernels.cpp
@@ -238,14 +238,12 @@ Tensor _histc_kernel(
     int64_t nbins,
     const Scalar& min,
     const Scalar& max) {
-  if (self.scalar_type() == ScalarType::Half) {
-    AT_ERROR("HalfTensor is not supported");
-  }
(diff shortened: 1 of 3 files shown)


src/ATen/native/xpu/sycl/SummaryOpsKernels.cpp
```diff
@@ -238,14 +238,12 @@ Tensor _histc_kernel(
...
-  if (self.scalar_type() == ScalarType::Half) {
-    AT_ERROR("HalfTensor is not supported");
-  }
-  return AT_DISPATCH_ALL_TYPES(self.scalar_type(), "_histc_xpu", [&] {
-    using bounds_t = at::acc_type_device<scalar_t, kXPU>;
-    return _histc_template<scalar_t>(
-        self, nbins, min.to<bounds_t>(), max.to<bounds_t>());
-  });
+  return AT_DISPATCH_ALL_TYPES_AND2(
+      kHalf, kBFloat16, self.scalar_type(), "_histc_xpu", [&] {
```
ZhiweiYan-96, line 242: maybe add bflo16 support is a little aggressive? currently no usage herd of

xytintel: BF16 is just another version of half-precision type format. Is cpu support bf16?

PR #1697, opened 2025-05-23
Temp fix for bug in scale-out profiling by adding env variable to disable profiler around allreduce

frost-intel:
Currently, a bug in PTI causes indefinite hang in all_reduce call when attempting to profile code on multiple nodes. If the upstream Kineto enables the toggle option for XPU PTI (https://github.com/pytorch/kineto/pull/1088), then this PR would allow users to profile distributed scale-out code.

This PR uses an env variable to disable profiling around the call to XPU allreduce. An additional change in pytorch would be necessary to enable `toggleCollectionDynamic` for XPU.

Ideally, a fix can be implemented in a new PTI version 0.12.3 or later before 2.8 deadline. If not, this fix would be required by customers.

@ashokei

### DIFF
--- a/src/xccl/Register.cpp
+++ b/src/xccl/Register.cpp
@@ -2,7 +2,12 @@
 #include <c10/util/intrusive_ptr.h>
 #include <torch/csrc/distributed/c10d/ProcessGroup.hpp>
 #include <torch/csrc/distributed/c10d/Types.hpp>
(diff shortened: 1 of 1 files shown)


frost-intel:
New oneAPI 2025.1.3 fixes the hang without needing to disable profiling around allreduce.

PR #1700, opened 2025-05-26
[CI] Remove using last known good torch commit for build

mengfei25:


mengfei25:
Distributed test failed is expected.

PR #1707, opened 2025-05-28
Modify extra PRs diff get method to gh

mengfei25:
Use gh cli to get PRs diff


chuanqi129 (approved):
Because we run those gh cli in the workflow, so we don't need provide github token specifically for it, is it true?

mengfei25:
setup via command line `gh auth login`

PR #1716, opened 2025-06-04
Add llama case for profile test

RUIJIEZHONG66166:
Add llama inference test for call number test when profiling


RUIJIEZHONG66166:
Ondemand test results: https://github.com/intel/torch-xpu-ops/actions/runs/16411411201/job/46370358970
Llama needs HF log in.

PR #1720, opened 2025-06-04
Enhanced fft_r2c overhead and reactivate fft_r2c on XPU

CuiYifeng:
# Motivation
Revert `fft_r2c` fallback after root causing the `TestInductorOpInfoXPU` failure. There is a memory layout mismatching between `aten::fft_r2c` and Inductor meta deducing. We will have another PR to correct Inductor meta deducing for XPU backend.

# Solution
Inconsistent strides should be resolved in `fft_r2c` meta calculation instead of `fft_r2c` XPU kernel. The change of this PR includes:
- Reduced `_fft_r2c_mkl` overhead and simplified kernel code
- Reactivate `fft_r2c` on XPU

### DIFF
--- a/src/ATen/native/xpu/mkl/SpectralOps.cpp
+++ b/src/ATen/native/xpu/mkl/SpectralOps.cpp
@@ -499,7 +499,6 @@ Tensor _fft_r2c_mkl(
 
   IntArrayRef out_sizes = onesided ? onesided_sizes : input_sizes;
 
-  auto sorted_dims = impl::_sort_dims(self, dim, /*exclude_last=*/true);
   auto out = at::empty(
       out_sizes, self.options().dtype(c10::toComplexType(self.scalar_type())));
 
@@ -515,9 +514,11 @@ Tensor _fft_r2c_mkl(
         self.options().dtype(c10::toComplexType(self.scalar_type())));
   }
 
-  sorted_dims.resize(sorted_dims.size() - 1);
+  DimVector sorted_dims(dim.begin(), dim.end() - 1);
 
   while (!sorted_dims.empty()) {
+    sorted_dims = impl::_sort_dims(self, sorted_dims);
+
     std::swap(out, working_tensor);
 
     const auto max_dims =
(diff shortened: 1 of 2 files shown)


fengyuan14:
So far, even XPU output's strides are different with CPU's, we have no evidence on CPU impl is correct. There is no definition in https://docs.pytorch.org/docs/stable/generated/torch.fft.irfft2.html#torch.fft.irfft2. We should not compromise here with performance regression.
My suggestion in the current stage is overriding the case.

CuiYifeng:
@fengyuan14 @chunhuanMeng This PR is redefined to optimize `fft_r2c` overhead and reactivate it on XPU. Below is the comparison between optimized `fft_r2c` and baseline on SPR 56c + Max 1550:
| Shape         | Ratio (%) |
|---------------|------------------|
| (5, 6, 7)     | 114.66%          |
| (16, 16)      | 132.38%          |
| (32, 32)      | 129.42%          |
| (32, 64)      | 123.62%          |
| (64, 64)      | 119.50%          |
| (128, 128)    | 101.52%          |
| (128, 512)    | 111.42%          |
| (512, 512)    | 106.37%          |
| (1024, 1024)  | 111.14%          |
| (2048, 2048)  | 105.01%          |
| (4096, 4096)  | 105.83%          |
| (512, 768, 16)| 116.72%          |

src/ATen/native/xpu/mkl/SpectralOps.cpp
```diff
@@ -560,7 +549,7 @@ Tensor _fft_r2c_mkl(
     at::native::_fft_fill_with_conjugate_symmetry_(out, dim);
   }
 
-  return impl::_fft_apply_normalization(out, normalization, input_sizes, dim);
+  return out;
```
fengyuan14, line 552:
```
out = std::move(working_tensor);
```
Should not override the original output tensor, as sometimes the output tensor is created by caller.

CuiYifeng: In function `_fft_r2c_mkl`, `Tensor out` is created by itself. Output tensor created by caller is only for `_fft_r2c_mkl_out`.

fengyuan14 (approved):

xytintel (approved):

PR #1723, opened 2025-06-06
[Nightly] Enhance XPU test workflows

mengfei25:
1. remove source oneapi in microbench to use pypi packages
2. fix windows test workflow issue in nightly
3. modify torchbench installation to reduce reinstalling torch
4. extended xpu ops ut timeout to 5 hours

### DIFF
--- a/.github/workflows/_linux_op_benchmark.yml
+++ b/.github/workflows/_linux_op_benchmark.yml
@@ -81,7 +81,6 @@ jobs:
       - name: Install Pytorch XPU
         run: |
           source activate xpu_op_${ZE_AFFINITY_MASK}
-          source .github/scripts/env.sh ${{ inputs.pytorch }}
           if [ "${{ inputs.pytorch }}" != "nightly_wheel" ]; then
             cd ../pytorch
             export CMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}:${CONDA_PREFIX:-"$(dirname $(which conda))/../"}
@@ -96,7 +95,6 @@ jobs:
       - name: Torch Config
         run: |
           source activate xpu_op_${ZE_AFFINITY_MASK}
-          source .github/scripts/env.sh ${{ inputs.pytorch }}
           python -c "import torch; print(torch.__config__.show())"
           python -c "import torch; print(torch.__config__.parallel_info())"
           python -c "import torch; print(torch.__config__.torch.xpu.device_count())"
@@ -108,7 +106,6 @@ jobs:
       - name: Run Torch XPU Op Benchmark
         if: ${{ inputs.driver == 'rolling' }} 
         run: |
-          source .github/scripts/env.sh ${{ inputs.pytorch }}
           source activate xpu_op_${ZE_AFFINITY_MASK}
           mkdir -p ${{ github.workspace }}/op_benchmark
           cd test/microbench
--- a/.github/actions/inductor-xpu-e2e-test/action.yml
+++ b/.github/actions/inductor-xpu-e2e-test/action.yml
@@ -53,31 +53,39 @@ runs:
         source activate e2e_ci
         if [[ ${{ inputs.suite }} == *"torchbench"* ]]; then
           if [ "${{ inputs.pytorch }}" != "nightly_wheel" ]; then
+            pip uninstall torchaudio torchvision -y
             cd ../ && rm -rf audio && git clone --single-branch -b main https://github.com/pytorch/audio.git
             cd audio && git checkout $TORCHAUDIO_COMMIT_ID
-            python setup.py bdist_wheel && pip uninstall torchaudio -y && pip install dist/*.whl
+            python setup.py install
             cd ../ && rm -rf vision && git clone --single-branch -b main https://github.com/pytorch/vision.git
             cd vision && git checkout $TORCHVISION_COMMIT_ID
-            python setup.py bdist_wheel && pip uninstall torchvision -y && pip install dist/*.whl
+            python setup.py install
           fi
           cd ../ && python -c "import torch, torchvision, torchaudio"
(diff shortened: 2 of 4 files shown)


.github/scripts/build.sh
```diff
@@ -85,6 +85,17 @@ rm -rf ./tmp
 bash third_party/torch-xpu-ops/.github/scripts/rpath.sh ${WORKSPACE}/pytorch/dist/torch*.whl
 python -m pip install --force-reinstall tmp/torch*.whl
 
+# Build torchvision torchaudio
```
chuanqi129, line 88: Please use a env var or parameter control this behavior, by default off unless the var/parameter has been set to on. 

Including triton build also. For pinned triton commit, we can use `make triton` directly under pytorch root dir. For customized triton commit, we can build by ourselves or leverage the scripts directly https://github.com/pytorch/pytorch/blob/main/.github/scripts/build_triton_wheel.py and refer usage https://github.com/chuanqi129/pytorch/blob/fix_triton_version_split/.github/workflows/build-triton-wheel.yml#L158-L160. Before this step, we need replace the pined triton xpu commit file content to customized one. 

cc: @RUIJIEZHONG66166

.github/workflows/_linux_accelerate.yml
```diff
@@ -60,6 +60,9 @@ jobs:
       HF_HUB_ETAG_TIMEOUT: 120
       HF_HUB_DOWNLOAD_TIMEOUT: 120
     steps:
+      - name: Cleanup Workspace
```
chuanqi129: Why we need this change? If this test failed, it should caused by other jobs don't clean workspace at the end of the workflow

.github/workflows/_linux_op_benchmark.yml
```diff
@@ -45,70 +45,35 @@ jobs:
       NEOReadDebugKeys: ${{ inputs.driver == 'rolling' && '1' || '0' }}
       DisableScratchPages: ${{ inputs.driver == 'rolling' && '1' || '0' }}
     steps:
-      - name: Checkout torch-xpu-ops
-        uses: actions/checkout@v4
-      - name: Prepare Stock Pytorch
+      - name: Cleanup Workspace
```
chuanqi129: Same here. And suggest to use a github action or define a github action directly for the workspace cleanup, such as https://github.com/pytorch/pytorch/blob/main/.github/workflows/_xpu-test.yml#L338 or https://github.com/pytorch/pytorch/blob/main/.github/workflows/_xpu-test.yml#L83

.github/scripts/setup_test_env.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+set -xe -o pipefail
```
chuanqi129, line 3: I feel like this script is over design somehow. Add some suggestions

- Try to define new github actions in stead of bash script directly, they both can reuse in different workflows.
-  Split this script as multiple parts. For example, source code preparation and requirements setup, env information collection and display etc. There is a good example https://github.com/intel/torch-xpu-ops/tree/main/.github/actions/print-environment
- Can we remove conda dependency?

.github/workflows/_linux_op_benchmark.yml
```diff
@@ -45,70 +45,35 @@ jobs:
...
-            pip install --force-reinstall ${{ github.workspace }}/torch*.whl
-            git clone https://github.com/pytorch/vision && cd vision && python setup.py install && cd ..
+          set -xe -o pipefail
+          PYTORCH_REPO="https://github.com/pytorch/pytorch.git"
+          if [ "${{ inputs.pytorch }}" == "nightly_wheel" ]; then
+            PYTORCH_VERSION="nightly"
           else
-            pip install torch torchvision torchaudio --pre --index-url https://download.pytorch.org/whl/nightly/xpu
-            TORCH_COMMIT_ID=$(python -c 'import torch; print(torch.version.git_version)')
+            PYTORCH_VERSION="main"
```
chuanqi129: Seem the `main` has ambiguity here, if we don't test main branch from source code build

.github/workflows/_linux_op_benchmark.yml
```diff
@@ -45,70 +45,35 @@ jobs:
...
-          source .github/scripts/env.sh ${{ inputs.pytorch }}
-          python -c "import torch; print(torch.__config__.show())"
-          python -c "import torch; print(torch.__config__.parallel_info())"
-          python -c "import torch; print(torch.__config__.torch.xpu.device_count())"
-
-          cd ..
-          python pytorch/torch/utils/collect_env.py
-          rm -rf /tmp/torchinductor_*
-          rm -rf ~/.triton/cache
+          ./.github/scripts/setup_test_env.sh \
```
chuanqi129: Commented in the scripts file directly

.github/scripts/setup_test_env.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# Prepare pytorch
+if [ "${PYTORCH_VERSION}" == "release" ];then
+    python -m pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/xpu
+elif [ "${PYTORCH_VERSION}" == "test" ];then
+    python -m pip install torch torchvision torchaudio --pre --index-url https://download.pytorch.org/whl/test/xpu
+elif [ "${PYTORCH_VERSION}" == "nightly" ];then
+    python -m pip install torch torchvision torchaudio --pre --index-url https://download.pytorch.org/whl/nightly/xpu
+else
+    python -m pip install ${WORKSPACE}/torch*.whl
```
chuanqi129, line 32: Will this step include torchvision and torchaudio installation?

mengfei25: Yes

PR #1738, opened 2025-06-13
[CI] Remove unnecessary token-permissions for workflow

mengfei25:
To address security risk in workflows 

disable_windows


chuanqi129 (approved):
Please launch an on-demand test to check whether it can add new comments on the issue

PR #1756, opened 2025-06-19
Add an option for NAN check for xccl

Chao1Han:
Refer from https://github.com/pytorch/pytorch/pull/125726, https://github.com/pytorch/pytorch/pull/135414. 
Add nan check for xccl.
why we need to stop communication from spreading NaNs?
"technically if we can be sure which rank (or, even which host) detected the first nan, then its OK to let the nan spread to some other hosts. but in practice i dont know if we have good enough way to align our logs on different hosts, so if we let the nan spread to a few other hosts we may lose track of which one was first”


zhangxiaoli73:
@Chao1Han Do all collectives have nan check by default?

src/xccl/NanCheck_XPU.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+  using Kernel = checkForNaN<T>;
+  auto kfn = Kernel(tensor.data_ptr<T>(), tensor.numel());
+
+  sycl_kernel_submit(global_range, local_range, stream.queue(), kfn);
+}
+
+// CHECK if a Tensor contains NAN in any of its element
+void checkForNan(const at::Tensor& tensor, at::xpu::XPUStream& stream) {
+  AT_DISPATCH_FLOATING_TYPES_AND4(
```
zhangxiaoli73, line 203: Does it mean that FP16 has no nan check?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -338,6 +339,7 @@ ProcessGroupXCCL::ProcessGroupXCCL(
       local_id_(process_group_id++) {
   logPrefix_ = createLogPrefix();
   blockingWait_ = getCvarBool(TORCH_XCCL_BLOCKING_WAIT, false);
+  enableNanCheck_ = getCvarBool(TORCH_XCCL_NAN_CHECK, false);
```
zhangxiaoli73, line 342: Please attention that we may need to expose this environment variable in some pytorch documentation.

zhangxiaoli73 (approved):

PR #1759, opened 2025-06-20
[Nightly] Modify triton build method

mengfei25:
Use https://github.com/pytorch/pytorch/blob/main/.github/scripts/build_triton_wheel.py to build triton
disable_e2e
disable_ut


chuanqi129 (approved):
LGTM, next step, move the triton wheel build into build job

PR #1780, opened 2025-06-25
Use std::string_view

cyyever:
Follows #1619 and uses std::string_view


EikanWang (approved):

chuanqi129:
The E2E test result check failure is not related to the PR change

PR #1781, opened 2025-06-25
softmax: adjust vectorization length according to shape

weishi-deng:
To optimize the performance of Softmax with smaller shapes, this PR aims to adjust the vectorization size adaptively according to the input shape.

# Performance data (platform BMG )

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/wdeng/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/wdeng/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">

 dtype | shape(dim=0) | main | PR | ratio %
-- | -- | -- | -- | --
float | [64, 64] | 5.989 | 4.365 | 137%
float | [8192, 8192] | 2.833 | 2.818 | 101%
float | [64, 8192] | 21.838 | 21.802 | 100%
float | [8192, 64] | 197.458 | 121.979 | 162%
float | [1024, 1024] | 33.041 | 21.896 | 151%
float | [16384, 16384] | 13.342 | 13.445 | 99%
half | [64, 64] | 8.463 | 4.547 | 186%
half | [8192, 8192] | 1.346 | 1.35 | 100%
half | [64, 8192] | 26.718 | 26.776 | 100%
half | [8192, 64] | 312.901 | 131.197 | 238%
half | [1024, 1024] | 57.859 | 23.713 | 244%
half | [16384, 16384] | 5.982 | 5.974 | 100%

</body>

</html>

### DIFF
--- a/src/ATen/native/xpu/sycl/SoftMaxKernels.cpp
+++ b/src/ATen/native/xpu/sycl/SoftMaxKernels.cpp
@@ -199,20 +199,29 @@ static inline void get_wgroup_size_spatial(
   GroupRow = std::min(GroupRow, int(dim_size));
 }
 
-static bool inline can_use_vec(
+static bool inline get_spatial_vec_choice(
+    int bs,
     int dim_size,
+    int inner_size,
     int scalar_size,
-    int max_vec_size,
-    bool is_same_dtype) {
+    int max_vec_size) {
   if (dim_size % max_vec_size != 0)
     return false;
-  if (is_same_dtype) {
-    if (dim_size <= 2048 && dim_size * scalar_size <= 8192)
-      return false;
-  } else {
-    if (dim_size <= 1024 && dim_size * scalar_size <= 4096)
-      return false;
+  int total_resource = syclMaxWorkItemsPerTile();
+  int maxWGSize = syclDeviceMaxWorkGroupSize();
+  int group_col_size = std::min(inner_size, SIMD32);
+  auto local_group_num = (inner_size + group_col_size - 1) / group_col_size;
+  int group_row_size = 1;
+  while (bs *group_row_size * local_group_num * group_col_size <
+         total_resource ) {
+    group_row_size = group_row_size << 1;
+    if (group_row_size * SIMD32 == maxWGSize)
+      break;
   }
+  group_row_size = std::min(group_row_size, int(dim_size));
+  if(bs *group_row_size * local_group_num * group_col_size <=
+         total_resource)
+	  return false;
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/SoftMaxKernels.cpp
```diff
@@ -199,6 +199,23 @@ static inline void get_wgroup_size_spatial(
...
+    bool is_same_dtype) {
+  if (dim_size % max_vec_size != 0)
+    return false;
+  if (is_same_dtype) {
+    if (dim_size <= 2048 && dim_size * scalar_size <= 8192)
+      return false;
+  } else {
+    if (dim_size <= 1024 && dim_size * scalar_size <= 4096)
+      return false;
+  }
```
EikanWang: @weishi-deng , could you help elaborate on these magic numbers, including 2048, 1024, 4096, and 8192? Why are these numbers critical for the performance of `SoftMax`? Meanwhile, pls. consider whether the optimization applies to other platforms.

EikanWang:
@weishi-deng , pls. add platform information to the PR description.

src/ATen/native/xpu/sycl/SoftMaxKernels.cpp
weishi-deng: The heuristics and magic numbers are from the PyTorch implementation: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/SoftMax.cu#L1096 and https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/SoftMax.cu#L1187. As discussed last week with Liangang and Yutao, we prefer to add adaptive choices for our vectorization in SoftMax, and we prefer to align the implementation first. Furthermore, to decide which magic number we should set for all platforms, we need to collect more perf results on different shapes and platforms as they did in this issue. https://github.com/pytorch/pytorch/issues/144645

EikanWang: @weishi-deng, according to the description at https://github.com/pytorch/pytorch/issues/144645, the optimization aimed to reduce memory access by leveraging registers. Do you observe the same behavior on XPU?

EikanWang: If you believe that the optimization of https://github.com/pytorch/pytorch/issues/144645 is applicable for XPU as well, pls. collect data to prove it. We need to obtain insights regarding the optimization on XPU rather than the performance result.

PR #1798, opened 2025-07-01
update the tensor size of the test case based on the number of devices

newtdms:


test/xpu/distributed/test_c10d_ops_xccl.py
```diff
@@ -199,10 +199,12 @@ def allreduce(tensors, op):
     def test_alltoall_ops_with_xpufree_race(self):
         pg = self.pg
         opts = c10d.AllToAllOptions()
+        num_devices = torch.xpu.device_count()
+        tensor_size = num_devices * 100
         local_device = f"xpu:{self.rank_to_GPU[self.rank][0]}"
         torch.xpu.set_device(local_device)
-        input = torch.rand(1000, 1000, device=local_device)
-        output = torch.rand(1000, 1000, device=local_device)
+        input = torch.rand(tensor_size, 1000, device=local_device)
```
daisyden, line 206: For example we used 4 ranks to run weekly test, original code tensor_size is 1000, now it could change to 400, right? Shall we keep the tensor_size the same as before?

newtdms: We can keep the tensor size 1000 for 4 ranks and change per-rank size to 250. I can test that for 12 ranks.

newtdms: I updated the base and tensor size. Please let me know if you have more comments.

PR #1815, opened 2025-07-07
Skipped UT since it will hang in testing

PenghuiCheng:
It will hang when the unit test is running.
This should be pytorch issue.
https://github.com/pytorch/pytorch/issues/79528
Need to follow up on this issue.


chuanqi129:
@mengfei25 Let's create a torch-xpu-ops for them also if we don't have.

PR #1777, opened 2025-06-24
[CI] Set read-all for workflows top-level token-permissions

mengfei25:
disable_all

PR #1816, opened 2025-07-07
ci: revert breaking changes in accelerate and transformers tests

dvrogozh:
This commit reverts some changes introduced in d9b81b8b. The changes:
* Introduce some undocumented mechanism to enable/disable CI checks
* Did not account that workflow might be triggered per schedule

The changes can be reintroduced if 1) documented, 2) work properly.

Fixes: d9b81b8b ("[CI] Set read-all for workflows top-level token-permissions (#1777)")

### DIFF
--- a/.github/workflows/_linux_accelerate.yml
+++ b/.github/workflows/_linux_accelerate.yml
@@ -46,30 +46,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  conditions-filter:
-    name: conditions-filter
-    if: ${{ github.event.pull_request.draft == false }}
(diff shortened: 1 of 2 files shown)


chuanqi129:
@mengfei25 Please help to fix it rather than revert it

dvrogozh:
> @mengfei25 Please help to fix it rather than revert it

@chuanqi129 : I can't help to fix it as it's not clear to me what you are trying to introduce and how it is supposed to work. As I said - it's not documented. This needs to be handled by original author.

chuanqi129:
@dvrogozh sure, no worry. @mengfei25 will fix it

Issue #1764, opened 2025-06-20
New kernels for concat


jianyizh:
### 🚀 The feature, motivation and pitch

Follow cuda on CatArrayBatchedCopy_alignedK_contig and other possible kernels.
For this example, BMG takes 4.15ms, 4080 takes 0.7ms
```python
import torch
import time
input_list = [torch.randn((2048,512,7,7),dtype=torch.float16).to(memory_format=torch.channels_last).to(torch.accelerator.current_accelerator())]
for _ in range(15):
    input_list.append(torch.randn((2048,32,7,7),dtype=torch.float16).to(memory_format=torch.channels_last).to(torch.accelerator.current_accelerator()))
r = torch.cat(input_list,1)
torch.accelerator.synchronize()
start = time.time()
for _ in range(20):
    r = torch.cat(input_list,1)
torch.accelerator.synchronize()
end = time.time()
print((end-start)/20*1000)

```

### Additional context

Need implement
1. CatArrayBatchedCopy
2. CatArrayBatchedCopy_contig
3. CatArrayBatchedCopy_alignedK_contig

PR #1821, opened 2025-07-08
Add CatArrayBatchedCopy_contig relative kernels

yucai-intel:
https://github.com/intel/torch-xpu-ops/issues/1764

### DIFF
--- a/src/ATen/native/xpu/sycl/Shape.cpp
+++ b/src/ATen/native/xpu/sycl/Shape.cpp
@@ -1,34 +1,79 @@
+#include <ATen/Dispatch.h>
 #include <ATen/Dispatch_v2.h>
 #include <ATen/NumericUtils.h>
 #include <ATen/WrapDimUtils.h>
+#include <ATen/ceil_div.h>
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/Shape.cpp
```diff
@@ -325,96 +578,105 @@ void cat_out_kernel(
...
+          }),
+          AT_EXPAND(AT_ALL_TYPES_AND_COMPLEX),
+          kComplexHalf,
+          kHalf,
+          kBool,
+          kBFloat16,
+          kFloat8_e4m3fn,
+          kFloat8_e4m3fnuz,
+          kFloat8_e5m2,
+          kFloat8_e5m2fnuz,
```
CuiYifeng: `kFloat8_e8m0fnu` is not supported here?

CuiYifeng (commented):
Do we have any UT to ensure the correctness?

src/ATen/native/xpu/sycl/Shape.cpp
yucai-pro: Done.

PR #1829, opened 2025-07-10
[SYCL] Port to event free APIs for kernel submission

KepingYan:


gujinghui:
Should we add build assertion to check SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS first?

KepingYan:
> Should we add build assertion to check SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS first?

Do we need to keep the original code path? If SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS is defined and set to 1, run the modified code; otherwise, use the original code. Or just add an assertion?

KepingYan:
Gentle ping @fengyuan14 .

guangyey:
Hi @chuanqi129, could you help assign the right person to validate whether this change may introduce any unintended performance regressions?

src/comm/SYCLHelpers.h
```diff
@@ -3,6 +3,12 @@
 #include <comm/Scalar.h>
 #include <sycl/sycl.hpp>
 
+#if !defined(SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS) || \
+    (SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS != 1)
+#error This change enables event-less APIs, \
+  please ensure SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS=1.
```
gujinghui: `ifdef SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS` is good enough, as the only value is 1 when defined.

BTW, can we provide more readable error message? For example,
`FATAL ERROR: The SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS is excepted to be defined to enqueue kernel without returning event. But the undef is detected!`

guangyey (approved):
LGTM.
No performance gain or regression. Anyway, land it first.

src/comm/SYCLHelpers.h
```diff
@@ -80,10 +87,10 @@ sycl_kernel_submit(
     ker_t ker) {
   auto cgf = [&](::sycl::handler& cgh) {
     ker.sycl_ker_config_convention(cgh);
-    cgh.parallel_for<ker_t>(
-        ::sycl::nd_range<dim>(global_range, local_range), ker);
+    ::sycl::ext::oneapi::experimental::nd_launch<ker_t>(
+        cgh, ::sycl::nd_range<dim>(global_range, local_range), ker);
   };
-  q.submit(cgf);
+  ::sycl::ext::oneapi::experimental::submit(q, cgf);
```
vinser52, line 93: it is better to omit handler-based API if possible. In this particular case it is better to call `sycl::ext::oneapi::experimental::nd_launch`.

Here is how it could be done:

```cpp
::sycl::ext::oneapi::experimental::nd_launch<ker_t>(q, ::sycl::nd_range<dim>(global_range, local_range), ker);
```

We in SYCL RT are working to optimize cases where creation of `sycl::handler` can be omitted.

PR #1849, opened 2025-07-15
[Nightly] Enable bisect search

mengfei25:
disable_all

### DIFF
--- a/.github/scripts/check_llama_baseline.sh
+++ b/.github/scripts/check_llama_baseline.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+file1="$1"
+file2="$2"
+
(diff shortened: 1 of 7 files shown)


chuanqi129:
Suggest to rename to bisect search, in case we may also need to search good commit in the future

.github/workflows/_bisect_search.yml
chuanqi129, line 1: Rename to `bisect_search.yml` directly

chuanqi129 (approved):

.github/workflows/_bisect_search.yml
mengfei25: https://github.com/intel/torch-xpu-ops/pull/1849/commits/6b2559a374974100642972ecd60548664e61f6f3

PR #1867, opened 2025-07-17
Add FlightRecorder support for ProcessGroupXCCL

frost-intel:
This PR provides initial support for FlightRecorder, which allows debug trace dumps for distributed jobs. 

Features added:
1. Heartbeat Monitor thread which regularly checks if a dump signal has been received via pipe files, and on trigger writes traces to file (for each rank seperately)
2. Logic to record XCCL work events

Compared to NCCL, we don't have some features. These could be added in a later PR:
1. No support for event timing/duration
2. No support for Watchdog thread which allows for debug dump on error or timeout, with other additional features (remote error detection, etc)


src/xccl/ProcessGroupXCCL.hpp
```diff
@@ -105,7 +171,44 @@ class TORCH_API ProcessGroupXCCL : public Backend {
...
+
+    static c10::intrusive_ptr<Options> create() {
+      return c10::make_intrusive<Options>();
+    }
+
+    std::vector<uint64_t> global_ranks_in_group;
+    std::string group_name;
+  };
+
+  class HeartbeatMonitor {
```
zhangxiaoli73: Do you think if it could be separated in a new file instead of binding with XCCL process? Put  all codes in `ProcessGroupXCCL` to make files too large and hard to maintain in future.

zhangxiaoli73:
@frost-intel Could you please add a UT to cover this new feature?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -687,6 +815,7 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::pointToPoint(
     }
   }
 
+  op_id_++;
```
zhangxiaoli73, line 818: Why this id need to increase here? Do you mean that we missed to increase originally?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -353,9 +380,63 @@ ProcessGroupXCCL::ProcessGroupXCCL(
             << ", TORCH_XCCL_BLOCKING_WAIT: " << blockingWait_
             << ", TORCH_DISTRIBUTED_DEBUG: " << torch_distributed_debug
             << ", TORCH_XCCL_NAN_CHECK: " << enableNanCheck_;
+
+  // Heartbeat monitor thread dumps debug info on write to pipe
+  heartbeatMonitor_ = std::make_unique<HeartbeatMonitorXCCL>(this);
```
zhangxiaoli73, line 385: Is it really needed for flight recorder?

frost-intel: The goal of the flight recorder is to keep a log of executed collectives which can be dumped if the process hangs, so this monitor thread is necessary since the record dump must be possible if the default PG thread has hung. With this initial FlightRecorder support, the heartbeat monitor is simpler than the complicated watchdog thread in PGNCCL, but is still required to dump the logs.

src/xccl/ProcessGroupXCCL.cpp
frost-intel: This is a new counter added in this PR, which is incremented each time we record some work. So when dumping the flight record, the logs will have increasing ID for each collective.

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -353,9 +380,63 @@ ProcessGroupXCCL::ProcessGroupXCCL(
...
+              << writer.getWriterTarget();
+    writer.write(xcclTrace);
+    LOG(INFO) << logPrefix() << "Flight Recorder trace successfully dumped.";
+    return true;
+  }
+  return false;
+}
+
+const std::vector<uint64_t>& ProcessGroupXCCL::groupRanks() const {
+  if (options_->global_ranks_in_group.empty()) {
```
guangyey: Do we need to check `local_id_ == 0` here?

test/xpu/distributed/test_c10d_xccl.py
```diff
@@ -627,6 +632,316 @@ def test_all_gather_into_tensor(self):
...
+        self.set_thread_name("fr_test_thread")
+        a = torch.full((3, 4), float(self.rank), device=device)
+        for _ in range(2):
+            f = pg.allreduce(a)
+        f.wait()
+        torch.xpu.synchronize(device=device)
+        # gah ok so now the duration_ms is populated best-effort since it can only happen outside "dump()" api
+        time.sleep(1)
+        t = pickle.loads(
+            torch._C._distributed_c10d._dump_xccl_trace(
```
guangyey: This will create a mutual dependency between the two PRs (one in PyTorch and the other in torch-xpu-ops).

frost-intel: Do you suggest pushing the test into a different PR after Pytorch PR can be upstreamed, to avoid this?

guangyey: It would be great if we could separate this UT in a new PR. We can then
1. land this current PR,
2. update pin, and land the PR in PyTorch,
3. land the UT

frost-intel: @zhangxiaoli73 UT has been moved to a new PR, #1917

src/xccl/ProcessGroupXCCLMonitor.cpp
```diff
@@ -0,0 +1,66 @@
...
+      std::unique_lock<std::mutex> lock(monitorMutex_);
+      if (monitorWakeUpCV_.wait_for(
+              lock, std::chrono::milliseconds(coordCheckIntervalMilSec_), [&] {
+                return terminateHeartbeatMonitorThread_.load();
+              })) {
+        return;
+      }
+      // Write to pipe files for all ranks to dump debug info
+      if (dumpPipe.has_value() && dumpPipe->shouldDump()) {
+        LOG(INFO) << pg_->logPrefix()
```
zhangxiaoli73, line 54: Will you always dump this as it's in a background thread?

src/xccl/ProcessGroupXCCL.cpp
zhangxiaoli73: Frost, I understand that you want to add some fault tolerance features like heartbeat detection, and I agree with that. However, could you please create a separate PR for those changes instead of combining them with this flight recorder PR?

src/xccl/FlightRecorderXCCL.cpp
```diff
@@ -0,0 +1,21 @@
+#ifdef USE_C10D_XCCL
+
+#include <torch/csrc/distributed/c10d/FlightRecorderDetail.hpp>
+#include <ATen/xpu/XPUEvent.h>
+#include <xccl/ProcessGroupXCCL.hpp>
+
+namespace c10d {
+
+template <>
+float getDurationFromEvent<at::xpu::XPUEvent>(
```
zhangxiaoli73, line 10: No code is calling this API, I think you should remove this file.

src/xccl/ProcessGroupXCCL.hpp
```diff
@@ -108,7 +117,22 @@ class TORCH_API ProcessGroupXCCL : public Backend {
...
-  ProcessGroupXCCL(const c10::intrusive_ptr<Store>& store, int rank, int size);
+  struct Options : public Backend::Options {
+    explicit Options();
+
+    static c10::intrusive_ptr<Options> create() {
+      return c10::make_intrusive<Options>();
+    }
+
+    std::vector<uint64_t> global_ranks_in_group;
+    std::string group_name;
```
zhangxiaoli73, line 128: I see you also bind those to python in public stock PR. I'm just wondering if it's more reasonable to move definitions to `Backend::Options` instead of keeping separately in each backend like NCCL/XCCL.

src/xccl/ProcessGroupXCCLMonitor.cpp
frost-intel: In the NCCL equivalent code, there are two methods of dumping the FlightRecorder info to file: 
1. Write to a pipe file, which triggers the dump in all ranks
2. Receive a dump signal from the heartbeat monitor, caused by collective timeout or other error.

In this initial FlightRecorder PR, I've only implemented the first method since implementing the timeout and other heartbeat functionality is part of the Watchdog PR to be done in a follow-up PR. `dumpPipe` handles the file descriptor and the heartbeat thread here is periodically checking if the pipe file has been written to.

src/xccl/ProcessGroupXCCL.cpp
frost-intel: In the PGNCCL implementation, there are two background threads:
1. HeartbeatMonitor
  -  Regularly checks if a specific pipe has been written to, which triggers the FlightRecorder log dumps
  -  Checks if the Watchdog thread has sent a dump signal, which will also dump FR logs
  - Checks if there has been a timeout error.
2. Watchdog
  - Monitors the list of in-flight collectives. The PG status is regularly updated to reflect the in-progress work. 
  - If there is any error from some Work, then an error/dump signal is sent to all the PG processes and heartbeatMonitors, instructing them to dump FlightRecorder logs and shut down.

In this PR, I am only implementing the first part of the HeartbeatMonitor, namely the part which checks if a FlightRecorder dump signal has been received from the pipe-writing trigger. This doesn't include any fault tolerance features, only a periodic check if the user is trying to dump the FR logs. Without this HeartbeatMonitor thread, there isn't any method to actually retrieve the logs saved by the FR.

src/xccl/FlightRecorderXCCL.cpp
frost-intel: Unfortunately, the template for FlightRecorder<XPUEvent> requires this definition or else you get an undefined symbol error.

src/xccl/ProcessGroupXCCL.hpp
frost-intel: @zhangxiaoli73 @guangyey In order to prevent a cyclic dependency on these PRs, I'll move these to `Backend::Options` in upstream, then I'll do a follow-up PR in `torch-xpu-ops` to remove `Options` from `ProcessGroupXCCL`.

guangyey:
Let's land this PR to prevent blocking upstream PR in PyTorch.

guangyey:
cc @frost-intel 
This PR will be included in https://github.com/pytorch/pytorch/pull/160062

src/xccl/FlightRecorderXCCL.cpp
zhangxiaoli73: Should we raise a change in stock pytorch? It's not a good idea to keep something that we never call in `torch-xpu-ops`.

PR #1868, opened 2025-07-18
fix NllLossForwardReduce2DKernelFunctor accuracy

jianyizh:
follow cuda and make NllLossForwardReduce2DKernelFunctor more accurate using accumulate type


xytintel (approved):

jianyizh:
We can use the existing ut. For nan case, using accumulate type can fix loss output, but total weight will still be inf.

PR #1889, opened 2025-07-30
Capture compiler stderr in FindSYCLToolkit.cmake

Maetveis:
FindSYCLToolkit exectutes the DPC++ compiler using the `execute_process` CMake command to extract information about the compiler. To help debugging it tries to capture the output of the compilation to a file named Compile.log. However before this patch only the standard output was captured, even though the compiler will log errors to the standard error, and will almost certainly leave stdout empty.

Capture both stderr and stdout to the same file, this is explicitly supported by CMake, `execute_process` documentation notes:
> Added in version 3.3: If the same <file> is named for both OUTPUT_FILE and ERROR_FILE then it will be used for both standard output and standard error pipes.


Maetveis:
It seems like the compile result and output is expected to be printed to the CMake Log, but it cannot be both captured to a file and a variable.

PR #1891, opened 2025-07-31
optimize embedding

jianyizh:
Current index select kernel uses index kernel config, which sets wrong batch and problem batch, result to wrong launch configs that may lead to low occupancy. Also, inside kernel, half of thread is skipped in the following case:
index: 56103808
embedding table: [1683, 1]
previous on pvc: 19ms
now on pvc: 6.7ms

We will try to improve index kernel config next step. @yucai-intel


EikanWang:
@jianyizh , has this PR been ready for review? Sorry for the late response.

jianyizh:
@EikanWang cai yu is going to re-implement all index ops and remove the usage of index kernel config. We can review and merge this pr if there is urgent request on performance of embedding

PR #1915, opened 2025-08-07
[Nightly] Set different time measure for microbench

RUIJIEZHONG66166:
Enhance the op microbench test with device, test method and iter number ondemanded control.


RUIJIEZHONG66166:
After rebasing, ondemand test results: https://github.com/intel/torch-xpu-ops/actions/runs/17570650066

PR #1916, opened 2025-08-07
[WA] Fix the lack of exceptions vector in getrf_batch

CuiYifeng:
This PR provides a work-around for the lack of exception vector within batch_error exception, otherwise segmentation fault would happen while accessing exceptions.

### DIFF
--- a/src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
+++ b/src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
@@ -38,15 +38,18 @@ static oneapi::mkl::transpose to_blas_(TransposeType trans) {
   TORCH_INTERNAL_ASSERT(false, "Invalid transpose type");
 }
 
-void error_handle(int32_t* infos, const oneapi::mkl::lapack::batch_error& be) {
+void error_handle(
+    int32_t* info_cpu,
+    const oneapi::mkl::lapack::batch_error& be) {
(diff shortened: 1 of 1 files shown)


CuiYifeng:
Plan to remove this WA after upgrading to oneMKL 2025.3.

chunhuanMeng (approved):

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -529,8 +540,8 @@ void lu_factor_mkl(
       "linalg.lu_factor: LU without pivoting is not implemented on the XPU");
 
   // handle the info
-  info.zero_();
-  int32_t* infos_data = info.data_ptr<int32_t>();
+  Tensor info_ = at::zeros_like(info, Device(at::kCPU));
+  int32_t* infos_data = info_.data_ptr<int32_t>();
```
jianyizh: can you provide more explanation here? why you create a zero tensor on cpu and pass it to apply_lu_xpu_？ and then you copy info_ to xpu? mkl will get seg fault when info is on xpu?

CuiYifeng: On stock Pytorch side, original info Tensor is on device instead of host.

jianyizh (approved):

Issue #1121, opened 2024-11-27
SYCL: Runtime: Roll back to original usage of sycl::get_kernel_bundle


fengyuan14:
### 🐛 Describe the bug

The straight forward thought is kernel bundle is not device specific under a specific platform context (Like GPU platform). So we should not use `dev` (Existing WA for the https://github.com/intel/llvm/issues/15127) as a hint.
```
 18   auto kid = ::sycl::get_kernel_id<KernelClass>();
 24   auto kbundle = ::sycl::get_kernel_bundle<::sycl::bundle_state::executable>(
 25       ctx, {dev}, {kid}); // Should not be device specific
 26
 27   ::sycl::kernel k = kbundle.get_kernel(kid);
 28   return k.get_info<::sycl::info::kernel_device_specific::work_group_size>(dev); // Device specific. In additional, launching a kernel on a specific device is device specific.
```

### Versions

The fixing will land on oneAPI25.2. PT2.8

chuanqi129:
@fengyuan14 as the dependency PR has been merge, do we need to target this issue for PT2.8?

PR #1935, opened 2025-08-15
Roll back to original usage of sycl::get_kernel_bundle

CuiYifeng:
To solve https://github.com/intel/torch-xpu-ops/issues/1121.

The original usage of `sycl::get_kernel_bundle` should not be device-specific. Since the fixing has landed on oneAPI 2025.2, kernel bundle is now created using only the kernel ID (`kid`) instead of both device and kernel ID, removing the workaround for device-specific kernel builds and associated comments.

### DIFF
--- a/src/comm/DeviceProperties.h
+++ b/src/comm/DeviceProperties.h
@@ -16,13 +16,8 @@ static int64_t syclMaxWorkGroupSize(
   auto dev = q.get_device();
 
   auto kid = ::sycl::get_kernel_id<KernelClass>();
-  // The kernel won't be built for devices except for the first device.
-  // Launching kernel on devices except for the first device will raise
-  // runtime error. Here is an alternative as a temporary solution to
-  // provide an extra hint to SYCL runtime.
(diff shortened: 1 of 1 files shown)


fengyuan14 (approved):

chunhuanMeng (approved):

newtdms:
@CuiYifeng This patch is causing regression for some of the E2E model for both single- and multi-rank distributed training. Does the new DLE version require removing the device id from the kernel bundle? Do you have any documentation?

CuiYifeng:
> @CuiYifeng This patch is causing regression for some of the E2E model for both single- and multi-rank distributed training. Does the new DLE version require removing the device id from the kernel bundle? Do you have any documentation?

@newtdms The device id of `sycl::get_kernel_bundle` is a work-around to fix a issue. After the device id is removed, some old DLE versions are incompatible. Could you check if the DLE version >= 2025.2.0 in your environment?
BTW, could you share your error logs?

PR #1940, opened 2025-08-19
Enable aten::linalg_inv and aten::linalg_inv_ex

CuiYifeng:
This PR introduces the `inverse`, `linalg_inv` and `linalg_inv_ex` operators for XPU.
Added skip entries for error and warning tests related to inverse operations for all dtypes, due to a temporary regression in oneMKL.


guangyey:
The mkl regression will impact functionality or performace? Could you give a description such as what error the `test_inv_errors_and_warnings_xpu_float64` would raise?

guangyey:
> Since exception vector of `oneapi::mkl::lapack::batch_error` exception was incorrectly removed in oneMKL 2025.2, exception handling cannot correctly get the index of pivot element with a value of 0 (to check if the input matrix is singular). The hard-code value 1 in the above AssertionError comes from a work-around, which is planned to be removed when oneMKL 2025.3 is available.

Thanks, I see. 
https://github.com/intel/torch-xpu-ops/blob/a6790a95063892d1f169e1fad334e5ce1693cec2/src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp#L409-L424
The root cause is that oneMKL 2025.2 can't handle line 422 and it will be fixed in 2025.3, right?

Issue #1465, opened 2025-03-14
RuntimeError: Non-uniform work-groups are not supported by the target device reported on BMG


daisyden:
### 🐛 Describe the bug

On BMG platform with oneAPI 2025.1, a unit test failed. This is not a block for oneAPI 2025.1.

pytest -o cache_dir=F:\daisy_cache  -v .\test_ops_xpu.py -k test_compare_cpu_nn_functional_interpolate_bicubic_xpu_float64

> f:\pytorch\test\test_ops.py(397)test_compare_cpu()
-> cuda_results = op(sample.input, *sample.args, **sample.kwargs)
(Pdb) p sample.input[0]
tensor([[[ 6.3624, -3.1650,  7.6229,  ..., -0.6583,  5.6562, -2.6834],
         [ 5.8487, -4.8281,  3.9124,  ...,  4.1770, -0.2367,  6.5664],
         [-3.8261,  1.6662, -7.3418,  ...,  7.6588,  0.4768,  3.2033],
         ...,
         [-2.6621,  1.6049, -4.1793,  ..., -4.3173,  1.4278,  5.7185],
         [ 4.3221,  1.9088,  1.2546,  ..., -2.2788, -7.1614,  8.2987],
         [-2.8176,  3.1966,  2.7872,  ...,  5.3593,  4.3372, -8.9420]],

        [[-0.8267, -5.4715,  5.8321,  ...,  2.4388,  8.5541,  7.9531],
         [-3.2353,  2.6728, -1.2041,  ...,  6.5705, -8.1450,  4.1592],
         [-4.3237,  3.1313, -1.1240,  ..., -3.6158, -4.5955,  7.7287],
         ...,
         [ 8.3621,  3.5703,  3.8345,  ..., -0.9044, -2.3004,  8.7097],
         [-0.2895, -8.3597, -0.6198,  ...,  2.8927,  2.4080, -7.8745],
         [-5.9602,  5.4839, -2.4273,  ...,  1.9575,  3.3089, -6.1973]],

        [[ 8.6392, -8.7811, -4.0175,  ..., -8.5162, -4.5784, -0.7337],
         [ 8.7840, -6.6767, -5.6158,  ..., -3.4711, -4.4496, -6.5952],
         [-7.1685,  5.9275, -2.8213,  ..., -3.9898, -5.6324,  1.8914],
         ...,
         [ 3.1870,  2.3287, -4.2834,  ...,  3.5331,  2.8657, -2.6220],
[...]

PR #1942, opened 2025-08-19
UpsampleBilinear2Daa: Correct the global range

chunhuanMeng:
Correct the global range. The global range should be multiples of work group size.
Fix https://github.com/intel/torch-xpu-ops/issues/1465


fengyuan14:
Please correct the PR comment.

PR #1971, opened 2025-08-27
Add FlightRecorder tests

frost-intel:
As a follow-up to #1867 , this PR includes tests for the FlightRecorder on XCCL, as well as moving some definitions from ProcessGroupXCCL::Options to Backend::Options.

These tests are largely based on `pytorch/test/distributed/test_c10d_nccl.py`, but doesn't include some tests:
- `test_short_json` since json dumps are not supported in ProcessGroupXCCL
- `test_trace_while_all_works_retired`: `_wait_for_pending_works` isn't supported by XCCL
- `test_trace_while_active`: XCCL hangs when op is called on only one rank
- `test_trace_while_stuck`: XCCL hangs when op is called on only one rank

### DIFF
--- a/src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
+++ b/src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
@@ -5,6 +5,7 @@
 #include <ATen/native/Pool.h>
 
 #include <ATen/native/xpu/sycl/AveragePool2dKernels.h>
+#include <ATen/native/xpu/sycl/KernelUtils.h>
 #include <comm/Runtime.h>
 #include <comm/SYCLContext.h>
 #include <comm/SYCLHelpers.h>
@@ -25,9 +26,7 @@ inline int max(int a, int b) {
 template <typename scalar_t, typename accscalar_t, typename index_t>
 struct AvgPool2dKernelFunctor {
   void operator()(sycl::nd_item<1> item) const {
-    index_t index = item.get_global_linear_id();
-
-    if (index < total_elements_) {
+    XPU_KERNEL_LOOP(item, index, total_elements_) {
       const int pw = index % pooled_width_;
       const int ph = (index / pooled_width_) % pooled_height_;
       const int c = (index / pooled_width_ / pooled_height_) % channels_;
@@ -73,19 +72,19 @@ struct AvgPool2dKernelFunctor {
   AvgPool2dKernelFunctor(
       scalar_t* top_data,
       const scalar_t* bottom_data,
-      index_t total_elements,
-      index_t channels,
-      index_t height,
-      index_t width,
-      int pooled_height,
-      int pooled_width,
-      int kernel_h,
-      int kernel_w,
-      int stride_h,
-      int stride_w,
-      int pad_h,
-      int pad_w,
-      int divisor_override,
+      const int total_elements,
+      const int64_t channels,
+      const int64_t height,
+      const int64_t width,
+      const int64_t pooled_height,
+      const int pooled_width,
+      const int kernel_h,
+      const int kernel_w,
+      const int stride_h,
+      const int stride_w,
+      const int pad_h,
+      const int pad_w,
+      const int divisor_override,
       bool count_include_pad,
       bool use_divisor)
       : top_data_(top_data),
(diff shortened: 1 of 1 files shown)


src/xccl/ProcessGroupXCCL.hpp
```diff
@@ -171,7 +168,8 @@ class TORCH_API ProcessGroupXCCL : public Backend {
       bool isP2P,
       const char* profilingTitle = nullptr,
       const std::vector<at::Tensor>& inputs = {},
-      const std::vector<at::Tensor>& outputs = {});
+      const std::vector<at::Tensor>& outputs = {},
+      bool record = false);
```
zhangxiaoli73, line 184: Does it follow the same logic like NCCL backend?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -660,16 +663,19 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::collective(
     const char* profilingTitle,
     bool nanCheck) {
   nanCheck &= enableNanCheck_;
-  seqCollective_++;
   auto device = inputs[0].device();
   const auto key = std::to_string(device.index());
   auto comm = getXCCLComm(key, device, opType);
 
+  if (!coalescing_state_) {
```
zhangxiaoli73, line 669: @Chao1Han Please help check why this condition is not needed previously.

Chao1Han: frost’s modification is correct, this is indeed a bug. The community fixed it ten months ago, but since we had never dumped `seqCollective_`, we failed to notice it.

Chao1Han:
Hi @frost-intel, could you apt install clang-format and format ProcessGroupXCCL.cpp/ProcessGroupXCCL.hpp.
Avoid periodic code cleanup like https://github.com/intel/torch-xpu-ops/pull/1960/files

src/xccl/ProcessGroupXCCL.hpp
frost-intel: Yes, NCCL uses the same logic. Sometimes `initWork` needs to be recorded, but in other cases it causes improper access in FlightRecorder.

guangyey:
@frost-intel The UT failure is related?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -433,6 +433,10 @@ void ProcessGroupXCCL::setCompletedPgStatus(
   pgStatus_->lastCompletedWorkName = opTypeToString(work->opType_);
   pgStatus_->lastCompletedNumelIn = work->numelIn_;
   pgStatus_->lastCompletedNumelOut = work->numelOut_;
+  while (!work->isCompleted()) {
+    std::this_thread::sleep_for(
+        std::chrono::milliseconds(kSynchronizeBusyWaitMillis));
+  }
```
frost-intel: @guangyey @zhangxiaoli73 This method is called by a Future callback. It looks like the issue found by the tests was that the Work endEvent wasn't always marked as complete when this callback is called. This is introducing a busy wait. In most cases, there isn't any wait, but I believe that this callback would be part of a separate Future thread? In the NCCL case they do this waiting inside of the continuously running watchdog thread. They don't wait specifically for the work to complete, they just check all works to see if any have completed, then retire the id from FlightRecorder.

Thoughts? We could just remove the one piece of the tests that is failing until I implement Watchdog for XCCL, or we could use this as written (assuming this is running in a non-blocking thread).

zhangxiaoli73: I think you cannot make a host blocking wait here. By default, works are non host blocking and only maintain data dependency instead of host waiting until GPU kernel done. If you add such code, then there will be big perf harm (CUDA also have if you write same code on CUDA).

frost-intel: Since I'm not sure of the threading behavior, I'll remove this wait and when I add watchdog I can fix the completed behavior by copying the CUDA implementation, where a separate thread watches the work.

frost-intel:
@zhangxiaoli73 This is also ready for merge. Do we need a 2nd reviewer?

test/xpu/distributed/test_c10d_xccl.py
```diff
@@ -733,6 +738,564 @@ def test_wait_tensor(self) -> None:
...
+        [
+            [(2, 3)],
+            [(2, 3), (5, 5), (1,)],
+        ],
+    )
+    @parametrize("timing_enabled", [False])
+    def test_individual_send_recv(self, op_sizes, timing_enabled):
+        """
+        'WorkEnqueue' was skipped for isendirecv, leading to segfault on dump_entries when update_state tried to use
+        a destructed Work obj's cuda events
```
guangyey:
```suggestion
        a destructed Work obj's xpu events
```

test/xpu/distributed/test_c10d_xccl.py
```diff
@@ -733,6 +738,564 @@ def test_wait_tensor(self) -> None:
...
+        [
+            [(2, 3)],
+            [(2, 3), (5, 5), (1,)],
+        ],
+    )
+    @parametrize("timing_enabled", [False])
+    def test_batched_send_recv(self, op_sizes_per_coalesce, timing_enabled):
+        """
+        'WorkEnqueue' was skipped for isendirecv, leading to segfault on dump_entries when update_state tried to use
+        a destructed Work obj's cuda events
```
guangyey:
```suggestion
        a destructed Work obj's xpu events
```

EikanWang (approved):

Issue #1059, opened 2024-11-07
SYCL RT: Using recommended shortcut API for kernel specific max work group size.


fengyuan14:
### 🚀 The feature, motivation and pitch

Existing: https://github.com/intel/torch-xpu-ops/blob/b1582e14f9e27bcbc0666ff636389d2be61783e4/src/comm/DeviceProperties.h#L14-L28
TODO: https://github.com/intel/llvm/pull/15650

chuanqi129:
@fengyuan14  @majing921201 as the dependency PR has been merged, any update for this issue?

CuiYifeng:
Need to retry with DLE 2026.0 and check if there is performance regression after using shortcut.

chuanqi129:
@CuiYifeng let's prepare a branch to verify it

CuiYifeng:
Here is the dev branch to be verified: https://github.com/intel/torch-xpu-ops/tree/yifeng/shortcut_sycl_api.

chuanqi129:
@mengfei25 please verify this dev branch with regular scope

PR #1989, opened 2025-08-29
Update to recommended shortcut API for kernel specific max work group size

CuiYifeng:
This PR refactors how the maximum work group size for a SYCL kernel is retrieved in the `syclMaxWorkGroupSize` function. The manual construction and querying of a kernel bundle is replaced with `sycl::ext::oneapi::get_kernel_info` for directly obtaining the work group size, simplifying the code and improving clarity.


CuiYifeng:
To solve #1059.

PR #1991, opened 2025-08-29
Move checks from nonzero kernel to operator

Silv3S:
- Early return if input to nonzero has no elements,
- Change pytorch `MAX_DIMS` to `XPU_MAX_TENSORINFO_DIMS` xpu equivalent,
- Rename `tensor` to `out` to match op schema.


src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -15,24 +15,24 @@ struct FlattenIdxtoRealIdxKernelFunctor {
     auto global_id = item_id.get_global_linear_id();
 
     if (global_id < N_) {
-      auto dim = global_id / num_nonzeros_;
-      auto index = global_id % num_nonzeros_;
-      tensor_begin_[global_id] =
+      auto index = global_id / num_dim_;
+      auto dim = global_id % num_dim_;
```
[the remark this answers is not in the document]
Silv3S:
```suggestion
      auto dim = global_id % num_dim_;
      auto index = global_id / num_dim_;
```

tsocha (approved):

src/ATen/native/xpu/Nonzero.cpp
```diff
@@ -24,11 +24,15 @@ Tensor& nonzero_out_xpu(const Tensor& self, Tensor& out) {
...
   TORCH_CHECK(
-      self.dim() <= MAX_DIMS,
+      self.dim() <= XPU_MAX_TENSORINFO_DIMS,
       "nonzero is not supported for tensor with more than ",
-      MAX_DIMS,
+      XPU_MAX_TENSORINFO_DIMS,
       " dimensions");
 
+  if (self.numel() == 0) {
+    out = at::detail::empty_xpu({0, self.dim()}, out.options());
```
CuiYifeng, line 33: The shape `{0, self.dim()}` is translated from the transpose of `{num_dim, N}`, right?

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -79,77 +79,74 @@ struct CopyIfFunc<bool> {
...
-void nonzero_template(const Tensor& self_, Tensor& tensor) {
+void nonzero_template(const Tensor& self_, Tensor& out) {
   Tensor self = self_.contiguous();
 
   const int64_t num_dim = self.dim();
   TORCH_CHECK(num_dim <= XPU_MAX_TENSORINFO_DIMS, "dim exceed max allowed dim");
 
   int64_t N = self.numel();
+  if (N <= 0) {
+    out = out.resize_({num_dim, N}).contiguous().t();
```
CuiYifeng: Since the check has been moved to operator, please consider whether this part of the code is redundant.

src/ATen/native/xpu/Nonzero.cpp
Silv3S: When input has no elements, like for example
```
torch.empty(0,3,5).nonzero()       # tensor([], size=(0, 3), dtype=torch.int64)
torch.empty(0,3,5,5,3).nonzero()   # tensor([], size=(0, 5), dtype=torch.int64)
```
then expected output has shape of `{0, self.dim()}`. As we create new `out` tensor, I believe there is no need for any transpositions, as we can explicitly set this shape.

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
Silv3S: Done

src/ATen/native/xpu/Nonzero.cpp
CuiYifeng: Thanks for the elaboration.

Issue #1853, opened 2025-07-16
MatMul Complex datatype support


CuiYifeng:
### 🚀 The feature, motivation and pitch

# Background
Complex datatype MatMul is not supported in oneDNN and needs to be implemented using oneMKL.
# Feature Request
Need to support complex datatype for the following aten ops and their variants:

- `aten::addmm`
- `aten::baddbmm`
- `aten::bmm`
- `aten::mm`

Current complex datatype scope:

- `std::complex<float>`
- `std::complex<double>`

pkourdis:
@tye1 @CuiYifeng please prioritize this as impacts Argonne workloads transitioning to non-IPEX toolchain.

PR #1992, opened 2025-08-29
Implementation of matmul for complex datatypes.

PawelSwider2000:
Implementation of kernels for complex datatype support for 4 ops: `mm`, `bmm`,`addmm`, `baddbmm` using OneMKL.

Current implementation of this ops for XPU is in `pytorch/aten/src/ATen/native/mkldnn/xpu/Blas.cpp`. Since OneMKL is a torch-xpu-ops dependency and is available only with USE_ONEMKL_XPU=ON (which is a default value). Implementation needs to be in torch-ops-xpu and kernels and TORCH_LIBRARY_IMPL are in ifdef macro to avoid complication error when OneMKL is not supported. Newly declared op will be called from existing torch implementation using c10::Dispatcher.

This is part of: https://github.com/intel/torch-xpu-ops/issues/1853

### DIFF
--- a/src/ATen/native/xpu/Blas.cpp
+++ b/src/ATen/native/xpu/Blas.cpp
@@ -1,6 +1,4 @@
 #include <ATen/ATen.h>
-#include <ATen/native/DispatchStub.h>
-#include <ATen/native/mkldnn/xpu/Blas.h>
 #include <ATen/xpu/XPUContext.h>
 #include <comm/Runtime.h>
 #include <oneapi/mkl/blas.hpp>
@@ -129,15 +127,15 @@ at::Tensor& bmm_complex_out_xpu(
 
 template <typename T>
 at::Tensor& addmm_complex_out_xpu_impl(
-    const Tensor& input,
+    const Tensor& self,
     const Tensor& mat1,
     const Tensor& mat2,
     const Scalar& beta,
     const Scalar& alpha,
-    Tensor& result) {
+    Tensor& out) {
   at::Tensor mat1_cont = resolveViewsAndConjugation(mat1);
   at::Tensor mat2_cont = resolveViewsAndConjugation(mat2);
-  at::Tensor input_cont = resolveViewsAndConjugation(input).clone().detach();
+  at::Tensor self_cont = resolveViewsAndConjugation(self).clone().detach();
 
   const int64_t m = mat1_cont.sizes().at(0);
   const int64_t n = mat2_cont.sizes().at(1);
@@ -145,27 +143,27 @@ at::Tensor& addmm_complex_out_xpu_impl(
 
   // Some paths in the code below do not handle multiplications of the form [n, 0] x [0, m]
   if (k == 0) {
-    if (result.numel() == 0) {
-      return result;
+    if (out.numel() == 0) {
+      return out;
     }
     if (beta.toComplexDouble() == 0.0) {
-      result.zero_();
+      out.zero_();
     } else {
-      if (!input.is_same(result)) {
-        result.copy_(input);
+      if (!self.is_same(out)) {
+        out.copy_(self);
       }
-      result.mul_(beta);
+      out.mul_(beta);
     }
-    return result;
+    return out;
   }
 
   if (m == 0 || n == 0) {
-    return result;
+    return out;
   }
 
   const std::vector<int64_t> mm_output_size = {m, n};
-  if (input_cont.sizes() != mm_output_size) {
-    input_cont = at::broadcast_to(input_cont, mm_output_size).contiguous();
+  if (self_cont.sizes() != mm_output_size) {
+    self_cont = at::broadcast_to(self_cont, mm_output_size).contiguous();
   }
 
 
@@ -187,46 +185,46 @@ at::Tensor& addmm_complex_out_xpu_impl(
       reinterpret_cast<const std::complex<T>*>(mat2_cont.const_data_ptr()),
       n,
       complex_beta,
-      reinterpret_cast<std::complex<T>*>(input_cont.data_ptr()),
+      reinterpret_cast<std::complex<T>*>(self_cont.data_ptr()),
       n);
 
-  if (result.sizes() == input_cont.sizes()) {
-    result.copy_(input_cont);
+  if (out.sizes() == self_cont.sizes()) {
+    out.copy_(self_cont);
   } else {
-    result.copy_(input_cont.view(result.sizes()));
+    out.copy_(self_cont.view(out.sizes()));
   }
 
-  return result;
+  return out;
 }
 
 at::Tensor& addmm_complex_out_xpu(
-    const Tensor& input,
+    const Tensor& self,
     const Tensor& mat1,
     const Tensor& mat2,
     const Scalar& beta,
     const Scalar& alpha,
-    Tensor& result) {
+    Tensor& out) {
 
-  AT_DISPATCH_COMPLEX_TYPES(input.scalar_type(), "addmm_complex_out_xpu", [&] {
+  AT_DISPATCH_COMPLEX_TYPES(self.scalar_type(), "addmm_complex_out_xpu", [&] {
     using underlying_t = typename c10::scalar_value_type<scalar_t>::type;
     addmm_complex_out_xpu_impl<underlying_t>(
-        input, mat1, mat2, beta, alpha, result);
+        self, mat1, mat2, beta, alpha, out);
   });
 
-  return result;
+  return out;
 }
 
 template <typename T>
 at::Tensor& baddbmm_complex_out_xpu_impl(
-    const Tensor& input,
+    const Tensor& self,
     const Tensor& batch1,
     const Tensor& batch2,
     const Scalar& beta,
     const Scalar& alpha,
-    Tensor& result) {
+    Tensor& out) {
   at::Tensor batch1_cont = resolveViewsAndConjugation(batch1);
   at::Tensor batch2_cont = resolveViewsAndConjugation(batch2);
-  at::Tensor input_cont = resolveViewsAndConjugation(input).clone().detach();
+  at::Tensor self_cont = resolveViewsAndConjugation(self).clone().detach();
 
   const int64_t batch_size = batch1_cont.sizes().at(0);
   const int64_t m = batch1_cont.sizes().at(1);
@@ -234,8 +232,8 @@ at::Tensor& baddbmm_complex_out_xpu_impl(
   const int64_t k = batch1_cont.sizes().at(2);
 
   const std::vector<int64_t> mm_output_size = {batch_size, m, n};
-  if (input_cont.sizes() != mm_output_size) {
-    input_cont = at::broadcast_to(input_cont, mm_output_size).contiguous();;
+  if (self_cont.sizes() != mm_output_size) {
+    self_cont = at::broadcast_to(self_cont, mm_output_size).contiguous();;
   }
 
   std::complex<T> complex_alpha =
@@ -258,41 +256,50 @@ at::Tensor& baddbmm_complex_out_xpu_impl(
       n,
       k * n,
       complex_beta,
-      reinterpret_cast<std::complex<T>*>(input_cont.data_ptr()),
+      reinterpret_cast<std::complex<T>*>(self_cont.data_ptr()),
       n,
       m * n,
       batch_size);
 
-  if (result.sizes() == input_cont.sizes()) {
-    result.copy_(input_cont);
+  if (out.sizes() == self_cont.sizes()) {
+    out.copy_(self_cont);
   } else {
-    result.copy_(input_cont.view(result.sizes()));
+    out.copy_(self_cont.view(out.sizes()));
   }
 
-  return result;
+  return out;
 }
 
 at::Tensor& baddbmm_complex_out_xpu(
-    const Tensor& input,
+    const Tensor& self,
     const Tensor& batch1,
     const Tensor& batch2,
     const Scalar& beta,
     const Scalar& alpha,
-    Tensor& result) {
+    Tensor& out) {
 
   AT_DISPATCH_COMPLEX_TYPES(
-      input.scalar_type(), "baddbmm_complex_out_xpu", [&] {
+      self.scalar_type(), "baddbmm_complex_out_xpu", [&] {
         using underlying_t = typename c10::scalar_value_type<scalar_t>::type;
         baddbmm_complex_out_xpu_impl<underlying_t>(
-            input, batch1, batch2, beta, alpha, result);
+            self, batch1, batch2, beta, alpha, out);
       });
 
-  return result;
+  return out;
 }
 
-REGISTER_XPU_DISPATCH(mm_complex_stub, &mm_complex_out_xpu)
-REGISTER_XPU_DISPATCH(bmm_complex_stub, &bmm_complex_out_xpu)
-REGISTER_XPU_DISPATCH(addmm_complex_stub, &addmm_complex_out_xpu)
-REGISTER_XPU_DISPATCH(baddbmm_complex_stub, &baddbmm_complex_out_xpu)
+TORCH_LIBRARY(xpu_mkl, m) {
+  m.def("xpu_mkl::mm(Tensor self, Tensor mat2, *, Tensor(a!) out) -> Tensor(a!)");
+  m.def("xpu_mkl::bmm(Tensor self, Tensor mat2, *, Tensor(a!) out) -> Tensor(a!)");
+  m.def("xpu_mkl::addmm(Tensor self, Tensor mat1, Tensor mat2, *, Scalar beta=1, Scalar alpha=1, Tensor(a!) out) -> Tensor(a!)");
+  m.def("xpu_mkl::baddbmm(Tensor self, Tensor batch1, Tensor batch2, *, Scalar beta=1, Scalar alpha=1, Tensor(a!) out) -> Tensor(a!)");
+}
+
+TORCH_LIBRARY_IMPL(xpu_mkl, XPU, m) {
+  m.impl("xpu_mkl::mm", mm_complex_out_xpu);
+  m.impl("xpu_mkl::bmm", bmm_complex_out_xpu);
+  m.impl("xpu_mkl::addmm", addmm_complex_out_xpu);
+  m.impl("xpu_mkl::baddbmm", baddbmm_complex_out_xpu);
+}
 
 } // namespace at::native
--- a/src/ATen/native/xpu/Blas.cpp
+++ b/src/ATen/native/xpu/Blas.cpp
@@ -6,13 +6,18 @@
 
 namespace at::native {
 
-inline at::Tensor resolveViewsAndConjugation(const at::Tensor& input) {
-  at::Tensor input_resolved = input.is_conj() ? input.resolve_conj() : input;
-  at::Tensor input_contiguous = input_resolved.is_contiguous()
-      ? input_resolved
-      : input_resolved.contiguous();
+#if defined(USE_ONEMKL_XPU)
 
-  return input_contiguous;
+at::Tensor& handle_output_copy(at::Tensor& out, const at::Tensor& result) {
+  if (!out.is_same(result)) {
+    if (out.sizes() == result.sizes()) {
+      out.copy_(result);
+    } else {
+      out.copy_(result.view(out.sizes()));
+    }
+  }
+
+  return out;
 }
 
 template <typename T>
@@ -20,16 +25,16 @@ at::Tensor& mm_complex_out_xpu_impl(
     const at::Tensor& self,
     const at::Tensor& mat2,
     at::Tensor& out) {
-  at::Tensor self_cont = resolveViewsAndConjugation(self);
-  at::Tensor mat2_cont = resolveViewsAndConjugation(mat2);
-  at::Tensor out_cont = resolveViewsAndConjugation(out);
+  at::Tensor self_cont = self.contiguous().resolve_conj();
+  at::Tensor mat2_cont = mat2.contiguous().resolve_conj();
+  at::Tensor out_cont = out.contiguous().resolve_conj();
 
   const int64_t m = self_cont.sizes().at(0);
   const int64_t n = mat2_cont.sizes().at(1);
   const int64_t k = self_cont.sizes().at(1);
 
-  constexpr std::complex<T> alpha = {T(1.0), T(0.0)};
-  constexpr std::complex<T> beta = {T(0.0), T(0.0)};
+  constexpr std::complex<T> alpha = {T(1), T(0)};
+  constexpr std::complex<T> beta = {T(0), T(0)};
 
   oneapi::mkl::blas::row_major::gemm(
       c10::xpu::getCurrentXPUStream().queue(),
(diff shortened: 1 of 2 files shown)


PawelSwider2000:
@CuiYifeng @kbinias Please review

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,306 @@
...
+    } else {
+      if (!self.is_same(out)) {
+        out.copy_(self);
+      }
+      out.mul_(beta);
+    }
+    return out;
+  }
+
+  if (m == 0 || n == 0) {
```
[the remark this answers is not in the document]
PawelSwider2000: When m == 0 || n == 0 then output is also an empty tensor (with 0 on one of its dimension) and applying beta have so sense.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,306 @@
...
+at::Tensor& addmm_complex_out_xpu_impl(
+    const Tensor& self,
+    const Tensor& mat1,
+    const Tensor& mat2,
+    const Scalar& beta,
+    const Scalar& alpha,
+    Tensor& out) {
+  at::Tensor mat1_cont = mat1.contiguous().resolve_conj();
+  at::Tensor mat2_cont = mat2.contiguous().resolve_conj();
+  at::Tensor self_cont = self.contiguous().resolve_conj().clone().detach();
```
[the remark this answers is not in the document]
PawelSwider2000: We are cloning self tensor here because `oneapi::mkl::blas::row_major::gemm` will modify it inplace, see: https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onemath/source/domains/blas/gemm#onemath-blas-gemm and self is expected to be const so we cannot modify it's content.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,306 @@
...
+  const int64_t n = mat2_cont.sizes().at(1);
+  const int64_t k = self_cont.sizes().at(1);
+
+  constexpr std::complex<T> alpha = {T(1), T(0)};
+  constexpr std::complex<T> beta = {T(0), T(0)};
+
+  oneapi::mkl::blas::row_major::gemm(
+      c10::xpu::getCurrentXPUStream().queue(),
+      oneapi::mkl::transpose::nontrans,
+      oneapi::mkl::transpose::nontrans,
```
CuiYifeng: Please note that the storage of input tensors and output tensor is not always row-major.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,306 @@
...
+  const int64_t n = mat2_cont.sizes().at(2);
+  const int64_t k = self_cont.sizes().at(2);
+
+  constexpr std::complex<T> alpha = {T(1), T(0)};
+  constexpr std::complex<T> beta = {T(0), T(0)};
+
+  oneapi::mkl::blas::row_major::gemm_batch(
+      c10::xpu::getCurrentXPUStream().queue(),
+      oneapi::mkl::transpose::nontrans,
+      oneapi::mkl::transpose::nontrans,
```
CuiYifeng: Ditto.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+  std::complex<T> complex_alpha =
+      static_cast<std::complex<T>>(alpha.toComplexDouble());
+  std::complex<T> complex_beta =
+      static_cast<std::complex<T>>(beta.toComplexDouble());
+
+  oneapi::mkl::blas::row_major::gemm(
+      c10::xpu::getCurrentXPUStream().queue(),
+      oneapi::mkl::transpose::nontrans,
+      oneapi::mkl::transpose::nontrans,
```
CuiYifeng: Ditto.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+  std::complex<T> complex_alpha =
+      static_cast<std::complex<T>>(alpha.toComplexDouble());
+  std::complex<T> complex_beta =
+      static_cast<std::complex<T>>(beta.toComplexDouble());
+
+  oneapi::mkl::blas::row_major::gemm_batch(
+      c10::xpu::getCurrentXPUStream().queue(),
+      oneapi::mkl::transpose::nontrans,
+      oneapi::mkl::transpose::nontrans,
```
CuiYifeng: Ditto.

CuiYifeng (changes requested):

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,306 @@
...
+  m.def(
+      "aten::_addmm_mkl.out(Tensor self, Tensor mat1, Tensor mat2, *, Scalar beta=1, Scalar alpha=1, Tensor(a!) out) -> Tensor(a!)");
+  m.def(
+      "aten::_baddbmm_mkl.out(Tensor self, Tensor batch1, Tensor batch2, *, Scalar beta=1, Scalar alpha=1, Tensor(a!) out) -> Tensor(a!)");
+}
+
+#if defined(USE_ONEMKL_XPU)
+
+TORCH_LIBRARY_IMPL(aten, XPU, m) {
+  m.impl("aten::_mm_mkl.out", mm_complex_out_xpu);
```
guangyey: All ops listed here require a `c10::DeviceGuard` to ensure the current device is expected...

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Yes, we could decide which algorithm (row-major, col-major) to use base on input/output tensors. The implementation however will be much more complicated since we need to make a decision based on strides and shapes. Example we could have one tensor as row-major and second as col-major then deciding what to use could be more complicated.

Using row-major along with transposition to this format leads to worse performance when comparing to row-major, however for contiguous, second_col_major, first_col_major they are still comparable. Only both_col_major is visibly worse.

I would suggest to make this performance improvements in subsequent PRs, as it is much better to have complex support at all. Also mm is the simplest out of these ops and for baddbmm algorithm for efficient selection between row-major and col-major could be more complicated.

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

guangyey: Thanks.

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Made more detailed comparison between existing reference on IPEX GPU found following performance issues:

1. Perf degradation when one/both inputs are not row-major
2. Perf degradation when one/both inputs are conjugated
3. Worse performance for smaller sizes

Observed differences in perf are big, for some cases current implementation is a few times slower than reference. 

@CuiYifeng do you know if there is a performance difference between row-major and column-major implementations?

The other issue that I notice is that for addmm, baddbmm some tests are failing like: `TestCommonXPU::test_noncontiguous_samples_addmm_xpu_complex64` which was passing with implementation proposed in this PR

PawelSwider2000: To be more precise about perf for 4096x4096 tensors perf for contiguous non conjugate tensors implementation speedup is: 1.028 and it is similar for both conjugate inputs.

whereas the same for 256x256 tensors is around 0.565

For both columns major tensors and 4096x4096 tensors we have 0.255.

when only one tensor is column major, then differences are smaller but still large.

CuiYifeng: Different memory layouts may lead to performance differences. In complex MatMul kernel, data reorder for different layouts may also introduce differences.

PawelSwider2000: Changed implementation to use column_major, with other performance improvements current implementation is a bit faster than reference (up to few %)

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,454 @@
...
+  if (c.strides()[ndim - 2] == c.strides()[ndim - 1] &&
+      c.strides()[ndim - 1] == 1) {
+    ldc = c.sizes()[is_transposed ? ndim - 1 : ndim - 2];
+  } else {
+    ldc = c.strides()[is_transposed ? ndim - 2 : ndim - 1];
+  }
+  return ldc;
+}
+
+static inline int64_t get_stridec(const Tensor c) {
```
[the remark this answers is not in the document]
PawelSwider2000: This function is called only from code that require that tensor have enough dimensions.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,454 @@
...
+    const Tensor& batch2_input,
+    const Scalar& beta,
+    const Scalar& alpha,
+    Tensor& out) {
+  Tensor result = prepare_result_tensor(self, batch1_input, batch2_input, true);
+  const auto result_sizes = result.sizes();
+
+  if (result.numel() == 0) {
+    return out;
+  } else if (batch1_input.sizes()[2] == 0) {
```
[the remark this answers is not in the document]
PawelSwider2000: This function is called only from code that require that tensor have enough dimensions.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+  const int64_t last_dim = ndim - 1;
+  const int64_t second_last_dim = ndim - 2;
+
+  Tensor c = result.resolve_conj();
+
+  // Check if already in column-major layout (first dimension has unit stride)
+  if (result_strides[second_last_dim] == 1 &&
+      (result_sizes[last_dim] == 1 ||
+       result_strides[last_dim] ==
+           std::max(int64_t{1}, result_sizes[second_last_dim]))) {
```
CuiYifeng: Assuming Tensor `result` is obtained by slicing a column-major Tensor with an original shape of `[m, n]` to a shape of `[m', n] (m' < m)`, the strides of `result` will still be `[1, m]`. Meanwhile, oneMKL `gemm` and `gemm_batch` support such cases. Then, one `contiguous` can be avoided.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+    return {c, false};
+  }
+
+  // Check if in row-major layout (second dimension has unit stride)
+  if (result_strides[last_dim] == 1 &&
+      (result_sizes[second_last_dim] == 1 ||
+       result_strides[second_last_dim] ==
+           std::max(int64_t{1}, result_sizes[last_dim]))) {
```
CuiYifeng: As mentioned above, row-major `result` Tensor has scenarios similar to column-major.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+  // Check if matrix is already in the preferred layout (column-major for BLAS)
+  if (stride_inner == 1 && stride_outer == std::max(int64_t{1}, first_dim)) {
+    return {m.resolve_conj(), false};
+  }
+
+  // Check if matrix needs transposition but has unit stride in the other
+  // dimension
+  if (stride_outer == 1 && stride_inner == std::max(int64_t{1}, second_dim)) {
+    return {m, true};
+  }
```
CuiYifeng: Input matrices have scenarios similar to `result` and such usage are supported by oneMKL `gemm` and `gemm_batch`.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+    if (beta.toComplexDouble() == 0.0) {
+      out.zero_();
+    } else {
+      if (!self.is_same(out)) {
+        out.copy_(self);
+      }
+      out.mul_(beta);
+    }
+    return out;
+  }
```
CuiYifeng: Can we move this part of logic to op level function like `addmm_complex_out_xpu`? This part of logic does not depend on the handling for sizes and strides.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+  if (m == 0 || n == 0) {
+    return out;
+  }
```
CuiYifeng: Please move this check to op level if possible.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+  Tensor result = prepare_result_tensor(self, batch1_input, batch2_input, true);
+  const auto result_sizes = result.sizes();
+
+  if (result.numel() == 0) {
+    return out;
+  } else if (batch1_input.sizes()[2] == 0) {
+    if (beta.to<c10::complex<double>>() == 0.0) {
+      result.zero_();
+    }
+  }
```
CuiYifeng: Please consider if this part of code can be moved to op level, too.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+  return out;
+}
+
+#else
+
+Tensor& mm_complex_out_xpu(
+    const at::Tensor& self,
+    const at::Tensor& mat2,
+    at::Tensor& out) {
+  TORCH_CHECK(false, "Complex datatype matmul is not supported in oneDNN");
```
CuiYifeng:
```suggestion
  TORCH_CHECK(false, "Complex datatype matmul is not supported in oneDNN. Please include oneMKL library in compilation.");
```

CuiYifeng: Please fix the format of suggestion if needed.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+}
+
+Tensor& bmm_complex_out_xpu(
+    const Tensor& self,
+    const Tensor& mat2,
+    Tensor& out) {
+  TORCH_CHECK(false, "Complex datatype matmul is not supported in oneDNN");
```
CuiYifeng: Ditto

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+}
+
+Tensor& addmm_complex_out_xpu(
+    const Tensor& self,
+    const Tensor& mat1,
+    const Tensor& mat2,
+    const Scalar& beta,
+    const Scalar& alpha,
+    Tensor& out) {
+  TORCH_CHECK(false, "Complex datatype matmul is not supported in oneDNN");
```
CuiYifeng: Ditto

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,461 @@
...
+}
+
+Tensor& baddbmm_complex_out_xpu(
+    const Tensor& self,
+    const Tensor& batch1,
+    const Tensor& batch2,
+    const Scalar& beta,
+    const Scalar& alpha,
+    Tensor& out) {
+  TORCH_CHECK(false, "Complex datatype matmul is not supported in oneDNN");
```
CuiYifeng: Ditto

src/ATen/native/xpu/Blas.cpp
CuiYifeng: Such scenarios have been handled by IPEX.

PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/mkl/BlasImpl.cpp
```diff
@@ -0,0 +1,317 @@
...
+    const int64_t dim_second,
+    const bool contiguous_batch,
+    const bool check_dim_first = true) {
+  return is_column_major(
+      stride_second,
+      stride_first,
+      dim_second,
+      dim_first,
+      contiguous_batch,
+      check_dim_first);
```
CuiYifeng: Reusing `is_column_major` in `is_row_major` confused me. Directly implementing `is_row_major` is better for readability, such as:
```
inline bool is_row_major(
    const int64_t stride_first,
    const int64_t stride_second,
    const int64_t dim_first,
    const int64_t dim_second,
    const bool contiguous_batch,
    const bool check_dim_first = true) {
  return contiguous_batch && stride_second == 1 &&
      ((dim_first == 1 && check_dim_first) ||
       stride_first >= std::max(int64_t{1}, dim_second));
}
```

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,122 @@
...
+    if (beta.toComplexDouble() == 0.0) {
+      out.zero_();
+    } else {
+      if (!self.is_same(out)) {
+        out.copy_(self);
+      }
+      out.mul_(beta);
+    }
+    return out;
+  }
```
CuiYifeng, line 79: Just for code style: Less if-else nesting would be better.

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,122 @@
...
```
CuiYifeng, line 116: Ditto.

src/ATen/native/xpu/Blas.h
```diff
@@ -0,0 +1,28 @@
+#include <ATen/core/Tensor.h>
+
+namespace at::native {
+
+Tensor& mm_complex_out_xpu(const Tensor& self, const Tensor& mat2, Tensor& out);
```
CuiYifeng: Could you check the calling status on Windows?

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/mkl/BlasImpl.cpp
```diff
@@ -0,0 +1,317 @@
...
+    const Tensor& result,
+    IntArrayRef result_sizes) {
+  const auto result_strides = result.strides();
+  const int64_t ndim = result_strides.size();
+  const int64_t last_dim = ndim - 1;
+  const int64_t second_last_dim = ndim - 2;
+
+  const bool contiguous_batch = ndim > 2
+      ? result_strides[0] == (result_sizes[1] * result_sizes[2])
+      : true;
```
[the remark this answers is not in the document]
PawelSwider2000: We will not have here not 3d tensors

src/ATen/native/xpu/mkl/BlasImpl.cpp
```diff
@@ -0,0 +1,317 @@
...
+    bool transpose_c,
+    int64_t first_dim,
+    int64_t second_dim) {
+  const auto m_strides = m.strides();
+  const int64_t ndim = m_strides.size();
+  const int64_t last_stride = m_strides[ndim - 1];
+  const int64_t second_last_stride = m_strides[ndim - 2];
+
+  const bool contiguous_batch =
+      ndim > 2 ? m_strides[0] == (m.sizes()[1] * m.sizes()[2]) : true;
```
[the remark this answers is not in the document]
PawelSwider2000: We will not have here not 3d tensors

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace at::native {
+
+Tensor& mm_complex_out_xpu(
+    const Tensor& self,
+    const Tensor& mat2,
+    Tensor& out) {
+  c10::DeviceGuard guard(self.device());
+  TORCH_CHECK(
+      self.is_complex(), "_mm_mkl expects self to be a complex datatype.");
```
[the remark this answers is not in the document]
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,122 @@
...
+#endif // USE_ONEMKL_XPU
+}
+
+Tensor& bmm_complex_out_xpu(
+    const Tensor& self,
+    const Tensor& mat2,
+    Tensor& out) {
+  c10::DeviceGuard guard(self.device());
+  TORCH_CHECK(
+      self.is_complex(), "_bmm_mkl expects self to be a complex datatype.");
```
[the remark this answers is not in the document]
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,122 @@
...
+Tensor& addmm_complex_out_xpu(
+    const Tensor& self,
+    const Tensor& mat1,
+    const Tensor& mat2,
+    const Scalar& beta,
+    const Scalar& alpha,
+    Tensor& out) {
+  c10::DeviceGuard guard(self.device());
+  TORCH_CHECK(
+      self.is_complex(), "_addmm_mkl expects self to be a complex datatype.");
```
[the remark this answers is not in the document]
PawelSwider2000: Done

src/ATen/native/xpu/Blas.cpp
```diff
@@ -0,0 +1,122 @@
...
+Tensor& baddbmm_complex_out_xpu(
+    const Tensor& self,
+    const Tensor& batch1,
+    const Tensor& batch2,
+    const Scalar& beta,
+    const Scalar& alpha,
+    Tensor& out) {
+  c10::DeviceGuard guard(self.device());
+  TORCH_CHECK(
+      self.is_complex(), "_baddbmm_mkl expects self to be a complex datatype.");
```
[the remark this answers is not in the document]
PawelSwider2000: Done

src/ATen/native/xpu/mkl/BlasImpl.cpp
PawelSwider2000: Yes, it could be confusing, I make it in such way to remote duplicated logic between this two functions. However it this case readiness is also important

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Convert: `} else if (mat1.numel() == 0) {` to separate if, other ifs are necessary.

src/ATen/native/xpu/Blas.cpp
PawelSwider2000: Done

src/ATen/native/xpu/Blas.h
PawelSwider2000: I added: TORCH_XPU_API before, CI and on-demand tests should detect issues with Windows

PR #1993, opened 2025-08-29
Tests for matmul for complex datatypes.

PawelSwider2000:
Unskip test that were passing after implementing matmuls for complex datatypes.

Some test still fails but with different errors, unrelated to matmuls like: `# NotImplementedError: The operator 'aten::cholesky_inverse.out' is not currently implemented for the XPU device.`

Created as separate PR (implementation: https://github.com/intel/torch-xpu-ops/pull/1992) as this will require additional PR in upstream torch


PawelSwider2000:
> Although there are many test cases require complex MatMul for assertion, we don't need to enable all these cases at once. Activating the test case for directly testing Complex MatMul is sufficient.

Sure, I think we should run this test on CI and check them, if they are passing, I do not see reason to now unskip them. However if they are failing and are not directly testing MatMul they could stay skipped for now.

PawelSwider2000:
For now, waiting for https://github.com/pytorch/pytorch/pull/160867 to be merged

PR #1966, opened 2025-08-27
[CI] Enable pytest parallel run

mengfei25:

PR #1862, opened 2025-07-16
[CI] Refactor CICD test workflows

mengfei25:
1. test in container
2. use local python instead of conda
3. split common steps

disable_winidows

PR #1999, opened 2025-09-01
[CI] Modify accelerate and transformers tests

mengfei25:
1. enable test in container
2. use local python instead of conda
3. enable pytest parallel run and continue-if-cash
4. use pytest-xdist to parallelize tests instead of pytest-shard on a 8 cards system
5. all tests on rolling driver

test accelerate and transformers only
disable_build
disable_ut
disable_e2e
disable_distributed

### DIFF
--- a/.github/actions/get-runner/action.yml
+++ b/.github/actions/get-runner/action.yml
@@ -25,7 +25,7 @@ runs:
       run: |
         # get test runner
         echo "runner_id=$(echo ${RUNNER_NAME} |sed 's/\-[0-9]$//')" |tee -a ${GITHUB_OUTPUT}
-        echo "user_id=$(id -u):$(id -g)" |tee -a ${GITHUB_OUTPUT}
+        echo "user_id=$(id -u)" |tee -a ${GITHUB_OUTPUT}
         echo "render_id=$(getent group render |cut -d: -f3)" |tee -a ${GITHUB_OUTPUT}
         echo "hostname=$(hostname)" |tee -a ${GITHUB_OUTPUT}
         # show host info


.github/workflows/_linux_transformers.yml
```diff
@@ -308,45 +236,72 @@ jobs:
           python -c 'import torch; exit(not torch.xpu.is_available())'
       - name: Run tests
         run: |
-          source activate $CONDA_ENV_NAME
+          set -x
+          set +e
           cd transformers
-          python3 -m pytest -rsf --make-reports=$TEST_CASE --junit-xml=reports/$TEST_CASE.xml \
-            -k "${{ matrix.test.filter}}" ${{ matrix.test.cmd }} || true
+          # Excluding tests due to:
```
dvrogozh: Is this a temporary change to remove a matrix job? It's essential actually as it allows to restart smaller chunk of the test on failure.

dvrogozh (changes requested):

.github/workflows/_linux_transformers.yml
mengfei25: Firstly, there will be not 1 card runner and all are the whole system with 8 cards, which means mount to github easily , maintain easily and to leverage all resources for E2E, UT and others.
I just considered to save time, and will rollback it. But parallel enabling also can save time, so 2 or 4 shards is enough to reduce test_models test time within 15 mins. Is it okay for you? or any other considerations?

chuanqi129:
Hi @dvrogozh , we're working on torch-xpu-ops CI/CD workflows refactor, mainly including 2 aspects:

1. Containerized build and test, please refer #1862
2. Parallelized UT tests by using `pytest-xdist`, please refer #1966

Those changes have a lot of benefits:

- Standardize build and tests
- Reduce single UT test job time cost
- No Conda dependency, use `setup-python` in containers directly, test env totally isolated
- Simplify the runner maintain effort, don't need to split one node as multiple runners
- No change for matrix tests support

Please help to review this PR, after this PR land, we'll remove all single card runners with `linux.idc.xpu` label

chuanqi129:
@mengfei25 please check the transformers test, there is no case to be executed actually.

.github/workflows/_linux_accelerate.yml
```diff
@@ -66,22 +70,48 @@ jobs:
...
+      options: --device=/dev/mem --device=/dev/dri --group-add video --security-opt seccomp=unconfined --cap-add=SYS_PTRACE --shm-size=8g
+              -u ${{ needs.prepare.outputs.user_id }}:${{ needs.prepare.outputs.render_id }}
+              -e ZE_AFFINITY_MASK
+      env:
+        WORK_DIR: 'accelerate'
+        PYTORCH_DEBUG_XPU_FALLBACK: 1
+        HF_HUB_ETAG_TIMEOUT: 120
+        HF_HUB_DOWNLOAD_TIMEOUT: 120
+        PARSE_JUNIT: ${{ github.workspace }}/torch-xpu-ops/.github/scripts/parse-junitxml.py
+        AGENT_TOOLSDIRECTORY: /tmp/xpu-tool
```
dvrogozh, line 107: You define `AGENT_TOOLSDIRECTORY`, but I don't see it's used anywhere. Is it actually needed? maybe we can drop it?

.github/workflows/_linux_accelerate.yml
```diff
@@ -93,25 +123,17 @@ jobs:
...
+        with:
+          python-version: ${{ env.python }}
+      - name: Install pytorch and deps
         run: |
-          echo "Using Conda ENV name: $CONDA_ENV_NAME"
-          conda remove --all -y -n $CONDA_ENV_NAME || rm -rf $(dirname ${CONDA_EXE})/../envs/$CONDA_ENV_NAME
-          conda create -y -n $CONDA_ENV_NAME python=${{ env.python }}
-          source activate $CONDA_ENV_NAME
-          pip install junitparser pytest-timeout
+          pip install junitparser pytest-timeout pytest-xdist
```
dvrogozh: Accelerate test takes around 5 minutes, so I don't think there is currently a need to run anything in it in parallel. Are there any other benefits from xdist? If not, I suggest to remove it for simplicity.

.github/workflows/_linux_transformers.yml
```diff
@@ -127,94 +134,35 @@ jobs:
...
-          # * Network proxy connection issue, reason unknown
-          # *'tests/utils/test_import_utils.py' invalidates state of the test engine causing
-          #   next tests to fail. See: https://github.com/huggingface/transformers/issues/36267
-          - test_case: 'tests_utils'
-            cmd: '--ignore=tests/utils/test_import_utils.py tests/utils'
-            filter: 'not test_load_img_url_timeout'
-    env:
-      PYTORCH_DEBUG_XPU_FALLBACK: '1'
-      TRANSFORMERS_TEST_DEVICE_SPEC: 'spec.py'
+        test_case: [tests_backbone,tests_py,tests_generation,tests_models_4_0,tests_models_4_1,tests_models_4_2,tests_models_4_3,tests_trainer,tests_utils]
```
dvrogozh: Why do you consider new version better than what was before? To me previous version was simpler - it was shorter and actual command lines were better visible.

.github/workflows/_linux_transformers.yml
```diff
@@ -299,54 +238,52 @@ jobs:
           xpu-smi discovery -y --json --dump -1
       - name: Sanity check installed packages
         run: |
-          source activate $CONDA_ENV_NAME
+          # Use latest pytest
+          pip install -U pytest pytest-timeout pytest-xdist pytest-shard
```
dvrogozh, line 285: What `xdist` is doing exactly? If it parallelizes execution, then how?

.github/workflows/_linux_accelerate.yml
```diff
@@ -66,22 +70,48 @@ jobs:
...
+    outputs:
+      runner_id: ${{ steps.runner-info.outputs.runner_id }}
+      user_id: ${{ steps.runner-info.outputs.user_id }}
+      render_id: ${{ steps.runner-info.outputs.render_id }}
+      hostname: ${{ steps.runner-info.outputs.hostname }}
+      pytest_extra_args: ${{ steps.runner-info.outputs.pytest_extra_args }}
+    steps:
+      - name: Cleanup workspace
+        run: |
+          sudo find ./ |grep -v "^\./$" |xargs sudo rm -rf
```
dvrogozh: This looks weird. Basically that's a hack since github actions are supposed to take care of that entirely. Effectively you increase a fetch time for the repos as you are deleting everything each time as it seems. It seems there is some mess with access rights. Can this be fixed?

dvrogozh (changes requested):
Transformers tests don't actually run anything after this change. I don't know the rootcause, but this needs to be debugged and fixed.

**Important.** Be extremely careful with parallelizing transformers tests. In the currently executing version parallelism of the tests is essentially switched off (see max parallel jobs settings set to 1). That was done for the reason that as seen as we are trying to parallelize we step into networking issues either on our side or on huggingface hub side. Last time we failed to overcome the problem and max parallel jobs was set to 1. The martix is still used to break the whole test into smaller chunks each around ~30mins or less. This allows to rerun smaller portion of the test on the failure instead of rerunning the whole suite.

dvrogozh:
> By the way, the accelerate test has 3 failures, I guess they are not related with this PR changes

These are new failures it seems. I see some test failures in this weekly as well and the a week old weekly was clean.

.github/workflows/_linux_accelerate.yml
```diff
@@ -66,22 +70,48 @@ jobs:
...
+  tests:
+    runs-on: ${{ needs.prepare.outputs.runner_id }}
+    needs: prepare
+    container:
+      image: mengfeili/intel-pvc-driver:1146-1136
+      volumes:
+        - ${{ github.workspace }}:${{ github.workspace }}
+      options: --device=/dev/mem --device=/dev/dri --group-add video --security-opt seccomp=unconfined --cap-add=SYS_PTRACE --shm-size=8g
+              -u ${{ needs.prepare.outputs.user_id }}:${{ needs.prepare.outputs.render_id }}
+              -e ZE_AFFINITY_MASK
```
dvrogozh, line 100: Why it's `-e ZE_AFFINITY_MASK` instead of `-e ZE_AFFINITY_MASK=${{env.ZE_AFFINITY_MASK}}`? it seems you set the variable without value. I don't know how oneAPI will behave on such a setting.

.github/workflows/_linux_accelerate.yml
mengfei25: It is used by setup-python, in order to get clean python env and avoid to reuse the exist in /__w/_tools

dvrogozh: Indeed, described here: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-setup-python-with-a-self-hosted-runner.

.github/workflows/_linux_accelerate.yml
```diff
@@ -66,22 +70,49 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Get runner
+        id: runner-info
+        uses: ./.github/actions/get-runner
+
+  tests:
+    runs-on: ${{ needs.prepare.outputs.runner_id }}
+    needs: prepare
+    container:
+      image: mengfeili/intel-pvc-driver:1146-1136
```
dvrogozh, line 94: Is using image under user account (`mengfeili`) per-design? or left-over from the debug?

dvrogozh (commented):
@mengfei25, the errors in transformers and accelerate tests look unrelated to this PR and correlate with the current version of the test. I am impressed by the reduction of the transformers test execution time. I was afraid that we will again step into networking issue, but test results indicate that we did not.

However, I have 2 requests before merging this test:
1. There are still few not answered questions from me in this PR. Can you, please, go thru them and reply?
2. I would like to be on the safe side making sure we really can parallelize transforms side. I suggest to force rerun test few times to verify that we don't step into the problem. We can do that once we will finish discussion on the questions not yet closed.

mengfei25:
@dvrogozh  Please help review it

.github/workflows/_linux_transformers.yml
mengfei25: Change back to the previous version

.github/workflows/_linux_accelerate.yml
mengfei25: Remove parallel but keep the crash restart with `-n 1` supplied by `pytest-xdist`

.github/workflows/_linux_transformers.yml
mengfei25: Change to previous version

.github/workflows/_linux_transformers.yml
mengfei25: `xdist` is for parallel and crash restart. with `--dist` and `--tx` to distribute CPUs and GPUs evenly.
And get the args from runner init `get-runner`, use the `PYTEST_ADDOPTS` to enable it.

.github/workflows/_linux_accelerate.yml
mengfei25: 1. Our CI has potential permission issue for WORKSPACE
2. A clean WORKSPACE is important for CI to make sure correct
3. Source code is lightweight, clone uses very less time

.github/workflows/_linux_accelerate.yml
mengfei25: `-e ZE_AFFINITY_MASK` has same behavior with host, and I tested it, works well

.github/workflows/_linux_accelerate.yml
mengfei25: No clean public image with driver only, so uses my image temporarily, consider to push such image to official hub later

mengfei25: @dvrogozh  Commented all your questions, please help review.

.github/workflows/_linux_accelerate.yml
dvrogozh: Ok, documented here: https://docs.docker.com/reference/cli/docker/container/run/#env.

.github/workflows/_linux_accelerate.yml
dvrogozh: ok

.github/workflows/_linux_accelerate.yml
```diff
@@ -66,22 +70,49 @@ jobs:
...
+      volumes:
+        - ${{ github.workspace }}:${{ github.workspace }}
+      options: --device=/dev/mem --device=/dev/dri --group-add video --security-opt seccomp=unconfined --cap-add=SYS_PTRACE --shm-size=8g
```
dvrogozh: Groups owning the device might differ depending on the OS. For modern Ubuntu that's `video` owning `/dev/dri/cardN` and `render` owning `/dev/dri/renderDN`. In our workflows we don't typically use `/dev/dri/cardN` and rely on `/dev/dri/renderDN`. This means that it's important to allow access to `render` group at the first place, but you have `video` instead. I suspect that `render` is hiding under `needs.prepare.outputs.render_id`. Is that right?

I think that this part of CI setup needs to be polished. I suggest you need the following:
1. Pass `render` and `video` explicitly here, i.e. `options: --group-add render --group-add video ...`
2. `-u user:group` is special. It's a key to managing permissions and is a reply to the above question (why you need `sudo rm -rf` to clean workspace). To fix that I suggest that you need to pass a `user:group` matching credentials of the user under which the runner runs. Thus should fix file permission issues. And you pass additional groups to access devices in addition explicitly.

.github/workflows/_linux_accelerate.yml
dvrogozh: See proposal on the resolution here: https://github.com/intel/torch-xpu-ops/pull/1999/files#r2330443890. We need to discuss if this can be done within this PR or later on.

.github/workflows/_linux_accelerate.yml
mengfei25: 1. Good point, fixed
2. Already with the runner user, removed the `rm` before checkout.

.github/workflows/_linux_accelerate.yml
mengfei25: Removed, will rm permissioned files in host manually if meets in the future

.github/workflows/_linux_transformers.yml
```diff
@@ -299,36 +293,63 @@ jobs:
...
-      - name: Run tests
+      - name: Run tests on ${{ needs.prepare.outputs.hostname }}
         run: |
-          source activate $CONDA_ENV_NAME
+          set -x
+          set +e
           cd transformers
-          python3 -m pytest -rsf --make-reports=$TEST_CASE --junit-xml=reports/$TEST_CASE.xml \
-            -k "${{ matrix.test.filter}}" ${{ matrix.test.cmd }} || true
+          xpu_list=($(echo ${ZE_AFFINITY_MASK} |sed 's/,/ /g'))
```
dvrogozh: These changes did not exist just yesterday. Why do we need them? I thought we have agreed that previous version was almost ready to go. I definitely has concerns with this new version. For example, output table became much bigger and complex: https://github.com/intel/torch-xpu-ops/actions/runs/17613758986?pr=1999. We don't need all these categories...

.github/actions/get-runner/action.yml
```diff
@@ -52,31 +57,60 @@ runs:
...
+      shell: bash -xe {0}
+      id: tests
+      run: |
+        # cpu number
         cpu_num="$(lscpu |grep -E 'Core\(s\) per socket:|Socket\(s\):' |awk 'BEGIN{sum=1}{sum*=$NF}END{printf sum}')"
-        xpu_num="$(clinfo --list |awk 'BEGIN{gpu=0;}{
-          if(gpu==1 && $0~/Platform/){gpu=0}; if(gpu==1){print $0}; if($0~/Platform.*Graphics/){gpu=1}
-        }' |wc -l)"
-        cpus_per_xpu="$(echo |awk -v c="${cpu_num}" -v x="${xpu_num}" '{printf c/x}')"
+        # available gpu card number
```
dvrogozh: These changes were not here yesterday. Are they absolutely necessary? I also see that few other workloads besides accelerate and transformers got changed. Is this necessary as well?

dvrogozh (changes requested):
I am not ok with the last changes in the PR. A lot of new changes appeared all of the sudden which require review the PR anew while yesterday version was almost ready to go with few minor changes. We need align again.

.github/workflows/_linux_e2e.yml
```diff
@@ -80,9 +79,12 @@ jobs:
       image: mengfeili/intel-pvc-driver:1146-1136
       volumes:
         - ${{ github.workspace }}:${{ github.workspace }}
-      options: --device=/dev/mem --device=/dev/dri --group-add video --security-opt seccomp=unconfined --cap-add=SYS_PTRACE --shm-size=8g
-              -u ${{ needs.runner.outputs.user_id }}:${{ needs.runner.outputs.render_id }}
+      options: --device=/dev/mem --device=/dev/dri --group-add video --group-add ${{ needs.runner.outputs.render_id }}
```
dvrogozh: Can this change (`rm-rf` for workspace) which affects multiple workloads be extracted into separate PR, please? This will reduce the scope to review. Besides this change needs to be validated across all affected workflows.

.github/workflows/_linux_e2e.yml
```diff
@@ -62,10 +62,9 @@ jobs:
       hostname: ${{ steps.runner-info.outputs.hostname }}
       xpu_num: ${{ steps.runner-info.outputs.xpu_num }}
       cpus_per_xpu: ${{ steps.runner-info.outputs.cpus_per_xpu }}
+      ZE_AFFINITY_MASK: ${{ steps.runner-info.outputs.ZE_AFFINITY_MASK }}
+      numactl_args: ${{ steps.runner-info.outputs.numactl_args }}
```
dvrogozh: How `numactl` relates to this PR? This seems unrelated change and must be done in separate PR.

.github/workflows/_linux_transformers.yml
mengfei25: Removed the overcomplicated changes

.github/actions/get-runner/action.yml
mengfei25: Removed the overcomplicated changes

.github/workflows/_linux_e2e.yml
mengfei25: Sure, Removed

.github/workflows/_linux_e2e.yml
mengfei25: Not related with this PR, removed

PR #2018, opened 2025-09-04
[CI] Modify test workflows

mengfei25:
1. use cache in container for datasets and models
2. fix np.bool8 issue in soft_actor_critic
3. fix microbench test reference issue
4. remove inductor test in nightly
5. use nightly wheel in CI if build not necessary

disable_build

### DIFF
--- a/.github/workflows/_linux_ut.yml
+++ b/.github/workflows/_linux_ut.yml
@@ -77,6 +77,11 @@ jobs:
           pytorch: ${{ inputs.pytorch }}
           torch_xpu_ops: ${{ inputs.torch_xpu_ops }}
           python: ${{ inputs.python }}
+      - name: Login huggingface
+        run: |
+          pip install huggingface-hub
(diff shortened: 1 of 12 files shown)


chuanqi129 (approved):

.github/workflows/_linux_ut.yml
```diff
@@ -139,7 +140,7 @@ jobs:
             sudo cp ${{ github.workspace }}/ptrace_scope.bk /proc/sys/kernel/yama/ptrace_scope
           fi
 
-  host:
+  test-in-host:
```
chuanqi129: Do we still need this `test-in-host` job?

mengfei25: Removed according to https://jira.devtools.intel.com/browse/GSD-11385
with oneCCL 2021.16.1, the issue is fixed

mengfei25: Looks like still need to run distributed on host, because xelink does not work in container

Issue #2025, opened 2025-09-09
70% Performance regression in single-rank CosmicTagger workload


XueHuCH:
### 🐛 Describe the bug

There were ~70% performance drop in single-tile Cosmic-tagger test using the nightly wheel torch==2.9.0.dev20250902+xpu. No regression with the nightly wheel torch==2.9.0.dev20250830+xpu. The regression is recovered by setting ZE_AFFINITY_MASK=0.

Regression observed between nightly wheel version:
- 2.9.0.dev20250902+xpu
  - Torch hash: commit 6737e2c996990024187ba620d2764f3b6f6add2c
  - Torch-xpu-ops hash: 8b58040ee32689487f660462f655085f31506dab

- 2.9.0.dev20250830+xpu
  - Torch hash: commit: f6368e934e6bef84211f7db82c22e3623038e43f
  - Torch-xpu-ops hash: 77cc792cd265179745d335579d233e6d4f9a2667

Regression is recovered by resetting Torch-xpu-ops hash to 77cc792cd265179745d335579d233e6d4f9a2667 and observed when Torch-xpu-ops hash: c091232d489597104baf29c0ba8a5f1c6149e593 is used.

### Versions

PyTorch version: 2.9.0.dev20250903+xpu
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: SUSE Linux Enterprise Server 15 SP4 (x86_64)
GCC version: (Spack GCC) 13.3.0
Clang version: Could not collect
CMake version: version 3.20.4
Libc version: glibc-2.31

Python version: 3.10.18 | packaged by conda-forge | (main, Jun  4 2025, 14:45:41) [GCC 13.3.0] (64-bit runtime)
Python platform: Linux-5.14.21-150400.24.55-default-x86_64-with-glibc2.31
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
Is XPU available: True
XPU used to build PyTorch: 20250201
Intel GPU driver version:
* intel-opencl: 25.18.33578.15-1146
* level-zero:   1.21.9.0-i1136
Intel GPU models onboard:
* Intel(R) Data Center GPU Max 1550
* Intel(R) Data Center GPU Max 1550
* Intel(R) Data Center GPU Max 1550
[...]

PR #769, opened 2024-08-16
SYCL: Alternative solution to avoid runtime error of launching kernel on xpu:1 when que…

fengyuan14:
…rying SYCL kernel bundle ahead
The kernel won't be built for devices except for the first device. Launching kernel on devices except for the first device will raise runtime error. Here is an alternative as a temporary solution to provide an extra hint to SYCL runtime. https://github.com/intel/llvm/issues/15127

PR #2026, opened 2025-09-10
Revert "Roll back to original usage of sycl::get_kernel_bundle (#1935)"

CuiYifeng:
To solve #2025.
This reverts commit 83a15554d50730ecd30814e8ce4ba7b72a444324 to quickly bypass the performance regression caused by usage of sycl::get_kernel_bundle .

### DIFF
--- a/src/comm/DeviceProperties.h
+++ b/src/comm/DeviceProperties.h
@@ -16,8 +16,13 @@ static int64_t syclMaxWorkGroupSize(
   auto dev = q.get_device();
 
   auto kid = ::sycl::get_kernel_id<KernelClass>();
-  auto kbundle =
-      ::sycl::get_kernel_bundle<::sycl::bundle_state::executable>(ctx, {kid});
+  // The kernel won't be built for devices except for the first device.
+  // Launching kernel on devices except for the first device will raise
+  // runtime error. Here is an alternative as a temporary solution to
+  // provide an extra hint to SYCL runtime.
+  // https://github.com/intel/llvm/issues/15127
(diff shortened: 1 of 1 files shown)


fengyuan14 (approved):

src/comm/DeviceProperties.h
```diff
@@ -16,8 +16,13 @@ static int64_t syclMaxWorkGroupSize(
   auto dev = q.get_device();
 
   auto kid = ::sycl::get_kernel_id<KernelClass>();
-  auto kbundle =
-      ::sycl::get_kernel_bundle<::sycl::bundle_state::executable>(ctx, {kid});
+  // The kernel won't be built for devices except for the first device.
```
gujinghui, line 19: what's the meaning of first device here?

CuiYifeng: @gujinghui The first device represents `xpu:0`.

gujinghui: then, this comment is confused. it's not related to first device here, it's related to the default device type of get_kernel_bundle. So we can say, the current device must be set in get_kernel_bundle API. Otherwise, the kernel for default device type will be returned, which will be different from current device sometime, and cause runtime error.

CuiYifeng: This comment comes from an old WA #769.

CuiYifeng: Please check hard code in https://github.com/intel/llvm/blob/bcc0fab8933a636a6856e3e2c52211041079ae7c/sycl/source/detail/program_manager/program_manager.cpp#L2508-L2513

EikanWang (approved):

gujinghui (approved):

PR #2030, opened 2025-09-11
[SYCL-TLA] Enable SYCL-TLA build

LuFinch:
This is a draft PR to enable SYCL-TLA build in torch-xpu-ops so that we can test SYCL-TLA kernels' accuracy/performance in Pytorch when SDPA/GEMM kernels are ready.

After discussion with Eikan, we decided to put build logic in torch-xpu-ops while put kernels source code in Pytorch in-tree. Please put your SYCL-TLA kernel source code in Pytorch and set its path as part of `ATen_XPU_SYCLTLA_SRCS` in `torch-xpu-ops/src/ATen/CMakeLists.txt`.

Since SYCL-TLA has different compilation options compared with normal SYCL kernels in torch-xpu-ops, I make the logic in `cmake/BuildFlags.cmake` as a macro so that I can reuse the common compilation options.

Since there is not a determined plan of how to import sycl-tla repo, I git clone the main branch in cmake for debug convinence. We can pin commit after sycl-tla has first release tag

Depend on g++ upgrading to gcc13, otherwise the sycltla kernel won't build


cmake/Modules/FindSYCL/run_sycl.cmake
```diff
@@ -72,7 +72,7 @@ foreach(def ${SYCL_compile_definitions})
 endforeach()
 
 # string(APPEND SYCL_host_compiler_flags "\"")
-set(SYCL_host_compiler "-fsycl-host-compiler=${SYCL_host_compiler}")
+set(SYCL_host_compiler "-fsycl-host-compiler=g++-13")
```
guangyey: I think we could check here if g++ is not 13 or newer, we could disable sycl-tla build and continue building torch-xpu-ops.

src/ATen/CMakeLists.txt
```diff
@@ -2,20 +2,23 @@
 
 file(GLOB xpu_cpp "xpu/*.cpp")
 file(GLOB xpu_mkl "native/xpu/mkl/*.cpp")
-file(GLOB xpu_native_cpp "native/xpu/*.cpp" "native/sparse/*.cpp" "native/sparse/xpu/*.cpp" "native/nested/*.cpp" "native/nested/xpu/*.cpp" "native/transformers/*.cpp" "native/quantized/*.cpp")
+file(GLOB xpu_native_cpp "native/xpu/*.cpp" "native/sparse/*.cpp" "native/sparse/xpu/*.cpp" "native/nested/*.cpp" "native/nested/xpu/*.cpp" "native/transformers/*.cpp" "native/quantized/*.cpp" ${TORCH_ROOT}/aten/src/ATen/native/transformers/xpu/flash_attn/*.cpp)
 file(GLOB xpu_sycl "native/xpu/sycl/*.cpp" "native/sparse/xpu/sycl/*.cpp" "native/nested/xpu/sycl/*.cpp" "native/transformers/sycl/*.cpp" "native/quantized/sycl/*.cpp")
+file(GLOB xpu_sycltla "${TORCH_ROOT}/aten/src/ATen/native/transformers/xpu/flash_attn/sycltla/*.cpp")
```
guangyey, line 6: I can't find the folder under `${TORCH_ROOT}/aten/src/ATen/native/transformers/xpu` in https://github.com/pytorch/pytorch/tree/main/aten/src/ATen/native/transformers, is there any dependency?

cmake/Modules/FindSYCL/run_sycl.cmake
LuFinch: Yes, this is a hard code for local build.  I can add check for host complier version like `if (CMAKE_CXX_COMPILER <= gcc13) USE_SYCLTLA=OFF`

src/ATen/CMakeLists.txt
LuFinch: The PR containing `${TORCH_ROOT}/aten/src/ATen/native/transformers/xpu` depends on this build enabling PR. I will create a new PR to put sycl-tla flash attention kernel in Pytorch after this PR merged.

cmake/BuildFlags.cmake
```diff
@@ -22,141 +22,162 @@ function(CHECK_SYCL_FLAG FLAG VARIABLE_NAME)
...
-  endif()
+    if(CMAKE_BUILD_TYPE MATCHES Debug)
+      list(APPEND SYCL_HOST_FLAGS -g -fno-omit-frame-pointer -O0)
+    elseif(CMAKE_BUILD_TYPE MATCHES RelWithDebInfo)
+      list(APPEND SYCL_HOST_FLAGS -g -O2)
+    endif()
+    if(USE_PER_OPERATOR_HEADERS)
+      list(APPEND SYCL_HOST_FLAGS -DAT_PER_OPERATOR_HEADERS)
+    endif()
+    list(APPEND SYCL_HOST_FLAGS -D__INTEL_LLVM_COMPILER_VERSION=${__INTEL_LLVM_COMPILER})
```
guangyey: This is an unused flag now.

LuFinch: I saw this commit https://github.com/intel/torch-xpu-ops/commit/2713310c37b22dd21845320e6fa8cb0493d15693 remove `set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -D__INTEL_LLVM_COMPILER_VERSION=${__INTEL_LLVM_COMPILER})` but not remove `list(APPEND SYCL_HOST_FLAGS -D__INTEL_LLVM_COMPILER_VERSION=${__INTEL_LLVM_COMPILER})`..

LuFinch: Should I remove it from SYCL_HOST_FLAGS, too?

guangyey: I think yes, we already have a macro that can do this `SYCL_COMPILER_VERSION`.

LuFinch: Done

cmake/Modules/FindSYCL/run_sycl.cmake
LuFinch: Done

src/ATen/CMakeLists.txt
guangyey: It’s a bit unusual that we are building some files from PyTorch and some from torch-xpu-ops into a single .so.
Could we decouple them? For example, we could keep the implementations in torch-xpu-ops and provide a header file for PyTorch. PyTorch would then only use the APIs exposed in the header file.

cmake/Modules/FindSYCL.cmake
```diff
@@ -383,7 +382,6 @@ macro(SYCL_LINK_DEVICE_OBJECTS output_file sycl_target)
     set(SYCL_device_link_flags
         ${link_type_flag}
         ${important_host_flags}
-        ${SYCL_FLAGS}
```
guangyey, line 391: If `SYCL_FLAGS` was removed here, where `-fsycl` would be added back to `SYCL_device_link_flags`

src/ATen/CMakeLists.txt
LuFinch: For example, we could keep the implementations in torch-xpu-ops and provide a header file for PyTorch. PyTorch would then only use the APIs exposed in the header file. 

=> I used to implement sycltla sdpa with this method: put kernel in torch-xpu-ops, expose a header file and call from Pytorch. However, it is hard to maintain the code after upstream. For example, we need to prepare a Pytorch PR and a torch-xpu-ops PR when we want to change the APIs after first upstream. The torch-xpu-ops PR can't be built in CI because the Pytorch PR hasn't merged yet. The Pytorch PR can't be built in CI because the torch-xpu-ops PR hasn't merged yet.

It is more reasonable to put all code in Pytorch only or in torch-xpu-ops only. Since the SDPA_overrideable is registered in Pytorch already, I think put sycltla kernel in Pytorch intree is more convinent.

cmake/Modules/FindSYCL.cmake
LuFinch: It is set by `set(SYCL_DEVICE_LINK_FLAGS ${SYCL_LINK_FLAGS})` in `BuildFlags.cmake` while `list(APPEND SYCL_LINK_FLAGS "-fsycl")` is set in `FindSYCLToolKit.cmake`

src/ATen/CMakeLists.txt
guangyey: It doesn’t seem reasonable that a file located in PyTorch is being built into a third-party library. This design doesn’t make much sense to me.

LuFinch:
@fengyuan14 @EikanWang Could you help review and give some comments?

EikanWang (approved):

src/ATen/CMakeLists.txt
guangyey: I see.

guangyey (approved):

LuFinch:
@EikanWang @guangyey  Can we merge this PR?

guangyey:
Of course, let's move forward. We can apply a minor fix if needed.

PR #2041, opened 2025-09-15
Add symmetric memory support on XPU device

zhangxiaoli73:
PyTorch provides symmetric memory support on CUDA device. 

Accordingly, we would like to provide similar feature on XPU device.


chuanqi129:
@copilot The Python lint (flake8/ruff) check failed. Please fix the lint errors in this PR.

**Important: Do NOT update the plan in the PR description directly. Use reply comments to update the status.**

Lint job log: https://github.com/intel/torch-xpu-ops/actions/runs/25904657285

Instructions:
1. Review the lint failure log via the link above
2. Fix all lint errors in the relevant source files
3. Ensure both Python lint and Clang format checks pass
4. Use reply comments to report progress - do NOT modify the PR description plan directly

zhangxiaoli73:
@gujinghui @Chao1Han Please review

src/xccl/Signal.cpp
```diff
@@ -0,0 +1,201 @@
+#include <ATen/xpu/XPUContext.h>
+#include <comm/SYCLContext.h>
+#include <xccl/Signal.hpp>
+#include <chrono>
+
+namespace c10d::symmetric_memory {
+
+struct barrierKernel {
```
gujinghui, line 8: As the 2026.0 toolchain is deployed in the CI, can we rewrite the kernels with free func kernel style, instead of functor?

Chao1Han (approved):

src/xccl/Signal.hpp
```diff
@@ -0,0 +1,88 @@
...
+  sycl::atomic_fence(sycl::memory_order::acquire, sycl::memory_scope::system);
+  uint32_t val = *addr;
+  return val;
+}
+
+// =============================================================================
+// Put signal: wait until addr == 0, then set to 1 (release semantics)
+// =============================================================================
+
+template <std::memory_order Sem>
```
gujinghui: Why the Sem is not used anywhere?

zhangxiaoli73:
@gujinghui May I know if you have some comments? We would like to merge this PR  ASAP to not block PyTorch PR.

src/xccl/Signal.hpp
```diff
@@ -0,0 +1,88 @@
...
+  sycl::atomic_fence(sycl::memory_order::release, sycl::memory_scope::system);
+}
+
+// Load value with acquire fence (for get_signal/wait_signal)
+// Order: acquire fence first, then load to see the latest value
+inline uint32_t load_acquire(uint32_t* addr) {
+  sycl::atomic_fence(sycl::memory_order::acquire, sycl::memory_scope::system);
+  uint32_t val = *addr;
+  return val;
+}
```
gujinghui, line 36: why the scope is system? Because the user scenario is cross devices?
please give more comments to explain.

src/xccl/XPUSymmetricMemoryTypes.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+namespace c10d::symmetric_memory {
+
+constexpr size_t signal_pad_size = 2048;
```
gujinghui, line 9: please add some comments to explain 2048.

src/xccl/XPUSymmetricMemory.cpp
```diff
@@ -0,0 +1,478 @@
...
+void* XPUSymmetricMemoryAllocator::alloc(
+    size_t size,
+    int device_idx,
+    const std::optional<std::string>& group_name) {
+  size_t signal_pad_offset = at::round_up(size, 16UL);
+  size_t block_size = signal_pad_offset + get_signal_pad_size();
+
+  c10::DeviceGuard device_guard(c10::Device(c10::DeviceType::XPU, device_idx));
+  sycl::queue current_queue = at::xpu::getCurrentXPUStream().queue();
+  void* ptr = sycl::malloc_device(block_size, current_queue);
```
gujinghui, line 248: why we do not alloc the buffer from caching allocator?

src/xccl/XPUSymmetricMemory.cpp
```diff
@@ -0,0 +1,478 @@
...
+  current_queue.memset(ptr, 0, block_size);
```
gujinghui, line 249: It's OK not to wait until memset finished, right?

src/xccl/XPUSymmetricMemory.cpp
```diff
@@ -0,0 +1,478 @@
...
+  TORCH_CHECK(
+      group != nullptr,
+      "XPUSymmetricMemory::rendezvous: Could not resolve process group '",
+      group_name_,
+      "'. This can happen if rendezvous() is called before the process "
+      "group is initialized or if the group name is incorrect.");
+  auto rank = group->getRank();
+  auto world_size = group->getSize();
+  auto store = group->getStore();
+  sycl::queue current_queue = at::xpu::getCurrentXPUStream().queue();
```
gujinghui: Here should be a queue reference?

gujinghui:
> @gujinghui May I know if you have some comments? We would like to merge this PR ASAP to not block PyTorch PR.

Done. The PR is too big... Please address my comments. Thanks.

src/xccl/Signal.hpp
zhangxiaoli73: Yes, we are using it for crossing devices. So system scope is needed.

src/xccl/XPUSymmetricMemoryTypes.hpp
zhangxiaoli73: Currently, we just align with cuda pad size.

src/xccl/XPUSymmetricMemory.cpp
zhangxiaoli73: 1) Align with cuda. 2) We don't want to be affected by expand memory, so allocate from sycl directly and exchange by common IPC. 3) Leave flexibility for future platform.  If someday multicast is supported, then we must exchange by physical handle and reconstruct multicast object for this physical handle. Such process cannot be supported in caching allocator.

src/xccl/XPUSymmetricMemory.cpp
zhangxiaoli73: Yes, no problem. Queue is in-order and no need to do host blocking.

src/xccl/Signal.cpp
zhangxiaoli73: I don't know quite more about free func. Then could we land this PR first and then separate another PR for free func kernel style.  We have several kernels under xccl, suppose we need to refine them with free func kernel style.

src/xccl/Signal.hpp
zhangxiaoli73: Good cache up. Previously, we try to implement like cuda with such template. But cuda implementation is not so suitable for our PCIE connected system. We move to our own implementations and seems like leave some unclear code.

src/xccl/Signal.cpp
gujinghui: Let's align with the progress of free func kernel for other existing kernels. If all utilities for free func kernel is landed, we'd better always write new kernel in free func kernel style. If not ready, let's merge this one first.

src/xccl/Signal.hpp
gujinghui: Any usage limitation from host threads and processes? Please give more details in code comments.

src/xccl/XPUSymmetricMemoryTypes.hpp
gujinghui: It's an experience value? We do not know why cuda use this size?

src/xccl/XPUSymmetricMemory.cpp
gujinghui: "expand memory" is reasonable. please add comments to explain it in code.

src/xccl/XPUSymmetricMemoryTypes.hpp
zhangxiaoli73: I think it maybe an experience value from Meta usage. And also, 2048 is just the default and users can adjust pad size from symmetric memory API.

src/xccl/Signal.hpp
zhangxiaoli73: Those APIs are just used under symmetric memory and not directly exposed to users.

src/xccl/XPUSymmetricMemory.cpp
zhangxiaoli73: You mean using queue ref to avoid some shallow copy?

src/xccl/Signal.cpp
Chao1Han: Free function helper not ready, let`s merge it firstly.

src/xccl/Signal.hpp
Chao1Han: updated

src/xccl/XPUSymmetricMemory.cpp
gujinghui: yes. https://github.com/pytorch/pytorch/blob/7a377e41a720c2b94b53c890e5a22d621a186f6a/c10/xpu/XPUStream.h#L122

Chao1Han: updated.

zhangxiaoli73:
> > @gujinghui May I know if you have some comments? We would like to merge this PR ASAP to not block PyTorch PR.
> 
> Done. The PR is too big... Please address my comments. Thanks.

@gujinghui  Thanks for your comments and we have addressed all your comments.

gujinghui (approved):

Chao1Han (approved):

PR #2044, opened 2025-09-15
Remove unused variables

Silv3S:
- remove unused variables,
- add compiler flag to prevent this in the future


CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(PROJECT_VERSION "2.3.0")
 if(NOT WIN32)
   string(APPEND CMAKE_CXX_FLAGS " -Wno-error")
 endif()
+string(APPEND CMAKE_CXX_FLAGS " -Werror=unused-variable")
```
Silv3S, line 29:
```suggestion
  string(APPEND CMAKE_CXX_FLAGS " -Wno-error")
  string(APPEND CMAKE_CXX_FLAGS " -Werror=unused-variable")
endif()
```
Restore for linux-only, as it breaks Windows CI

CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ set(PROJECT_VERSION "2.3.0")
 # Avoid SYCL compiler error
 if(NOT WIN32)
   string(APPEND CMAKE_CXX_FLAGS " -Wno-error")
+  string(APPEND CMAKE_CXX_FLAGS " -Werror=unused-variable")
```
CuiYifeng: Enabling this option is good for the final code quality, but it also requires every compilation in development to follow this rule. Such checks would be better placed in CI.

Silv3S: Ok, I've passed this flag as env variable in linux_build CI pipeline

.github/workflows/_linux_build.yml
```diff
@@ -94,6 +94,7 @@ jobs:
       - name: Build Pytorch on ${{ needs.runner.outputs.hostname }}
         run: |
           export USE_XCCL=1
+          export CXXFLAGS="${CXXFLAGS} -Werror=unused-variable"
```
guangyey: Exporting this flag would propagate it to the entire PyTorch build, which may cause build failures in third-party components.

Silv3S: I'd opt for keeping this in CMake, because even if it affects each dev-build, it's beneficial to get rid of unused code as soon as possible. Other possibility is to have some `is CI` flag, so it would be executed only for CI builds

CMakeLists.txt
```diff
@@ -25,6 +25,9 @@ set(PROJECT_VERSION "2.3.0")
 # Avoid SYCL compiler error
 if(NOT WIN32)
   string(APPEND CMAKE_CXX_FLAGS " -Wno-error")
+  if("$ENV{IS_XPU_CI}" STREQUAL "1")
+    string(APPEND CMAKE_CXX_FLAGS " -Werror=unused-variable")
```
EikanWang, line 29: Should we always turn on this build option?

EikanWang (approved):
LGTM

.github/workflows/_linux_build.yml
EikanWang: @CuiYifeng, I think we should always follow the rule during daily development, as I don't quite understand the logic behind unused variables. What's the motivation in your mind to bypass the logic for regular development?

guangyey (approved):

CuiYifeng (approved):
LGTM.

chuanqi129 (approved):

.github/workflows/_linux_build.yml
CuiYifeng: During the development or debugging process, unused variables might be temporary in local and will be used soon. I agree that submitted code should follow this rule, but I do not recommend enforcing it locally. Since unused variables can be prevented by CI, leaving some space for the development process would be more flexible. AFAK, PyTorch currently does not require checking unused variables in each compilation locally.

CMakeLists.txt
CuiYifeng: Personally, I do not recommend always turning on this build option for the reason I mentioned.

Silv3S: I got your point @CuiYifeng - let's just keep this flag for CI builds

.github/workflows/_linux_build.yml
EikanWang: > unused variables might be temporary in local and will be used soon. Leaving some space for the development process would be more flexible.

I don't believe so. Pls. follow the process. Any unused variables should not be used.

PR #2054, opened 2025-09-17
[CI] Add deps for new LLM models

mengfei25:
Add new LLM models deps `accelerate`

disable_ut
disable_distributed

### DIFF
--- a/.github/actions/linux-testenv/action.yml
+++ b/.github/actions/linux-testenv/action.yml
@@ -57,7 +57,7 @@ runs:
       run: |
         # install pytorch
         if [ $(echo "${{ inputs.pytorch }}" |grep -w "release_wheel" -c) -ne 0 ];then
-          pip install torch torchvision torchaudio --pre --index-url https://download.pytorch.org/whl/xpu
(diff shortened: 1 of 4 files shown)


.github/actions/linux-testenv/action.yml
```diff
@@ -115,6 +115,8 @@ runs:
         pip install pandas psutil scipy pyyaml
         cd pytorch
         if [[ "${{ inputs.suite }}" == *"huggingface"* ]];then
+          # for new LLM models
+          pip install accelerate
```
chuanqi129, line 119: Do we need to fix the version?

chuanqi129 (approved):
Do we need to update the accuracy baseline files?

.github/actions/linux-testenv/action.yml
mengfei25: Looks like not, if transformers does need a specific version, it will reinstall the one needed I think.
And torchbench also does not use a specific version.

chuanqi129 (approved):

PR #2065, opened 2025-09-18
fbgemm async complete cumsum op, jagged and dense conversion ops

jiafuzha:
supported below four fbgemm ops on xpu.

fbgemm::asynchronous_complete_cumsum
fbgemm::jagged_to_padded_dense_forward
fbgemm::jagged_to_padded_dense
fbgemm::dense_to_jagged_forward 

Please make sure you have below env vars set correctly for running the UT.

```
# make sure ONEAPI_ROOT is set since it's referenced in umf's vars.sh. Otherwise, you may not able to see any device.
export ONEAPI_ROOT=.../intel/oneapi/
# DPCPP 2025.3
source .../DPCPP/env/vars.sh
source ~/intel/oneapi/mkl/latest/env/vars.sh
source .../pti_0.12/env/vars.sh
source .../umf/1.0.2/env/vars.sh
export BUILD_SEPARATE_OPS=ON
export BUILD_WITH_CPU=ON
export TORCH_XPU_ARCH_LIST='pvc'
export USE_PTI=ON
export USE_KINETO=ON
export USE_XETLA=OFF
```

### DIFF
--- a/src/ATen/native/xpu/FbgemmOps.cpp
+++ b/src/ATen/native/xpu/FbgemmOps.cpp
@@ -1,0 +1,270 @@
+#include <ATen/ATen.h>
+#include <ATen/DeviceGuard.h>
+#include <ATen/record_function.h>
+#include <torch/csrc/autograd/custom_function.h>
+#include <torch/library.h>
+#include <torch/torch.h>
+
+#include <ATen/native/xpu/sycl/FbgemmKernels.h>
+
+namespace at {
+namespace native {
+namespace xpu {
+
+#define XPU_DEVICE_GUARD(TENSOR) \
+  const OptionalDeviceGuard device_guard(device_of(TENSOR));
+
+Tensor asynchronous_complete_cumsum_xpu(const Tensor& t_in) {
+  TORCH_CHECK(t_in.is_contiguous());
+  TORCH_CHECK(t_in.dtype() == at::kInt || t_in.dtype() == at::kLong);
+  TORCH_CHECK(t_in.dim() == 1 || t_in.dim() == 2);
+
+  if (t_in.dim() == 1) {
+    Tensor t_out = at::zeros({t_in.numel() + 1}, t_in.options());
+    auto r_out = t_out.slice(0, 1);
+    at::cumsum_out(r_out, t_in, 0);
+    return t_out;
+  }
+
+  Tensor t_out = at::zeros({t_in.size(0), t_in.size(1) + 1}, t_in.options());
+  auto r_out = t_out.slice(1, 1);
+  at::cumsum_out(r_out, t_in, 1);
+  return t_out;
+}
+
+Tensor dense_to_jagged_forward_xpu(
+    const Tensor& dense,
+    const std::vector<Tensor>& offsets,
+    std::optional<at::SymInt> total_L) {
+  TORCH_CHECK(dense.is_xpu(), "value must be a xpu tensor");
+  for (auto& offset : offsets) {
+    TORCH_CHECK(offset.is_xpu(), "offset must be a xpu tensor");
+  }
+
+  const int num_jagged_dim = dense.dim() - 2;
+  TORCH_CHECK(
+      offsets.size() == static_cast<size_t>(num_jagged_dim),
+      "x_offsets.size(), ",
+      offsets.size(),
+      " != num_jagged_dim, ",
+      num_jagged_dim);
+
+  // D is the embedding dimension
+  auto D = dense.size(-1);
+
+  // If total_L is not given then compute it
+  at::SymInt total_L_computed;
+  if (total_L.has_value()) {
+    total_L_computed = total_L.value();
+  } else {
+    total_L_computed = (int64_t)offsets.back().max().item<int64_t>();
+  }
+  auto values = at::empty_symint({total_L_computed, D}, dense.options());
+  auto output = at::empty_like(values); // not used
+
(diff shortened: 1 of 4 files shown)


jiafuzha:
@majing921201 @fengyuan14 @gujinghui , All necessary fbgemm ops are supported in xpu. please help review.

src/ATen/native/xpu/FbgemmOps.cpp
```diff
@@ -0,0 +1,552 @@
...
+
+namespace at {
+namespace native {
+namespace xpu {
+
+#define XPU_DEVICE_GUARD(TENSOR) \
+  const OptionalDeviceGuard device_guard(device_of(TENSOR));
+
+Tensor asynchronous_complete_cumsum_xpu(const Tensor& t_in) {
+  TORCH_CHECK(t_in.is_contiguous());
```
fengyuan14, line 18: Is the limitation in XPU implementation aligned with CUDA?

jiafuzha: which limitation do you refer to?

src/ATen/native/xpu/FbgemmOps.cpp
```diff
@@ -0,0 +1,552 @@
...
+}
+
+} // namespace xpu
+} // namespace native
+} // namespace at
+
+namespace {
+
+TORCH_LIBRARY(fbgemm, m) {
+  m.def("asynchronous_complete_cumsum(Tensor t_in) -> Tensor");
```
fengyuan14: Is there conflict here? when FBGEMM define the same symbol?

src/ATen/native/xpu/FbgemmOps.cpp
```diff
@@ -0,0 +1,552 @@
...
+      "jagged_dense_elementwise_add_jagged_output(Tensor values, Tensor[] offsets, Tensor y) -> (Tensor, Tensor[])");
+  m.def(
+      "reorder_batched_ad_lengths(Tensor cat_ad_lengths, Tensor batch_offsets, int num_ads_in_batch, bool broadcast_lengths, int max_batch_size=0) -> Tensor");
+  m.def(
+      "reorder_batched_ad_indices(Tensor cat_ad_offsets, Tensor cat_ad_indices, Tensor reordered_cat_ad_offsets, Tensor batch_offsets, int num_ads_in_batch, bool broadcast_indices, int num_indices_after_broadcast) -> Tensor");
+  m.def(
+      "permute_2D_sparse_data(Tensor permute, Tensor lengths, Tensor indices, Tensor? weights=None, int? permuted_lengths_sum=None) -> (Tensor, Tensor, Tensor?)");
+}
+
+TORCH_LIBRARY_IMPL(fbgemm, XPU, m) {
```
fengyuan14, line 482: Please use one `TORCH_LIBRARY_IMPL` to contain all impl definitions.

test/xpu/test_fbgemm_ops_xpu.py
```diff
@@ -0,0 +1,1122 @@
+# Owner(s): ["module: intel"]
```
majing921201, line 1: Why not take CPU result as ref like PyTorch UT ?

jiafuzha: Then, we need to install fbgemm cpu which is not desired.

src/ATen/native/xpu/FbgemmOps.cpp
jiafuzha: Will do

src/ATen/native/xpu/FbgemmOps.cpp
jiafuzha: yes, it could be a problem. I'll verify and figure it out.

test/xpu/test_fbgemm_ops_xpu.py
majing921201: if you remove schema define as @fengyuan14 suggestion, then you must install fbgemm

jiafuzha: correct. But it also means our torch-xpu-ops repo has dependency to fbgemm. sounds good?

src/ATen/native/xpu/FbgemmOps.cpp
jiafuzha: As verified, fbgemm ops schema cannot be registered here. Otherwise, it'll fail fbgemm lib. And it's not proper adding fbgemm lib dep here. My solution is to add schema definition in test_fbgemm_ops_xpu.py, like below, 

`lib = torch.library.Library("fbgemm", "DEF")

lib.define("asynchronous_complete_cumsum(Tensor t_in) -> Tensor")
...`

Then, I can run all fbgemm ops test successfully.

In user side, they need to 'import fbgemm_gpu' as normal. Then they can use these xpu ops. Considering the schemas unlikely changing, it should be good for us to use official fbgemm's schema.

test/xpu/test_fbgemm_ops_xpu.py
jiafuzha: see my reply to yuanfeng's comments.

src/ATen/native/xpu/FbgemmOps.cpp
jiafuzha: done

flezaalv:
Hi @jiafuzha
 I noticed that the PR in the public repo wasn't approved or merged. Was this work incorporated into a new PR or a different branch instead?

PR #2073, opened 2025-09-18
Fix error handling for BatchLinearAlgebra Ops

CuiYifeng:
This PR is to fix the access to exception vector in error handling. Original implementation may cause a segmentation fault due to out-of-bounds access.


EikanWang:
@CuiYifeng , pls. add test cases.

majing921201 (approved):

PR #2074, opened 2025-09-18
Fix accuracy issues with CTC loss

### LINKED ISSUE #1737: [LNL][windows]Accuracy issue for test_nn.py TestNN.test_CTCLoss_long_targets on windows
### 🐛 Describe the bug

Accuracy issue for test_nn.py TestNN.test_CTCLoss_long_targets on windows

reproduce step:
pytest -v test/test_nn.py::TestNN::test_CTCLoss_long_targets

error message:
  File "C:\pytorch\third_party\torch-xpu-ops\test\xpu\../../../../test\test_nn.py", line 2751, in test_CTCLoss_long_targets
    self.assertEqual(res_cpu, res_gpu, atol=1e-4, rtol=0)
  File "C:\Users\dcai01\miniforge3\envs\xpu-test\lib\site-packages\torch\testing\_internal\common_utils.py", line 4094, in assertEqual
    raise error_metas.pop()[0].to_error(  # type: ignore[index]
AssertionError: Scalars are not close!

Expected 5601.796223907546 but got 8208.032377407448.
Absolute difference: 2606.236153499902 (up to 0.0001 allowed)
Relative difference: 0.46525008217487707 (up to 0 allowed)

To execute this test, run the following from the base repo dir:
    PYTORCH_TEST_WITH_SLOW=1 python ..\..\test\test_nn.py TestNN.test_CTCLoss_long_targets

This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0

### Versions

PyTorch: 975b4d9707e92abdc8ba2f6a33fd2b1e41114ad4
Torch XPU Ops: 4e027f1e1c560d7dc7db7eb41e48bdee5fc00707
Os: Microsoft Windows 11 Pro
CPU: Intel(R) Core(TM) Ultra 7 258V | Intel(R) Arc(TM) 140V
GPU (16GB):  32.0.101.6790
oneapi: intel.oneapi.win.deep-learning-essentials.product=2025.0.1+20

kdrozd-dev:
Fixes #1737 and the flaky backward accuracy issue uncovered by fixing the forward pass, using sycl_global_and_local_fence to sync data in global memory - log_alpha_data_.


src/ATen/native/xpu/sycl/LossCTCKernels.cpp
```diff
@@ -115,7 +115,7 @@ struct CTCLossLogAlphaKernelFunctor {
         have_three = false;
       }
       for (int64_t t = 1; t < max_input_length_; t++) {
-        item.barrier(sycl_local_fence);
+        item.barrier(sycl_global_fence);
```
EikanWang:
```suggestion
        item.barrier(sycl_global_and_local_fence);
```

src/ATen/native/xpu/sycl/LossCTCKernels.cpp
```diff
@@ -502,7 +502,7 @@ struct CTCLossBackwardLogBetaKernelFunctor {
       // now go backward in t. Note that we need to skip the last timestep that
       // we did above.
       for (int64_t t = max_input_length_ - 2; t >= 0; t--) {
-        item.barrier(sycl_local_fence);
+        item.barrier(sycl_global_fence);
```
EikanWang:
```suggestion
        item.barrier(sycl_global_and_local_fence);
```

EikanWang (approved):
CUDA uses [__syncthreads](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#synchronization-functions), which means waits until all threads in the thread block have reached this point and all global and shared memory accesses made by these threads prior to `__syncthreads()` are visible to all threads in the block.

From the semantics perspective, it is equivalent to `item.barrier(sycl_global_and_local_fence)`. 

Back to this PR, the kernel does not use `shared local memory`. So we can use `item.barrier(sycl_global_fence)` directly. It is a proper fixing.

@SanityRemnants , could you help collect the performance data to evaluate the performance impact of this PR?

fengyuan14 (approved):

fengyuan14:
@SanityRemnants The fixing is good to me. The performance drop is expected, as the global memory fence is introduced. Please correct the commit log and elaborate it, "Using global fence to sync data in global memory, `log_alpha_data_`".

SanityRemnants:
> CUDA uses [__syncthreads](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#synchronization-functions), which means waits until all threads in the thread block have reached this point and all global and shared memory accesses made by these threads prior to `__syncthreads()` are visible to all threads in the block.
> 
> From the semantics perspective, it is equivalent to `item.barrier(sycl_global_and_local_fence)`.
> 
> Back to this PR, the kernel does not use `shared local memory`. So we can use `item.barrier(sycl_global_fence)` directly. It is a proper fixing as well. If there is no performance variance between the two different fence spaces, let's use `sycl_global_and_local_fence` to save the maintenance effort.
> 
> @SanityRemnants , could you help collect the performance data to evaluate the performance impact of this PR?

There seems to be no noticeable performance difference between the 2 types so I switched to the  sycl_global_and_local_fence

PR #2076, opened 2025-09-18
Revert tracking of Work status for FlightRecorder in ProcessGroupXCCL 

frost-intel:
The callback used to track the work status in ProcessGroupXCCL was causing an unintended memory leak by maintaining the work objects and therefor the stashed tensors. For now, I'm removing the callback and I have added a unit test to ensure this memory leak isn't returning.

Fix https://github.com/intel/torch-xpu-ops/issues/2084


frost-intel:
@zhangxiaoli73 @EikanWang Can you also approve this PR for main branch?

PR #2077, opened 2025-09-18
[release/2.9] Revert tracking of Work status for FlightRecorder in ProcessGroupXCCL

frost-intel:
See #2076 , this is a cherry-pick for 2.9 release

This is a high impact bug, in many distributed applications this is a large memory leak resulting in OoM error (see https://github.com/intel/torch-xpu-ops/issues/2084)

### DIFF
--- a/src/xccl/ProcessGroupXCCL.cpp
+++ b/src/xccl/ProcessGroupXCCL.cpp
@@ -756,6 +756,12 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::collective(
   work->future_ = c10::make_intrusive<at::ivalue::Future>(
       c10::ListType::create(c10::TensorType::get()), devices);
   work->future_->markCompleted(at::IValue(*work->outputs_));
+  auto id = work->trace_id_;
+  work->future_->addCallback(
+      [id](at::ivalue::Future&) {
+        FlightRecorderXCCL::get()->retire_id(id, /*compute_duration*/ false);
+      },
+      /*use_future*/ false);
   work->blockingWait_ = blockingWait_;
 
   work->numelIn_ = 0;
@@ -866,6 +872,12 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::pointToPoint(
     work->future_ = c10::make_intrusive<at::ivalue::Future>(
         c10::ListType::create(c10::TensorType::get()), devices);
     work->future_->markCompleted(at::IValue(*work->outputs_));
+    auto id = work->trace_id_;
+    work->future_->addCallback(
+        [id](at::ivalue::Future&) {
+          FlightRecorderXCCL::get()->retire_id(id, /*compute_duration*/ false);
+        },
+        /*use_future*/ false);
 
     work->numelIn_ = work->numelOut_ = tensor.numel();
     setEnqueuedPgStatus(work);
--- a/src/xccl/ProcessGroupXCCL.cpp
+++ b/src/xccl/ProcessGroupXCCL.cpp
@@ -427,17 +427,6 @@ void ProcessGroupXCCL::setEnqueuedPgStatus(
   pgStatus_->lastEnqueuedNumelOut = work->numelOut_;
 }
 
-void ProcessGroupXCCL::setCompletedPgStatus(
-    c10::intrusive_ptr<ProcessGroupXCCL::WorkXCCL> work) {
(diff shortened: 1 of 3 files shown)


riverliuintel:
@CuiYifeng  add this fix as torch-xpu-ops cherry-pick fix list for PT2.9

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -427,17 +427,6 @@ void ProcessGroupXCCL::setEnqueuedPgStatus(
   pgStatus_->lastEnqueuedNumelOut = work->numelOut_;
 }
 
-void ProcessGroupXCCL::setCompletedPgStatus(
```
EikanWang, line 430: Why does this function lead to a memory leak?

frost-intel: It's not this function itself, rather using it in the callback. It made it so the WorkXCCL object still had a reference so it wasn't deleted properly, which meant that the tensors stashed for the allocator were still reserved.

EikanWang: Which line holds a reference?

frost-intel: Like 770 in the old code,  for example. https://github.com/intel/torch-xpu-ops/pull/2077/files#diff-29271b6f1608f7ad940c9cd242ce24dcc68bba932348e39dc0b524604cc78c6aL770

The context of the callback includes `this, work` which results in the work not being cleaned up on return.

CuiYifeng:
> @CuiYifeng add this fix as torch-xpu-ops cherry-pick fix list for PT2.9

Added.

zhangxiaoli73:
@frost-intel The issue happened as: 1)  you use work in you callback. 2) you use value in Future argument in your callback, and then output tensor will be reference again. 

For 1) I think you don't need whole work but some work status. As work status is already updated before return, so you don't need to put it to callback.
https://github.com/intel/torch-xpu-ops/blob/77cc792cd265179745d335579d233e6d4f9a2667/src/xccl/ProcessGroupXCCL.cpp#L784 

For 2) PyTorch provide option "use_future" to decide if you really need Future argument in callback.

So my suggested change is only:
```
  auto id = work->trace_id_;
  work->future_->addCallback(
      [id](at::ivalue::Future&) {
        FlightRecorderXCCL::get()->retire_id(id, /*compute_duration*/ false);
      },
       /*use_future*/ false);
```

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -767,8 +756,6 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::collective(
   work->future_ = c10::make_intrusive<at::ivalue::Future>(
       c10::ListType::create(c10::TensorType::get()), devices);
   work->future_->markCompleted(at::IValue(*work->outputs_));
-  work->future_->addCallback(
```
zhangxiaoli73, line 770: Suggested change: 
`auto id = work->trace_id_;
  work->future_->addCallback(
      [id](at::ivalue::Future&) {
        FlightRecorderXCCL::get()->retire_id(id, /*compute_duration*/ false);
      },
       /*use_future*/ false);`

src/xccl/ProcessGroupXCCL.cpp
zhangxiaoli73, line 430: Remove this API if you follow suggested change.

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -879,9 +866,6 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::pointToPoint(
     work->future_ = c10::make_intrusive<at::ivalue::Future>(
         c10::ListType::create(c10::TensorType::get()), devices);
     work->future_->markCompleted(at::IValue(*work->outputs_));
-    work->future_->addCallback([this, work](at::ivalue::Future&) {
```
zhangxiaoli73, line 882: Suggested change:

`auto id = work->trace_id_; work->future_->addCallback( [id](at::ivalue::Future&) { FlightRecorderXCCL::get()->retire_id(id, /*compute_duration*/ false); }, /*use_future*/ false);`

zhangxiaoli73 (approved):

PR #2090, opened 2025-09-22
separate comm init from getXCClComm

Chao1Han:


src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -824,7 +830,10 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::pointToPoint(
   }
 
   op_id_++;
-  auto comm = getXCCLComm(key, device, opType, p2pRank, isSendRecvSelf);
+  std::shared_ptr<xcclComm_t> comm = getXCCLComm(key);
```
zhangxiaoli73, line 856: Seems like no difference with previous code, then why to separate as two APIs/steps?

Chao1Han: Just for align nccl, To facilitate future feature integration, we can first align this part.

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -494,6 +494,16 @@ c10::intrusive_ptr<ProcessGroupXCCL::WorkXCCL> ProcessGroupXCCL::initWork(
 }
 
 std::shared_ptr<xcclComm_t> ProcessGroupXCCL::getXCCLComm(
+    const std::string& deviceKey) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (devXCCLCommMap_.find(deviceKey) != devXCCLCommMap_.end()) {
+    // Reuse the cached communicator if there is one.
+    return devXCCLCommMap_[deviceKey];
+  }
```
dvrogozh: This code searches the map twice - first time in the if condition and then to return the value. That's inefficient from performance stand point. Search first then check result and return value pointed to by iterator: 
```suggestion
  auto it = devXCCLCommMap_.find(deviceKey);
  if (it != devXCCLCommMap_.end()) {
    // Reuse the cached communicator if there is one.
    return it->second;
  }
```

src/xccl/ProcessGroupXCCL.cpp
dvrogozh: To me updated code is better structured. I am supportive of the change.

dvrogozh (approved):
LGTM

PR #2095, opened 2025-09-22
use SYCL queue.memcpy instead of copy kernel

### LINKED ISSUE #3566: Support copy engine in XPU PyTorch ATen copy OP.

Please support copy engine for PyTorch ATen copy op, it's very important for copy engine and compute overlap.

### Alternatives

_No response_

### Additional context

_No response_

jianyizh:
depend on jira GSD-3944. It's expected that d2d copy through copy engine and kernel should be the same performance in level zero 2, but I still see gaps.

Note `memcpyAsync` is only called when `memcpy_eligible`, means no cast, and all contiguous, so it's just copy without other operations. Previous implementation in `copy_kernel` calls two implementation by loops kernel: `gpu_kernel(iter, CopyScalarFunc<scalar_t>());` and `float8_copy_kernel_xpu`, since we won't do cast, they both equivalent to `queue.memcpy`


CuiYifeng (commented):
One question: this PR will introduce D2D copy performance drop on PVC (and the other old GPUs?), which is identified as a expected gap but a bug, right?

jianyizh:
> One question: this PR will introduce D2D copy performance drop on PVC (and the other old GPUs?), which is identified as a expected gap but a bug, right?

Yes, we donot care PVC

pkourdis:
> > One question: this PR will introduce D2D copy performance drop on PVC (and the other old GPUs?), which is identified as a expected gap but a bug, right?
> 

@jianyizh @CuiYifeng please don't merge yet. We need to discuss the customer impact.

src/ATen/native/xpu/Copy.cpp
```diff
@@ -83,14 +83,27 @@ void memcpyAsync(
...
+  auto q = copy_stream.queue();
+  auto device_architecture =
+      q.get_device()
+          .get_info<
+              sycl::ext::oneapi::experimental::info::device::architecture>();
+  constexpr auto unsupported_architectures =
+      std::array<sycl::ext::oneapi::experimental::architecture, 2>{
+          sycl::ext::oneapi::experimental::architecture::intel_gpu_pvc,
+          sycl::ext::oneapi::experimental::architecture::intel_gpu_pvc_vg};
+  if (std::find(
```
guangyey: Let's try use this:
```cpp
auto device_architecture = at::xpu::getDeviceProperties(copy_stream.device()).architecture;
device_architecture >= syclex::architecture::intel_gpu_pvc &&
    device_architecture <= syclex::architecture::intel_gpu_pvc_vg
```

guangyey: if (dev.ext_oneapi_architecture_is(syclex::architecture::intel_gpu_pvc) ||
    dev.ext_oneapi_architecture_is(syclex::architecture::intel_gpu_pvc_vg)) {
  // ...
}
This is better.

pkourdis:
@jianyizh thank you for addressing the PVC issue promptly.

Can we introduce an environment variable such as `TORCH_XPU_USE_COPY_ENGINE` (or similar) to explicitly enable the use of SYCL `queue.memcpy`, independent of device architecture?

While some platforms (e.g., PVC) may experience a regression in pure D2D copy performance when using `queue.memcpy`, it is still important to provide a mechanism to force this path. In particular, this enables evaluation of end-to-end model performance under scenarios where compute and data movement can overlap, allowing us to assess the true net effect beyond isolated copy benchmarks, such as in realistic workloads.

jianyizh:
> @jianyizh thank you for addressing the PVC issue promptly.
> 
> Can we introduce an environment variable such as `TORCH_XPU_USE_SYCL_MEMCPY` (or similar) to explicitly enable the use of SYCL `queue.memcpy`, independent of device architecture?
> 
> While some platforms (e.g., PVC) may experience a regression in pure D2D copy performance when using `queue.memcpy`, it is still important to provide a mechanism to force this path. In particular, this enables evaluation of end-to-end model performance under scenarios where compute and data movement can overlap, allowing us to assess the true net effect beyond isolated copy benchmarks, such as in realistic workloads.

This PR removes workaround for BMG and future device. Actually, the right thing to do is using level zero v2 or fixed jira GSD-3944 for copy engine. You can add a build flag for that in another PR. Currently, as far as I know, we don't have any env variable to control kernel behaviors for different device

pkourdis:
> > @jianyizh thank you for addressing the PVC issue promptly.
> > Can we introduce an environment variable such as `TORCH_XPU_USE_SYCL_MEMCPY` (or similar) to explicitly enable the use of SYCL `queue.memcpy`, independent of device architecture?
> > While some platforms (e.g., PVC) may experience a regression in pure D2D copy performance when using `queue.memcpy`, it is still important to provide a mechanism to force this path. In particular, this enables evaluation of end-to-end model performance under scenarios where compute and data movement can overlap, allowing us to assess the true net effect beyond isolated copy benchmarks, such as in realistic workloads.
> 
> This PR removes workaround for BMG and future device. Actually, the right thing to do is using level zero v2 or fixed jira GSD-3944 for copy engine. You can add a build flag for that in another PR. Currently, as far as I know, we don't have any env variable to control kernel behaviors for different device

Addressing Level Zero (L0) performance issues on PVC is very difficult. In some workload scenarios, it is important to evaluate whether an overall end-to-end (E2E) performance benefit can still be achieved, even when there is a drop in raw D2D copy performance.

Our main PVC customer deploys a single PyTorch build as part of the system image across compute nodes in a large-scale supercomputer, validated to scale to thousands of nodes. Maintaining multiple PyTorch builds to experiment with different behaviors is not a viable option for production releases.

A well-documented runtime flag provides a practical and flexible mechanism to enable such experimentation without requiring changes to the deployed software stack. This allows users to explore potential E2E performance gains in real workloads while preserving deployment simplicity and consistency.

src/ATen/native/xpu/Copy.cpp
jianyizh: done

guangyey (approved):

test/xpu/test_torch_xpu.py
```diff
@@ -12650,16 +12650,14 @@ def test_swap_basic(self):
             if t1.is_floating_point():
                 t3 = t1.clone().detach().requires_grad_(True)
                 out = t3 * 2
+                torch.utils.swap_tensors(t3, t2)
```
guangyey, line 12653: Is this code change unintended? It seems unrelated to this PR.

jianyizh: I see UT fail in CI and it is different compare to its pytorch version. So I changed to align pytorch

guangyey: Thanks, I got it. I remember there is a PR already submitted to fix this issue, but I can't find it due to too many PRs created recently. Anyway, the latter PR wants to land needs to rebase.

LuFinch (approved):

CuiYifeng (approved):

chuanqi129:

jianyizh:
> > > @jianyizh thank you for addressing the PVC issue promptly.
> > > Can we introduce an environment variable such as `TORCH_XPU_USE_SYCL_MEMCPY` (or similar) to explicitly enable the use of SYCL `queue.memcpy`, independent of device architecture?
> > > While some platforms (e.g., PVC) may experience a regression in pure D2D copy performance when using `queue.memcpy`, it is still important to provide a mechanism to force this path. In particular, this enables evaluation of end-to-end model performance under scenarios where compute and data movement can overlap, allowing us to assess the true net effect beyond isolated copy benchmarks, such as in realistic workloads.
> > 
> > 
> > This PR removes workaround for BMG and future device. Actually, the right thing to do is using level zero v2 or fixed jira GSD-3944 for copy engine. You can add a build flag for that in another PR. Currently, as far as I know, we don't have any env variable to control kernel behaviors for different device
> 
> Addressing Level Zero (L0) performance issues on PVC is very difficult. In some workload scenarios, it is important to evaluate whether an overall end-to-end (E2E) performance benefit can still be achieved, even when there is a drop in raw D2D copy performance.
> 
> Our main PVC customer deploys a single PyTorch build as part of the system image across compute nodes in a large-scale supercomputer, validated to scale to thousands of nodes. Maintaining multiple PyTorch builds to experiment with different behaviors is not a viable option for production releases.
> 
> A well-documented runtime flag provides a practical and flexible mechanism to enable such experimentation without requiring changes to the deployed software stack. This allows users to explore potential E2E performance gains in real workloads while preserving deployment simplicity and consistency.

@pkourdis Got it. Please create PR for runtime env flag and document it

Issue #2086, opened 2025-09-19
nd_item::barrier has been deprecated


EikanWang:
As the title, the barrier function of the nd_item class has been deprecated in the SYCL spec 2020. In torch-xpu-ops, we need to replace `nd_item::barrier` with `group_barrier`. By the way, we need to check if there is any compile option to expose all the deprecated functions.
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#group-barrier

cc @fengyuan14

dvrogozh:
Prepared https://github.com/intel/torch-xpu-ops/pull/2096. I don't think I see any warnings from DPC++ compiler for `nd_item::barrier`. Do you know if they are actually throwing deprecations for this one? and in which version?

CuiYifeng:
Closing this issue since deprecated `nd_item::barrier` has been resolved by https://github.com/intel/torch-xpu-ops/pull/2576.

PR #2096, opened 2025-09-22
build: enable SYCL warnings on Linux

dvrogozh:
Some versions of DPC++ compiler pass paths to SYCL headers as user include paths (`-I`) rather than system paths (`-isystem`). This makes host compiler to report warnings encountered in the SYCL headers, such as deprecated warnings, even if warned API is not actually used in the program. We expect that this issue will be addressed in the later version of DPC++ compiler. To workaround the issue we wrap paths to SYCL headers in `-isystem`.

disable_ut
disable_e2e
disable_distributed


dvrogozh:
@EikanWang, this PR is to illustrate discussion in https://github.com/intel/torch-xpu-ops/issues/2086#issuecomment-3313747817. This adds a WA to suppress SYCL header warnings on Linux, but leave SYCL warnings for our own code which we might actually wish to look into. Note that I still left suppression for `-Wno-sign-compare` as there are too many of these coming from our own code. There are few `-Wdeprecated` we are stepping into and quite few `-Wattributes`. Please, take a look if we want to proceed with this PR and start fixing these warnings in our code.

PR #2105, opened 2025-09-24
Add exception to float->int conversion test case

ppapierk:
Downcasting float32 values beyond target type value range is an undefined behaviour according to torch documentation.


ppapierk:
@daisyden I have added unskip, and I am waiting for review.

PR #2108, opened 2025-09-25
[Sycl free func] Refactor index_select OP

NeoZhangJianyu:
Refactor index_select OP by SYCL free function style.
Unit test is passed in local cases.


NeoZhangJianyu:
This PR just refactor the current implementation by sycl free func style.
When merge to main branch, it will be rebased.

Close this PR due to wrong target branch.

Thank you for your review!

PR #2139, opened 2025-10-09
[sycl-free-func] refactor for arange OP

NeoZhangJianyu:
Refactor arange OP by SYCL free function style.
Unit test is passed in local cases.


NeoZhangJianyu:
cancel

PR #2156, opened 2025-10-10
Unskip working torch.special.logit unit test

Silv3S:
`test_compare_cpu_logit_xpu_bfloat16` test for `torch.special.logit` bfloat16/float16 has permissive tolerance set in [upstream test config](https://github.com/pytorch/pytorch/blob/d67b2799dc84d0f96208a9bc28bf19d712518737/torch/testing/_internal/common_methods_invocations.py#L21457). It was previously skipped due to accuracy difference between XPU and CPU, but now it passes. There was no need to patch kernel, as differences are within tolerance. Test passes locally

### DIFF
--- a/.github/actions/linux-uttest/action.yml
+++ b/.github/actions/linux-uttest/action.yml
@@ -154,9 +154,6 @@ runs:
           echo -e "[ERROR] XCCL is not enabled"
           exit 1
         fi
-        export CCL_ROOT=$(dirname $(which python))/../
-        export PATH="${CCL_ROOT}/bin/libfabric:${PATH}"
-        export LD_LIBRARY_PATH="${CCL_ROOT}/lib:${LD_LIBRARY_PATH}"
         python run_distributed.py \
           2> ${{ github.workspace }}/ut_log/xpu_distributed/xpu_distributed_test_error.log | \
           tee ${{ github.workspace }}/ut_log/xpu_distributed/xpu_distributed_test.log
--- a/.github/scripts/lintrunner.sh
+++ b/.github/scripts/lintrunner.sh
@@ -8,7 +8,7 @@ if ! uv --help > /dev/null 2>&1; then
 fi
 uv venv lint --python 3.12 --clear
 source lint/bin/activate
-uv pip install -U pip setuptools wheel
+uv pip install -U pip setuptools wheel pyyaml typing-extensions
 
 # Use uv to speed up lintrunner init
 uv pip install ruamel.yaml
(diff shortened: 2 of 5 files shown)


CuiYifeng:
@Silv3S I noticed that community also perceives the inconsistency between CPU and CUDA but there is currently no clear solution. This PR is a temporary solution, right?

Silv3S:
@EikanWang Our current kernel is aligned with CUDA, but it feels wrong to execute bf16/fp16 operator in fp32 precision. We lose benefits of low precision by using extra casts and computing it as normal floats. In my opinion proper solution is to either 1) calculate everything in original precision or 2) keep the T_ACC for sub and div but do log in original precision.

graphviz (1)


@CuiYifeng Thanks for pointing to community PR https://github.com/pytorch/pytorch/issues/140252, I wasn't aware that someone else encountered this bug. I think we should align to CPU regarding expected accuracy

riverliuintel:
@EikanWang @CuiYifeng  please review this PR again?

CuiYifeng:
This PR can be landed in the following two situations:

1. as a temporary solution when the accuracy difference between CPU and XPU causes breaking problems
2. as a formal solution when CUDA implementation is proven to be incorrect and CPU implementation is proven to be the ground truth.

Pending.

Silv3S:
Thanks for feedback. My question was, why would we take CUDA as ground truth for accuracy if all tests compare XPU to CPU.

1. It's not needed to be merged as **temporary** solution, because these are not breaking changes. It only affects one unit test, when it's configured for strict tolerance like 1e-5.

2. What kind of proof would be needed to accept this patch as formal solution? I think that attached tables are showing the accuracy issue clearly. The problem is (regardless of using CUDA or XPU) that wrapper for `log(x/(1-x))` gives different results than calling subtraction, division and log operators individually. This is caused by casting intermediate tensors to higher precision.

Mentioned CUDA issue https://github.com/pytorch/pytorch/issues/140252 has the same casting problem and it's not resolved yet. I can gather more data and submit patch to upstream kernel to get their opinion. Then (if accepted) all 3 backends would have same accuracy for lower precision torch.logit.

Silv3S:
Thank you for explanation @CuiYifeng, @EikanWang
As discussed I reverted changes to logit kernel. We can enable the upstream test even without any patches as it has very permissive tolerance `5e-1` set in [upstream test config](https://github.com/pytorch/pytorch/blob/d67b2799dc84d0f96208a9bc28bf19d712518737/torch/testing/_internal/common_methods_invocations.py#L21457)

Silv3S:
Done, please merge

PR #2175, opened 2025-10-15
add time event support

Chao1Han:
Refer https://github.com/pytorch/pytorch/commit/ab557421a47, Implemented the event cache, added the start event, and enabled support for event timing (as event timing no longer causes hangs after version 2021.17). Its usage scenario lies in `getDurationFromEvent`, which is part of the FR feature triggered by the watchdog thread; therefore, no test case is currently available.

Usage
```
    pg = dist.distributed_c10d._get_default_group()
    pg._enable_collectives_timing()
    x = torch.ones([2, 2]).to(device)
    num_repeats = 10
    for _ in range(num_repeats):
        dist.all_reduce(x)
    time.sleep(1)
    t = pickle.loads(torch._C._distributed_c10d._dump_xccl_trace())
    for seq in range(num_repeats):
        duration = t["entries"][seq]["duration_ms"]
        print(duration)

### DIFF
--- a/src/xccl/ProcessGroupXCCL.cpp
+++ b/src/xccl/ProcessGroupXCCL.cpp
@@ -273,8 +273,9 @@ ProcessGroupXCCL::WorkXCCL::WorkXCCL(
         : nullptr;
     xcclEndEvent_ = XPUEventCache::get(device.index())->create(enableTiming);
   } else {
-    xcclStartEvent_ =
-        enableTiming ? std::make_shared<at::xpu::XPUEvent>(enableTiming ) : nullptr;
+    xcclStartEvent_ = enableTiming
+        ? std::make_shared<at::xpu::XPUEvent>(enableTiming)
+        : nullptr;
     xcclEndEvent_ = std::make_shared<at::xpu::XPUEvent>(enableTiming);
   }
   stashed_for_allocator_safety_ = std::make_shared<TensorShelf>();
@@ -927,7 +928,22 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::pointToPoint(
   syncStream(device, xcclEventsMap_[key], stream);
 
   c10::intrusive_ptr<ProcessGroupXCCL::WorkXCCL> work;
-  if (!coalescing_state_) {
+  if (coalescing_state_) {
+    FlightRecorderXCCL::get()->record(
+        local_id_,
+        std::make_tuple(pg_uid_, pg_desc_), // PG name tuple
+        seqCollective_,
+        seqP2P_,
+        op_id_,
+        profilingTitle,
+        {tensor},
+        {tensor},
+        nullptr,
+        nullptr,
+        options_->timeout,
+        pgStatus_,
+        true);
+  } else {
     work = initWork(device, rank_, opType, true, profilingTitle, {tensor}, {});
     work->outputs_ = std::make_shared<std::vector<at::Tensor>>();
     work->outputs_->push_back(tensor);
@@ -943,22 +959,6 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::pointToPoint(
         {tensor},
         work->xcclStartEvent_.get(),
         work->xcclEndEvent_.get(),
-        options_->timeout,
-        pgStatus_,
-        true);
-
-  } else {
-    FlightRecorderXCCL::get()->record(
-        local_id_,
-        std::make_tuple(pg_uid_, pg_desc_), // PG name tuple
-        seqCollective_,
-        seqP2P_,
-        op_id_,
-        profilingTitle,
-        {tensor},
-        {tensor},
-        nullptr,
-        nullptr,
         options_->timeout,
         pgStatus_,
         true);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/xccl/ProcessGroupXCCL.cpp [hunk 1]
```diff
@@ -258,25 +258,39 @@ ProcessGroupXCCL::WorkXCCL::WorkXCCL(
...
+      timingEnabled_(enableTiming) {
+  if (xpuEventCacheEnabled) {
+    xcclStartEvent_ = enableTiming
+        ? XPUEventCache::get(device.index())->create(enableTiming)
+        : nullptr;
+    xcclEndEvent_ = XPUEventCache::get(device.index())->create(enableTiming);
+  } else {
+    xcclStartEvent_ =
+        enableTiming ? std::make_shared<at::xpu::XPUEvent>(1) : nullptr;
+    xcclEndEvent_ = std::make_shared<at::xpu::XPUEvent>(enableTiming ? 1 : 0);
```
guangyey:
```suggestion
    xcclEndEvent_ = std::make_shared<at::xpu::XPUEvent>(enableTiming);
```

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -258,25 +258,39 @@ ProcessGroupXCCL::WorkXCCL::WorkXCCL(
...
+      isP2P_(isP2P),
+      timingEnabled_(enableTiming) {
+  if (xpuEventCacheEnabled) {
+    xcclStartEvent_ = enableTiming
+        ? XPUEventCache::get(device.index())->create(enableTiming)
+        : nullptr;
+    xcclEndEvent_ = XPUEventCache::get(device.index())->create(enableTiming);
+  } else {
+    xcclStartEvent_ =
+        enableTiming ? std::make_shared<at::xpu::XPUEvent>(1) : nullptr;
```
guangyey:
```suggestion
        enableTiming ? std::make_shared<at::xpu::XPUEvent>(enableTiming ) : nullptr;
```

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -2059,8 +2112,8 @@ c10::DeviceIndex ProcessGroupXCCL::guessDeviceId() const {
   } else if (!usedDeviceIdxs_.empty()) {
     return *usedDeviceIdxs_.begin();
   }
-  int devIdx =
-      static_cast<int16_t>(globalRank() % at::detail::getXPUHooks().getNumGPUs());
+  int devIdx = static_cast<int16_t>(
+      globalRank() % at::detail::getXPUHooks().getNumGPUs());
```
guangyey:
```suggestion
      globalRank() % at::xpu::device_count());
```

src/xccl/XPUEventCache.cpp
```diff
@@ -0,0 +1,43 @@
...
+  {
+    std::lock_guard<std::mutex> lock(cacheMutex_);
+    auto& events = eventsArray_[timing ? 1 : 0];
+    // If we still have events in the cache, we reuse it. Otherwise, we create a
+    // new one.
+    if (!events.empty()) {
+      event = events.front();
+      events.pop_front();
+    } else {
+      event = new at::xpu::XPUEvent(timing ? 1 : 0);
```
guangyey:
```suggestion
      event = new at::xpu::XPUEvent(timing);
```

guangyey (commented):
I’m curious about how much performance improvement we could get by using the event cache. I guess nothing...
Overall LGTM.
I need some more time to go through the code.

src/xccl/XPUEventCache.cpp
```diff
@@ -0,0 +1,43 @@
...
+  {
+    std::lock_guard<std::mutex> lock(cacheMutex_);
+    auto& events = eventsArray_[timing ? 1 : 0];
+    // If we still have events in the cache, we reuse it. Otherwise, we create a
+    // new one.
+    if (!events.empty()) {
+      event = events.front();
+      events.pop_front();
+    } else {
+      event = new at::xpu::XPUEvent(timing);
```
frost-intel, line 26: Are there any plans to use a generic or templatized EventCache? This looks like there is only a single-line difference from CUDAEventCache.hpp/cpp. Even if it's not for this PR.

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -881,13 +926,9 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::pointToPoint(
   auto cclstream = xcclStreamsMap_.at(key).second;
   syncStream(device, xcclEventsMap_[key], stream);
 
-  if (enableNanCheck_ && opType == OpType::SEND) {
-    checkForNan(tensor, stream);
-  }
-
+  c10::intrusive_ptr<ProcessGroupXCCL::WorkXCCL> work;
   if (!coalescing_state_) {
```
frost-intel: Could we swap the order of the if/else block, so we check `if (coalescing_state_)`, just to match ProcessGroupNCCL?

frost-intel (approved):
One minor nit recommended, but other than that LGTM.

src/xccl/ProcessGroupXCCL.cpp
Chao1Han: of course, let us match ProcessGroupNCCL.

src/xccl/XPUEventCache.cpp
Chao1Han: You mean using a shared event cache for both CUDA and XPU, right? I initially planned to do that, but @guangyey  told me that `c10::Event` cannot be converted into either `at::xpu::XPUEvent` or `at::cuda::CUDAEvent`.

PR #2190, opened 2025-10-20
Support FP8 in op flip, index_put, and index.Tensor

chunhuanMeng:
To solve https://github.com/intel/torch-xpu-ops/issues/2207
Extends support for float8 data types across various XPU tensor indexing and transformation kernels, ensuring these operations are compatible with the new types. It also adds a regression test for flipping float8 tensors and removes the skip for float8 indexing tests.

**Float8 type support:**

* Updated dispatch macros in `XPUScalar.cpp` and `Indexing.cpp` to include `AT_FLOAT8_TYPES`, enabling float8 support in scalar extraction, indexing, index_put, and deterministic index_put kernels.
* Modified `flip_kernel` in `TensorTransformationsKernels.cpp` to support float8 and barebones unsigned types, updating the dispatch mechanism accordingly.
* Included the new dispatch header `Dispatch_v2.h` for the updated dispatch macros.

**Testing improvements:**

* Added a regression test for flipping float8 tensors in `test_index_and_index_put.py` to verify correctness of the operation on XPU.
* Removed the skip for float8 tests in `test_indexing_xpu.py`, re-enabling these tests now that support is implemented.

### DIFF
--- a/src/ATen/native/xpu/sycl/Indexing.cpp
+++ b/src/ATen/native/xpu/sycl/Indexing.cpp
@@ -713,7 +713,12 @@ void index_put_deterministic_kernel(
                 accumulate);
           }),
           AT_EXPAND(AT_ALL_TYPES_AND_COMPLEX),
-          AT_EXPAND(AT_FLOAT8_TYPES),
+          // TODO: Enable AT_FLOAT8_DTYPES after accumulation behavior is
+          // cleared for float8 dtypes.
+          kFloat8_e4m3fn,
+          kFloat8_e5m2,
+          kFloat8_e4m3fnuz,
+          kFloat8_e5m2fnuz,
           kComplexHalf,
           kHalf,
           kBool,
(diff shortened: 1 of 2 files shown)


src/ATen/native/xpu/sycl/Indexing.h
```diff
@@ -835,7 +835,7 @@ struct IndexPutDeterministicKernelFunctor {
       int64_t idx_orig = indices_[inner_idx];
       int64_t v_gid = idx_orig * stride_ + v_stride;
       if (accumulate_) {
-        acc += (accscalar_t)c10::load(&value_[v_gid]);
+        acc = acc + (accscalar_t)c10::load(&value_[v_gid]);
```
CuiYifeng: This change is for `Float8_e8m0fnu`, right? My suggestion is not to enable `Float8_e8m0fnu` for `index_put_deterministic_kernel` at the time being.

src/ATen/native/xpu/sycl/Indexing.cpp
```diff
@@ -711,17 +711,19 @@ void index_put_deterministic_kernel(
                 strideBefore,
                 nElemBefore,
                 accumulate);
-          });
+          }),
+          AT_EXPAND(AT_ALL_TYPES_AND_COMPLEX),
+          AT_EXPAND(AT_FLOAT8_TYPES),
```
CuiYifeng: For the change about "acc +=":
```suggestion
          kFloat8_e4m3fn,
          kFloat8_e5m2,
          kFloat8_e4m3fnuz,
          kFloat8_e5m2fnuz,
```

yucai-intel (approved):

PR #2203, opened 2025-10-21
Fix ProcessGroupXCCL options initialization

frost-intel:
Group UID needs to be set before creating the LogPrefix, and the initialization list order is fixed now.

In addition, we remove the redefinition of certain members in ProcessGroupXCCL.hpp which then allows the existing python bindings to work properly.

### DIFF
--- a/src/xccl/ProcessGroupXCCL.cpp
+++ b/src/xccl/ProcessGroupXCCL.cpp
@@ -369,7 +369,7 @@ ProcessGroupXCCL::ProcessGroupXCCL(
     int size,
     c10::intrusive_ptr<Options> options)
     : Backend(rank, size),
-      store_(store),
+      store_(std::move(store)),
       xcclCommCounter_(0),
       local_id_(process_group_id++),
       options_(std::move(options)) {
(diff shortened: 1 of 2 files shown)


dvrogozh (commented):
> Upstream [PR](https://github.com/pytorch/pytorch/pull/166022/files) will cause options to be properly set.

@frost-intel, can you, please, clarify how your PR relates with the upstream PR mentioned above? Is upstream PR a prerequisite to this one or vice versa? i.e. which one should be merged first or they are independent?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -370,14 +370,14 @@ ProcessGroupXCCL::ProcessGroupXCCL(
     c10::intrusive_ptr<Options> options)
     : Backend(rank, size),
       store_(store),
```
guangyey: Could we use `std::move` here?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -370,14 +370,14 @@ ProcessGroupXCCL::ProcessGroupXCCL(
...
-      options_(std::move(options)),
       xcclCommCounter_(0),
-      local_id_(process_group_id++) {
+      local_id_(process_group_id++),
+      options_(std::move(options)) {
```
guangyey, line 375: May I know if the order of `options_` doesn't matter here, I guess yes.

frost-intel: Order shouldn't matter here, just moving to suppress compiler warning.

PR #2204, opened 2025-10-22
Few CMake improvements

Silv3S:
- align torch-xpu-ops version to match target PyTorch version
- simplify install_xpu_headers macro and add `CONFIGURE_DEPENDS` to track these headers


guangyey (commented):
Overall LGTM. The only thing we need to discuss is the removal of `BUILD_SPLIT_KERNEL_LIB`. As I understand it, this flag was originally used to
1. Work around the binary size when too many AOT targets were enabled, however, this is no longer necessary since the compiler now supports device code suppression feature.
2. Facilitate debugging - some developers might still use this flag to debug. It is useful for reducing incremental build time, especially when debugging ATen XPU ops code.

Personally, I haven't used `BUILD_SPLIT_KERNEL_LIB` recently, so I can't decide whether it should be removed. Let's ask for opinions from others.
I suggest you split this PR into two separate ones. I can help land the other changes, and we can keep the `BUILD_SPLIT_KERNEL_LIB` removal PR for further discussion.

Silv3S:
Thanks for review. I agree that it will be better to split this PR and continue discussion about `BUILD_SPLIT_KERNEL_LIB` removal. I assumed it's not necessary since new compilers don't have such limitations and also noticed that this flag can't be used for debug builds without customizing CMake.
```CMake
if(CMAKE_BUILD_TYPE MATCHES "(Debug|RelWithDebInfo)")
  set(BUILD_SEPARATE_OPS TRUE)
endif()
```
For debug builds `BUILD_SEPARATE_OPS` takes precedence, making `BUILD_SPLIT_KERNEL_LIB` block unreachable
```CMake
if(BUILD_SEPARATE_OPS)
  ...
elseif(BUILD_SPLIT_KERNEL_LIB OR __INTEL_LLVM_COMPILER LESS 20250004 OR ICX_DATE LESS 20241205)
  ...
else()
  ...
endif()
```

Continued in https://github.com/intel/torch-xpu-ops/pull/2208

guangyey:
Land this due to the low risk.

PR #2205, opened 2025-10-22
Fused RMSNorm implementation

guangyey:
# Motivation
Refer to https://github.com/pytorch/pytorch/pull/153666, enable fused RMSNorm support on XPU.

# Additional Context
## Registered
These two ops (`_fused_rms_norm_xpu` and `_fused_rms_norm_backward_xpu`)  must be registered upstream in PyTorch, not in torch-xpu-ops, because registering them in torch-xpu-ops would cause the AOTI C shim generator to emit `aoti_torch_xpu__fused_rms_norm`, which conflicts with the symbol already generated by PyTorch for the same op.
- The first symbol is code-generated by PyTorch at [Codegen.cmake#L312-L320](https://github.com/pytorch/pytorch/blob/0b2f7f46e295421fbaef38554cf9a4242efd9e69/cmake/Codegen.cmake#L312-L320). At that time, XPU did not yet have its implementation, so it fall back to `rms_norm_composite`, see [native_functions.yaml#L3349-L3353]( https://github.com/pytorch/pytorch/blob/0b2f7f46e295421fbaef38554cf9a4242efd9e69/aten/src/ATen/native/native_functions.yaml#L3349-L3353) The codegend symbol (leveraging **compositeimplicitautograd::_fused_rms_norm**) in `c_shim_xpu.cpp` looks like
```cpp
AOTITorchError aoti_torch_xpu__fused_rms_norm(AtenTensorHandle input, const int64_t* normalized_shape, int64_t normalized_shape_len_, AtenTensorHandle* weight, double* eps, AtenTensorHandle* ret0, AtenTensorHandle* ret1) {
    AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE({
        auto tmp_result = at::compositeimplicitautograd::_fused_rms_norm(
            resolve_tensor_dispatch_flags(input), pointer_to_list<int64_t>(normalized_shape, normalized_shape_len_), resolve_tensor_dispatch_flags(weight), pointer_to_optional<double>(eps)
        );
		*ret0 = new_tensor_handle(std::move(std::get<0>(tmp_result)));
		*ret1 = new_tensor_handle(std::move(std::get<1>(tmp_result)));
    });
}
```
- The second symbol is code-generated by torch-xpu-ops at [Codegen.cmake#L107-L133](https://github.com/intel/torch-xpu-ops/blob/722f7ee0417bc9761c5c845d24429c3fc99d628a/cmake/Codegen.cmake#L107-L133). If these two ops are registed in torch-xpu-ops, the extended `c_shim_xpu.cpp` (the extended verion generated separately in torch-xpu-ops) would generate a duplicate symbol (leveraging **xpu::_fused_rms_norm_xpu**):
```cpp
AOTITorchError aoti_torch_xpu__fused_rms_norm(AtenTensorHandle input, const int64_t* normalized_shape, int64_t normalized_shape_len_, AtenTensorHandle* weight, double* eps, AtenTensorHandle* ret0, AtenTensorHandle* ret1) {
    AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE({
        auto tmp_result = at::xpu::_fused_rms_norm_xpu(
            resolve_tensor_dispatch_flags(input), pointer_to_list<int64_t>(normalized_shape, normalized_shape_len_), resolve_tensor_dispatch_flags(weight), pointer_to_optional<double>(eps)
        );
		*ret0 = new_tensor_handle(std::move(std::get<0>(tmp_result)));
		*ret1 = new_tensor_handle(std::move(std::get<1>(tmp_result)));
    });
}
```
To avoid symbol collisions, these two ops (`_fused_rms_norm_xpu` and `_fused_rms_norm_backward_xpu`) must be registered in PyTorch.

## Test Case
We add the UT in https://github.com/pytorch/pytorch/pull/177721
See https://github.com/pytorch/pytorch/pull/177721/changes#diff-90dbc2d4fedeae4036c450fc822bb422e97b4a8e576138d6811827a0c8ceeb13R2774-R2845

## Calculation: RMSNorm vs LayerNorm
RMSNorm drops mean-centering and the bias term, normalizing by root-mean-square only, which makes it cheaper and simpler than LayerNorm. This is why modern LLMs (LLaMA, Mistral, etc.) prefer RMSNorm.

###  LayerNorm forward:

$$
\mu = \frac{1}{N} \sum x_i
$$

$$
\mathrm{Var}(x) = \frac{1}{N} \sum_{i=1}^{N} (x_i - \mu)^2
$$

$$
\mathrm{rstd} = \frac{1}{\sqrt{{\mathrm{Var}(x) + \epsilon}}}
$$

$$
y_i = \gamma_i \cdot ({x_i - \mu}) \cdot \mathrm{rstd} + \beta_i
$$

### RMSNorm forward:

$$
\mathrm{rstd} = \frac{1}{\sqrt{\frac{1}{N} \sum_{i=1}^{N} x_i^2 + \epsilon}}
$$

$$
y_i = \gamma_i \cdot x_i \cdot \mathrm{rstd}
$$

### LayerNorm backward:

$$
\frac{\partial L}{\partial x_i} = \frac{\mathrm{rstd}}{N} \Big(N\gamma_i \frac{\partial L}{\partial y_i} - \mathrm{stats_{x1}} - (x_i - \mu) \mathrm{rstd} \cdot \mathrm{stats_{x2}}\Big)
$$

$$
\mathrm{stats_{x1}} = \sum_j \gamma_j \frac{\partial L}{\partial y_j}
$$

$$
\mathrm{stats_{x2}} = \sum_j \gamma_j \frac{\partial L}{\partial y_j} \cdot (x_j - \mu) \cdot \mathrm{rstd}
$$

$$
\frac{\partial L}{\partial \beta_j} = \sum_{\text{batch}} \frac{\partial L}{\partial y_{ij}}
$$

$$
\frac{\partial L}{\partial \gamma_j} = \sum_{\text{batch}} \frac{\partial L}{\partial y_{ij}} \cdot \(x_{ij} - \mu_i) \cdot \mathrm{rstd_i}
$$

### RMSNorm backward:

$$
\frac{\partial L}{\partial x_i} = \frac{\mathrm{rstd}}{N} \Big(N\gamma_i \frac{\partial L}{\partial y_i} - x_i \cdot \mathrm{rstd} \cdot \mathrm{stats_{x2}}\Big)
$$

$$
\mathrm{stats_{x2}} = \sum_j \gamma_j \frac{\partial L}{\partial y_j} \cdot x_j  \cdot \mathrm{rstd}
$$

$$
\frac{\partial L}{\partial \gamma_j} = \sum_{\text{batch}} \frac{\partial L}{\partial y_{ij}} \cdot x_{ij} \cdot \mathrm{rstd_i}
$$


src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -214,8 +225,12 @@ struct RowwiseMomentsFunctor : public __SYCL_KER_CONFIG_CONVENTION__ {
       T_ACC m1;
       T_ACC m2;
       std::tie(m2, m1) = welford_op.project(val);
-      mean_[i] = m1;
-      rstd_[i] = c10::xpu::compat::rsqrt(m2 + eps_);
+      if constexpr (!rms_norm) {
+        mean_[i] = m1;
+        rstd_[i] = c10::xpu::compat::rsqrt(m2 + eps_);
```
jianyizh, line 230: Does CUDA also use c10 compat rsqrt?

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -333,46 +354,54 @@ struct WelfordDataLN {
...
-                                 // accurate but noticeably faster
-  return {
-      static_cast<float>(new_mean),
-      static_cast<float>(curr_sum.sigma2 + delta * (val - new_mean)),
-      static_cast<float>(new_count)};
+  if constexpr (!rms_norm) {
+    U delta = val - curr_sum.mean;
+    U new_count = curr_sum.count + 1.f;
+    // proper division is slow, this is less accurate but noticeably faster
+    U new_mean = curr_sum.mean + delta * (1.f / new_count);
```
jianyizh, line 363: `delta / new_count` is slower than `delta * (1.f / new_count)`? Does it calls math.inv instruction? We should align accuracy with cuda first.

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
guangyey: CUDA uses `c10::cuda::compat`, refer to https://github.com/pytorch/pytorch/blob/f91dbda116d7a2672ae7a6386b1baebaa25708d8/aten/src/ATen/native/cuda/layer_norm_kernel.cu#L89-L94

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
guangyey: Refer to CUDA https://github.com/pytorch/pytorch/blob/f91dbda116d7a2672ae7a6386b1baebaa25708d8/aten/src/ATen/native/cuda/layer_norm_kernel.cu#L145
I guess (1.f/new_count) will call reciprocal directly, and it would run faster than division. I will write a simple reproducer to see if it is as I imagined.

src/ATen/native/xpu/LayerNorm.cpp
```diff
@@ -197,6 +197,115 @@ REGISTER_XPU_DISPATCH(LayerNormKernel, &xpu::layer_norm_kernel);
...
+                         LEGACY_CONTIGUOUS_MEMORY_FORMAT);
+  }
+
+  if (M > 0 && N > 0) {
+    // rstd is stored with stat_shape (e.g. [B, 1, 1]); flatten to [M] for
+    // kernel
+    Tensor rstd_flattened = rstd.contiguous().view({M});
+    xpu::rms_norm_backward_kernel(
+        dY, *X, rstd_flattened, *gamma, M, N, &dX, &dgamma);
+  }
```
CuiYifeng, line 301: Why we need double checking for `M > 0 && N > 0` here?

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -1308,12 +1433,67 @@ void layer_norm_backward_kernel(
...
+void rms_norm_backward_kernel(
+    const Tensor& dY,
+    const Tensor& X,
+    const Tensor& rstd,
+    const Tensor& gamma,
+    int64_t M,
+    int64_t N,
+    Tensor* dX,
+    Tensor* dgamma) {
+  if (M > 0 && N > 0) {
```
CuiYifeng: Please note this checking as mentioned.

src/ATen/native/xpu/LayerNorm.cpp
guangyey: Good catch! Updated.

PR #2224, opened 2025-10-28
[CI] Add CI gating based on E2E performance test results check

mengfei25:
1. use nightly as reference
2. performance gates PR
3. fix torchbench deps

disable_build
disable_ut
disable_distributed


chuanqi129 (approved):

mengfei25:
Performance is not very stable, there are some fluctuation, I test several times and some models maybe dropped ~15%.
Such as https://github.com/intel/torch-xpu-ops/actions/runs/18935790442

chuanqi129:
> Performance is not very stable, there are some fluctuation, I test several times and some models maybe dropped ~15%. Such as https://github.com/intel/torch-xpu-ops/actions/runs/18935790442

How much models? Can we ignore the unstable models in the check? @mengfei25

Issue #2207, opened 2025-10-23
Enable FP8/MXFP8 Ops with requests and CUDA alignment


CuiYifeng:
## Summary
This is a tracking issue to enable FP8 and MXFP8 operations on XPU, covering memory ops (fill, flip, index_put, gather, cat, eq, ne, where, empty/zeros/ones, to, copy, clone, add/sub/mul/div, compare, normal) and GEMM ops (_scaled_mm and _scaled_grouped_mm) across multiple FP8 variants (e4m3fn, e4m3fnuz, e5m2, e5m2fnuz, e8m0fnu). The issue tracks support status with green/yellow/red indicators and links to related PRs, targeting PyTorch releases from 2.10 through 2.16.

## Type
- **Category:** feature-gap
- **Platform:** xpu
- **Related Components:** FP8/MXFP8 kernels, memory ops (fill, flip, index_put, index_select, gather, cat, eq, ne, where, empty/zeros/ones, to, copy, clone, add/sub/mul/div, compare, normal), GEMM ops (_scaled_mm, _scaled_grouped_mm)

## Objective
Enable FP8/MXFP8 support for a set of memory and GEMM operations across all FP8 dtype variants (e4m3fn, e4m3fnuz, e5m2, e5m2fnuz, e8m0fnu), with CUDA alignment for scaled matrix multiply ops.

## Current Status
| Memory Op                         | e4m3fn | e4m3fnuz | e5m2 | e5m2fnuz | e8m0fnu | PR Link |
|------------------------------------|--------|----------|------|----------|---------|----------|
| fill/fill_                         | 🟢     | 🟢       | 🟢   | 🟢       | 🟢      |         |
| flip/fliplr/flipud                 | 🟢     | 🟢       | 🟢   | 🟢       | 🟢      |   #2190 |
| index_put/index_put_               | 🟢     | 🟢       | 🟢   | 🟢       | ❌      |   #2190 |
| index.Tensor/index.Tensor_out      | 🟢     | 🟢       | 🟢   | 🟢       | 🟢      |   #2190 |
| index_select/index_select.out      | 🟢     | 🟢       | 🟢   | 🟢       | 🟢      |         |
| gather/gather.out                  | 🟢     | 🟢       | 🟢   | 🟢       | 🟢      |         |
| cat/cat.out                        | 🟢     | 🟢       | 🟢   | 🟢       | 🟢      |  #2152  |
[...]

PR #2258, opened 2025-10-31
Add Float8_e8m0fnu support for copy

CuiYifeng:
To solve #2207
This PR is to support Float8_e8m0fnu copy on XPU backend.


CuiYifeng:
Depends on #2190.

PR #2260, opened 2025-10-31
add CMAKE_SYCL_COMPILER_LAUNCHER for sccache.

xuhancn:
Add `CMAKE_SYCL_COMPILER_LAUNCHER` for sccache enabling, please reference to https://github.com/pytorch/pytorch/blob/7d67a41db4801632d5862e99c79b766284dfd307/.ci/pytorch/windows/build_pytorch.bat#L91C13-L91C41

When we setup `CMAKE_SYCL_COMPILER_LAUNCHER` to sccache binary, it will enable sccache.
When we didn't set `CMAKE_SYCL_COMPILER_LAUNCHER`, it should be blank, and didn't change original logical.

Sccache usage:
https://github.com/mozilla/sccache?tab=readme-ov-file#usage


cmake/Modules/FindSYCL.cmake
```diff
@@ -35,6 +35,11 @@
 #
 #  SYCL_ADD_LIBRARY
 
+if(NOT CMAKE_SYCL_COMPILER_LAUNCHER AND DEFINED ENV{CMAKE_SYCL_COMPILER_LAUNCHER})
+  set(CMAKE_SYCL_COMPILER_LAUNCHER "$ENV{CMAKE_SYCL_COMPILER_LAUNCHER}"
```
fengyuan14, line 39: Do we need check whether the tool is valid?

xuhancn: @fengyuan14 I think it is enough, if tool is not valid, build failed message will show detailed info.

fengyuan14 (approved):

cmake/Modules/FindSYCL.cmake
```diff
@@ -35,6 +35,11 @@
...
+    CACHE STRING "Compiler launcher for SYCL.")
+endif()
```
guangyey, line 41:
```suggestion
if(USE_CCACHE)
  find_program(SCCACHE_EXECUTABLE sccache)
  set(CMAKE_SYCL_COMPILER_LAUNCHER
        "${SCCACHE_EXECUTABLE }"
        CACHE STRING "SYCL compiler launcher")
endif()
```
Is this better. refer to https://github.com/pytorch/pytorch/blob/e0791fc11dc0024a828495985898b29120dcc4c1/CMakeLists.txt#L403-L421 and https://github.com/pytorch/pytorch/blob/9501405de686068c5d7d0e4eab21e3ef0b95cc73/cmake/Dependencies.cmake#L67

guangyey: And is it better to move this code block to `torch-xpu-ops/CMakeLists.txt`?

guangyey (approved):

Issue #2023, opened 2025-09-09
F.max_pool3d().names are None


daisyden:
### 🐛 Describe the bug

Platform: PVC Linux
Case:
test_namedtensor.py::TestNamedTensor.test_max_pooling || op_ut

pytest -v test/test_namedtensor.py -k test_max_pooling 

```
______________________________________________________________________________ TestNamedTensor.test_max_pooling _______________________________________________________________________________
Traceback (most recent call last):                                                                                                                                                             
  File "/home/daisyden/upstream/test_nn_s1_ci/test/test_namedtensor.py", line 297, in test_max_pooling                                                                                         
    self.assertEqual(F.max_pool3d(named_tensor_3d, [2, 2, 2]).names, named_tensor_3d.names)                                                                                                    
  File "/home/daisyden/miniforge3/envs/test_nn/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 4179, in assertEqual                                                
    raise error_metas.pop()[0].to_error(  # type: ignore[index]                                                                                                                                
AssertionError: None mismatch: None is not A                                                                                                                                                   
                                                                                                                                                                                               
The failure occurred for item [0]                                                                                                                                                              
                                                                                                                                                                                               
To execute this test, run the following from the base repo dir:                                                                                                                                
    python test/test_namedtensor.py TestNamedTensor.test_max_pooling                                                                                                                           
                                                                                                                                                                                               
This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0    
```

```
(Pdb) l
294                 named_tensor_2d = torch.zeros(2, 3, 5, 7, device=device, names=list('ABCD'))
295                 named_tensor_3d = torch.zeros(2, 3, 5, 7, 9, device=device, names=list('ABCDE'))
296  
297                 self.assertEqual(F.max_pool1d(named_tensor_1d, 2).names, named_tensor_1d.names)
298                 self.assertEqual(F.max_pool2d(named_tensor_2d, [2, 2]).names, named_tensor_2d.names)
299  ->             self.assertEqual(F.max_pool3d(named_tensor_3d, [2, 2, 2]).names, named_tensor_3d.names)
300  
301                 check_tuple_return(F.max_pool1d_with_indices, [named_tensor_1d, 2], named_tensor_1d.names)
302                 check_tuple_return(F.max_pool2d_with_indices, [named_tensor_2d, [2, 2]], named_tensor_2d.names)
[...]

PR #2281, opened 2025-11-04
Ensured that output tensors (output and indices) correctly propagate the names from the input tensor

chunhuanMeng:
Fix https://github.com/intel/torch-xpu-ops/issues/2023
Makes improvements to the `max_pool3d_with_indices_xpu` implementation by enhancing support for named tensors and ensuring proper propagation of tensor names. The changes also include a minor code organization update.

Enhancements for named tensor support:

* Added an include for `ATen/NamedTensorUtils.h` in `DilatedMaxPool3d.cpp` to enable named tensor utilities.
* Ensured that output tensors (`output` and `indices`) correctly propagate the names from the input tensor by calling `namedinference::propagate_names` for both.


CuiYifeng (approved):
LGTM. Please indicate which test cases will be fixed by this PR.

src/ATen/native/xpu/DilatedMaxPool3d.cpp
```diff
@@ -26,6 +27,8 @@ std::tuple<Tensor, Tensor> max_pool3d_with_indices_xpu(
       ceil_mode,
       output,
       indices);
+  namedinference::propagate_names(output, input);
+  namedinference::propagate_names(indices, input);
```
guangyey, line 33: Doesn't here need `NoNamesGuard guard;` and `guard.reset();`

guangyey (approved):

PR #2282, opened 2025-11-04
layernorm: enlarge the range for 2-pass reduction

weishi-deng:
From the OOB models, there are some shapes still below the performance expectation with large M but small N.
Simple shapes:
[128, 197, 384]
[64,784, 256]
[(64, 28, 28, 256]
[256, 197, 256]
[128, 196, 384]
After enlarging the range for 2-pass reduction, these models can benefit an average of 10-20ms model execution time and optimize the geomean performance of eager training in timm models from 0.835 to 0.842.


src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -1063,8 +1063,9 @@ void _layer_norm_backward_kernel(
       norm_config_global_size / syclMaxSubGroupSize() * 2 <= thread_slots;
   // cuda uses condition M > 64 * 1024 && N / 32 < sm_count / 2 to parallelize
   // in the M dimension
-  if (use_two_stage_col_reduction && M > 64 * 1024 &&
-      N / 32 < syclGpuEuCount() / syclGpuEUCountPerSubslice() / 2) {
+  int subslice_count = syclGpuEuCount() / syclGpuEUCountPerSubslice();
```
EikanWang:
```suggestion
  int xe_core_count = syclGpuEuCount() / syclGpuEUCountPerSubslice();
```

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -1063,8 +1063,9 @@ void _layer_norm_backward_kernel(
...
+  if (use_two_stage_col_reduction && M > subslice_count * 1024 &&
+      N / 32 < subslice_count) {
```
EikanWang: How could we correlate `N / 32 < subslice_count` to the performance insights? Does the `32` mean simd32?

weishi-deng: Updated.

PR #2284, opened 2025-11-04
[UT] Skip more unit test cases on LNL due to OOM

kaileiyx:
This PR adds unit test skip cases for Windows LNL platform to handle two specific UR (Unified Runtime) backend errors: UR_RESULT_ERROR_OUT_OF_RESOURCES and UR_RESULT_ERROR_OUT_OF_DEVICE_MEMORY.

disable_all


chuanqi129 (approved):
@kaileiyx could we double check whether those OOM cases also failed on Linux LNL?

kaileiyx:
@chuanqi129  Yes, after filtering out these OOM cases, the behaviors of Windows and Linux are consistent on LNL machine

PR #2286, opened 2025-11-04
[AOT] Add PTL in the default AOT target list for both Win and Lin

chuanqi129:
This PR adds PTL (Panther Lake) architecture support to the Ahead-of-Time (AOT) compilation target lists for both Windows and Linux platforms in the CMake build configuration.

- Windows builds now include ptl target
- Linux builds now include ptl-h and ptl-u targets


chuanqi129:
Windows UT failures are not related to this PR changes

PR #2306, opened 2025-11-07
[Nightly] Sync benchmark list with pytorch for nightly tests

mengfei25:
disable_all


.ci/benchmarks/torchbench.yaml
```diff
@@ -216,6 +192,9 @@ skip:
 
     cuda: []
 
+    xpu:
+      - *DETECTRON2_MODELS
```
chuanqi129: Do we want to test those detectron2 series models in our side?

mengfei25: Modified and will run detectron2 in our weekly

PR #2308, opened 2025-11-07
Prevent complex<Half> accumulation overflow in sum_functor specialization

yucai-intel:
**Problem Description**: 
The XPU backend did not correctly enable the high-precision accumulator rule when processing the sum reduction of complexHalf, causing premature overflow of low-precision accumulators in the kernel's internal accumulation operation.

**Specific Modifications and Effects**:
- Modify the reduce_dispatch function: The original scheduler logic omitted explicit high-precision boosting for ComplexHalf. The fix forces the acc_t to be boosted to complex<float> at the scheduling entry point, ensuring the correct precision is used for reduction.
- Modify the sum_functor specialization: The kernel's underlying implementation may incorrectly downgrade the accumulator type to the low-precision out_t. The fix forces the out_t parameter of gpu_reduce_kernel to be boosted to complex<float> to avoid the type downgrading bug within the kernel.

### DIFF
--- a/.github/workflows/_linux_build.yml
+++ b/.github/workflows/_linux_build.yml
@@ -59,7 +59,7 @@ jobs:
     if: ${{ ! endsWith(inputs.pytorch, '_wheel') }}
     runs-on: ${{ needs.runner.outputs.runner_id }}
     container:
-      image: 'pytorch/manylinux2_28-builder:xpu-2.9'
+      image: 'pytorch/manylinux2_28-builder:xpu-main'
       volumes:
         - ${{ github.workspace }}:${{ github.workspace }}
       env:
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/ReduceSumProdKernels.cpp
```diff
@@ -48,23 +48,24 @@ struct SumFunctor {
...
 template <
     typename scalar_t,
     typename acc_t = scalar_t,
     typename out_t = scalar_t>
 struct sum_functor {
   void operator()(TensorIterator& iter) {
-    gpu_reduce_kernel<scalar_t, out_t>(
-        iter, func_wrapper<out_t>(SumFunctor<acc_t>()));
+    gpu_reduce_kernel<scalar_t, scalar_t>(
+        iter, func_wrapper<scalar_t>(SumFunctor<acc_t>()));
```
CuiYifeng: `out_t` cannot be used in any case here even if `out_t` is specified by upper level calls to `sum_functor`?

CuiYifeng (approved):

EikanWang (commented):
Please add regression test cases.

yucai-intel:
UT result:
image

chunhuanMeng (approved):

PR #2312, opened 2025-11-10
[BE] Upgrade GCC version to 13 for Linux build

chuanqi129:
Follows https://github.com/pytorch/pytorch/pull/162474


mengfei25 (approved):

chuanqi129:
> @chuanqi129 Need we change the test under gcc 13 together? I had uploaded an image named "intelgpu/ubuntu-24.04-lts2:2523.31"

@mengfei25 good capture, after the sycl-tla integrated, we should switch the test docker image based on ubuntu 24.04 I think. Please follow it at then.

zxd1997066 (approved):

Issue #2015, opened 2025-09-04
inf is returned by nn.TransformerEncoderLayer


daisyden:
### 🐛 Describe the bug

These cases are in skip list before due to AssertionError, now they got accuracy issue. 

test/test_nn.py::TestNNDeviceTypeXPU::test_transformerencoderlayer_gelu_xpu_float16
FAILED [0.6178s]   test/test_nn.py::TestNNDeviceTypeXPU::test_transformerencoderlayer_xpu_float16
FAILED [0.0251s]   test/test_nn.py::TestNNDeviceTypeXPU::test_transformerencoderlayer_xpu_float32
FAILED [2.7030s]   test/test_nn.py::TestNNDeviceTypeXPU::test_transformerencoderlayer_xpu_float64

```
______ TestNNDeviceTypeXPU.test_transformerencoderlayer_gelu_xpu_float16 _______
Traceback (most recent call last):
  File "/home/daisyden/upstream/test_nn/test/test_nn.py", line 13304, in test_transformerencoderlayer_gelu
    _test(activation=activation, batch_first=batch_first, training=training)
  File "/home/daisyden/upstream/test_nn/test/test_nn.py", line 13296, in _test
    torch.testing.assert_close(result, ref_output, rtol=rtol, atol=atol)
  File "/home/daisyden/miniforge3/envs/test_nn/lib/python3.10/site-packages/torch-2.9.0a0+git97f1b3e-py3.10-linux-x86_64.egg/torch/testing/_comparison.py", line 1589, in assert_close
    raise error_metas[0].to_error(msg)
AssertionError: Tensor-likes are not close!

Mismatched elements: 35 / 40 (87.5%)
Greatest absolute difference: nan at index (0, 0, 0) (up to 0.001 allowed)
Greatest relative difference: nan at index (0, 0, 0) (up to 0.01 allowed)
```
```
_________ TestNNDeviceTypeXPU.test_transformerencoderlayer_xpu_float16 _________
Traceback (most recent call last):
  File "/home/daisyden/upstream/test_nn/test/test_nn.py", line 13188, in test_transformerencoderlayer
    _test(batch_first=batch_first, training=training, atol=atol, rtol=rtol)
  File "/home/daisyden/upstream/test_nn/test/test_nn.py", line 13061, in _test
    self.assertTrue(np.isnan(result).all())
  File "/home/daisyden/miniforge3/envs/test_nn/lib/python3.10/unittest/case.py", line 687, in assertTrue
[...]

PR #2336, opened 2025-11-11
Fix TransformerEncoderLayer Full Mask UT Failure on XPU

yucai-intel:
**The Problem Solved**
The nn.TransformerEncoderLayer unit test failed on XPU devices when checking the expected output for a fully masked input (Sequence Length=1, Mask=[[True]]). https://github.com/intel/torch-xpu-ops/issues/2015
Root Cause: 
Due to a missing device entry in the core PyTorch's Fast Path whitelist, XPU execution was incorrectly forced into the slow Python Fallback path, instead of the optimized C++ fused kernel. https://github.com/pytorch/pytorch/pull/168234
Semantic Mismatch: 
The XPU implementation executing the Fallback path produces a non-NaN finite value for fully masked inputs (reflecting the robust mathematical result of X+Attention(0)).
Assertion Error: 
However, the unit test's internal logic erroneously categorized the XPU execution as belonging to a Fast Path, which expected a NaN result. The actual non-NaN finite result did not match the expected NaN assertion, causing the test failure.

**Fix Implementation**
By modifying the internal UT logic to explicitly track the device's correct execution semantics (Non-NaN for XPU), we switch the test's assertion from the incorrect self.assertTrue(np.isnan(result).all()) to the correct self.assertTrue(not np.isnan(result).any()).

### DIFF
--- a/test/xpu/skip_list_common.py
+++ b/test/xpu/skip_list_common.py
@@ -359,7 +359,12 @@ skip_dict = {
     "test_legacy_vmap_xpu.py": None,
     "test_utils_xpu.py": None,
     "functorch/test_vmap_xpu.py": None,
-    "dynamo/test_ctx_manager_xpu.py": None,
+    "dynamo/test_ctx_manager_xpu.py": (
+        # Autocast to float64 is CUDA-specific and not supported on XPU.
+        # More details in https://github.com/pytorch/pytorch/pull/179141
+        "test_autocast_float64",
+        "test_cuda_amp_autocast",
+    ),
     "functorch/test_control_flow_xpu.py": None,
     "functorch/test_aot_joint_with_descriptors_xpu.py": None,
     "profiler/test_memory_profiler.py": None,


test/xpu/test_nn_xpu.py
```diff
@@ -31,6 +31,8 @@
     TEST_WITH_CROSSREF,
 )
 
+TEST_WITH_CROSSREF = 1  # noqa: F811
```
CuiYifeng: I noticed that CUDA passed these cases with `TEST_WITH_CROSSREF = 0`. Meanwhile, `Fast Paths` works with CUDA. We need to check the related XPU kernel.

CuiYifeng (changes requested):

test/xpu/test_nn_xpu.py
yucai-pro: I ran this test on the A100. CUDA took the non-fast path, and the output of xpu was consistent with CUDA.

yucai-pro: On XPU：
res---------  [[[ 2.213   0.1344 -0.721   0.1926]]]
exp---------  False
On CUDA：
test_transformerencoderlayer---------
Non Fast Paths---------
res---------  [[[ 2.213   0.1344 -0.721   0.1926]]]
exp---------  False

CuiYifeng:
@yucai-intel Could you resolve conflicts and check if this fix is still valid?

tye1:
@yucai-intel pls. fix the conflicts

yucai-pro:
The XPU TransformerEncoderLayer test failures and skips reported in issue #2015 have been fully addressed by PR #3399. This PR introduced proper dtype coverage for XPU (float16, float32, float64) via @dtypesIfXPU, and corrected the fast_path_device logic to align XPU behavior with CPU/CUDA.
Validation
After updating both PyTorch and torch-xpu-ops to the latest main and rebuilding, all relevant XPU test cases now pass:
test_transformerencoderlayer_xpu_float16 — PASSED
test_transformerencoderlayer_xpu_float32 — PASSED
test_transformerencoderlayer_xpu_float64 — PASSED
No XPU dtype cases are skipped. Only fast_path_xpu_float64 remains skipped by design (CPU-only).
Conclusion
This issue is fully resolved in upstream main. No further patch is required for PR #2336. All XPU TransformerEncoderLayer dtype paths are now robustly validated and passing.

Issue #2240, opened 2025-10-30
RuntimeError: Trying to set a forward gradient that has a different size than that of the original Tensor, this is not supported. in test/functorch/test_ops.py


zxd1997066:
### 🐛 Describe the bug

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_functorch_xpu.TestOperatorsXPU,test_vmapjvpall_has_batch_rule_nn_functional_logsigmoid_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_functorch_xpu.TestOperatorsXPU,test_vmapjvpall_nn_functional_logsigmoid_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_functorch_xpu.TestOperatorsXPU,test_vmapvjp_has_batch_rule_nn_functional_conv3d_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_functorch_xpu.TestOperatorsXPU,test_vmapjvpvjp_nn_functional_logsigmoid_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vmapjvpall_has_batch_rule_nn_functional_logsigmoid_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vmapjvpall_nn_functional_logsigmoid_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vmapvjp_has_batch_rule_nn_functional_conv3d_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vmapjvpvjp_nn_functional_logsigmoid_xpu_float32
```
_ TestOperatorsXPU.test_vmapjvpall_has_batch_rule_nn_functional_logsigmoid_xpu_float32 _
[gw5] linux -- Python 3.10.19 /tmp/xpu-tool/Python/3.10.19/x64/bin/python
Traceback (most recent call last):
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 1151, in test_wrapper
    return test(*args, **kwargs)
[...]

PR #2373, opened 2025-11-17
Fix logsigmoid buffer shape for vmap+JVP

Silv3S:
Fixes https://github.com/intel/torch-xpu-ops/issues/2240

### DIFF
--- a/cmake/SYCLTLA.cmake
+++ b/cmake/SYCLTLA.cmake
@@ -3,6 +3,8 @@ macro(replace_cmake_build_flags)
   set(CMAKE_CXX_FLAGS_BK "${CMAKE_CXX_FLAGS}")
   string(REPLACE "-Werror=format" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
   string(REPLACE "-Werror=format" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+  string(REPLACE "-Werror=unused-variable" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+  string(REPLACE "-Werror=unused-variable" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 endmacro()
 
 macro(restore_cmake_build_flags)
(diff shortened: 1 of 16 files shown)


PawelSwider2000 (approved):

guangyey:
May I know if there is the same issue existing on CUDA. I see CUDA use `empty` as well
```cpp
std::tuple<Tensor, Tensor> log_sigmoid_forward_cuda(const Tensor& input) {
  auto result = at::empty_like(input);
  auto buffer = at::empty({0}, input.options());
  log_sigmoid_forward_out_cuda(input, result, buffer);
  return std::forward_as_tuple(result, buffer);
}
```

EikanWang (approved):
Could you please add a regression test?

EikanWang:
> Thanks, @Silv3S. Regarding the logic in https://github.com/pytorch/pytorch/blob/2f023bf7b962e69c0de01b89af197388d9df27cc/aten/src/ATen/functorch/BatchRulesBinaryOps.cpp#L320-L331, I think it may be worth changing the logic here. @EikanWang, what’s your opinion?

Sorry for the late response. I think the changes of this PR is related to https://github.com/pytorch/pytorch/blob/2f023bf7b962e69c0de01b89af197388d9df27cc/aten/src/ATen/native/cuda/Activation.cpp#L89-L94, right?

Silv3S:
@EikanWang, CUDA would have same issue, but they don't evaluate buffer rank in `log_sigmoid_backward_batch_rule()`, so it doesn't matter what shape it has.
```c++
if (!grad.is_cuda() && !self.is_cuda() && !buffer.is_cuda()) {
    out_logical_rank = std::max(out_logical_rank, rankWithoutBatchDim(buffer, buffer_bdim)); 
```
XPU fails because we have hardcoded buffer shape that is later used in `log_sigmoid_backward_batch_rule()`. 

We could either
1) patch torch-xpu-ops - Keep buffer shape same as input shape
2) patch upstream - Add `is_xpu()` to existing `is_cuda()` conditions in `log_sigmoid_backward_batch_rule`, so we won't evaluate buffer rank

Issue #2219, opened 2025-10-27
float8_e4m3fn precision overflow


jiqing-feng:
### 🐛 Describe the bug

```python
import torch

value = torch.load("clamped_value.pt").to(0)
print(value.to(torch.float8_e4m3fn).to(torch.float16).max())
```

[clamped_value.zip](https://github.com/user-attachments/files/23157648/clamped_value.zip)

XPU output:
```
tensor(nan, device='xpu:0', dtype=torch.float16)
```

A100 output:
```
tensor(288., device='cuda:0', dtype=torch.float16)
```

### Versions

[pip3] torch==2.10.0.dev20251026+xpu
[pip3] torchao==0.15.0.dev20251027+xpu
[pip3] torchaudio==2.10.0.dev20251026+xpu
[pip3] torchdata==0.12.0.dev20250220
[pip3] torchvision==0.25.0.dev20251026+xpu
[pip3] triton==3.5.0

jiqing-feng:
If the value in zip is not available, please reach me via Teams to get the value. Thanks!

yucai-intel:
The error occurs because a NaN is generated when converting the tensor to float8, and this is subsequently retained in the result. 
https://github.com/intel/torch-xpu-ops/blob/main/src/ATen/native/xpu/sycl/ReduceMaxValuesKernels.cpp#L17
Specific improvements are still under investigation.

EikanWang:
Although this issue has been workarounded by the PR https://github.com/intel/torch-xpu-ops/pull/2387. However, it doesn't mean the PR is the right fix. So, I reopened this issue to keep track of the SYCL compiler's fixes. We need to revert the PR when the fixing of the SYCL compiler is ready.

PR #2387, opened 2025-11-21
Temporary Fix for FP16 -> FP8 conversion failure on -0.0

yucai-intel:
To resolve https://github.com/intel/torch-xpu-ops/issues/2219
This PR is to temporarily work around the issue where FP16's -0.0 is erroneously converted to NaN during certain fusion passes (fp16 -> fp32 -> fp8), we are currently avoiding the use of the **sycl::half** data type in the intermediate conversion steps to prevent the problematic fusion from occurring.

### DIFF
--- a/src/ATen/native/xpu/sycl/CopyKernel.cpp
+++ b/src/ATen/native/xpu/sycl/CopyKernel.cpp
@@ -30,8 +30,9 @@ struct CastScalarFunc {
 // be removed once the compiler/driver error is resolved.
 template <typename Float8DataType>
 struct CastScalarFunc<Half, Float8DataType> {
-  C10_HOST_DEVICE Float8DataType operator()(Half src_val) const {
-    return Float8DataType(c10::detail::fp16_ieee_to_fp32_value(src_val.x));
+  Float8DataType operator()(Half src_val) const {
+    Half val = src_val == Half(-0.0) ? Half(0.0) : src_val;
+    return Float8DataType(val);
   }
 };
 


CuiYifeng (approved):

EikanWang (approved):
LGTM

src/ATen/native/xpu/sycl/CopyKernel.cpp
```diff
@@ -25,6 +25,16 @@ struct CastScalarFunc {
   }
 };
 
+// TODO: Avoid using sycl::half to prevent the fp16->fp32->fp8 fusion
+// from incorrectly converting -0.0 to NaN. This temporary fix should
+// be removed once the compiler/driver error is resolved.
+template <typename Float8DataType>
+struct CastScalarFunc<Half, Float8DataType> {
+  C10_HOST_DEVICE Float8DataType operator()(Half src_val) const {
+    return Float8DataType(c10::detail::fp16_ieee_to_fp32_value(src_val.x));
```
EikanWang: `fp16_ieee_to_fp32_value` is a host-only functionon? Why do we need to invoke it here? By the way, why is this function `C10_HOST_DEVICE`?

EikanWang: Could you help try the code block as you proposed in the previous commits?
```c++
auto val = src_val == sycl::half(-0.0) ? sycl::half(0.0) : src_val;
```

EikanWang: My bad. `fp16_ieee_to_fp32_value` is not marked host-only.

CuiYifeng: @EikanWang Updated work-around based on your suggestion. Please note that FP16 -0.0 will be converted to FP8 +0.0 in this version.

EikanWang (approved):

PR #2399, opened 2025-11-24
Add native XPU support for `linalg.qr`, `geqrf`, and `householder_product`

### LINKED ISSUE #1900: implement torch.linalg.qr xpu backend

`torch.linalg.qr` is needed by PEFT LoRA EVA([here](https://github.com/huggingface/peft/tree/main/examples/eva_finetuning)) for weight initialization and update, but currently we can find it fallback to CPU, and slow, as below log.

Pls plan the XPU backend implementation for this op, thx.

>  UserWarning: Aten Op fallback from XPU to CPU happends. This may have performance implications. If need debug the fallback ops please set environment variable `PYTORCH_DEBUG_XPU_FALLBACK=1`  (Triggered internally at /pytorch/build/xpu/ATen/RegisterXPU_0.cpp:54525.)
>   Q = torch.linalg.qr(X).Q

### Alternatives

_No response_

### Additional context

_No response_

mwiktor-intel:
Implement `torch.linalg.qr`, `torch.geqrf`, and `torch.linalg.householder_product` on XPU by registering `geqrf_stub` and `orgqr_stub`, which allows reusing PyTorch's existing QR decomposition logic.

Implementation details:
- `orgqr`/`ungqr` (Q matrix recovery): uses oneMKL LAPACK when built with `USE_ONEMKL_XPU`, with CPU fallback otherwise. Uses the batched API (`orgqr_batch`/`ungqr_batch`) when `k == n`, and a per-item loop otherwise
- `geqrf` (QR factorization): uses CPU fallback for now (will be switch to oneMKL once XPU performance is optimized)
- supports both real floating-point and complex types

Changes:
- Register `geqrf_stub` and `orgqr_stub` XPU dispatchers in `BatchLinearAlgebra.cpp`
- Add `orgqr_mkl` and `ungqr_mkl` oneMKL kernel implementations in `xpu/mkl/BatchLinearAlgebra.cpp`
- Register `linalg_qr.out`, `linalg_householder_product{,.out}`, and `geqrf{,.a}` ops for XPU in `native_functions.yaml`
- Remove `geqrf`, `linalg_householder_product`, and `linalg_qr.out` from the XPU CPU-fallback list
- Add XPU-specific tests for QR modes (`r, reduced, complete`) covering batched inputs, various dimensions, and float/complex dtypes

Fixes https://github.com/intel/torch-xpu-ops/issues/1900


src/ATen/native/xpu/BatchLinearAlgebra.cpp
```diff
@@ -64,4 +66,21 @@ void lu_factor_kernel_xpu(
...
+  xpu::linalg_qr_kernel(A, mode, Q, R);
+#else
+  auto A_cpu = A.to(at::kCPU);
+  auto Q_cpu = at::empty_like(Q, at::kCPU);
+  auto R_cpu = at::empty_like(R, at::kCPU);
+  at::cpu::linalg_qr_out(Q_cpu, R_cpu, A_cpu, mode);
+  Q.copy_(Q_cpu);
+  R.copy_(R_cpu);
+#endif // USE_ONEMKL_XPU
+}
```
CuiYifeng: My suggestion is to register `geqrf_kerenl_xpu`/`orgqr_kernel_xpu` to `geqrf_stub`/`orgqr_stub`, which allows us to reuse op level code in stock Pytorch and reuse these two kernels in future.

yaml/native/native_functions.yaml
```diff
@@ -9443,6 +9443,17 @@
...
+  python_module: linalg
+  variants: function
+  structured_delegate: linalg_qr.out
+
+- func: linalg_qr.out(Tensor A, str mode='reduced', *, Tensor(a!) Q, Tensor(b!) R) -> (Tensor(a!) Q, Tensor(b!) R)
+  python_module: linalg
+  structured: True
+  dispatch:
+    XPU: linalg_qr_xpu_out
+
```
CuiYifeng: Please refer to https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L14623 and consider stub mentioned above.

CuiYifeng (changes requested):
Please check comments and fix related failed cases.

yaml/native/native_functions.yaml
mwiktor-intel: Using existing PT stub for implementing QR has both pros and cons. At the side of pros, we take the most of the existing infrastructure with simplified flow on sycl/xpu side. The implementation proposed here has, however also several strengths. Namely:
1) Although QR is a composition of geqrf and orgqr, the initial request was for QR, not a wrappers for mentioned mkl funciotns intended to fit QR meta-operator.
2) MKL functions are matrix oriented and can operate only on matrices. Thus, every subsequent call would require slicing a tensor and sending its parts to the kernel. It would require forcing contiquous function separately for each slice, instead of once, at the beginning of the proposed kernel. Since the proposed kernel uses direct pointers it can go through batch simply by manipulating pointers, no slicing and memory reorganization needed.
3) The biggest possible efficiency trap lays in matrix layout differences between PT and LAPACK. The former stores data column wise the latter row wise. In order to properly call mkl functions, the data must be transposed.
In fused version, as proposed in this PR, the transposition is called twice, at the beginning and exit. Splitting this kernel into two separate calls would require fitting what user expect calling geqrf and orgqr, which means, the data must have been transposed twice per call. Saying transposed means real reordering data in memory, since unlike PT functions, MKL does not support strides.

Thus, although provitidn these two kernels separately is a natural development path, I would prioritize it for later, after QR and two pending requests will be ready.

CuiYifeng: @mwiktor-intel For layout differences, could you indicate in MKL document the reason why LAPACK is row-major? https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2025-2/geqrf-usm-version.html Please pay attention to leading dimension `lda` of input matrix. Suppose matrix A is col-major storage and shape is (m, n), we can set `lda=m` to indicate correct layout.

mwiktor-intel: LDA cannot be replacement of stride. LDA can be used to skip some elements in matrix, but with LDA we cannot force difference data orientation. Specifically: say we have 3x2 matrix. Lapack expects this 6 element array in memory to represent the data as follows:
[ 0 2 4
  1 3 5]
LDA here is 2, Setting LDA 3 does not change anything here The purpose of LDA is when, say, we have for any reason some bigger structure in memory, say,
 0  3 6 
1  4 7
2 5 8 ]
and we want to take only 2 upper rows to process. LDA here should be 3, giving information to the function, that apart from logical number of rows equal to 2, the procedure should skip 3 values to get to the data for 2nd column.

Some BLAS function have row_major of col_major order, but it is not the case within LAPACK.
Note, in the documentation, LDA cannot be set to less than number of rows in the processed matrix.

CuiYifeng: Thanks for the elaboration. Let me rephase your example 1: given 3x2 matrix stored as [0, 2, 4, 1, 3, 5] in memory, the logic order is [[0, 1], [2, 3], [4, 5]] if this matrix is col-major, right? You mentioned "LDA here is 2", but why LDA here can be smaller than the number of rows?
In your example 2: given memory buffer [0, 3, 6, 1, 4, 7, 2, 5, 8], the logic order of a 3x3 col-major matrix is [[0, 1, 2], [3, 4, 5], [6, 7, 8]]. We can use shape 2x3 and LDA=3 to get col-major matrix [[0, 1, 2], [3, 4, 5]] from this buffer. In this example, I agree with your understanding.
LDA is the distance in memory between the start of consecutive columns (for column-major storage) or consecutive rows (for row-major storage). In other words, LDA means stride[0] for row-major or stride[1] for col-major in buffer (not stride in matrix, as the bigger memory you mentioned).
Let's return to the API of geqrf. Given a buffer, shape (m, n) and LDA tell API how to get data from buffer. Furthermore, since you have noticed that `The leading dimension of a, at least max(1, m)`, why not `at least max(1, n)`? In my understanding, this implies that API geqrf requires col-major input matrix.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -561,4 +561,133 @@ void lu_factor_mkl(
...
+  int64_t m = a_contig.sizes().at(range - 1);
+  int64_t mn = int64_t(m * n);
+  int64_t b = numel ==0 ? 0 : numel / mn;
+
+
+  if (b==0 && mode=="complete" && n>0) {
+    b=1;
+    for (int dimension=0; dimension<range-2; dimension++) {
+      b*=a_contig.sizes().at(dimension);
+    }
```
CuiYifeng: `at::native::batchCount` can help you get flattened batch size. https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/LinearAlgebraUtils.h#L112

CuiYifeng (changes requested):
Please check comments (including Copilot's). This PR needs to be reviewed again after fixing a series of failed cases.
Suggest to refresh code format before review. Furthermore, please also check performance compared to CPU.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -561,4 +561,133 @@ void lu_factor_mkl(
...
+
+  int numel = a_contig.numel();
+  int range = a_contig.dim();
+  int64_t n = a_contig.sizes().at(range - 2);
+  int64_t m = a_contig.sizes().at(range - 1);
+  int64_t mn = int64_t(m * n);
+  int64_t b = numel ==0 ? 0 : numel / mn;
+
+
+  if (b==0 && mode=="complete" && n>0) {
```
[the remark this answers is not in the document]
CuiYifeng: Please refer to this comment.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -571,4 +571,143 @@ void lu_factor_mkl(
...
+  int64_t n = a_contig.sizes().at(range - 2);
+  int64_t m = a_contig.sizes().at(range - 1);
+  int64_t mn = int64_t(m * n);
+  int64_t b = numel == 0 ? 0 : numel / mn;
+
+  // correct R matrix  dimensions if needed
+  if (numel == 0 && mode != "complete") {
+    std::vector r(dimensions.begin(), dimensions.end());
+    if (r[range-1] == 0)
+      r[range-2]=0;
```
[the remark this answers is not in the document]
CuiYifeng: Please refer to this comment.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -571,4 +571,143 @@ void lu_factor_mkl(
...
+  int numel = a_contig.numel();
+  int range = a_contig.dim();
+  int64_t n = a_contig.sizes().at(range - 2);
+  int64_t m = a_contig.sizes().at(range - 1);
+  int64_t mn = int64_t(m * n);
+  int64_t b = numel == 0 ? 0 : numel / mn;
+
+  // correct R matrix  dimensions if needed
+  if (numel == 0 && mode != "complete") {
+    std::vector r(dimensions.begin(), dimensions.end());
```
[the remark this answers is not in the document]
CuiYifeng: Please try to make your code self-explanatory.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -571,4 +571,143 @@ void lu_factor_mkl(
...
+    result_r =
+        result_r
+            .index(
+                {"...", at::indexing::Slice(0, n), at::indexing::Slice(0, m)})
+            .contiguous();
+  }
+
+  // normal case, non-zero dimensions
+  if (mode!="r") {
+    result_q.transpose_(-2, -1).contiguous();
```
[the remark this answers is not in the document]
CuiYifeng: Please note that returned contiguous tensor was discarded here. For reference: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/TensorProperties.cpp#L119-L128.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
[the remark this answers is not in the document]
pbielak: Fixed

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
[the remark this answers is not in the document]
pbielak: Fixed

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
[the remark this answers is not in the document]
pbielak: Fixed

pbielak:
Initially I refactored the base implementation from @mwiktor-intel, but finally I ended up registering the `geqrf` and `orgqr` stubs, which allows to reuse the QR implementation from PyTorch. There is no observable performance gap between these two implementations, so I decided to remove Michał's implementation and use the function stubs. 

As oneMKL supports also complex dtypes for these kernels, I enabled the call path here too, while updating the tests to include complex dtypes as well.

Please review again @CuiYifeng

src/ATen/native/xpu/XPUFallback.template
```diff
@@ -230,23 +230,18 @@ TORCH_LIBRARY_IMPL(aten, XPU, m) {
...
-    "geqrf.a",
     "hash_tensor.out",
     "linalg_cholesky_ex.L",
     "linalg_eig",
     "linalg_eig.out",
     "_linalg_eigvals",
     "linalg_eigvals.out",
     "_linalg_eigh.eigenvalues",
-    "linalg_householder_product",
-    "linalg_householder_product.out",
```
CuiYifeng: Please also add `geqrf`, `geqrf.a`, `linalg_householder_product` and `linalg_householder_product.out` into `yaml/native/native_functions.yaml`.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -479,4 +479,156 @@ void triangular_solve_mkl(
       });
 }
 
+Tensor& ungqr_mkl(Tensor& result, const Tensor& tau) {
+  if (result.numel() == 0) {
+    return result;
+  }
+
+  auto& queue = c10::xpu::getCurrentXPUStream().queue();
```
CuiYifeng:
```suggestion
  auto& queue = at::xpu::getCurrentSYCLQueue();
```

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -479,4 +479,156 @@ void triangular_solve_mkl(
...
+
+  return result;
+}
+
+Tensor& orgqr_mkl(Tensor& result, const Tensor& tau) {
+  if (result.numel() == 0) {
+    return result;
+  }
+
+  auto& queue = c10::xpu::getCurrentXPUStream().queue();
```
CuiYifeng:
```suggestion
  auto& queue = at::xpu::getCurrentSYCLQueue();
```

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -479,4 +479,156 @@ void triangular_solve_mkl(
...
+      const int64_t scratchpad_size =
+          oneapi::mkl::lapack::ungqr_batch_scratchpad_size<T>(
+              queue,
+              m,
+              n,
+              k,
+              lda,
+              std::max<int64_t>(1, matrix_stride),
+              batch_tau_stride,
+              std::max<int64_t>(1, batch_size));
```
CuiYifeng: With the early-return of `result.numel() == 0`, why can we meet batch_size < 1 here?

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -479,4 +479,156 @@ void triangular_solve_mkl(
...
+      const int64_t scratchpad_size =
+          oneapi::mkl::lapack::orgqr_batch_scratchpad_size<T>(
+              queue,
+              m,
+              n,
+              k,
+              lda,
+              std::max<int64_t>(1, matrix_stride),
+              batch_tau_stride,
+              std::max<int64_t>(1, batch_size));
```
CuiYifeng: Ditto.

src/ATen/native/xpu/BatchLinearAlgebra.cpp
```diff
@@ -134,4 +137,163 @@ void triangular_solve_kernel_xpu(
...
+
+TORCH_LIBRARY_IMPL(aten, XPU, m) {
+  m.impl("geqrf", TORCH_FN(at::native::geqrf_xpu));
+  m.impl(
+      "linalg_householder_product",
+      TORCH_FN(at::native::linalg_householder_product_xpu));
+  m.impl(
+      "linalg_householder_product.out",
+      TORCH_FN(at::native::linalg_householder_product_out_xpu));
+}
```
CuiYifeng: Since `geqrf_stub` and `orgqr_stub` have been registered for XPU kernel, why not reuse `linalg_householder_product_out`, `geqrf` and `geqrf_out` instead of using `TORCH_LIBRARY_IMPL`?

CuiYifeng (changes requested):
@pbielak Please also resolve the code conflict and check the following new failed cases:

- op_ut,third_party.torch-xpu-ops.test.xpu.test_ops_xpu.TestCommonXPU,test_out_warning_geqrf_xpu
- op_ut,third_party.torch-xpu-ops.test.xpu.test_ops_xpu.TestCompositeComplianceXPU,test_cow_input_linalg_householder_product_xpu_float32

src/ATen/native/xpu/BatchLinearAlgebra.cpp
pbielak: Done

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
pbielak: Must have overlooked after adding the early-return - removed the `std::max` calls

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
pbielak: Removed the `std::max` calls

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
pbielak: Applied

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
pbielak: Applied

src/ATen/native/xpu/XPUFallback.template
pbielak: Added the entries to the `native_functions.yaml`

CuiYifeng (approved):

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -479,4 +479,142 @@ void triangular_solve_mkl(
...
+Tensor& ungqr_mkl(Tensor& result, const Tensor& tau) {
+  if (result.numel() == 0) {
+    return result;
+  }
+
+  auto& queue = at::xpu::getCurrentSYCLQueue();
+  const int64_t m = result.size(-2);
+  const int64_t n = result.size(-1);
+  const int64_t k = tau.size(-1);
+  const int64_t lda = std::max<int64_t>(1, m);
```
BBBela: It looks, like the `std::max<int64_t>(1, m);` is always equal to `m` in this line. If `m` is zero, then `result.numel() == 0` is true and function returns early. Just for consideration, can stay as it is.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -479,4 +479,142 @@ void triangular_solve_mkl(
...
+  const int64_t batch_size = native::batchCount(result);
+  const int64_t matrix_stride = native::matrixStride(result);
+  const int64_t tau_stride = std::max<int64_t>(1, k);
```
BBBela: It looks, like the `std::max<int64_t>(1, k);` is not necessary. It should work well when `k == tau_stride == 0` as mkl doesn't use the calculated `tau_batch_ptr` if `k == 0` anyway. Just for consideration, can stay as it is.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -479,4 +479,142 @@ void triangular_solve_mkl(
...
+
+  AT_DISPATCH_COMPLEX_TYPES(result.scalar_type(), "ungqr_mkl_xpu", [&] {
+    using T = get_mkl_type<scalar_t>::type;
+    auto* result_data = reinterpret_cast<T*>(result.data_ptr());
+    const auto* tau_data = reinterpret_cast<const T*>(tau.const_data_ptr());
+
+    if (k == n) {
+      const int64_t batch_tau_stride = std::max<int64_t>(1, std::min(m, n));
```
BBBela: Here also the `std::max<int64_t>` seems not necessarily required as the `m` or `n` equal `0` case would return early. Just for consideration, can stay as it is.

BBBela (commented):
Generally it looks good to me.

I have just three minor comments below, but fell free to resolve them if you don't find them relevant as they are not of high importance. Just wanted to point them for consideration.

I have just one bigger question - do we need all the newly added test cases? I mean, the operators implemented by this PR are already used in PyTorch by at least CPU and CUDA, so I suppose that they are tested in the upstream for correctness, aren't they? Does those test cases added in this PR perform some additional or more extensive checks that are not performed in the upstream? Aren't they duplications?

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
pbielak: Right - I simplified this line

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
pbielak: Right - I simplified this line

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
pbielak: Right - I simplified this line

pbielak:
> Generally it looks good to me.
> 
> I have just three minor comments below, but fell free to resolve them if you don't find them relevant as they are not of high importance. Just wanted to point them for consideration.
> 
> I have just one bigger question - do we need all the newly added test cases? I mean, the operators implemented by this PR are already used in PyTorch by at least CPU and CUDA, so I suppose that they are tested in the upstream for correctness, aren't they? Does those test cases added in this PR perform some additional or more extensive checks that are not performed in the upstream? Aren't they duplications?

@BBBela I did not inspect these tests as they were added in the initial PR draft by @mwiktor-intel However, now I reevaluated their scope and they actually analogous cases as the upstream tests. Hence, I removed the tests added by this PR, as all the cases are covered by upstream tests. 

Thanks for all the review comments!

BBBela (approved):
Looks good to me, thanks!

pbielak:
Due to updated build process (`native_functions.yaml` is now only available in PyTorch), this commit only provides the XPU implementations for geqrf oraz orgqr/ungqr. After merging, we need to land https://github.com/pytorch/pytorch/pull/187564 and finally the PR for removing the entries from the XPUFallback.cpp file https://github.com/intel/torch-xpu-ops/pull/4043 (so we can test the changes)

pbielak:
@CuiYifeng @BBBela I had to change `at::empty` to `at::zeros` for scratchpad allocation, as in another test we found a regression, where re-using memory (provided via the XPU caching allocator), caused the QR kernel to return wrong values. After the inspection of the MKL code, I found that the memory must be zero-initialized.

Re-running the benchmark showed no performance regressions when using zeros instead of empty.

CuiYifeng:
> @CuiYifeng @BBBela I had to change `at::empty` to `at::zeros` for scratchpad allocation, as in another test we found a regression, where re-using memory (provided via the XPU caching allocator), caused the QR kernel to return wrong values. After the inspection of the MKL code, I found that the memory must be zero-initialized.
> 
> Re-running the benchmark showed no performance regressions when using zeros instead of empty.

Thanks for the elaboration. May I know in which test we found the regression in?

pbielak:
@CuiYifeng The test case: `test_linalg_xpu.py::TestLinalgXPU::test_det_logdet_slogdet_xpu_float64`. I reported the observation to MKL and it turned out to be a but, but it should have been fixed by now. Nonetheless, I would keep `at::zeros` as it doesn't introduce any noticeable overhead.

CuiYifeng:
> @CuiYifeng The test case: `test_linalg_xpu.py::TestLinalgXPU::test_det_logdet_slogdet_xpu_float64`. I reported the observation to MKL and it turned out to be a but, but it should have been fixed by now. Nonetheless, I would keep `at::zeros` as it doesn't introduce any noticeable overhead.

Agree with you.

pbielak:
CI now uses DLE2026.1 / MKL 2026.1 - rebasing and merging

pbielak:

PR #2407, opened 2025-11-25
Reduce build warnings: fix field reordering and conditionally drop isystem w/a

Silv3S:
- Conditionally remove `isystem` flags, as they don't silence any warnings and also throw extra warnings per each kernel:
```
c++: warning: /opt/intel/oneapi/compiler/2025.3/include/sycl: linker input file unused because linking not done
```
- Fix no-reorder warning

### DIFF
--- a/.github/actions/linux-testenv/action.yml
+++ b/.github/actions/linux-testenv/action.yml
@@ -35,6 +35,7 @@ runs:
         fi
         sudo apt update
         sudo apt install -y jq numactl
+        rm -rf Torch-XPU-Wheel-*
     - name: Setup python-${{ inputs.python }}
       uses: actions/setup-python@v5
       with:
@@ -65,7 +66,8 @@ runs:
         elif [ $(echo "${{ inputs.pytorch }}" |grep -w "nightly_wheel" -c) -ne 0 ];then
           pip install torch torchvision torchaudio --pre --index-url https://download.pytorch.org/whl/nightly/xpu
         else
-          pip install --force-reinstall $(find ${{ github.workspace }}/ -name "*torch*.whl")
+          latest_dir="$(find . -type d -name "Torch-XPU-Wheel-*" |sort -V |tail -n 1)"
+          pip install --force-reinstall $(find ${latest_dir} -name "*torch*.whl")
         fi
         TORCH_COMMIT_ID=$(python -c 'import torch; print(torch.version.git_version)')
         if [[ "${{ inputs.pytorch }}" == *"https://"* ]];then
--- a/.github/ci_expected_accuracy/check_expected.py
+++ b/.github/ci_expected_accuracy/check_expected.py
@@ -1,98 +1,229 @@
+#!/usr/bin/env python3
+"""
+Accuracy Check Script
+Compares test results against reference data and calculates pass rates.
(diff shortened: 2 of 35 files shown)


cmake/BuildFlags.cmake
```diff
@@ -50,17 +50,11 @@ macro(set_build_flags)
...
-      # We expect that this issue will be addressed in the later version of DPC++ compiler. To workaround
-      # the issue we wrap paths to SYCL headers in `-isystem`.
-      foreach(FLAGS IN LISTS SYCL_INCLUDE_DIR)
-        list(APPEND SYCL_HOST_FLAGS "-isystem ${FLAGS}")
-      endforeach()
       # Excluding warnings which flood the compilation output
       # TODO: fix warnings in the source code and then reenable them in compilation
       list(APPEND SYCL_HOST_FLAGS -Wno-sign-compare)
+      list(APPEND SYCL_HOST_FLAGS -Wno-attributes)
+      list(APPEND SYCL_HOST_FLAGS -Wno-reorder)
```
dvrogozh: `-Wno-order` is a standard and fixable warning if it's against code hosted in this repo. Please, don't add to to ignore list - fix it. Same for attributes if it's a warning for our sources.

Silv3S: Right, done

dvrogozh:
> Regarding warnings which I wanted to silence - they are visible with and without `-isystem`

Are you using oneapi 2025.3? If yes, then the behavior you observe is expected as this is the first version which does not need the `-isystem` workaround. Earlier versions of the compiler, i.e. oneapi 2025.2, 2025.1 and below, has an issue that it flooded output with warnings on SYCL headers themselves which are unrelated to our sycl code. The workaround to fix that is to add `-isystem`.

If you are using 2025.3, then I suggest that we should protect adding `-isystem` with the oneapi version check and avoid adding it for 2025.3+.

As for the warning examples you've posted, I believe these are legitimate warnings for our code in this repository and such warnings needs to be fixed rather than silenced. If we can't fix them for whatever reason we first need to raise issue with oneapi team and only then consider silencing them.

dvrogozh (approved):
LGTM

cmake/BuildFlags.cmake
```diff
@@ -55,9 +55,11 @@ macro(set_build_flags)
       # SYCL headers, such as deprecated warnings, even if warned API is not actually used in the program.
       # We expect that this issue will be addressed in the later version of DPC++ compiler. To workaround
       # the issue we wrap paths to SYCL headers in `-isystem`.
-      foreach(FLAGS IN LISTS SYCL_INCLUDE_DIR)
-        list(APPEND SYCL_HOST_FLAGS "-isystem ${FLAGS}")
-      endforeach()
+      if(__INTEL_LLVM_COMPILER LESS 20250300)
```
guangyey:
```suggestion
      if(SYCL_COMPILER_VERSION VERSION_LESS 20250300)
```
We already have a definition `SYCL_COMPILER_VERSION` in PyTorch

Silv3S:
@guangyey could you please merge?

guangyey:
OK, why not.

Issue #1895, opened 2025-07-31
div_trunk_rounding accuracy gap on float64


daisyden:
### 🐛 Describe the bug

Accuracy issue of div_trunk_rounding cases in test_ops_xpu.py

test_python_ref__refs_div_trunc_rounding_xpu_float64
test_python_ref_executor__refs_div_trunc_rounding_executor_aten_xpu_float64
test_python_ref_torch_fallback__refs_div_trunc_rounding_xpu_float64

PYTORCH_OPINFO_SAMPLE_INPUT_INDEX=16 PYTORCH_TEST_WITH_SLOW=1 pytest -v test_ops.py test_python_ref_torch_fallback__refs_div_trunc_rounding_xpu_float64

Traceback (most recent call last):
  File "/home/daisyden/miniforge3/envs/test_ops/lib/python3.10/site-packages/torch-2.9.0a0+gitf3913ea-py3.10-linux-x86_64.egg/torch/testing/_internal/common_device_type.py", line 1135, in test_wrapper
    return test(*args, **kwargs)
  File "/home/daisyden/miniforge3/envs/test_ops/lib/python3.10/site-packages/torch-2.9.0a0+gitf3913ea-py3.10-linux-x86_64.egg/torch/testing/_internal/common_device_type.py", line 1451, in only_fn
    return fn(self, *args, **kwargs)
  File "/home/daisyden/miniforge3/envs/test_ops/lib/python3.10/site-packages/torch-2.9.0a0+gitf3913ea-py3.10-linux-x86_64.egg/torch/testing/_internal/common_utils.py", line 1665, in wrapper
    fn(*args, **kwargs)
  File "/home/daisyden/upstream/test_ops/third_party/torch-xpu-ops/test/xpu/../../../../test/test_ops.py", line 616, in test_python_ref_torch_fallback
    self._ref_test_helper(contextlib.nullcontext, device, dtype, op)
  File "/home/daisyden/upstream/test_ops/third_party/torch-xpu-ops/test/xpu/../../../../test/test_ops.py", line 518, in _ref_test_helper
    raise e
  File "/home/daisyden/upstream/test_ops/third_party/torch-xpu-ops/test/xpu/../../../../test/test_ops.py", line 506, in _ref_test_helper
    self.assertEqual(
[...]

PR #2413, opened 2025-11-26
Enable  -fno-reciprocal-math to fix div accuracy.

kdrozd-dev:
Added  -fno-reciprocal-math to SYCL_KERNEL_OPTIONS to eliminate issues with division of the number by itself returning non 1 output which caused: https://github.com/intel/torch-xpu-ops/issues/1895.


riverliuintel:
@CuiYifeng

cmake/BuildFlags.cmake
```diff
@@ -96,6 +96,7 @@ macro(set_build_flags)
     # gcc -shared host.o kernel.o device-code.o -o libxxx.so
     set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -fno-sycl-unnamed-lambda)
     set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -sycl-std=2020)
+    set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -fno-reciprocal-math)
```
EikanWang, line 107: Let's study the NVCC behaviors first. I think `reciprocal-math` is a general optimization, and most the compilers enable it by default.

kdrozd-dev: @EikanWang  I believe ```--prec-div``` https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html?highlight=reciprocal#prec-div-true-false-prec-div is the corresponding NVCC flag by default set to True unless ```--use_fast_math``` (https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html?highlight=reciprocal#use-fast-math-use-fast-math) is set to True. I do not believe pytorch is built with ```--use_fast_math``` correct me if I'm wrong.

EikanWang: Thanks for the information. Some HW platforms like CPU turn on this option by default. For CUDA, it keeps its default, saying `use_fast_math` is `False`. So, it makes sense to add `no-reciprocal-math`. I'm curious why SYCL turns on this option by default.

@fengyuan14 do you know the background?
https://github.com/intel/torch-xpu-ops/blob/20baf08891aae0d246cc66c58203a238f5c59edb/cmake/BuildFlags.cmake#L77

EikanWang: Per my understanding, we should disable `fast-math` to align with CUDA

NeoZhangJianyu:
> > There are a few options regarding `#pragma clang fp reciprocal(off)` (https://clang.llvm.org/docs/LanguageExtensions.html#extensions-to-specify-floating-point-flags) with the least invasive being adding it only to the BinaryDivTrueKernel.cpp file. However, I'm not sure whether it's worth it to sacrifice some performance for accuracy in this oddly specific use case (calling div first and than trunc instead of div_trunc) for which we have no known business need outside the failing UT. Both the div and div_trunc results are within needed norms. It might be better to modify the UT. What is your opinion on that @EikanWang?
> 
> @SanityRemnants , Although the performance is the king, I'd prefer to keep a high bar of maintenance when we have no concrete performance impact.
> 
> * @SanityRemnants since it is close to the PyTorch 2.10 branch cut date, I'd like to defer this PR landing.
> * @weishi-deng , @chuanqi129 , could you help collect the performance data w/ this PR? My expectation is that this PR only impacts a few micro-benchmarks and there should be no impact on E2E models. Priority-wise, the collections is lower than PT 2.10.

There is no such issue on CUDA.
IPEX has added some similar parameters to fix the accuracy issue.
Looks like they are not upstream to Pytorch.

This parameter is neither in IPEX or in PyTorch.
It only impact the OPs including divide().

We are waiting for this PR to fix same issue in other products.
Please speed up the review!

Thank you!

EikanWang:
@weishi-deng , I'm curious why the option impacts the other operations besides div. Let understand why.

EikanWang:
@weishi-deng , `cat, eq, ne, ne, mul, repeat_interleave` - Have you checked why this PR impacts these operations?

EikanWang (approved):

Issue #2182, opened 2025-10-16
test_transform_bias_rescale_qkv_nested_xpu_float32 failed with AssertionError: Scalars are not equal!


wincent8:
### 🐛 Describe the bug

when we enable test/test_native_mha.py for xpu， it will meet below issues, 
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_native_mha_xpu.TestMHADeviceTypeXPU,test_transform_bias_rescale_qkv_nested_xpu_float32
```
pytest -sv test/test_native_mha.py::TestMHADeviceTypeXPU::test_transform_bias_rescale_qkv_nested_xpu_float32
================================================================================================================================ test session starts =================================================================================================================================
platform linux -- Python 3.10.0, pytest-8.4.2, pluggy-1.6.0 -- /home/sdp/miniforge-pypy3/envs/wliao2/bin/python3.10
cachedir: .pytest_cache
rootdir: /home/sdp/wliao2/pytorch
configfile: pytest.ini
collected 1 item
Running 1 items in this shard

test/test_native_mha.py::TestMHADeviceTypeXPU::test_transform_bias_rescale_qkv_nested_xpu_float32 /home/sdp/miniforge-pypy3/envs/wliao2/lib/python3.10/site-packages/torch/nested/__init__.py:250: UserWarning: The PyTorch API of nested tensors is in prototype stage and will change in the near future. We recommend specifying layout=torch.jagged when constructing a nested tensor, as this layout receives active development, has better operator coverage, and works with torch.compile. (Triggered internally at /pytorch/aten/src/ATen/NestedTensorImpl.cpp:178.)
  return _nested.nested_tensor(
FAILED [0.3817s]

====================================================================================================================================== FAILURES ======================================================================================================================================
______________________________________________________________________________________________________ TestMHADeviceTypeXPU.test_transform_bias_rescale_qkv_nested_xpu_float32 _______________________________________________________________________________________________________
Traceback (most recent call last):
  File "/home/sdp/miniforge-pypy3/envs/wliao2/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/home/sdp/miniforge-pypy3/envs/wliao2/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/home/sdp/miniforge-pypy3/envs/wliao2/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
[...]

PR #2450, opened 2025-12-02
Fix incorrect Tensor Size for NestedTensor QKV Transform

yucai-intel:
To solve https://github.com/intel/torch-xpu-ops/issues/2182 : Q (Query) tensor output size from torch.transform_bias_rescale_qkv was mismatched against the expected reference size in test cases involving Nested Tensors where the sequence length (T) was not a multiple of 8 after implicit padding.

**Resolution**: The resolution involved introducing logic within the C++ function transform_bias_rescale_qkv_xpu specifically for the Nested Tensor case to explicitly use the calculated sequence length T to resize the output q, k, and v tensors, thereby ensuring their final size matches the shape derived by the Python reference implementation.

### DIFF
--- a/src/ATen/native/transformers/Attention.cpp
+++ b/src/ATen/native/transformers/Attention.cpp
@@ -54,6 +54,16 @@ std::tuple<Tensor, Tensor, Tensor> transform_bias_rescale_qkv_xpu(
   auto T = qkv.is_nested() ? native::NestedTensor_get_max_size(
                                  *native::get_nested_tensor_impl(qkv))[0]
                            : qkv.size(1);
+  if (qkv.is_nested()) {
+    // Don't mess with non-nested case for now since it's not set up to fiddle
+    // with mask size.
+
+    // Round T up to next multiple of 8 so as to be able to utilize Tensor
+    // cores. Otherwise, sometimes with padding, *no* row will have the maximum
+    // sequence length and so we'll have a non-divisible-by-8 dimension even if
+    // the model author chose a multiple of 8.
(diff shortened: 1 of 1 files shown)


CuiYifeng (approved):
LGTM.

src/ATen/native/transformers/Attention.cpp
```diff
@@ -38,6 +38,16 @@ std::tuple<Tensor, Tensor, Tensor> transform_bias_rescale_qkv_xpu(
...
                            : qkv.size(1);
+  if (qkv.is_nested()) {
+    // Don't mess with non-nested case for now since it's not set up to fiddle
+    // with mask size.
+
+    // Round T up to next multiple of 8 so as to be able to utilize Tensor
+    // cores. Otherwise, sometimes with padding, *no* row will have the maximum
+    // sequence length and so we'll have a non-divisible-by-8 dimension even if
+    // the model author chose a multiple of 8.
+    T = T + (8 - (T % 8)) % 8;
```
liangan1, line 65: The dpas should not have limitation, the m can be 1~8, why we need to change it here?

yucai-pro: For **Nested Tensors**, since T is derived from the **max_size**, a non-aligned T would cause row-starts to be misaligned in the contiguous q_k_v buffer, breaking memory coalescing and potentially causing illegal memory access in fused kernels.
We round up T here to ensure that every row in the packed buffer satisfies the alignment constraints required by the XPU Tensor Core templates used in bmm and SDP.

liangan1 (approved):

Issue #2313, opened 2025-11-10
[xpu][test]profiler can not get the key of multi_tensor_apply_kernel


wincent8:
### 🐛 Describe the bug

when we enable the test_foreach.py, refer to the cuda path, we added the xpu path
```python
elif (
            is_xpu
            and torch.autograd.kineto_available()
            and torch.profiler.ProfilerActivity.XPU
            in torch.profiler.supported_activities()
        ):
            with torch.profiler.profile() as p:
                actual = self.func(*inputs, **kwargs)
                # synchronize within the profiler context to make sure events happen before exiting
                torch.xpu.synchronize()
            keys = tuple([e.key for e in p.key_averages()])
            mta_called = any("multi_tensor_apply_kernel" in k for k in keys)
 
            assert mta_called == (expect_fastpath and (not zero_size)), (
                f"{mta_called=}, {expect_fastpath=}, {zero_size=}, {self.func.__name__=}, {keys=}"
            )
``` 
about 1400 case will failed with AssertionError like below.
```
test/test_foreach.py::TestForeachXPU::test_parity__foreach_zero_fastpath_outplace_xpu_uint8 FAILED [0.2256s]
 
==================================================================================== FAILURES =====================================================================================
______________________________________________________ TestForeachXPU.test_parity__foreach_zero_fastpath_outplace_xpu_uint8 _______________________________________________________
Traceback (most recent call last):
  File "/home/sdp/wliao2/pytorch/test/test_foreach.py", line 262, in test_parity
    actual = func(
  File "/home/sdp/wliao2/pytorch/test/test_foreach.py", line 121, in __call__
    assert mta_called == (expect_fastpath and (not zero_size)), (
AssertionError: mta_called=False, expect_fastpath=True, zero_size=False, self.func.__name__='_foreach_zero_', keys=('aten::_foreach_zero_', 'aten::empty', 'urUSMHostAlloc', 'urEnqueueUSMMemcpy', 'Memcpy H2D (HOST -> DEVICE)', 'urEnqueueKernelLaunch', 'at::native::xpu::MultiTensorApplyKernelFunctor<at::native::xpu::TLMetaForAddress<1>*, at::native::xpu::TLMetaForWG*, at::native::xpu::ZeroFunctor<unsigned char, 1, 1, 0>>')
 
During handling of the above exception, another exception occurred:
 
Traceback (most recent call last):
[...]

PR #2455, opened 2025-12-03
Change MTA logic for foreach_copy

PawelSwider2000:
On XPU, `foreach_copy` operator follows similar logic for switching between multi tensor apply and slower mode. For CUDA the behavior is different as in other foreach ops as it enable MTA path when source and destination datatypes are different. This commit is aligning XPU logic with CUDA.

To do so:
- create `can_use_fast_route_for_copy` matching CUDA logic for selecting fast MTA path.
- create `CopyFunctor` and modify `Copy` to handle different dtypes and perform casts instead of Identity operation
- change `foreach_copy_list_kernel_` dispatch logic to take into account both src and dst types.

PR fixes issues in: https://github.com/intel/torch-xpu-ops/issues/2313 

Tests for the issue are being enabled in issue mention above and also we could use existing `foreach_copy` tests.

### DIFF
--- a/src/ATen/native/xpu/ForeachOpList.cpp
+++ b/src/ATen/native/xpu/ForeachOpList.cpp
@@ -40,9 +40,15 @@ bool can_use_fast_route_for_copy(TensorList self, TensorList src) {
             [&](const auto& t) -> bool {
               return t.dtype() == src[0].dtype();
             });
+  const bool is_all_self_same_dtype = std::all_of(
+            self.cbegin(),
+            self.cend(),
+            [&](const auto& t) -> bool {
+              return t.dtype() == self[0].dtype();
+            });
   return _check_tensors_share_device_and_dtype(
             {self, src}, /* skip_dtype_check */ true) &&
-        is_all_src_same_dtype &&
+        is_all_src_same_dtype && is_all_self_same_dtype &&
         _check_tensors_share_sizes_and_strides({self, src});
 }
 


src/ATen/native/xpu/ForeachOpList.cpp
```diff
@@ -21,6 +21,22 @@
...
+
+bool can_use_fast_route_for_copy(TensorList self, TensorList src) {
+  return _check_tensors_share_device_and_dtype(
+            {self, src}, /* skip_dtype_check */ true) &&
+        std::all_of(
+            src.cbegin(),
+            src.cend(),
+            [&](const auto& t) -> bool {
+              return t.dtype() == src[0].dtype();
+            }) &&
```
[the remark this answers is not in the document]
PawelSwider2000: This is done in: check_foreach_api_restrictions

tsocha: You can save the result to helper variable, it will be easier to read.

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
```diff
@@ -7,13 +7,95 @@
...
+template <typename dst_t>
+constexpr bool is_complex_dtype() {
+  return std::is_same_v<dst_t, c10::complex<float>> ||
+      std::is_same_v<dst_t, c10::complex<double>>;
+}
+
+template <typename dst_t, typename src_t>
+struct Copy {
+  dst_t operator()(const src_t& x) {
+    if constexpr (is_complex_dtype<src_t>() && !is_complex_dtype<dst_t>()) {
```
tsocha, line 45: You can probably use `std::enable_if` for it but this solution looks ok as well.

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
```diff
@@ -7,13 +7,95 @@
...
+    typename src_t,
+    int depth,
+    int r_args_depth,
+    int res_arg_index>
+struct CopyFunctor {
+  static_assert(
+      depth == 2 && r_args_depth == 1 && res_arg_index == 1,
+      "CopyFunctor only supports depth=2, r_args_depth=1, res_arg_index=1");
+  template <typename TLA, typename TLW, typename Op>
+  void operator()(
```
tsocha, line 67: give it some space :D
```suggestion
      "CopyFunctor only supports depth=2, r_args_depth=1, res_arg_index=1");

  template <typename TLA, typename TLW, typename Op>
  void operator()(
```

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
```diff
@@ -7,13 +7,95 @@
...
+    const auto item_range = item_id.get_local_range(0);
+    const auto group_idx = item_id.get_group(0);
+    const int tensor_loc = tlWGMeta[group_idx].wg_to_tensor;
+    const int chunk_idx = tlWGMeta[group_idx].wg_to_chunk;
+    const int64_t n =
+        tlAddress[tensor_loc].numel_to_tensor - chunk_idx * chunk_size;
+
+    src_t* src_ptr = (src_t*)tlAddress[tensor_loc].addresses[0];
+    src_ptr += chunk_idx * chunk_size;
+    dst_t* dst_ptr = (dst_t*)tlAddress[tensor_loc].addresses[1];
```
tsocha: Don't use c casts please.

tsocha (changes requested):

src/ATen/native/xpu/ForeachOpList.cpp
PawelSwider2000: Done

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
PawelSwider2000: I prefer to keep this in single function for now

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
PawelSwider2000: Space given

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
PawelSwider2000: Done

Silv3S (approved):

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
```diff
@@ -24,15 +107,27 @@ void foreach_copy_list_kernel_(TensorList self, TensorList src) {
...
-            Identity<opmath_t>());
+        using dst_t = scalar_t;
+        using dst_opmath_t = at::opmath_type<dst_t>;
+        AT_DISPATCH_ALL_TYPES_AND_COMPLEX_AND3(
+            at::ScalarType::Half,
+            at::ScalarType::BFloat16,
+            at::ScalarType::Bool,
+            src[0].scalar_type(),
+            "foreach_tensor_copy",
+            [&]() {
```
EikanWang, line 140: Why does this PR need the nested `AT_DISPATCH_ALL_TYPES_AND_COMPLEX_AND3`?

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
```diff
@@ -24,15 +107,27 @@ void foreach_copy_list_kernel_(TensorList self, TensorList src) {
...
+              using src_opmath_t = at::opmath_type<src_t>;
+              multi_tensor_apply<2>(
+                  tensor_lists,
+                  CopyFunctor<
+                      dst_t,
+                      src_t,
+                      /* depth */ 2,
+                      /* r_args_depth */ 1,
+                      /* res_arg_index */ 1>(),
+                  Copy<dst_opmath_t, src_opmath_t>());
```
EikanWang: Since it is a copy function, why are `dst_opmath_t` and `src_opmath_t` required?

PawelSwider2000: It's since PyTorch supports that `torch.copy` (and foreach version as well) supports copying between different dtype if we want to have support here for such cases we need to have both src and dst dtypes. Thats also the reason why we are doing a nested `AT_DISPATCH_ALL_TYPES_AND_COMPLEX_AND3`

EikanWang: I suppose you are addressing the above comment.

https://github.com/intel/torch-xpu-ops/pull/2455/changes/BASE..18ab57c4ee1faefd76b003818ac20bd8c0aab103#diff-346dedcf59d4381e08b92e700ab3354c84bc13e5a13097ee55f7872cfd86e3a7R103-R108
```c++
  // Outer dispatch
  AT_DISPATCH_ALL_TYPES_AND_COMPLEX_AND3(
      at::ScalarType::Half,
      at::ScalarType::BFloat16,
      at::ScalarType::Bool,
      self[0].scalar_type(),
      "foreach_tensor_copy",
      [&]() {
        using dst_t = scalar_t;
        using dst_opmath_t = at::opmath_type<dst_t>;
        
        // Inner dispatch (newly added)
        AT_DISPATCH_ALL_TYPES_AND_COMPLEX_AND3(
            at::ScalarType::Half,
            at::ScalarType::BFloat16,
            at::ScalarType::Bool,
            src[0].scalar_type(),
            "foreach_tensor_copy",
```

The outer dispatch has decided the data type. Why is the inner dispatch necessary? @PawelSwider2000

EikanWang:
> but it could handle copying from one dtype to another like which allows us to use MTA logic in more amount of cases.

Yes. However, `UnaryOpFunctor` can handle it as well.
https://github.com/intel/torch-xpu-ops/blob/de582bafd338de1251711eb1769fa4145c20cf7a/src/ATen/native/xpu/sycl/ForeachFunctors.h#L151-L154

> On average we see an perf increase, which is bigger for tensorlists with many small tensors and small for few tensorlists with big tensors.

I need to take time to understand the behavior because it cannot be well explained. I may miss something.

PawelSwider2000:
https://github.com/intel/torch-xpu-ops/blob/de582bafd338de1251711eb1769fa4145c20cf7a/src/ATen/native/xpu/sycl/ForeachFunctors.h#L151-L154

In this code the `opmath_t` and `T` are related to each other and they could not be in foreach_copy. I will try to analyze and debug it in the meantime also.

src/ATen/native/xpu/sycl/ForeachCopyKernels.cpp
PawelSwider2000: Here, dispatch needs to decide based on src and dst datatypes so this is the reason for outer and inner dispatch. We want to have such dispatching:

```
AT_DISPATCH_CASE((float32, float32))
AT_DISPATCH_CASE((float32, bfloat16))
etc...
```

EikanWang: Make sense.

EikanWang (approved):

PR #2458, opened 2025-12-04
[CI] Use Ubuntu 24.04 and LTS2 2523.40 for CICD tests

mengfei25:
Follow gcc 13 build https://github.com/intel/torch-xpu-ops/pull/2312
Use gcc 13 for tests

disable_transformers
disable_accelerate


mengfei25:
Transformers failures are same as https://github.com/intel/torch-xpu-ops/actions/runs/19597929924
Accelerator failures are same as https://github.com/intel/torch-xpu-ops/actions/runs/20396801255
disable them in CI tests

Issue #1818, opened 2025-07-08
[BMG-Windows][PT2.8]Torch-xpu-ops UT got accuracy issue


libohao1201:
### 🐛 Describe the bug

The following torch-xpu-ops uts got `AssertionError: Scalars are not close!` or `AssertionError: Tensor-likes are not close!`.

```
# AssertionError: Scalars are not close!
test_nn_xpu.py::TestNNDeviceTypeXPU::test_GroupNorm_general_xpu
test_reductions_xpu.py::TestReductionsXPU::test_noncontiguous_expanded__refs_prod_xpu_complex128
test_reductions_xpu.py::TestReductionsXPU::test_noncontiguous_expanded_masked_prod_xpu_complex128
test_reductions_xpu.py::TestReductionsXPU::test_noncontiguous_transposed__refs_prod_xpu_complex128
test_reductions_xpu.py::TestReductionsXPU::test_noncontiguous_transposed_masked_prod_xpu_complex128
test_reductions_xpu.py::TestReductionsXPU::test_ref_duplicate_values__refs_prod_xpu_complex128
test_reductions_xpu.py::TestReductionsXPU::test_ref_duplicate_values_masked_prod_xpu_complex128
test_reductions_xpu.py::TestReductionsXPU::test_ref_small_input__refs_prod_xpu_complex128
test_reductions_xpu.py::TestReductionsXPU::test_ref_small_input_masked_prod_xpu_complex128
test_reductions_xpu.py::TestReductionsXPU::test_reference_masked_masked_prod_xpu_complex128



# AssertionError: Tensor-likes are not close!
FAILED test_reductions_xpu.py::TestReductionsXPU::test_ref_duplicate_values_prod_xpu_complex128

```

Reproducer:
```
set HTTP_PROXY=http://proxy.ims.intel.com:911
set HTTPS_PROXY=http://proxy.ims.intel.com:911

conda create -n pt28_rc1 python=3.10 -y
conda activate pt28_rc1
pip install --pre torch==2.8.0 --index-url https://download.pytorch.org/whl/test/xpu 

git clone https://github.com/pytorch/pytorch.git
cd pytorch
git checkout 03eb1e40f9ddf09cb9eef86ace74332e87f11a79

cd third_party
git clone https://github.com/intel/torch-xpu-ops.git
cd torch-xpu-ops
git checkout 3a9419c8bb6a98dd3e3cd473c36691fb4abeae40

pip install -r requirements.txt
[...]

PR #2477, opened 2025-12-08
Fix incorrect num_items in reduction for small kernels.

kdrozd-dev:
Ensures num_items is recalculated after updating group_width when num_items is less than max_sg_sz, preventing access to wrong memory later and fixes prod reduction for small kernels. Fixes some UTs from https://github.com/intel/torch-xpu-ops/issues/1818.

### DIFF
--- a/test/xpu/skip_list_common.py
+++ b/test/xpu/skip_list_common.py
@@ -104,6 +104,7 @@ skip_dict = {
         "test_scaled_dot_product_attention_3D_input_dim_2D_attn_mask_dropout_p_0_5_xpu",
         "test_scaled_dot_product_attention_3D_input_dim_2D_attn_mask_dropout_p_0_2_xpu",
     ),
+    "test_schema_check.py": None,
     "test_complex_xpu.py": None,
     "test_modules_xpu.py": (
         # oneDNN issues
--- a/test/xpu/test_schema_check.py
+++ b/test/xpu/test_schema_check.py
@@ -1,0 +1,554 @@
+# Owner(s): ["oncall: jit"]
+# ruff: noqa: F841
+
+import os
+import sys
+import unittest
+
+import torch
+from torch._subclasses.schema_check_mode import SchemaCheckMode
+from torch.fx.operator_schemas import normalize_function
+from torch.testing._internal.common_device_type import (
+    instantiate_device_type_tests,
+    OpDTypes,
+    ops,
+)
+from torch.testing._internal.common_methods_invocations import op_db
+from torch.testing._internal.common_utils import (
+    IS_WINDOWS,
+    run_tests,
+    slowTestIf,
+    TEST_WITH_TORCHDYNAMO,
+)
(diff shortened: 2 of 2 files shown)


src/ATen/native/xpu/sycl/Reduce.h
```diff
@@ -269,7 +269,10 @@ struct ReduceConfig {
     num_items = group_width * group_height;
 
     if (num_items < max_sg_sz)
+    {
       group_width = max_sg_sz;
+      num_items = group_width * group_height;
```
guangyey, line 273: Is it better to move line 269 behind the `if condition scope`.

src/ATen/native/xpu/sycl/Reduce.h
```diff
@@ -269,7 +269,10 @@ struct ReduceConfig {
     num_items = group_width * group_height;
 
     if (num_items < max_sg_sz)
+    {
```
EikanWang:
```suggestion
if (num_items < max_sg_sz) {
```

src/ATen/native/xpu/sycl/Reduce.h
EikanWang: The re-calculated `num_items` may be greater than `max_num_items`. right?

kdrozd-dev: @guangyey we cannot move line 269 as it would interfere with the condition evaluation.

EikanWang: > The re-calculated `num_items` may be greater than `max_num_items`. right?

If it is impossible, we can add an `assert` statement.

kdrozd-dev: It shouldn't be possible as the recalculation only happens for small kernels, but I think adding an assert is the right choice.

kdrozd-dev: @EikanWang does this look good to you with the added assert?

EikanWang: LGTM

src/ATen/native/xpu/sycl/Reduce.h
```diff
@@ -268,8 +268,11 @@ struct ReduceConfig {
     group_width = std::min(dim0_pow2, int(max_num_items / group_height));
     num_items = group_width * group_height;
 
-    if (num_items < max_sg_sz)
+    if (num_items < max_sg_sz) {
       group_width = max_sg_sz;
+      num_items = group_width * group_height;
+      assert(num_items <= max_num_items);
```
guangyey:
```suggestion
      TORCH_INTERNAL_ASSERT(num_items <= max_num_items);
```

guangyey (approved):
LGTM for this issue. However, I still have some questions about the logic used to select `group_width` and `group_height`. I will discuss this with @EikanWang later.

PR #2524, opened 2025-12-11
fix cri build error

jianyizh:
current compiler only support cri target only, user should set TORCH_XPU_ARCH_LIST=cri

disable_all


CuiYifeng (approved):
LGTM.

PR #2561, opened 2025-12-12
Remove compilation warnings

moksiuc:


riverliuintel:
Yes, it should be. we need to enable this warning as error build flag. @CuiYifeng have we clean all warning till now?

PR #2583, opened 2025-12-16
Bump version and unskip compiler warnings

Silv3S:
- bump torch-xpu-ops version to match PyTorch,
- unskip `-Wreturn-type` in kernels - these are not reproducing locally,

### DIFF
--- a/cmake/BuildFlags.cmake
+++ b/cmake/BuildFlags.cmake
@@ -58,6 +58,16 @@ macro(set_build_flags)
       list(APPEND SYCL_HOST_FLAGS -std=${CPP_STD})
       list(APPEND SYCL_HOST_FLAGS -Wunused-variable)
       list(APPEND SYCL_HOST_FLAGS -Wno-interference-size)
+      # Some versions of DPC++ compiler pass paths to SYCL headers as user include paths (`-I`) rather
+      # than system paths (`-isystem`). This makes host compiler to report warnings encountered in the
+      # SYCL headers, such as deprecated warnings, even if warned API is not actually used in the program.
+      # We expect that this issue will be addressed in the later version of DPC++ compiler. To workaround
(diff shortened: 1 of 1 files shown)


cmake/BuildFlags.cmake
```diff
@@ -58,16 +58,6 @@ macro(set_build_flags)
       list(APPEND SYCL_HOST_FLAGS -std=${CPP_STD})
       list(APPEND SYCL_HOST_FLAGS -Wunused-variable)
       list(APPEND SYCL_HOST_FLAGS -Wno-interference-size)
-      # Some versions of DPC++ compiler pass paths to SYCL headers as user include paths (`-I`) rather
-      # than system paths (`-isystem`). This makes host compiler to report warnings encountered in the
-      # SYCL headers, such as deprecated warnings, even if warned API is not actually used in the program.
-      # We expect that this issue will be addressed in the later version of DPC++ compiler. To workaround
-      # the issue we wrap paths to SYCL headers in `-isystem`.
-      if(SYCL_COMPILER_VERSION VERSION_LESS 20250300)
```
Silv3S, line 66: @dvrogozh is it ok to remove this check? It was added recently in https://github.com/intel/torch-xpu-ops/pull/2407, but now `main` branch is already targeted for PT2.11, so oneAPI will be 2026.0 (currently 2025.3), making this check obsolete

guangyey: We can consider removing this check after upgrading to the 2026.0 toolchain. For now, we should keep it so that PyTorch remains buildable with two compiler versions (2025.2, 2025.3).

guangyey (approved):

cmake/BuildFlags.cmake
Silv3S: Sure. I restored this check

PR #2586, opened 2025-12-16
Fix `dtypesIf` decorator patching

pbielak:
Currently, if a function has been decorated by both `dtypesIfCUDA` and `dtypesIfXPU`, the patch applied by `XPUPatchForImport` causes the function to be decorated by `dtypesIfXPU` twice, which triggers the following assertion error: `dtypes redefinition for xpu`

This commit rewrites the patched decorator with a more generic implementation, which takes into account other applications of the decorator. Detailed behaviour has been explained in the docstring.

Provides fixes for currently failing tests:
- op_ut,,third_party.torch-xpu-ops.test.xpu.test_indexing_xpu
- op_ut,,third_party.torch-xpu-ops.test.xpu.test_native_mha_xpu
- op_ut,,third_party.torch-xpu-ops.test.xpu.nn.test_embedding_xpu

(see issue: https://github.com/intel/torch-xpu-ops/issues/2366) 

disable_e2e
disable_distributed

### DIFF
--- a/test/xpu/xpu_test_utils.py
+++ b/test/xpu/xpu_test_utils.py
@@ -14,6 +14,7 @@
 
 
 import copy
+import logging
 import os
 import sys
 import unittest
@@ -31,6 +32,8 @@ from torch.testing._internal.common_modules import module_db
 from torch.testing._internal.common_nn import CriterionTest, ModuleTest
 from torch.testing._internal.common_utils import set_default_dtype
 from torch.testing._internal.opinfo.core import DecorateInfo
+
+log = logging.getLogger(__name__)
 
 _xpu_computation_op_list = [
     "empty",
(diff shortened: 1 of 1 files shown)


test/xpu/xpu_test_utils.py
```diff
@@ -845,6 +848,123 @@ def _skipXPU(obj):
...
+                # unless we're overriding a CUDA-derived mocked value.
+                if existing is None:
+                    d[_XPU_KEY] = self.args
+                    setattr(fn, _XPU_DTYPES_SOURCE_ATTR, _SRC_XPU)
+                    fn.dtypes = d
+                    return fn
+
+                # (B) Override only if the existing value came from mocked CUDA
+                if existing_src == _SRC_CUDA:
+                    log.warning(
```
EikanWang: Will it trigger a log storm?

pbielak: From what I saw in [https://github.com/pytorch/pytorch/pull/165886/](https://github.com/pytorch/pytorch/pull/165886/files#diff-4c6d0607e741ed4481effe7d1c18a6c73e307cef1579240ea8c3a9289d7d6be5R319), the `dtypesIfXPU` is always placed right after the `dtypesIfCUDA` decorator - so `dtypesIfXPU` will be applied first. Hence, we should not see any of those warnings, but surely if it will be applied the other way around, there will be some logs (one for each test function).

Also I will change it to debug level (to match the the other log call).

Issue #2384, opened 2025-11-21
[upstream_ut] Fallback op failures


daisyden:
### 🐛 Describe the bug

Reproduce command:

PYTORCH_TEST_WITH_SLOW=1 pytest test/test_ops.py -k test_errors_dot_xpu

FAILED [0.0080s] test/test_ops.py::TestCommonXPU::test_errors_dot_xpu - Exception: RuntimeError not raised
FAILED [0.0075s] test/test_ops.py::TestCommonXPU::test_errors_vdot_xpu - Exception: RuntimeError not raised
FAILED [0.2754s] test/test_ops.py::TestCommonXPU::test_out_ormqr_xpu_float32


```
________________________________________________________________________________________________________ TestCommonXPU.test_errors_dot_xpu _________________________________________________________________________________________________________
Traceback (most recent call last):
  File "/home/daisyden/miniforge3/envs/torch_ops_s2/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 1151, in test_wrapper
    return test(*args, **kwargs)                                                                                                                                                                                                                    
  File "/home/daisyden/miniforge3/envs/torch_ops_s2/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 1231, in dep_fn
    return fn(slf, *args, **kwargs)                                                                                       
  File "/home/daisyden/miniforge3/envs/torch_ops_s2/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 1490, in only_fn
    return fn(self, *args, **kwargs)
  File "/home/daisyden/upstream/torch_ops_s2/test/test_ops.py", line 800, in test_errors
    with self.assertRaisesRegex(ei.error_type, ei.error_regex):                                                                                                                                                                                     
  File "/home/daisyden/miniforge3/envs/torch_ops_s2/lib/python3.10/unittest/case.py", line 226, in __exit__
    self._raiseFailure("{} not raised".format(exc_name))                                                                                                                                                                                            
  File "/home/daisyden/miniforge3/envs/torch_ops_s2/lib/python3.10/unittest/case.py", line 163, in _raiseFailure
    raise self.test_case.failureException(msg)
AssertionError: RuntimeError not raised
```

More cases failed due to vdot

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_meta_xpu.TestMetaXPU,test_dispatch_symbolic_meta_outplace_vdot_xpu_complex128
[...]

PR #2627, opened 2025-12-22
Add device consistency check for XPU Fallback Ops

lchen2331:
Fix issue: #2384 by adding device consistency validation

### DIFF
--- a/src/ATen/native/xpu/XPUFallback.template
+++ b/src/ATen/native/xpu/XPUFallback.template
@@ -14,20 +14,20 @@ static bool DEBUG_XPU_FALLBACK = false;
 static void check_device_consistency(
     const torch::jit::Stack* stack,
     const c10::OperatorHandle& op) {
-  std::optional<c10::DeviceType> first_device;
+  std::optional<c10::DeviceType> reference_device;
 
   auto check_tensor = [&](const at::Tensor& t) {
     if (!t.defined()) return;
 
     auto cur_device = t.device().type();
-    if (!first_device) {
-      first_device = cur_device;
+    if (!reference_device) {
+      reference_device = cur_device;
       return;
     }
 
-    TORCH_CHECK(*first_device == cur_device,
+    TORCH_CHECK(*reference_device == cur_device,
       "Expected all tensors to be on the same device, but found at least two devices, ",
-      *first_device, " and ", cur_device,
+      *reference_device, " and ", cur_device,
       "! (Operator ", op.schema().operator_name().name, ")");
   };
 


src/ATen/native/xpu/XPUFallback.template
```diff
@@ -5,9 +5,53 @@ namespace at {
...
+  bool has_xpu = false;
+  bool has_non_xpu = false;
+
+  // Check all inputs device
+  for (const auto& iv : *stack) {
+    check_device_consistency(iv, has_xpu, has_non_xpu);
+    if (has_xpu && has_non_xpu) break;
+  }
+
+  if (has_xpu && has_non_xpu) {
```
CuiYifeng: Since `has_xpu` and `has_non_xpu` are always used to calculate `has_xpu && has_non_xpu` out of `check_device_consistency`, how about having the checking function to return a bool to represent consistent or not?

src/ATen/native/xpu/XPUFallback.template
```diff
@@ -5,9 +5,53 @@ namespace at {
...
+    const c10::IValue& iv,
+    bool& has_xpu,
+    bool& has_non_xpu) {
+  if (iv.isTensor()){
+    const auto& t = iv.toTensor();
+    if (t.defined())
+      if (t.device().type() == c10::DeviceType::XPU)
+        has_xpu = true;
+      else
+        has_non_xpu = true;
```
CuiYifeng: Can we reuse this part of logic?

src/ATen/native/xpu/XPUFallback.template
lchen2331: Sure.

lchen2331:
@CuiYifeng Thanks for your feedback. I have refined the implementation, could you please review it

src/ATen/native/xpu/XPUFallback.template
```diff
@@ -5,9 +5,44 @@ namespace at {
...
+  bool has_non_xpu = false;
+
+  auto check_tensor = [&](const at::Tensor& t) {
+    if (t.defined()) {
+      if (t.device().type() == c10::DeviceType::XPU) {
+        has_xpu = true;
+      } else {
+        has_non_xpu = true;
+      }
+    }
```
CuiYifeng:
```suggestion
    if (!t.defined()) return;
    if (t.device().type() == c10::DeviceType::XPU) {
      has_xpu = true;
    } else {
      has_non_xpu = true;
    }
```
Less indentation here is better.

src/ATen/native/xpu/XPUFallback.template
```diff
@@ -5,9 +5,44 @@ namespace at {
...
+  return false;
+}
+
 static void xpu_fallback_impl(
     const c10::OperatorHandle& op,
     torch::jit::Stack* stack) {
+
+  if (check_device_inconsistency(stack)) {
+     std::string msg = "Expected all tensors to be on the same device, but found at least two devices, xpu and cpu! (Operator " +
+                        op.schema().operator_name().name + ")";
```
CuiYifeng: Two different devices may not be `xpu` and `cpu`. We'd better provide a more general device check.

src/ATen/native/xpu/XPUFallback.template
```diff
@@ -5,9 +5,44 @@ namespace at {
 
 static bool DEBUG_XPU_FALLBACK = false;
 
+static bool check_device_inconsistency(const torch::jit::Stack* stack) {
+  bool has_xpu = false;
+  bool has_non_xpu = false;
```
CuiYifeng: I just realized that there's a corner case: both `has_xpu` and `has_non_xpu` are `false` (such as all input tensors are undefined). In this case, we cannot directly throw the error message of device inconsistency.

src/ATen/native/xpu/XPUFallback.template
```diff
@@ -5,9 +5,44 @@ namespace at {
 
 static bool DEBUG_XPU_FALLBACK = false;
 
+static bool check_device_inconsistency(const torch::jit::Stack* stack) {
```
CuiYifeng: Just a suggestion: `has_device_inconsistency` may have better self explanation and imply that the return value is `bool`.

CuiYifeng (changes requested):

lchen2331:
> Hi, without this check, there should be error raised from fallbacked CPU impl? If yes, we may not need this check.

Hi, without this check, the CPU fallback impl will not raise error, because it simply moves all inputs to the CPU for computation and then moves the results back to the original device, without checking device consistency. For example:
PYTORCH_TEST_WITH_SLOW=1 pytest test/test_ops.py -k test_errors_dot_xpu
```
________________________________________________________________________________________________________ TestCommonXPU.test_errors_dot_xpu _________________________________________________________________________________________________________
  File "/home1/lchen/miniforge3/envs/itorch/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/home1/lchen/miniforge3/envs/itorch/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/home1/lchen/miniforge3/envs/itorch/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/home1/lchen/miniforge3/envs/itorch/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 3331, in wrapper
    method(*args, **kwargs)
  File "/home1/lchen/miniforge3/envs/itorch/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 428, in instantiated_test
    result = test(self, **param_kwargs)
  File "/home1/lchen/miniforge3/envs/itorch/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 1736, in wrapper
    fn(*args, **kwargs)
  File "/home1/lchen/miniforge3/envs/itorch/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 1163, in test_wrapper
    raise e_tracked from e
Exception: RuntimeError not raised
Caused by error input at index 3: SampleInput(input=Tensor[size=(3,), device="xpu:0", dtype=torch.float32], args=TensorList[Tensor[size=(3,), device="cpu", dtype=torch.float32]], kwargs={}, broadcasts_input=False, name='')
```

etaf (approved):

PR #2628, opened 2025-12-22
Enable view_as_real and view_as_complex for SparseXPU

CuiYifeng:
This PR adds `SparseXPU` dispatch for `view_as_real` and `view_as_complex`.
Reusing common implementation `view_as_real_sparse` and `view_as_complex_sparse` to handle sparse tensors.

### DIFF
--- a/yaml/native/native_functions.yaml
+++ b/yaml/native/native_functions.yaml
@@ -1495,11 +1495,13 @@
   variants: function
   dispatch:
     XPU: view_as_real
+    SparseXPU: view_as_real_sparse
 
 - func: view_as_complex(Tensor(a) self) -> Tensor(a)
   variants: function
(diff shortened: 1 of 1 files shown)


yaml/native/native_functions.yaml
```diff
@@ -1495,11 +1495,13 @@
   variants: function
   dispatch:
     XPU: view_as_real
+    SparseXPU: view_as_real_sparse
```
guangyey, line 1498: Why CUDA doesn't have this.

CuiYifeng: Added recently for CUDA: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L421.

PR #2636, opened 2025-12-24
[WIP]enable werror

chunhuanMeng:
This pull request makes several improvements to SYCL/XPU backend code, primarily focusing on compiler warning management, build flag enhancements, and minor bug fixes. The most significant changes are the consistent handling of compiler warnings across platforms (especially for MSVC), stricter build flags to improve code quality, and some targeted bug fixes and code cleanups.

**Compiler warning management and portability:**

* Added MSVC-specific pragma directives to suppress warning C4715 and ensure proper push/pop handling in all SYCL-related kernel source files, improving cross-platform build stability.

**Build flag improvements:**

* Updated `cmake/BuildFlags.cmake` to treat warnings as errors (`/WX` for MSVC, `-Werror` for GCC), and suppressed specific SYCL attribute warnings for GCC, enforcing stricter code quality during builds.
* Added `-Wno-absolute-value` to kernel build options to suppress irrelevant warnings, and removed redundant flag setting. 

**Bug fixes and code cleanups:**

* Fixed a bug in `SpectralOps.cpp` by clarifying the contiguous check condition in `_fft_with_size`.
* Removed unused or incorrect DFT conjugate-even storage configuration in `_mkl_dft`, simplifying the code.
* Corrected the right-shift functor to return the correct value when the shift amount is out of range, preventing undefined behavior.

### DIFF
--- a/.github/actions/linux-testenv/action.yml
+++ b/.github/actions/linux-testenv/action.yml
@@ -130,6 +130,14 @@ runs:
           cd benchmark
           git checkout ${TORCHBENCH_COMMIT_ID}
           sed -i 's/^ *pynvml.*//' requirements.txt
+          # Work around legacy deps (e.g. visdom) that still import
+          # pkg_resources during build on newer Python toolchains.
+          CONSTRAINT_FILE=/tmp/pip-constraints.txt
+          echo 'setuptools<81' > ${CONSTRAINT_FILE}
+          export PIP_CONSTRAINT=${CONSTRAINT_FILE}
+          export PIP_NO_BUILD_ISOLATION=1
+          pip install -U "pip" "setuptools<81" "wheel"
+          # Follow with standard installation path
           pip install -r requirements.txt
           if [ "${{ github.event_name }}" == "pull_request" ];then
             while read line; do
--- a/src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
+++ b/src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
@@ -8,14 +8,23 @@
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
+#include <ATen/SparseCsrTensorUtils.h>
 #include <ATen/native/sparse/SparseStubs.h>
 #include <ATen/native/sparse/xpu/sycl/SparseCsrTensorMathKernels.h>
 #include <ATen/ops/_convert_indices_from_coo_to_csr_native.h>
 #include <ATen/ops/_convert_indices_from_csr_to_coo_native.h>
 
+#ifndef AT_PER_OPERATOR_HEADERS
+#include <ATen/Functions.h>
+#include <ATen/NativeFunctions.h>
+#else
+#include <ATen/ops/add.h>
+#endif
+
 namespace at::native {
 
 using namespace at::sparse;
+using namespace at::sparse_csr;
 
 TORCH_IMPL_FUNC(_convert_indices_from_coo_to_csr_structured_xpu)
 (const Tensor& input,
@@ -51,6 +60,47 @@ Tensor _sparse_csr_prod_xpu(
     std::optional<ScalarType> dtype) {
   return xpu::_sparse_csr_prod_xpu_kernel(
       input, dims_to_reduce, keepdim, dtype);
+}
+
+Tensor& add_out_sparse_compressed_xpu(
+    const Tensor& self,
+    const SparseCsrTensor& other,
+    const Scalar& alpha,
+    SparseCsrTensor& out) {
+  if (self.layout() == kStrided) {
+    at::add_out(out, self, other.to_dense(), alpha);
+    return out;
+  } else if (other.layout() == kStrided) {
+    at::add_out(out, other, self.to_dense(), alpha);
+    return out;
+  } else {
+    TORCH_CHECK(
+        self.sizes().equals(other.sizes()),
+        "torch.add: Expected input tensors to have the same shape, but got tensor `self` with shape ",
+        self.sizes(),
+        " and tensor `other` with shape ",
+        other.sizes());
+    TORCH_CHECK(
+        self.is_xpu(),
+        "add: expected 'self' to be XPU tensor, but got tensor on device: ",
+        self.device());
+    TORCH_CHECK(
+        other.is_xpu(),
+        "add: expected 'other' to be XPU tensor, but got tensor on device: ",
+        other.device());
+    TORCH_CHECK(
+        out.is_xpu(),
+        "add: expected 'out' to be XPU tensor, but got tensor on device: ",
+        out.device());
+
+    if (only_sparse_compressed_add_trivial_cases(self, other, alpha, out)) {
+      return out;
+    }
+
+    Tensor out_dense = at::add(self.to_dense(), other.to_dense(), alpha);
+    out = out_dense.to_sparse_csr();
+  }
+  return out;
 }
 
 } // namespace at::native
--- a/src/ATen/native/xpu/sycl/BinaryMiscOpsKernels.cpp
+++ b/src/ATen/native/xpu/sycl/BinaryMiscOpsKernels.cpp
@@ -74,12 +74,37 @@ struct HuberFunctor {
   scalar_t delta_val_;
 };
 
+template <>
+struct HuberFunctor<at::Half> {
+  at::Half operator()(at::Half a, at::Half b) const {
+    // Compute internally in float32 - following CPU implementation.
+    float af = static_cast<float>(a);
+    float bf = static_cast<float>(b);
+    float z = std::abs(af - bf);
+    float out = z < delta_val_ ? 0.5f * z * z
+                               : delta_val_ * (z - 0.5f * delta_val_);
+    return static_cast<at::Half>(out);
+  }
+  HuberFunctor(float delta_val) : delta_val_(delta_val) {}
+
+ private:
+  float delta_val_;
+};
+
 void huber_kernel(TensorIterator& iter, double delta) {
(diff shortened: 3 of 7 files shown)


src/ATen/native/xpu/mkl/SpectralOps.cpp
```diff
@@ -94,10 +94,6 @@ void _mkl_dft(
     desc.set_value(config_param::BWD_STRIDES, input_strides);
   }
 
-  if (!complex_input || !complex_output) {
-    desc.set_value(config_param::CONJUGATE_EVEN_STORAGE, DFTI_COMPLEX_COMPLEX);
```
chunhuanMeng, line 98: This parameter is deprecated and does not need to be set.

src/ATen/native/xpu/mkl/SpectralOps.cpp
```diff
@@ -143,7 +139,7 @@ void _fft_with_size(
     const auto strides = input_.strides();
     bool need_contiguous = strides.back() != 1;
     for (int64_t i = 0; !need_contiguous && i <= signal_ndim; i++) {
-      need_contiguous |= strides[i] % 2;
+      need_contiguous |= (strides[i] % 2 != 0);
```
chunhuanMeng, line 142: fix warning C4805: '|=': unsafe mix of type 'bool' and type 'int'

src/ATen/native/xpu/sycl/BinaryShiftOpsKernels.cpp
```diff
@@ -45,7 +45,7 @@ struct RshiftFunctor {
         sizeof(scalar_t) * CHAR_BIT - std::is_signed_v<scalar_t>;
     if ((static_cast<std::make_signed_t<scalar_t>>(b) < 0) ||
         (b >= max_shift)) {
-      return a >> max_shift;
+      return a < 0 ? scalar_t(-1) : scalar_t(0);
```
chunhuanMeng, line 48: Fix undefined behavior warning on MSVC

src/ATen/native/xpu/sycl/LinearInt4.cpp
```diff
@@ -86,8 +86,8 @@ struct LinearInt4KernelFunctor : public __SYCL_KER_CONFIG_CONVENTION__ {
           for (int ikk = 0; ikk < TileK; ikk += 2) {
             scalarx2_t tmpA = *(scalarx2_t*)(aptr + sg_id * TileK + ikk);
             scalarx2_t tmpB = {
-                static_cast<int8_t>((tmps8[ikk / 2] & 0x0f) - 8),
-                static_cast<int8_t>((tmps8[ikk / 2] >> 4) - 8)};
+                static_cast<scalar_t>((tmps8[ikk / 2] & 0x0f) - 8),
```
chunhuanMeng, line 89: Fix warning narrowing conversion

src/ATen/native/xpu/sycl/SortingRadixSort.h
```diff
@@ -120,13 +120,17 @@ class GroupRadixSort {
             KeyTraits<KeyT>::convert(c10::load(&keys_group_in[offset]));
       } else {
         KeyTraitsT padding_key;
-        if (IS_DESCENDING) {
-          padding_key = 0;
+        if constexpr (std::is_same<KeyTraitsT, bool>::value) {
```
chunhuanMeng, line 123: fix warning unsafe mix of type 'bool' and 'int', `padding_key` is `bool` and `padding_key - 1` is `int`

CuiYifeng (changes requested):
@chunhuanMeng The changes in `SpectralOps.cpp` looks good to me. Please resolve conflict first.

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd.h
```diff
@@ -9,11 +9,23 @@
  */
 
 #pragma once
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunknown-warning-option"
+#pragma clang diagnostic ignored "-Wchanges-meaning"
+#pragma clang diagnostic ignored "-Wsycl-strict"
+#pragma clang diagnostic ignored "-Wunused-local-typedefs"
```
dvrogozh, line 16: That's quite a list of warnings being suppressed. Why just not fix some, for example unused variables or typedefs? For variables there are special compiler treatments such as `[[maybe_unused]]` to mark unused variables in functions.
* https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.html

src/ATen/native/xpu/sycl/SortingRadixSort.h
dvrogozh: I don't think you actually fixed that as you did not change the code which evaluates `padding_key - 1`. What you did you introduced a new `constexpr` condition which is a compile time check. This means that if condition evaluates to `true` to compile time only "then" path will be considered and "else" will be skipped. If you don't hit your warning anymore this makes me think that "else" path was never actually executed.

src/ATen/native/quantized/sycl/QuantizedMaxPool2d.cpp
```diff
@@ -8,6 +8,12 @@
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
+
+#ifdef _MSC_VER
+  #pragma warning(push)
+  #pragma warning(disable : 4715)
```
dvrogozh, line 19: * https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4715?view=msvc-170
* which is: `'function' : not all control paths return a value`

Actually that's a valid warning and useful warning. We can lose ton of time debugging the issue when someone just missed a return statement in a function. I don't think that's a good idea to drop it. I also don't see that it is getting suppressed in the core pytorch. Can we consider to actually fix code to avoid these warnings?

src/ATen/native/xpu/sycl/SortingRadixSort.h
chunhuanMeng: This is intentional.
The bit-manipulation logic used in the generic path is not semantically
meaningful for bool. For KeyTraitsT = bool, I explicitly define the
padding value and exclude the generic computation at compile time.
The MSVC warning exposed this issue, but the change is a semantic
specialization rather than a workaround.

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd.h
chunhuanMeng: The plan is to enable -Werror first and then gate torch-xpu-ops PR. The suppressed warnings originate from SYCL TLA. We will create an issue in SYCL TLA to track this and remove the suppressions once the underlying issue is resolved.

src/ATen/native/quantized/sycl/QuantizedMaxPool2d.cpp
chunhuanMeng: Align with the Linux approach: the suppressed warnings originate from SYCL header files. We will create a tracking issue with the SYCL team and remove the suppressions once the underlying issue is resolved.

cmake/BuildFlags.cmake
```diff
@@ -53,11 +53,15 @@ macro(set_build_flags)
...
       list(APPEND SYCL_HOST_FLAGS /wd4996) # allow usage of deprecated functions
       list(APPEND SYCL_HOST_FLAGS /wd4018) # allow signed and unsigned comparison
+      list(APPEND SYCL_HOST_FLAGS /WX) # treat warnings as errors
     elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
       list(APPEND SYCL_HOST_FLAGS -fPIC)
       list(APPEND SYCL_HOST_FLAGS -std=${CPP_STD})
       list(APPEND SYCL_HOST_FLAGS -Wunused-variable)
       list(APPEND SYCL_HOST_FLAGS -Wno-interference-size)
+      # Disable warnings for unsupported SYCL attributes in GCC
+      list(APPEND SYCL_HOST_FLAGS -Wno-attributes)
```
dvrogozh: These are warnings which can be fixed on torch-xpu-ops side. Please, do not just disable them without an attempt. See:
* https://github.com/intel/torch-xpu-ops/pull/2834
* https://github.com/intel/torch-xpu-ops/pull/2835

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd.h
dvrogozh: I disagree with this approach. It is not good to massively silence warnings. That's bad route as this can grow uncontrollably. Especially in the case of SYCL TLA which is actively developed project and our dependency. We should be able to report and contribute and get the fix soon enough to avoid silencing on our side. If we can't assure quality of SYCL TLA then we will need to periodically silence other warnings as well.

src/ATen/native/quantized/sycl/QuantizedMaxPool2d.cpp
dvrogozh: Why we don't hit similar warnings on Linux?

That's too massive change in our sources. And I believe it's harmful. Can we, please, have a discussion with the compiler team on this issue? I at least would like to make sure that this is really their issue and there is no workaround or fix possible on our side.

chunhuanMeng: Linux also employs a similar warning suppression mechanism.

cmake/BuildFlags.cmake
chunhuanMeng: Thanks, I will remove related warning suppress options

PR #2644, opened 2025-12-24
[WIP]Init support for grouped_mm

liangan1:


laifenxiawucha:
Requested in [this mention](https://github.com/intel/torch-xpu-ops/pull/2644#issuecomment-4212582779).

## Summary: Likely bug (missing int64 implementation)

The issue clearly reports a missing implementation: `dot_xpu_mkl` is not implemented for `int64` (Long) dtype on XPU. The upstream PyTorch test suite expects this to work given the sample inputs being tested (`test_ops.py::TestCommonXPU.test_noncontiguous_samples_inner_xpu_int64`).

## Next step

Investigate the dispatch/registration for `dot_xpu_mkl` and add support for `torch.int64`. If `dot_xpu_mkl` is MKL-specific, check whether MKL supports int64 dot products; if not, a CPU or generic XPU fallback may be needed.

## Missing information needed

1. Confirmation of which backend paths `dot_xpu_mkl` currently supports (check the dispatch table or kernel registration files).
2. Whether MKL/oneMKL provides an int64 path for dot products or if this should fall back to a different implementation.
3. The minimal repro script (can be extracted from the upstream test call shown in the traceback if needed).

PR #2691, opened 2026-01-06
using sum/world_size to replace avg

songhappy:
fix https://jira.devtools.intel.com/browse/PYTORCHDGQ-6106.
oneccl does not allow avg and throw error while checking, so replace it using sum/world_size. nothing else is changed

### DIFF
--- a/src/xccl/ProcessGroupXCCL.cpp
+++ b/src/xccl/ProcessGroupXCCL.cpp
@@ -34,6 +34,21 @@ bool checkSameSize(const std::vector<at::Tensor>& input_tensors) {
   return true;
 }
 
+bool shouldEmulateAvg(const ReduceOp& reduceOp) {
+  // oneCCL AVG isn't supported on all scheduler paths; emulate via SUM + div.
+  return reduceOp == ReduceOp::AVG;
+}
+
+ReduceOp toXcclReduceOp(const ReduceOp& reduceOp, const at::Tensor& input) {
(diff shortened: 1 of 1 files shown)


Chao1Han:
oneCCL supports AVG directly over the XeLink path. Simulating everything with SUM followed by division would incur a performance penalty. Do we really have a strong justification to mandate that all AVG operations be emulated in this way?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -1198,17 +1213,16 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::allreduce_impl(
           xcclComm_t& comm,
           at::xpu::XPUStream& stream,
           ccl::stream& xcclStream) {
+        auto reduceOp = toXcclReduceOp(opts.reduceOp, input);
         xccl::onecclAllReduce(
-            input, output, comm, opts.reduceOp, xcclStream, stream);
-#if !defined(XCCL_HAS_AVG)
-        if (opts.reduceOp == ReduceOp::AVG) {
+            input, output, comm, reduceOp, xcclStream, stream);
+        if (shouldEmulateAvg(opts.reduceOp)) {
```
zhangxiaoli73, line 1219: Have you evaluated the perf impact? If you find some cases unsupported for avg, you should ask oneCCL for support instead of replacing avg by sum.

PR #2750, opened 2026-01-19
Add missing expect files for HasDecompTest test_aten_core_operators and test_has_decomposition

gplutop7:
This PR adds the missing .expect files for XPU decomposition tests.

The tests:

- HasDecompTest.test_aten_core_operators
- HasDecompTest.test_has_decomposition

No functionality was changed.
Only the expected reference files were generated and added.


EikanWang (approved):
The two files are generated automatically, right?

PR #2755, opened 2026-01-21
Add NaN check to prevent false singular matrix errors

lchen2331:
Fix issue: #2667 
The issue is caused by oneMKL receiving `NaN` inputs, which can lead to out-of-bounds memory access and convert `NaN` to `0`, causing the `singular check` to fail. A `NaN` check is added before oneMKL computation, and `NaN` outputs are returned when `NaN` inputs are detected to align with cuda results.

Testing:
```
pytest -sv test/regressions/test_linalg.py
```
Output:
```
test/regressions/test_linalg.py::TestLinalg::test_solve_batch_mixed_nan PASSED
test/regressions/test_linalg.py::TestLinalg::test_solve_cayley_transform PASSED
test/regressions/test_linalg.py::TestLinalg::test_solve_nan_variants PASSED
```


src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -545,6 +556,22 @@ void lu_factor_mkl(
       pivot,
       "linalg.lu_factor: LU without pivoting is not implemented on the XPU");
 
+  // NaN check: if input contains NaN, return NaN output with valid pivots
+  if (at::isnan(LU).any().item<bool>()) {
+    info.zero_();
+    // Fill pivots with default sequence [1, 2, 3, ..., min(m, n)]
+    int64_t min_mn = std::min(LU.size(-2), LU.size(-1));
```
CuiYifeng: Does this mean that `pivots` will always follow the default sequence when there is `nan` in `LU`?

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -553,7 +567,30 @@ void lu_factor_mkl(
...
+      int64_t batch_size = native::batchCount(LU);
+      int64_t m = LU.size(-2);
+      int64_t n = LU.size(-1);
+
+      // Detect NaN per-batch
+      auto nan_mask_batch = at::isnan(LU).reshape({batch_size, m * n}).any(/*dim=*/1);
+
+      // Replace NaN batches with identity matrix to avoid MKL crash
+      auto identity = at::eye(m, n, LU.options()).unsqueeze(0).expand({batch_size, m, n});
+      auto nan_mask_expanded = nan_mask_batch.unsqueeze(-1).unsqueeze(-1).expand({batch_size, m, n});
```
CuiYifeng: The shape of `nan_mask_batch` is `{batch_size}`, right? If so, can we directly reshape or view `nan_mask_batch` as `{batch_size, 1, 1}`?

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -553,7 +567,30 @@ void lu_factor_mkl(
...
+      LU.copy_(at::where(nan_mask_expanded, identity, LU));
+
+      apply_lu_xpu_<scalar_t>(LU, pivots_, info_data);
+
+      // Restore NaN for batches that originally had NaN
+      auto nan_mask_LU = nan_mask_batch.unsqueeze(-1).unsqueeze(-1)
+          .expand({batch_size, m, n});
```
CuiYifeng: Seems `nan_mask_LU` is equivalent to `nan_mask_expanded`. Please check.

test/regressions/test_linalg_solve_nan.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Owner(s): ["module: intel"]
+import torch
+from torch.testing._internal.common_utils import TestCase
+
+cpu_device = torch.device("cpu")
+xpu_device = torch.device("xpu")
+
+
+class TestLinalgSolveNaN(TestCase):
```
CuiYifeng: How about using `class TestLinalg(TestCase):`? Then, we can add the other customized linalg cases here in the future.

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
lchen2331: sure

Issue #2757, opened 2026-01-21
[BMG] test_tril got UR_RESULT_ERROR_DEVICE_LOST


mengfei25:
### 🐛 Describe the bug

Once test_tril got UR_RESULT_ERROR_DEVICE_LOST and all the next cases will be failed too.

Cases:
op_regression,third_party.torch-xpu-ops.test.regressions.test_tril.TestSimpleBinary,test_tril

```
__________________________ TestSimpleBinary.test_tril __________________________
[gw7] linux -- Python 3.12.12 /__w/torch-xpu-ops/torch-xpu-ops/.venv/bin/python
Traceback (most recent call last):
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/regressions/test_tril.py", line 13, in test_tril
    torch.xpu.synchronize()
  File "/__w/torch-xpu-ops/torch-xpu-ops/.venv/lib/python3.12/site-packages/torch/xpu/__init__.py", line 451, in synchronize
    return torch._C._xpu_synchronize(device)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: level_zero backend failed with error: 20 (UR_RESULT_ERROR_DEVICE_LOST)

To execute this test, run the following from the base repo dir:
    python test/regressions/test_tril.py TestSimpleBinary.test_tril

This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0
```

### Versions

BMG B60

PR #2760, opened 2026-01-22
[CI] Enhance unit test to rerun crash cases and clean xpu memory

mengfei25:
1. Clean XPU memory
2. Rerun the test worker to skip #2757 to run the remaining test cases

disable_e2e
disable_distributed

### DIFF
--- a/.github/actions/linux-testenv/action.yml
+++ b/.github/actions/linux-testenv/action.yml
@@ -67,7 +67,7 @@ runs:
           pip install torch torchvision torchaudio --pre --index-url https://download.pytorch.org/whl/nightly/xpu
         else
           latest_dir="$(find . -type d -name "Torch-XPU-Wheel-*" |sort -V |tail -n 1)"
-          pip install --force-reinstall $(find ${latest_dir} -name "*torch*.whl")
+          pip install --force-reinstall $(find ${latest_dir} -name "*.whl")
         fi
         TORCH_COMMIT_ID=$(python -c 'import torch; print(torch.version.git_version)')
         if [[ "${{ inputs.pytorch }}" == *"https://"* ]];then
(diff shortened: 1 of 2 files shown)


.github/scripts/conftest.py
```diff
@@ -0,0 +1,505 @@
...
+        # Run the test
+        try:
+            # Call the original runtest protocol
+            yield
+            break  # Test passed, exit loop
+
+        except Exception as e:
+            # Test failed, check if we should rerun
+            error_msg = str(e)
+            should_rerun, pattern_name, max_retries = TestRerunManager.should_rerun_test(item, error_msg)
```
chuanqi129: So we only according to the error msg to decision whether the case need rerun or not? If the know failures list has a lot of error in the RERUN _PATTERNS, will we rerun all those cases without any skip?

.github/actions/linux-uttest/action.yml
```diff
@@ -14,6 +14,9 @@ runs:
       run: |
         which python && python -V
         which pip && pip list
+        if [ "${{ github.event_name }}" != "pull_request" ];then
+          cp ${GITHUB_WORKSPACE}/.github/scripts/conftest.py pytorch/third_party/torch-xpu-ops/
+        fi
```
chuanqi129: I think we also need rerun new failed cases in PR CI test, how do you think?

mengfei25: I added the rerun in another PR #2813 , this PR is for resume mainly

.github/scripts/conftest.py
mengfei25: Will move failures rerun in another PR #2813

chuanqi129 (approved):

Issue #2772, opened 2026-01-26
Add ldexp implementation on XPU


xiaowangintel:
### 🚀 The feature, motivation and pitch

The ldexp's implementation have been changed in https://github.com/pytorch/pytorch/pull/168357. The new implementation involves adding a kernel to replace the decomposition.

PR #2773, opened 2026-01-26
Add ldexp implementation on XPU

xiaowangintel:
Related to https://github.com/intel/torch-xpu-ops/issues/2772

The ldexp's implementation have been changed in https://github.com/pytorch/pytorch/pull/168357. The new implementation involves adding a stub kernel to replace the decomposition. And we meets error as 'DispatchStub: missing kernel for xpu' when use ldexp operator with stock pytorch.

### DIFF
--- a/src/ATen/native/xpu/BinaryOps.cpp
+++ b/src/ATen/native/xpu/BinaryOps.cpp
@@ -114,7 +114,6 @@ REGISTER_XPU_DISPATCH(
 REGISTER_XPU_DISPATCH(
     shifted_chebyshev_polynomial_w_stub,
     &xpu::shifted_chebyshev_polynomial_w_kernel);
-
 REGISTER_XPU_DISPATCH(ldexp_stub, &xpu::ldexp_kernel);
 
 TORCH_IMPL_FUNC(add_out_xpu)


src/ATen/native/xpu/BinaryOps.cpp
```diff
@@ -115,6 +115,8 @@ REGISTER_XPU_DISPATCH(
     shifted_chebyshev_polynomial_w_stub,
     &xpu::shifted_chebyshev_polynomial_w_kernel);
```
CuiYifeng: Please remove this blank line.

src/ATen/native/xpu/sycl/BinaryMiscOpsKernels.cpp
```diff
@@ -126,4 +126,20 @@ void xlog1py_kernel(TensorIteratorBase& iter) {
       [&]() { gpu_kernel_with_scalars(iter, Xlog1pyFunctor<scalar_t>()); });
 }
 
+template <typename scalar_t>
+struct LdexpFunctor {
+  scalar_t operator()(scalar_t x, int exp) const {
+    return ::ldexp(x, exp);
```
CuiYifeng, line 132: One more question: which component provides `::ldexp`?

dvrogozh: This statement pulls `ldexp` from the global scope and unless PyTorch defines its own version it would be pulling C implementation from `#include <math.h>` which supports only `double`, i.e. we have implicit conversion here:
* https://en.cppreference.com/w/c/numeric/math/ldexp.html

I think `std::ldexp` might be preferable as it supports overloads for different data types, i.e. we can avoid data type conversions and get faster performance due to specialized implementations.
* https://en.cppreference.com/w/cpp/numeric/math/ldexp.html

CuiYifeng:
Close this old PR since `ldexp` has been implemented by #3248.

Issue #2736, opened 2026-01-13
[distributed] AssertionError: Scalars are not equal! with test_distributed_spawn.py


zxd1997066:
### 🐛 Describe the bug

please get wheels from https://github.com/intel/torch-xpu-ops/actions/runs/20895313621 or use gh download
```
gh run download 20895313621 --repo intel/torch-xpu-ops --name Torch-XPU-Wheel-1826-20895313621-1 --dir path --pattern "*.zip"
git clone -b distributed_2.10 https://github.com/daisyden/pytorch.git
cd pytorch
pip install -r requirements.txt
pip install pytest expecttest
export TEMP_DIR=/tmp
export BACKEND="xccl"
export WORLD_SIZE=4
pytest -v test/distributed/test_distributed_spawn.py::TestDistBackendWithSpawn::test_all_to_all_single_equal_split_cuda
pytest -v test/distributed/test_distributed_spawn.py::TestDistBackendWithSpawn::test_all_to_all_single_equal_split_cuda_complex
pytest -v test/distributed/test_distributed_spawn.py::TestDistBackendWithSpawn::test_all_to_all_single_equal_split_full_group_cuda
```
```
______________________________ TestDistBackendWithSpawn.test_all_to_all_single_equal_split_full_group_cuda _______________________________
Traceback (most recent call last):
  File "/home/sdp/miniforge3/envs/xccl_test/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/home/sdp/miniforge3/envs/xccl_test/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/home/sdp/miniforge3/envs/xccl_test/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/home/sdp/miniforge3/envs/xccl_test/lib/python3.10/site-packages/torch/testing/_internal/common_distributed.py", line 814, in wrapper
    self._join_processes(fn)
  File "/home/sdp/miniforge3/envs/xccl_test/lib/python3.10/site-packages/torch/testing/_internal/common_distributed.py", line 1083, in _join_processes
    self._check_return_codes(fn, elapsed_time)
  File "/home/sdp/miniforge3/envs/xccl_test/lib/python3.10/site-packages/torch/testing/_internal/common_distributed.py", line 1123, in _check_return_codes
    raise RuntimeError(error)
RuntimeError: Process 3 exited with error code 10 and exception:
[...]

PR #2782, opened 2026-01-27
Fix missing async param in AllToAll

frost-intel:
The async parameter was missing from the recent AllToAll implementation change in #1579 .

The missing param was causing alltoall errors in #2736

### DIFF
--- a/src/xccl/ProcessGroupXCCL.cpp
+++ b/src/xccl/ProcessGroupXCCL.cpp
@@ -2007,6 +2007,7 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::alltoall_base(
         return;
       },
       OpType::ALLTOALL_BASE,
+      opts.asyncOp,
       "xccl:all_to_all");
 }
 


src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -2007,6 +2007,7 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::alltoall_base(
         return;
       },
       OpType::ALLTOALL_BASE,
+      opts.asyncOp,
```
guangyey, line 2010: Good catch!
@Chao1Han, I wonder why doesn't raise compilation error when build pytorch.

Chao1Han: > Good catch! [@Chao1Han](https://github.com/Chao1Han), I wonder why doesn't raise compilation error when build pytorch. 

There are numerous overloads of collectives defined in `ProcessGroupXCCL.hpp`; it is possible that the string `xccl::all_to_all` was mistakenly interpreted as a boolean, which I overlooked in my earlier changes.

PR #2803, opened 2026-01-30
Support call oneccl all2all directly

Chao1Han:
This PR introduces a direct call path when oneCCL supports `onecclAllToAll`. XCCL currently provides three variants of **all-to-all**:

1. **Tensor-list all-to-all**, i.e. `ProcessGroupXCCL::alltoall`.
   It assumes the input and output tensor lists have the same length. The current implementation uses grouped send/recv, aligning with CUDA.

2. **`ProcessGroupXCCL::alltoall_base` with empty `outputSplitSizes` / `inputSplitSizes`**, i.e. equal partitioning, corresponding to the standard all-to-all semantics.
   When oneCCL directly supports all-to-all, it calls the oneCCL all-to-all API; otherwise, it falls back to grouped send/recv, aligning with CUDA.

3. **`ProcessGroupXCCL::alltoall_base` with non-empty `outputSplitSizes` / `inputSplitSizes`**, i.e. uneven partitioning, corresponding to all-to-all-v semantics.
   oneCCL does not provide this API, so it is implemented using grouped send/recv.

disable_e2e
disable_ut

### DIFF
--- a/src/xccl/xccl.cpp
+++ b/src/xccl/xccl.cpp
@@ -424,13 +424,13 @@ void onecclAllToAll(
     xcclComm_t& comm,
     ccl::stream& xcclStream,
     at::xpu::XPUStream& stream) {
-
-  auto [isUniform, uniformCount] = checkUniformAllToAll(
-      sendcounts, senddispls, recvcounts, recvdispls, numranks);
   if (isCCLV2EnabledCached()) {
     auto xcclDataType = getXcclDataTypeV2(dataType, false);
     int numranks = 0;
     onecclCommCount(comm.onecclComm, &numranks);
+
+    auto [isUniform, uniformCount] = checkUniformAllToAll(
+        sendcounts, senddispls, recvcounts, recvdispls, numranks);
 
     if (isUniform) {
       // Use native onecclAllToAll for uniform case
@@ -470,6 +470,9 @@ void onecclAllToAll(
   } else {
     auto xcclDataType = getXcclDataTypeV1(dataType, false);
     int numranks = comm.cclComm->size();
+
+    auto [isUniform, uniformCount] = checkUniformAllToAll(
+        sendcounts, senddispls, recvcounts, recvdispls, numranks);
 
     if (isUniform) {
       // Use native ccl::alltoall for uniform case


src/xccl/xccl.cpp
```diff
@@ -401,11 +401,43 @@ void onecclAllToAll(
     xcclComm_t& comm,
     ccl::stream& xcclStream,
     at::xpu::XPUStream& stream) {
-  xccl::oneccl_group_start();
   if (isCCLV2EnabledCached()) {
     auto xcclDataType = getXcclDataTypeV2(dataType, false);
     int numranks = 0;
     onecclCommCount(comm.onecclComm, &numranks);
+
+#if defined(ENABLE_XCCL_ALLTOALL_SUPPORT)
```
zhangxiaoli73: Suppose `onecclAllToAll` is always there, right? Then, why do you need this macro?

Chao1Han: I think so, let me remove it.

src/xccl/xccl.cpp
```diff
@@ -401,11 +401,41 @@ void onecclAllToAll(
...
-  xccl::oneccl_group_start();
   if (isCCLV2EnabledCached()) {
     auto xcclDataType = getXcclDataTypeV2(dataType, false);
     int numranks = 0;
     onecclCommCount(comm.onecclComm, &numranks);
+
+    // Check if this is a uniform alltoall (equal counts for all ranks)
+    // and contiguous displacements, in which case we can use the native
+    // onecclAllToAll API for better performance
+    bool isUniform = numranks > 0;
```
zhangxiaoli73: duplicate code for both ccl v1 and ccl v2?

Chao1Han: Yes, maybe I should wrap it as `checkUnify` func.

src/xccl/xccl.cpp
```diff
@@ -426,9 +466,28 @@ void onecclAllToAll(
             &stream.queue());
       }
     }
+    xccl::oneccl_group_end();
   } else {
     auto xcclDataType = getXcclDataTypeV1(dataType, false);
     int numranks = comm.cclComm->size();
+
+    auto [isUniform, uniformCount] = checkUniformAllToAll(
```
zhangxiaoli73, line 474: why don't you check before `isCCLV2EnabledCached` check?

Chao1Han: Sure, move it before version check.

zhangxiaoli73 (approved):

PR #2807, opened 2026-01-30
Implement complex logaddexp for the XPU backend

lchen2331:
To solve: #2376 : this PR adds support for logaddexp with complex dtypes on the XPU backend

### DIFF
--- a/src/ATen/native/xpu/sycl/LogAddExpKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LogAddExpKernels.cpp
@@ -31,9 +31,9 @@ c10::complex<scalar_t> _logaddexp_minmax(
     const c10::complex<scalar_t>& y) {
   scalar_t xr = std::real(x);
   scalar_t yr = std::real(y);
-  if (at::isnan(yr) || (at::isnan(std::imag(y)))) {
+  if (std::isnan(yr) || (std::isnan(std::imag(y)))) {
     return y;
-  } else if (at::isnan(xr) || (at::isnan(std::imag(x)))) {
+  } else if (std::isnan(xr) || (std::isnan(std::imag(x)))) {
     return x;
   } else if (min) {
     return (xr < yr) ? x : y;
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/LogAddExpKernels.cpp
```diff
@@ -10,16 +10,108 @@
...
+// min/max for complex numbers based on real part, with proper NaN propagation
+template <typename scalar_t, bool min>
+c10::complex<scalar_t> _logaddexp_minmax(
+    const c10::complex<scalar_t>& x,
+    const c10::complex<scalar_t>& y) {
+  scalar_t xr = std::real(x);
+  scalar_t yr = std::real(y);
+  if (at::_isnan(yr) || (at::_isnan(std::imag(y)))) {
+    return y;
+  } else if (at::_isnan(xr) || (at::_isnan(std::imag(x)))) {
```
CuiYifeng: Please replace `at::_isnan` with `std::isnan`.

src/ATen/native/xpu/sycl/LogAddExpKernels.cpp
```diff
@@ -10,16 +10,108 @@
...
+    const c10::complex<scalar_t>& y) {
+  c10::complex<scalar_t> min =
+      _logaddexp_minmax<scalar_t, /*min=*/true>(x, y);
+  c10::complex<scalar_t> max =
+      _logaddexp_minmax<scalar_t, /*min=*/false>(x, y);
+  scalar_t min_real = std::real(min);
+  scalar_t max_real = std::real(max);
+
+  // Handle NaN propagation
+  if (at::_isnan(min_real) || at::_isnan(std::imag(min))) {
```
CuiYifeng: Ditto.

CuiYifeng (changes requested):
Main part of this PR looks good to me. @daisyden I noticed that some of cases in #2376 is unavailable, could you provide a method to test these cases?

guangyey (approved):
I recall that constants in device kernel code should use `float` to avoid relying on `double`, which is not supported on Arc series GPUs. I did notice a few double-precision constants here, and I’m not entirely sure why this hasn’t caused issues on Arc so far.

That said, overall LGTM. We can adjust the constant dtypes to `float` if this turns out to be an issue.

PR #2808, opened 2026-01-30
`test_unary_ufuncs_xpu.py`: Import instead of copy

pbielak:
Currently, the `test_unary_ufuncs_xpu.py` code was copied 1:1 from the PyTorch repository. For some tests, this caused using wrong tolerances, while another test was xfailed in upstream but worked just fine on XPU. This was caused by not applying the `XPUPatchForImport` patching class. To properly apply it, this commit removes most of the copied test code, keeping only tests that need modification within the test function. All other tests are imported from the PyTorch repository and patched using the mentioned class.

After re-running the test suite, some tests were now skipped, due to skip declarations in `torch/testing/_internal/common_methods_invocations.py`. These tests were working fine on XPU, so additionally this commit adds a `_cuda_skip_xpu_run` list for tests to be "un-skipped" on XPU.

Old stats: 42 failed, 24064 passed, 946 skipped, 26 xfailed
New stats: 33 failed, 24071 passed, 810 skipped, 26 xfailed, 138 deselected

The deselected tests are due to `_jiterator_` not being available on XPU, so the `skip_list_common.py` was updated.

disable_e2e
disable_distributed


EikanWang:
@chuanqi129 , pls. help review this PR.

test/xpu/xpu_test_utils.py
```diff
@@ -466,6 +466,20 @@
     },
 }
 
+# Some cases are skipped on CUDA but work just fine on XPU
+_cuda_skip_xpu_run = [
+    ("_refs.sigmoid", "test_reference_numerics_extremal"),
```
daisyden: Do you mean cuda skipped test_reference_numerics_large in [op_db](https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/common_methods_invocations.py#L24464) ? But this should not apply to xpu since it used device_type='cuda'. 
```
DecorateInfo(unittest.skip("Skipped!"), 'TestUnaryUfuncs',
                         'test_reference_numerics_large',
                         dtypes=(torch.cdouble,), device_type='cuda'),
```

Issue #2427, opened 2025-11-27
[upstream_ut]  failed on setup with "worker 'gw0' crashed while running 'third_party/torch-xpu-ops/test/xpu/test_n



daisyden:
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_5d_size_128_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_5d_size_2_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_5d_size_32_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_5d_size_4_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_size_2_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_edge_case_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_size_1023_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_size_1024_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_size_128_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_size_256_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_size_32_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorAutogradXPU,test_layer_norm_backward_size_4_xpu
[...]

PR #2854, opened 2026-02-05
[#2427] Adjust layer_norm_backward_kernel interface to match pytorch's.

AKloniecki:
This is a replacement for previous implementation from this PR: https://github.com/intel/torch-xpu-ops/pull/2804 

Along with additional change in pytorch, (https://github.com/pytorch/pytorch/pull/174385), this allows to tie into pytorch's LayerNormBackwardKernel dispatch, to handle nested tensors via their pathway, and only substitute our kernel implementation.

This is part of resolution for #2427


src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -1173,9 +1172,9 @@ void _layer_norm_backward_kernel(
            static_cast<size_t>(tile_size_n < SIMD ? tile_size_n : SIMD)},
           getCurrentSYCLQueue(),
           kfn);
-      dgamma = dgamma_blocks.sum(0);
-      dbeta = dbeta_blocks.sum(0);
-    } else if (dgamma.defined() && !dbeta.defined()) {
+      dgamma->set_data(dgamma_blocks.sum(0));
```
guangyey: why here not `*dgamma = dgamma_blocks.sum(0);`

AKloniecki: changed

PR #2859, opened 2026-02-09
Refine complex matmul fallback support when USE_ONEMKL_XPU=0

lchen2331:
This PR refines complex matmul fallback support when `USE_ONEMKL_XPU=0`:
- Fixes writes to conjugated output tensors by using a `ConjPhysicalGuard` + `_view_as_real_physical`, preserving the conj flag while writing to the correct storage.
- Avoids `0 * NaN` propagation in the fallback path by not scaling `self` when `beta==0`, and by skipping the matmul term when `alpha==0`.
- Handles `addmv`-style shape cases by reshaping computed real/imag results to match the output view before copying.

### DIFF
--- a/src/ATen/native/xpu/Blas.cpp
+++ b/src/ATen/native/xpu/Blas.cpp
@@ -30,8 +30,7 @@ namespace {
 class ConjPhysicalGuard final {
  public:
   explicit ConjPhysicalGuard(Tensor& out)
-      : out_(out),
-        was_conj_(out.is_conj()) {
+      : out_(out), was_conj_(out.is_conj()) {
     // When the output tensor is conjugated, avoid mutating its conj state.
     // Instead, operate on a non-conjugated temporary and copy the conjugated
     // result back in the destructor.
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/Blas.cpp
```diff
@@ -9,6 +9,7 @@
  */
 
 #include <ATen/ATen.h>
+#include <ATen/native/ComplexHelper.h>
```
CuiYifeng, line 12: Which function we use depends this header file?

src/ATen/native/xpu/Blas.cpp
```diff
@@ -25,18 +26,50 @@ namespace at::native {
...
+      tmp_ = out_;
+    }
+  }
+
+  ~ConjPhysicalGuard() {
+    if (was_conj_) {
+      // Copy the result back into the original conjugated tensor.
+      out_.copy_(tmp_);
+    }
+  }
```
CuiYifeng, line 52: Could you add more detailed comments to explain the behavior of `~ConjPhysicalGuard()`?
There seem to be two implicit logics:
1. `copy_` will follow the semantics of the target tensor `out_`.
2. `~ConjPhysicalGuard()` will be executed before the function `*_complex_fallback` returns.

src/ATen/native/xpu/Blas.cpp
lchen2331: Hi, `at::view_as_real()`  depends on it.

src/ATen/native/xpu/Blas.cpp
lchen2331: Thanks for the suggestion，I have refine the comments.

PR #2866, opened 2026-02-09
Fix Metadata Inconsistency in nonzero Op for XPU Backend to Support Sparse Tensor Construction

yucai-intel:
1. Problem Statement
https://github.com/intel/torch-xpu-ops/issues/2209
When performing sparse tensor conversions (e.g., dense.to_sparse_csr()) on the XPU backend, a RuntimeError: expected col_indices to be a contiguous tensor per batch is triggered if the input tensor contains zero elements (specifically when at least one dimension is zero, such as size [1, 0, 3]).

2. Root Cause Analysis
The failure originates from a metadata mismatch during the interaction between the nonzero operator and subsequent tensor slicing operations:

- Operator Output: For an input with size [1, 0], nonzero_xpu returns an index tensor nz of size [0, 2]. By default, the XPU allocator assigns a stride of [2, 1] to this tensor.

- View Generation: The sparse conversion logic calls col_indices = nz.select(1, 1). Per PyTorch’s Dispatcher logic for strided views:new_storage_offset = original_offset + (index * stride[1]) -> 0 + (1 * 1) = 1.The resulting col_indices (size [0]) inherits the stride of the first dimension of nz, which is 2.

- Validation Conflict: The native sparse validation layer in SparseCsrTensor.cpp enforces a strict continuity check: TORCH_CHECK(plain_indices.stride(-1) == 1, ...). Since col_indices has a stride of 2 and a non-zero storage offset, the check fails, even though the tensor is logically empty.

3. Technical Solution
Metadata Alignment via Physical Transposition. The fix is implemented within nonzero_out_xpu to ensure that empty result tensors possess a memory layout that is invariant to slicing.Instead of a direct allocation of {0, dim}, we utilize a transposed allocation strategy:

- Mechanism: Allocate an intermediate tensor with dimensions {dim, 0} and then transpose it to {0, dim} before binding it to the output.

- Resulting Stride Physics: An empty tensor allocated as {dim, 0} typically has strides {0, 1}. Upon transposition to {0, dim}, the strides become {1, 0}.

- Invariant Slicing: When select(1, 1) is called on a tensor with stride {1, 0}, the offset calculation becomes 0 + 1 * 0 = 0, and the resulting stride for the 1D slice is 1. This satisfies the downstream stride(-1) == 1 requirement.

### DIFF
--- a/test/regressions/test_conversion.py
+++ b/test/regressions/test_conversion.py
@@ -44,6 +44,7 @@ class TestSimpleConversion(TestCase):
         y = x.nonzero()
         self.assertEqual(y.stride(), (1, 1))
 
+
 instantiate_device_type_tests(
     TestSimpleConversion, globals(), only_for="xpu", allow_xpu=True
 )


src/ATen/native/xpu/Nonzero.cpp
```diff
@@ -45,7 +45,9 @@ Tensor& nonzero_out_xpu(const Tensor& self, Tensor& out) {
       out.dtype());
   nonzero_common_checks(self, out, "nonzero");
   if (self.numel() == 0) {
-    out = at::detail::empty_xpu({0, self.dim()}, out.options());
+    at::Tensor out_temp = at::detail::empty_xpu({self.dim(), 0}, out.options());
+    at::Tensor out_final = out_temp.t();
+    out.set_(out_final);
```
CuiYifeng, line 50:
```suggestion
    out.as_strided_({0, self.dim()}, {1, 1});
```

liangan1: Both your changes should get the same result. I think you should use the original changes from caiyu. The non-zero is not the hardware-specific operator and we should follow cuda's implementation to refactor this operator.

CuiYifeng: This suggestion is for code-style. We can merge this PR once CI passed. Agree with refactoring this kernel based on CUDA's implementation in another PR.

CuiYifeng:
@liangan1 Please take another look, thanks.

PR #2874, opened 2026-02-10
Wrapping an anonymous namespace to clean warnings

chunhuanMeng:
* Wrapped the anonymous namespace and its complex matrix multiplication implementations with `#if !defined(USE_ONEMKL_XPU)` and `#endif` preprocessor directives to exclude this code when oneMKL is enabled. [[1]](diffhunk://#diff-39081611578e65fafbe069bf91c19784f4d90cacc7296181886da17ccb86e822R26) [[2]](diffhunk://#diff-39081611578e65fafbe069bf91c19784f4d90cacc7296181886da17ccb86e822R177)


src/ATen/native/xpu/Blas.cpp
```diff
@@ -23,6 +23,7 @@
 
 namespace at::native {
 
+#if defined(USE_ONEMKL_XPU)
```
CuiYifeng: @chunhuanMeng Please note that `mm_complex_fallback` is only used when `USE_ONEMKL_XPU` is not defined.
```
#if defined(USE_ONEMKL_XPU)
  return at::native::xpu::mm_complex_out_xpu_mkl(self, mat2, out);
#else
  return mm_complex_fallback(self, mat2, out);
#endif // USE_ONEMKL_XPU
}
```

chunhuanMeng: Thanks, I have updated the code.

CuiYifeng (approved):

PR #2884, opened 2026-02-11
Allow `nonzero` execution for large inputs

pbielak:
The `nonzero` op is disabled for large inputs (number of elements exceeding int32), however the implementation allows handling such large input tensors. This commit removes the `TORCH_CHECK` for large input tensors.

disable_e2e
disable_distributed

### DIFF
--- a/test/xpu/test_unary_ufuncs_xpu.py
+++ b/test/xpu/test_unary_ufuncs_xpu.py
@@ -54,6 +54,7 @@ from torch.testing._internal.common_utils import (
     IS_WINDOWS,
     numpy_to_torch_dtype_dict,
     run_tests,
+    serialTest,
     skipIfNoSciPy,
     slowTest,
     slowTestIf,
@@ -1662,6 +1663,7 @@ class TestUnaryUfuncs(TestCase):
         self.assertEqual(1, len(z))
         self.assertEqual(torch.empty(0, dtype=torch.long), z[0])
 
+    @serialTest()
     @dtypes(torch.int8)
     @largeTensorTest("8GB")
     def test_nonzero_large(self, device, dtype):
--- a/src/ATen/native/xpu/Nonzero.cpp
+++ b/src/ATen/native/xpu/Nonzero.cpp
@@ -23,11 +23,6 @@ void nonzero_common_checks(
     Tensor& out,
     const std::string& op_name) {
   TORCH_CHECK(
-      self.numel() < std::numeric_limits<int>::max(),
-      op_name,
-      " is not supported for tensors with more than INT_MAX elements, \
-      See https://github.com/pytorch/pytorch/issues/51871");
-  TORCH_CHECK(
       self.device() == out.device(),
       "expected self and out to be on the same device, but got out on ",
       out.device(),


EikanWang (approved):

tszulist-hbn:
**Finding 1** 🟡 **Suggestion** · `test/xpu/test_unary_ufuncs_xpu.py:L1666`

`@largeTensorTest("8GB")` significantly underestimates the actual peak memory usage.

The test creates a tensor with N ≈ 4.3B elements (int8, ~4GB), but `nonzero_template` allocates `idx_flat` of shape `{N}` with `dtype=kLong` (~34GB). Then `copy_if` allocates two more N-element int64 tensors (`global_mask`, `target_pos`) simultaneously (~68GB). **Peak ≈ 106GB**, not 8GB.

**Suggestion:** Update the guard to reflect the true peak (or add a comment explaining that the XPU driver supports memory oversubscription / shared virtual memory, which is why 8GB is sufficient in practice).

```python
@largeTensorTest("20GB")  # or higher to account for O(3N) int64 scratch space
```

---
*[- Reviewed by Awesome ChlOpus]*

tszulist-hbn:
**Finding 2** 🟡 **Suggestion** · `src/ATen/native/xpu/Nonzero.cpp:L22`

The removed `TORCH_CHECK` was the only user-facing error gate before the kernel attempts O(3N) int64 scratch allocations. For **dense** large tensors (e.g., all-ones with N > INT_MAX), removing the check means the failure mode changes from a clear, helpful error:

> `nonzero is not supported for tensors with more than INT_MAX elements`

...to a silent OOM or an opaque SYCL driver error.

The XPU kernel correctly uses `int64_t` indices throughout (unlike the CUDA path that triggered the original pytorch/pytorch issue 51871), so the correctness concern is gone. But the memory concern is real.

**Suggestion:** Replace the old check with a comment (or a new memory-awareness check) explaining why large inputs are now supported and what the memory cost is:

```cpp
// XPU uses int64_t indices throughout; correctness constraint from
// pytorch/pytorch issue 51871 does not apply here.
// Note: intermediate buffers are O(3*N) int64, so large dense tensors
// may require significant device memory beyond the input size.
```

---
*[- Reviewed by Awesome ChlOpus]*

tszulist-hbn:
**Finding 3** 🟢 **Nit** · `test/xpu/test_unary_ufuncs_xpu.py:L1665`

`test_nonzero_large` only covers `torch.int8`. This is a pre-existing gap, but since this PR is specifically enabling the large-input path, it's a good opportunity to extend `@dtypes` to at least one floating-point type (e.g., `torch.float16` or `torch.bfloat16`) to broaden coverage of the newly enabled path.

**Suggestion:**
```python
@dtypes(torch.int8, torch.bfloat16)
```
(or add a separate focused test)

---
*[- Reviewed by Awesome ChlOpus]*

test/xpu/test_unary_ufuncs_xpu.py
```diff
@@ -1662,6 +1663,7 @@ def assert_tuple_empty(tup, dim):
         self.assertEqual(1, len(z))
         self.assertEqual(torch.empty(0, dtype=torch.long), z[0])
 
+    @serialTest()
     @dtypes(torch.int8)
     @largeTensorTest("8GB")
```
tszulist-hbn, line 1668: Worth considering: I wonder if 8GB is enough. It seems nonzero kernel implementation may allocate much more.

tszulist-hbn (approved):

pbielak:
Waiting for re-implementation of nonzero kernel - PR https://github.com/intel/torch-xpu-ops/pull/3374 

Once approved, current PR can be closed

pbielak:
Closing in favour of https://github.com/intel/torch-xpu-ops/pull/3374

Issue #2781, opened 2026-01-27
Accuracy small failures for instance_norm


PawelSwider2000:
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_decomp_xpu.TestDecompXPU,test_comprehensive_nn_functional_instance_norm_xpu_float64

BBBela:
It is related to #2269 
Resolving #2269 should make 4 out of 5 test cases from this issue work.

Only one issue seems to have different root cause:
ops.test.xpu.test_decomp_xpu.TestDecompXPU,test_comprehensive_nn_functional_instance_norm_xpu_float64

astachowiczhabana:
I've moved:

op_ut,third_party.torch-xpu-ops.test.xpu.test_decomp_xpu.TestDecompXPU,test_comprehensive_baddbmm_xpu_float64
op_ut,third_party.torch-xpu-ops.test.xpu.test_decomp_xpu.TestDecompXPU,test_quick_baddbmm_xpu_float64
op_ut,third_party.torch-xpu-ops.test.xpu.test_decomp_xpu.TestDecompXPU,test_quick_core_backward_baddbmm_xpu_float64
op_ut,third_party.torch-xpu-ops.test.xpu.test_decomp_xpu.TestDecompXPU,test_quick_addmv_xpu_float64

to https://github.com/intel/torch-xpu-ops/issues/2269

PR #2892, opened 2026-02-16
Fix factor calculation for double

PawelSwider2000:
Fixes: #2781 

Improve `factor` calculation for double datatypes.

Previously `factor` was always calculated with float precision and then upcasted to double. This leads to small differences in `var` and `mean` (double->float precision error).

With this fix we are calculating `factor` with `stat_accscalar_t` precision retaining behavior for float and lower precision computations and achieving proper accuracy for double datatype.

### DIFF
--- a/src/ATen/native/xpu/sycl/BatchNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/BatchNormKernels.cpp
@@ -471,8 +471,7 @@ struct BatchNormCollectStatisticsKernelFunctor
       stat_accscalar_t o_avg = sycl::permute_group_by_xor(sg, avg, i);
       int o_n = sycl::permute_group_by_xor(sg, n, i);
       stat_accscalar_t factor = static_cast<stat_accscalar_t>(1.0) /
-          fmax(static_cast<stat_accscalar_t>(1.0),
-               static_cast<stat_accscalar_t>(n + o_n));
+          static_cast<stat_accscalar_t>(std::max(1, n + o_n));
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/BatchNormKernels.cpp
```diff
@@ -470,7 +470,9 @@ struct BatchNormCollectStatisticsKernelFunctor
     for (int i = 1; i < SIMD; i <<= 1) {
       stat_accscalar_t o_avg = sycl::permute_group_by_xor(sg, avg, i);
       int o_n = sycl::permute_group_by_xor(sg, n, i);
-      stat_accscalar_t factor = 1.0 / fmaxf(1.0, n + o_n);
+      stat_accscalar_t factor = static_cast<stat_accscalar_t>(1.0) /
+          fmax(static_cast<stat_accscalar_t>(1.0),
+               static_cast<stat_accscalar_t>(n + o_n));
```
CuiYifeng: @PawelSwider2000 Could you try `stat_accscalar_t factor = static_cast<stat_accscalar_t>(1.0) / static_cast<stat_accscalar_t>(std::max(1, n + o_n))` in local? The accuracy gap may ultimately be introduced by `div` instead of `fmaxf`. Since both `n` and `o_n` are int, `std::max + static_cast` may be a better choice.

jianyizh: agree

Issue #2838, opened 2026-02-04
index_put and index_copy erroneous results


kdrozd-dev:
### 🐛 Describe the bug

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_indexing_xpu.TestIndexingXPU,test_index_copy_deterministic_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_indexing_xpu.TestIndexingXPU,test_index_put_non_accumulate_deterministic_xpu

Errors:

- test_index_copy_deterministic_xpu:
```Python
E           AssertionError: Tensor-likes are not equal!
E           Mismatched elements: 24000 / 24000 (100.0%)
E           Greatest absolute difference: 0.9936878681182861 at index (103, 0, 2)
E           Greatest relative difference: 21759.560546875 at index (687, 3, 2)
E           To execute this test, run the following from the base repo dir:
E               PYTORCH_TEST_WITH_SLOW=1 python test/test_indexing.py TestIndexingXPU.test_index_copy_deterministic_xpu
E           This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0
```
- test_index_put_non_accumulate_deterministic_xpu:
```Python
E           AssertionError: Tensor-likes are not close!
E           Mismatched elements: 17 / 17 (100.0%)
E           Greatest absolute difference: 0.8402017951011658 at index (8,) (up to 1e-05 allowed)
E           Greatest relative difference: 21.065969467163086 at index (0,) (up to 1.3e-06 allowed)
E           To execute this test, run the following from the base repo dir:
E               PYTORCH_TEST_WITH_SLOW=1 python test/test_indexing.py TestIndexingXPU.test_index_put_non_accumulate_deterministic_xpu
E           This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0
```

### Versions

PyTorch2.11.0.dev20260203+xpu

PR #2895, opened 2026-02-18
Adjust index_put_deterministic to match CUDA and CPU

kdrozd-dev:
Fixes https://github.com/intel/torch-xpu-ops/issues/2838.

Match deterministic index_put behavior with CPU and CUDA - instead of taking the first of duplicate indices as the one used for index_put take the last one. Also split the deterministic functor into two versions(accumulate and non accumulate) to avoid unnecessary checks in the functors.


src/ATen/native/xpu/sycl/Indexing.h
```diff
@@ -839,16 +840,21 @@ struct IndexPutDeterministicKernelFunctor {
...
+         inner_idx <= id.glb_batch && sorted_indices_[inner_idx] == idx;
          inner_idx++) {
       int64_t idx_orig = indices_[inner_idx];
       int64_t v_gid = idx_orig * stride_ + v_stride;
       if (accumulate_) {
         acc += (accscalar_t)c10::load(&value_[v_gid]);
       } else {
+        // For non-accumulate, keep overwriting — last iteration wins
         self_[s_gid] = c10::load(&value_[v_gid]);
-        break;
```
CuiYifeng: For work-items reached the current loop body, `id.glb_batch` is the last `inner_idx`. Since the goal is to follow "last iteration wins", how about overwriting `self_[s_gid]` without loop for non-accumulate? Then, we may avoid some redundant load/store.

src/ATen/native/xpu/sycl/Indexing.h
```diff
@@ -839,20 +840,24 @@ struct IndexPutDeterministicKernelFunctor {
...
-    for (int64_t inner_idx = id.glb_batch;
-         inner_idx < cfg_.problem_batch_ && sorted_indices_[inner_idx] == idx;
-         inner_idx++) {
-      int64_t idx_orig = indices_[inner_idx];
-      int64_t v_gid = idx_orig * stride_ + v_stride;
-      if (accumulate_) {
+
+    int64_t start_idx = id.glb_batch;
+    while (start_idx > 0 && sorted_indices_[start_idx - 1] == idx)
+      --start_idx;
```
CuiYifeng: `start_idx` is only used when `accumulate_ == true`, so this part of calculation will be redundant to `accumulate_ == false`. How about reversing the direction of the loop of `accumulate_`?

kdrozd-dev: sure

kdrozd-dev: It seems that reversing the loop somehow causes the test_index_put_large_indices_xpu to fail. I'm not sure why.

kdrozd-dev: It is probably the order of float additions that's causing the problem. I have a solution in mind - will push it on monday

kdrozd-dev: @CuiYifeng to avoid unnecessary checks ```if(accumulate_)``` I've split the functor into accumulate version and non accumulate version and which one runs gets decided once at the launch instead of checking in each functor. This solves both the deterministic put issue and accumulate accuracy issue mentioned earlier which was caused by the reverse loop direction.

CuiYifeng: LGTM.

CuiYifeng (approved):

Issue #2542, opened 2025-12-11
Title: [upstream_ut]  RuntimeError: reflection_pad2d_backward_xpu does not have a deterministic implementation, but you s


daisyden:
Cases:
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_ReflectionPad2d_large_deterministic_xpu~~

pytest_command:
cd <pytorch>  && PYTORCH_TEST_WITH_SLOW=1 pytest -v third_party/torch-xpu-ops/test/xpu/test_nn_xpu.py -k test_ReflectionPad2d_large_deterministic_xpu

Error Message:
 RuntimeError: reflection_pad2d_backward_xpu does not have a deterministic implementation, but you set 'torch.use_deterministic_algorithms(True)'. You can turn off determinism just for this operation, or you can use the 'warn_only=True' option, if that's acceptable for your application. You can also file an issue at https://github.com/pytorch/pytorch/issues to help us prioritize adding deterministic support for this operation. 

Trace Example:

```
Command: cd <pytorch>  && PYTORCH_TEST_WITH_SLOW=1 pytest -v third_party/torch-xpu-ops/test/xpu/test_nn_xpu.py -k test_ReflectionPad2d_large_deterministic_xpu
Traceback (most recent call last):
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 3329, in wrapper
    method(*args, **kwargs)
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 448, in instantiated_test
    raise rte
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 428, in instantiated_test
    result = test(self, **param_kwargs)
[...]

PR #2901, opened 2026-02-19
Implement deterministic ReflectionPad2dBackward kernel

pbielak:
The `pytorch_CycleGAN_and_pix2pix` model (end2end tests) currently fails due to `eager_two_runs_differ`, which means that the training is non-deterministic. When trying to set the deterministic flag, the script fails as the `ReflectionPad2d`'s backward kernel implementation is non-deterministic. This commit implements a deterministic version of this backward kernel, allowing the model to pass the end2end tests. The implementation is based on the existing CUDA implementation.

### DIFF
--- a/src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
+++ b/src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
@@ -370,6 +370,184 @@ void reflection_pad2d_backward_template(
       kfn);
 }
 
+template <typename scalar_t>
+struct ReflectionPad2dBackwardDetKernelFunctor {
+  void operator()(sycl::nd_item<1> item) const {
+    const int64_t tid = static_cast<int64_t>(item.get_global_id(0));
+
+    const int64_t width = input_dim_x_ + pad_left_ + pad_right_;
+    const int64_t height = input_dim_y_ + pad_top_ + pad_bottom_;
+    const int64_t N = height * width;
+
+    const int64_t end = nbatch_ * nplane_ * input_dim_y_ * input_dim_x_;
+
+    if (tid >= end) {
+      return;
+    }
+    // linear index over B*C*H*W (contiguous)
+    const int64_t pos_xy = tid % (input_dim_x_ * input_dim_y_);
+    const int64_t inp_row = pos_xy / input_dim_x_;
+    const int64_t inp_col = pos_xy % input_dim_x_;
+
+    const int64_t bottom_row = input_dim_y_ - 1;
+    const int64_t rightmost_col = input_dim_x_ - 1;
+    const int64_t dist_from_bottom = std::abs(inp_row - bottom_row);
+    const int64_t dist_from_right = std::abs(inp_col - rightmost_col);
+
+    const bool is_top = (inp_row >= 1) && (inp_row <= pad_top_);
+    const bool is_bottom =
+        (inp_row < bottom_row) && (inp_row >= bottom_row - pad_bottom_);
+    const bool is_left = (inp_col >= 1) && (inp_col <= pad_left_);
+    const bool is_right =
+        (inp_col < rightmost_col) && (inp_col >= rightmost_col - pad_right_);
+
+    // Accumulate grad_output at reflected output position (row, col)
+    // into partial, if the linearized index falls within [0, N).
+    scalar_t partial = static_cast<scalar_t>(0);
+
+    const int64_t batch_idx = tid / (nplane_ * input_dim_x_ * input_dim_y_);
+    const int64_t channel_idx = (tid / (input_dim_x_ * input_dim_y_)) % nplane_;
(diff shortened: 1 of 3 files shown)


src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
```diff
@@ -370,6 +370,182 @@ void reflection_pad2d_backward_template(
...
+      const int64_t inp_col = pos_xy % input_dim_x_;
+
+      const bool is_top = (inp_row >= 1) && (inp_row <= pad_t_);
+      const bool is_bottom =
+          (inp_row < input_dim_y_ - 1) && (inp_row >= input_dim_y_ - pad_b_ - 1);
+      const bool is_left = (inp_col >= 1) && (inp_col <= pad_l_);
+      const bool is_right =
+          (inp_col < input_dim_x_ - 1) && (inp_col >= input_dim_x_ - pad_r_ - 1);
+
+      const int64_t bottom_row = input_dim_y_ - 1;
```
CuiYifeng: Since `bottom_row`  defined to simplify code and improve readability, how about using `bottom_row` for `is_bottom`?

src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
```diff
@@ -370,6 +370,182 @@ void reflection_pad2d_backward_template(
...
+      const int64_t dist_from_bottom = std::abs(inp_row - bottom_row);
+      const int64_t dist_from_right = std::abs(inp_col - (input_dim_x_ - 1));
```
CuiYifeng: For the consistency of style, please refer to the definition like `const int64_t rightmost_col = input_dim_x_ - 1`.

src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
```diff
@@ -370,6 +370,182 @@ void reflection_pad2d_backward_template(
...
+    const int64_t tid = static_cast<int64_t>(item.get_global_id(2));
+    const int64_t stride = static_cast<int64_t>(item.get_global_range(2));
+
+    const int64_t width = input_dim_x_ + pad_l_ + pad_r_;
+    const int64_t height = input_dim_y_ + pad_t_ + pad_b_;
+    const int64_t N = height * width;
+
+    const int64_t end = nbatch_ * nplane_ * input_dim_y_ * input_dim_x_;
+
+    for (int64_t input_xy = tid; input_xy < end; input_xy += stride) {
```
CuiYifeng: Please note that each work-item will only execute this loop body at most once because `stride >= end`, where `stride = work_group_size * ceil_div(total_elements, work_group_size)` and `end = total_elements`.

src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
```diff
@@ -370,6 +370,182 @@ void reflection_pad2d_backward_template(
...
+      pad_l,
+      pad_r,
+      nbatch,
+      nplane);
+
+  // Use a flat 1D range (not 3D nbatch x nplane x ...) because the kernel
+  // iterates over all input elements with a grid-stride loop to avoid atomics.
+  sycl_kernel_submit(
+      sycl::range<3>(1, 1, work_group_size * work_group_num),
+      sycl::range<3>(1, 1, work_group_size),
```
CuiYifeng: Just a kindly reminder: `sycl::range<1>(X)` is also an option here. Seems using `sycl::range<3>(1, 1, X)` instead of `sycl::range<1>(X)` is for readability?

src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
pbielak: Updated

src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
pbielak: Updated

src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
pbielak: Code refactored

src/ATen/native/xpu/sycl/ReflectionPadKernels.cpp
pbielak: Thanks, I updated the code to use `sycl::range<1>`

pbielak:
@CuiYifeng There is already a test case for this kernel: **test_ReflectionPad2d_large_deterministic_xpu** - until now, it failed as noted in issue: https://github.com/intel/torch-xpu-ops/issues/2542 I checked the CI logs and it started passing.

Note that this specific test checks **large inputs**, so if you'd like I can add another similar test case with smaller inputs or update this one to include more input cases.

PR #2911, opened 2026-02-23
Fix E2E model setup

pbielak:
Currently, some models fail (e.g., `pytorch_CycleGAN_and_pix2pix`) due to `ModuleNotFoundError: No module named 'pkg_resources'`. This is a known-issue in setuptools on Python 3.12. We need to pin the setuptools version and disable build isolation. This commit introduces the mentioned changes.

disable_ut
disable_distributed


pbielak:
@CuiYifeng Run finished - now (with this PR's patch), the `pkg_resources` error is eliminated.

mengfei25 (approved):

Issue #2006, opened 2025-09-02
work-item/workgroup issue in softmax/unsampling/nonzero


daisyden:
### 🐛 Describe the bug
Platfrom: PVC Linux
Reproduce command:
```
cd pytorch; pip install -r .ci/docker/requirements-ci.txt; cd third_party/torch-xpu-ops/xpu; PYTORCH_TEST_WITH_SLOW=1 pytest -v test_nn_xpu.py::TestNNDeviceTypeXPU -k 'test_upsampling_64bit_indexing_channels_last_xpu_bfloat16 or test_upsampling_64bit_indexing_channels_last_xpu_float16 or test_softmax_backward_64bit_indexing_xpu or test_warp_softmax_64bit_indexing_xpu_float16 or test_warp_softmax_64bit_indexing_xpu_float32'; PYTORCH_TEST_WITH_SLOW=1 pytest -v test_unary_ufuncs_xpu.py::TestUnaryUfuncsXPU -k test_nonzero_large_xpu_int8
```
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_upsampling_64bit_indexing_channels_last_xpu_bfloat16
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_upsampling_64bit_indexing_channels_last_xpu_float16
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_softmax_backward_64bit_indexing_xpu~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_warp_softmax_64bit_indexing_xpu_float16~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_warp_softmax_64bit_indexing_xpu_float32~~
op_ut,third_party.torch-xpu-ops.test.xpu.test_unary_ufuncs_xpu.TestUnaryUfuncsXPU,test_nonzero_large_xpu_int8

- [x] test_nn_xpu.py::TestNNDeviceTypeXPU::test_softmax_results_xpu_float16  || Fixed

```
2025-09-01T23:24:56.1436175Z _________ TestNNDeviceTypeXPU.test_softmax_backward_64bit_indexing_xpu _________
[...]

PR #2949, opened 2026-02-27
Fix 64-bit indexing overflow in Softmax backward kernel for XPU

yucai-intel:
**Description**
This PR fixes a critical integer overflow issue in the softmax_backward_kernel on XPU (Intel GPU). 

**The Problem**
https://github.com/intel/torch-xpu-ops/issues/2006 test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_softmax_backward_64bit_indexing_xpu
When dealing with "large tensors" (where numel > 2^31 - 1), the dim_size parameter in the softmax backward kernel would truncate due to being declared as a 32-bit int.
This caused a chain reaction of failures:
- Invalid Work-group Size: The calculation of local_range derived from dim_size would overflow, resulting in nonsensical values (e.g., ~268M threads) that exceed the hardware limit of 1024. This triggered the RuntimeError: The number of work-items in each dimension of a work-group cannot exceed {1024, 1024, 1024}.
- Numerical Incorrectness: Even if the kernel launched, the truncated dim_size led to incorrect memory offsets and incomplete reductions, causing wrong gradients.

**The Fix**
Promoted the dim_size parameter from int to int64_t in the softmax_backward_kernel signature. This ensures that:
Intermediate calculations like (dim_size + vec_size - 1) / vec_size are performed in 64-bit precision.
The std::min comparison with syclMaxWorkGroupSize works correctly, properly capping the local_range at 1024.

**Testing**
Verified with test/test_nn.py::test_softmax_backward_64bit_indexing.
Before: RuntimeError and incorrect local_range (268435454).
After: Kernel launches successfully with local_range correctly capped at 1024, and precision matches CPU results.


yucai-intel:
> Please fix build error first:
> 
> ```
> third_party/torch-xpu-ops/src/ATen/native/xpu/sycl/SoftMaxKernels.cpp:1350:20: error: no matching function for call to 'min'
>  1350 |   int local_size = std::min(
>       |                    ^~~~~~~~
> ```


Done.

Issue #2515, opened 2025-12-10
[upstream_ut]  RuntimeError: iter.device(arg).is_xpu() INTERNAL ASSERT FAILED at "/pytorch/third_party/torch-xpu-o



libohao1201:
Cases:
~~op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_addcmul_cuda_errors_with_cpu_scalars_xpu~~
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestTorchDeviceTypeXPU,test_addcmul_cuda_errors_with_cpu_scalars_xpu

pytest_command:
cd <pytorch> && PYTORCH_TEST_WITH_SLOW=1 pytest -v test_torch_xpu.py -k test_addcmul_cuda_errors_with_cpu_scalars_xpu

Error Message:
 RuntimeError: iter.device(arg).is_xpu() INTERNAL ASSERT FAILED at "/pytorch/third_party/torch-xpu-ops/src/ATen/native/xpu/sycl/Loops.h":626, please report a bug to PyTorch. argument 2: expected an XPU device but found cpu ; AssertionError: "CPU Scalar support for tensor1 argument" does not match "iter.device(arg).is_xpu() INTERNAL ASSERT FAILED at "/pytorch/third_party/torch-xpu-ops/src/ATen/native/xpu/sycl/Loops.h":626, please report a bug to PyTorch. argument 2: expected an XPU device but found cpu" 

Trace Example:

Silv3S:
Minimal repro
```Python
import torch

input = torch.rand((2, 2), device="xpu")
tensor1 = torch.rand([], device="cpu")
tensor2 = torch.rand((2, 2), device="xpu")
torch.addcmul(input, tensor1, tensor2, value=0.5)
```
The problem with XPU op are missing TORCH_CHECKs that would early exit before evaluating tensor device placement

PR #2985, opened 2026-03-04
addcmul | Early exit if inputs are on different devices (CPU and XPU)

Silv3S:
Fixes https://github.com/intel/torch-xpu-ops/issues/2515

This patch introduces early exit with user-friendly error message. Without this TORCH_CHECK, the functionality doesn't work anyway, but the error message is generic

```Python
RuntimeError: iter.device(arg).is_xpu() INTERNAL ASSERT FAILED
```
from kernel.


Silv3S:
XPU was missing input device placement guards. This patch introduces early exit with user-friendly error message. Without this `TORCH_CHECK` at operator level, the functionality doesn't work anyway,  but the error message is generic
```Python
RuntimeError: iter.device(arg).is_xpu() INTERNAL ASSERT FAILED
```
from kernel level.

Issue #2490, opened 2025-12-10
[upstream_ut]  AssertionError: 'XPU.used\t\t true' not found in '[TORCH_VITAL] Dataloader.enabled\t\t True\n[TORCH



libohao1201:
Cases:
op_ut,test_torch_xpu.TestVitalSignsCudaXPU,test_cuda_vitals_gpu_only_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestVitalSignsCudaXPU,test_cuda_vitals_gpu_only_xpu

pytest_command:
cd <pytorch> && PYTORCH_TEST_WITH_SLOW=1 pytest -v test_torch_xpu.py -k test_cuda_vitals_gpu_only_xpu

Error Message:
 AssertionError: 'XPU.used\t\t true' not found in '[TORCH_VITAL] Dataloader.enabled\t\t True\n[TORCH_VITAL] Dataloader.basic_unit_test\t\t TEST_VALUE_STRING\n[TORCH_VITAL] CUDA.used\t\t False\n' 

Trace Example:

PR #2987, opened 2026-03-04
Disable TestVitalSignsCuda::test_cuda_vitals_gpu_only on XPU

pbielak:
The VitalsAPI is not longer used and XPU will not support it. Hence,
let's skip this test -- now running only on CUDA.

Fixes https://github.com/intel/torch-xpu-ops/issues/2490

### DIFF
--- a/test/xpu/test_torch_xpu.py
+++ b/test/xpu/test_torch_xpu.py
@@ -269,14 +269,10 @@ class TestBasicVitalSigns(TestCase):
 
 # FIXME: document or deprecate whatever this is
 class TestVitalSignsCuda(TestCase):
-    @onlyXPU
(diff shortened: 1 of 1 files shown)


test/xpu/test_torch_xpu.py
```diff
@@ -268,14 +269,10 @@ def test_dataloader_vitals(self):
 
 # FIXME: document or deprecate whatever this is
 class TestVitalSignsCuda(TestCase):
-    @onlyXPU
+    @onlyCUDA
```
guangyey:
```suggestion
    @onlyCUDA # VitalsAPI has been deactivated and will remain disabled unless a valid use case is identified.
```

pbielak: Done

Issue #2239, opened 2025-10-30
Exception: could not create a primitive descriptor for the deconvolution forward propagation primitive. in test/functorch/test_ops.py


zxd1997066:
### 🐛 Describe the bug

Cases:
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vjp_nn_functional_conv_transpose2d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vjp_nn_functional_conv_transpose3d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vmapjvpall_has_batch_rule_nn_functional_conv_transpose2d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vmapjvpall_has_batch_rule_nn_functional_conv_transpose3d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_grad_nn_functional_conv_transpose2d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_grad_nn_functional_conv_transpose3d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_jvpvjp_nn_functional_conv_transpose2d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_jvpvjp_nn_functional_conv_transpose3d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vjpvmap_nn_functional_conv3d_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vjpvmap_nn_functional_conv_transpose2d_xpu_float32~~
[...]

PR #2995, opened 2026-03-05
Increase assert tolerance for non-deterministic ops in test_vjpvmap

wpietka:
Fixes op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vjpvmap_nn_functional_conv_transpose3d_xpu_float32 from #2239

Fixes op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vjpvmap_nn_functional_conv2d_xpu_float32
from #2729 

Basically the same issue and same fix as in #2988 

Test fails because backward propagation in vjp() is non-deterministic by default. We should expect outputs differ to some degree. This PR increases tolerance for tranpose3d and conv2d in failing test


test/xpu/functorch/test_ops_xpu.py
```diff
@@ -1743,13 +1743,18 @@ def test_vjpvmap(self, device, dtype, op):
...
-                _, vjp_fn = vjp(fn, *primals)
-                result_vjps = vjp_fn(cotangents)
+                for deterministic in [False, True]:
+                    torch.use_deterministic_algorithms(deterministic)
+                    _, vjp_fn = vjp(fn, *primals)
+                    result_vjps = vjp_fn(cotangents)
 
-                _, vjp_fn = ref_vjp(fn, *primals)
-                expected_vjps = vjp_fn(cotangents)
+                    _, vjp_fn = ref_vjp(fn, *primals)
```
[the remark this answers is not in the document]
wpietka: PR description has been edited

test/xpu/functorch/test_ops_xpu.py
```diff
@@ -1699,6 +1699,13 @@ def test():
             }
         ),
     )
+    @opsToleranceOverride(
+        "TestOperators",
+        "test_vjpvmap",
+        (
+            tol1("nn.functional.conv_transpose3d", {torch.float32: tol(atol=1e-04, rtol=1.3e-06)}),
+        ),
+    )
```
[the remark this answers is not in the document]
wpietka: outdated

test/xpu/functorch/test_ops_xpu.py
```diff
@@ -1718,6 +1718,22 @@ def test():
...
+        "test_vjpvmap",
+        (
+            tol1(
+                "nn.functional.conv_transpose3d",
+                {torch.float32: tol(atol=1e-04, rtol=1.3e-06)},
+                device_type=device_type,
+            ),
+            tol1(
+                "nn.functional.conv2d",
+                {torch.float32: tol(atol=1e-04, rtol=1.3e-06)},
```
CuiYifeng: I noticed the following error of `test_vjpvmap_nn_functional_conv_transpose3d_xpu_float32` in local:
```
Mismatched elements: 1 / 960 (0.1%)
Greatest absolute difference: 1.1444091796875e-05 at index (1, 1, 3, 4, 4, 0) (up to 1e-05 allowed)
Greatest relative difference: 1.859928852354642e-05 at index (1, 1, 3, 4, 4, 0) (up to 1.3e-06 allowed)
```
1.3e-06 is the original rtol. Don't we need to adjust it?
And, the atol may not need to be increased by an order of magnitude. How about using 2e-05?

wpietka: Test passes for me in 100% runs, but it might depend on specific device. I've adjusted atol and rtol to match CUDA https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/common_methods_invocations.py#L16070C20-L16070C82

etaf:
Maybe we should upstream it like #2988 and https://github.com/pytorch/pytorch/pull/177069/ , I'm not sure.

CuiYifeng:
> Maybe we should upstream it like #2988 and [pytorch/pytorch#177069](https://github.com/pytorch/pytorch/pull/177069) , I'm not sure.

Seems these cases also failed due to non-deterministic oneDNN behavior.

BBBela:
Hi @etaf @CuiYifeng,
yes, the cases from https://github.com/intel/torch-xpu-ops/pull/2988 and https://github.com/pytorch/pytorch/pull/177069 are failing due to oneDNN nondeterminism. I also initially created PR to increase tolerance in the `torch-xpu-ops`, but was advised by @guangyey to move those overrides to PyTorch directly. That is why I prepared the https://github.com/pytorch/pytorch/pull/177069 - but it waits a long time to be approved by the Community.

Maybe it would be good to settle on some common approach for all tolerance overrides - what do you think?
So it is clear for everyone if we should add tolerance overrides to PyTorch directly or to torch-xpu-ops.

etaf (approved):

CuiYifeng:
> Hi @etaf @CuiYifeng, yes, the cases from #2988 and [pytorch/pytorch#177069](https://github.com/pytorch/pytorch/pull/177069) are failing due to oneDNN nondeterminism. I also initially created PR to increase tolerance in the `torch-xpu-ops`, but was advised by @guangyey to move those overrides to PyTorch directly. That is why I prepared the [pytorch/pytorch#177069](https://github.com/pytorch/pytorch/pull/177069) - but it waits a long time to be approved by the Community.
> 
> Maybe it would be good to settle on some common approach for all tolerance overrides - what do you think? So it is clear for everyone if we should add tolerance overrides to PyTorch directly or to torch-xpu-ops.

Sorry for the late response. In my understanding, adding tolerance overrides to PyTorch or torch-xpu-ops is decided by the status of upstream. @daisyden Could you provide guideline since you are leading UT upstream?

About waiting for Community's review, personally, I do not oppose temporarily making changes (such as changing decorator) in torch-xpu-ops to avoid frequent waiting for community. @daisyden Please also provide detailed guideline, thanks.

CuiYifeng (approved):

daisyden:
> > Hi @etaf @CuiYifeng, yes, the cases from #2988 and [pytorch/pytorch#177069](https://github.com/pytorch/pytorch/pull/177069) are failing due to oneDNN nondeterminism. I also initially created PR to increase tolerance in the `torch-xpu-ops`, but was advised by @guangyey to move those overrides to PyTorch directly. That is why I prepared the [pytorch/pytorch#177069](https://github.com/pytorch/pytorch/pull/177069) - but it waits a long time to be approved by the Community.
> > Maybe it would be good to settle on some common approach for all tolerance overrides - what do you think? So it is clear for everyone if we should add tolerance overrides to PyTorch directly or to torch-xpu-ops.
> 
> Sorry for the late response. In my understanding, adding tolerance overrides to PyTorch or torch-xpu-ops is decided by the status of upstream. @daisyden Could you provide guideline since you are leading UT upstream?
> 
> About waiting for Community's review, personally, I do not oppose temporarily making changes (such as changing decorator) in torch-xpu-ops to avoid frequent waiting for community. @daisyden Please also provide detailed guideline, thanks.

Yes, let's submit PR to torch-xpu-ops first to fix the tolerance issue. At the same time we can submit PR to upstream and waiting for the review. For both PR please also add comments to justify this update with the torch-xpu-ops issue linked.

test/xpu/functorch/test_ops_xpu.py
```diff
@@ -1706,6 +1706,22 @@ def test():
...
         ),
     )
+    @opsToleranceOverride(
+        "TestOperators",
+        "test_vjpvmap",
+        (
+            tol1(
+                "nn.functional.conv_transpose3d",
+                {torch.float32: tol(atol=5e-5, rtol=5e-6)},
+                device_type=device_type,
```
BBBela:
```suggestion
                device_type="xpu",
```
I think we should use the `"xpu"` directly so it is clear that we want to override the tolerance only for this one device type. We do not want to override it for CUDA or CPU by accident during porting tests to PyTorch.

test/xpu/functorch/test_ops_xpu.py
```diff
@@ -1706,6 +1706,22 @@ def test():
...
+            ),
+            tol1(
+                "nn.functional.conv2d",
+                {torch.float32: tol(atol=5e-5, rtol=5e-6)},
+                device_type=device_type,
```
BBBela:
```suggestion
                device_type="xpu",
```
I think we should use the `"xpu"` directly so it is clear that we want to override the tolerance only for this one device type. We do not want to override it for CUDA or CPU by accident during porting tests to PyTorch.

test/xpu/functorch/test_ops_xpu.py
BartoszKokoszko: done

test/xpu/functorch/test_ops_xpu.py
BartoszKokoszko: done

BBBela (approved):
Looks good to me, thanks!

BartoszKokoszko:
Adding "disable_all" label as this PR changes only test tolerance.

chuanqi129:

vpirogov:
@wpietka, @BBBela, could you please help me understand what is referred as "non-deterministic behavior" here?

BBBela:
Hello @vpirogov,
my understanding regarding "non-deterministic behavior" is that calling the same operator with exactly the same inputs, can produce slightly different results due to different order of accumulation operations for example. I mean in general due to floating point numbers operations being not associative.
But please correct me if my understanding is wrong.

vpirogov:
> my understanding regarding "non-deterministic behavior" is that calling the same operator with exactly the same inputs, can produce slightly different results due to different order of accumulation operations for example.

This matches my understanding. If you see non-deterministic behavior with oneDNN operations in deterministic mode please report it. All oneDNN functionality must be deterministic when deterministic mode is enabled.

BBBela:
> > my understanding regarding "non-deterministic behavior" is that calling the same operator with exactly the same inputs, can produce slightly different results due to different order of accumulation operations for example.
> 
> This matches my understanding. If you see non-deterministic behavior with oneDNN operations in deterministic mode please report it. All oneDNN functionality must be deterministic when deterministic mode is enabled.

Thank you for this information. Sure, I will report any non-deterministic behavior in deterministic mode to oneDNN team.
We are adding those tolerance overrides as PyTorch tests in most cases use the default mode which is the non-deterministic mode. When I encounter similar issues, my first check is if test still fails when deterministic mode is enabled. And with deterministic mode enabled, everything always worked for me so far.

vpirogov:
Got it! Thanks, @BBBela!

PR #3002, opened 2026-03-06
Fix nn.functional.embedding operator by ensuring proper indices offsets.

BBBela:
Is part of the solution for #2238 

It fixes test case:
> op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_ops_xpu.TestOperatorsXPU,test_vmapvjp_nn_functional_embedding_functorch_xpu_float32

## Problem description

The XPU implementation of `nn.functional.embedding` operator includes the [count_by_segment()](https://github.com/intel/torch-xpu-ops/blob/08b58cfbf2869f7cd7bef8f7fc48a0118b2a83cf/src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h#L1071-L1128)  function call, which was producing wrong results, especially the [inclusive_scan()](https://github.com/intel/torch-xpu-ops/blob/08b58cfbf2869f7cd7bef8f7fc48a0118b2a83cf/src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h#L1101) part.
The `inclusive_scan` was producing the group indices starting from 1. Later on, they are used as c-style array indices, which start from 0. So basically, there was offset equal to real index plus one. This error was propagating and resulting in wrong output.
It is a problem of indexing from 0 vs indexing from 1.

### Example walkthrough

Example input tensor with already sorted embedding indices: `[2, 5, 5, 5, 7, 7, 8, 9, 9]`

**Step 1 - mask generation**

Creates mask, where:
1 = new group start
0 = same group continues

Mask: `[1, 1, 0, 0, 1, 0, 1, 1, 0]`

**Step 2 - position calculation**

It basically creates new tensor, where the values start from initial value and increase by `1` for every occurrence of `1` in the `mask`.

So for the `mask`: `[1, 1, 0, 0, 1, 0, 1, 1, 0]`
And the original starting value: `0`
We get the positions: `[1, 2, 2, 2, 3, 3, 4, 5, 5]`

**Step 3 - relies only on mask, so the bug wasn't affecting this step**

It calculates the number of elements in each group.

Group size: `[1, 3, 2, 1, 2]`

**Step 4 - where the bug propagates**

This step copies the calculated group sizes to each element using the positions from Step 2 as indices:

```cpp
positions = [1, 2, 2, 2, 3, 3, 4, 5, 5]
group_sizes = [1, 3, 2, 1, 2, ?, ?, ?, ?]  // 0-indexed array

result[i] = group_sizes[positions[i]]
```

The `positions[i]` is used as array index, so it should start from `0`, not from `1` to produce correct results.

### Summary

In short, the problem is that current solution is inconsistent in terms of enumerating. In one place the first group has index `0` and in the second place the first group has index `1`.

## Proposed solution

First element of `mask` is always `1` because the first array element always start a new group.
So the positions in current solution also always start from 1, because they are calculated as:
`positions[0] = initial_value + mask[0]`

So we can functionally subtract `1` from all the elements of the `positions` array or set the `initial_value` to `-1`, so the `positions` array will be created since the beginning with correct 0-based indices.

I propose the second solution - to simply specify the initial value as `-1`, so we get the 0-based indices in the first place.


src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h
```diff
@@ -1098,7 +1098,7 @@ OutputIt count_by_segment(
   sycl_kernel_submit(sycl::range<1>(N), q, kfn1);
 
   // 2. get target positions with inclusive_scan
```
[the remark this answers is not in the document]
BBBela: I added a constexpr variable with descriptive name instead of extending the comment.

src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h
```diff
@@ -1098,7 +1098,8 @@ OutputIt count_by_segment(
...
-  inclusive_scan(gmask_ptr, gmask_ptr + N, tpos_ptr, static_cast<index_t>(0));
+  constexpr index_t ZERO_BASED_INDEX_OFFSET = static_cast<index_t>(-1);
+  inclusive_scan(gmask_ptr, gmask_ptr + N, tpos_ptr, ZERO_BASED_INDEX_OFFSET);
```
guangyey, line 1102:
```suggestion
  exclusive_scan(gmask_ptr, gmask_ptr + N, tpos_ptr, static_cast<index_t>(0));
```

guangyey: When gmask[0] is 1, the `exclusive_scan`'s result is based on 0.

BBBela: The `exclusive_scan` with initial value `0` does not work the same as `inclusive_scan` with initial value `-1`.
For example:
```cpp
mask = [1, 1, 0, 0, 1, 0, 1, 1, 0]
inclusive_scan(mask, 0) = [1, 2, 2, 2, 3, 3, 4, 5, 5]
inclusive_scan(mask, -1) = [0, 1, 1, 1, 2, 2, 3, 4, 4]
exclusive_scan(mask, 0) = [0, 1, 2, 2, 2, 3, 3, 4, 5]
```

The effect of `exclusive_scan(mask, 0)` when compared to `inclusive_scan(mask, 0)` is that it shifts the elements of the array by one place, while we need to subtract `1` from every element to get the 0-based indices.

To make sure, I compiled and tested code in all those configurations, and confirmed that changing the `inclusive_scan` to `exclusive_scan` does not resolve the issue.

PR #3046, opened 2026-03-12
Fix Windows Build error LNK2005

CuiYifeng:
This PR is to fix Windows Build error:
```
torch_cpu.lib(torch_cpu.dll) : error LNK2005: "class at::Tensor __cdecl at::native::view_as_complex(class at::Tensor const &)" (?view_as_complex@native@at@@YA?AVTensor@2@AEBV32@@Z) already defined in torch_xpu_ops.lib(Blas.cpp.obj)
torch_cpu.lib(torch_cpu.dll) : error LNK2005: "class at::Tensor __cdecl at::native::view_as_real(class at::Tensor const &)" (?view_as_real@native@at@@YA?AVTensor@2@AEBV32@@Z) already defined in torch_xpu_ops.lib(Blas.cpp.obj)
```
Root Cause:
`ComplexHelper.h` contains non-inline functions and should be only included from ONE cpp file. However, this header was incorrectly included by `src/ATen/native/xpu/Blas.cpp‎`.

The changes include:

- Remove `ATen/native/ComplexHelper.h`
- Use `ATen/ops/view_as_real_native.h` only when `USE_ONEMKL_XPU` is undefined.
- Replace `at::view_as_real` with `at::native::view_as_real`

### DIFF
--- a/src/ATen/native/xpu/Blas.cpp
+++ b/src/ATen/native/xpu/Blas.cpp
@@ -9,7 +9,6 @@
  */
 
 #include <ATen/ATen.h>
-#include <ATen/native/ComplexHelper.h>
 #include <ATen/native/xpu/Blas.h>
 #if defined(USE_ONEMKL_XPU)
 #include <ATen/native/xpu/mkl/BlasImpl.h>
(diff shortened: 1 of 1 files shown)


guangyey (approved):
Sounds good!

PR #3049, opened 2026-03-13
Using `2^(y * log2(x))` instead of `pow` achieves higher precision

Chao1Han:
According to driver team comments "because GPUs usually don't support x^y natively, but only 2^y—so it must be calculated with 2^(y*log2( x ))." And sglang team also report RotaryEmbedding accuracy gap caused by `pow`, Add exp implement here.


Chao1Han:
clear not needed pr.

PR #3053, opened 2026-03-13
Update CI bot to cross out skipped tests

pbielak:


.github/scripts/check-ut.py
mengfei25, line 298: Looks like we just need add skipped here together with passed to check 'failed -> skipped or passed' cases, then cross out it in `ut_result_check.sh` according to the status

PR #3063, opened 2026-03-16
Add fused gate+up GEMM + SiLU kernel (sycl-tla)

xuhancn:
Add a fused SYCL kernel for the Llama MLP gate+up projection pattern:  output = SiLU(input @ gate_weight^T) * (input @ up_weight^T)

The kernel concatenates gate_weight and up_weight into a single [2N, K] matrix, performs one GEMM via sycl-tla CUTLASS for Intel Xe), then applies SiLU activation and element-wise multiply in a fused SYCL kernel.
This eliminates the second GEMM dispatch and intermediate memory traffic.

Kernel details:
- TileN=64 with M-based tile dispatch (Small/Medium/Large)
- Per-SG tile 16x32 = 512 accumulators, K=32, PipelineStages=3
- SiLU*mul kernel uses 2D nd_range (avoids int div for non-power-of-2 N)
- fp16 and bf16 support, fp32 accumulation

Further plan:
Registered as torch.ops.xpu._fused_gate_up_silu with XPU dispatch key.
Includes is_fused_gate_up_silu_available() (USE_SYCLTLA guard) and torch.ops.xpu._is_fused_gate_up_silu_available() for Inductor extra_check.

Authored with Claude.


xuhancn:
close as https://github.com/intel/torch-xpu-ops/pull/3091, sycl-tla is upgrade to v0.7.

Issue #2828, opened 2026-02-03
[Bug Skip]: test_foreach_norm_empty_tensor_inf_error__foreach_norm_xpu got failed


mengfei25:
### 🐛 Describe the bug with skip template

New test cases test_foreach_norm_empty_tensor_inf_error__foreach_norm_xpu  got failed

Cases:
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_foreach_xpu.TestForeachXPU,test_foreach_norm_empty_tensor_inf_error__foreach_norm_xpu_bfloat16~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_foreach_xpu.TestForeachXPU,test_foreach_norm_empty_tensor_inf_error__foreach_norm_xpu_float16~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_foreach_xpu.TestForeachXPU,test_foreach_norm_empty_tensor_inf_error__foreach_norm_xpu_float32~~
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_foreach_xpu.TestForeachXPU,test_foreach_norm_empty_tensor_inf_error__foreach_norm_xpu_float64~~

```
_ TestForeachXPU.test_foreach_norm_empty_tensor_inf_error__foreach_norm_xpu_float64 _
[gw1] linux -- Python 3.10.19 /tmp/xpu-tool/Python/3.10.19/x64/bin/python
Traceback (most recent call last):
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/../../../../test/test_foreach.py", line 1140, in test_foreach_norm_empty_tensor_inf_error
    with self.assertRaisesRegex(
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 226, in __exit__
    self._raiseFailure("{} not raised".format(exc_name))
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 163, in _raiseFailure
    raise self.test_case.failureException(msg)
AssertionError: RuntimeError not raised

To execute this test, run the following from the base repo dir:
    PYTORCH_TEST_WITH_SLOW=1 python ../../test/test_foreach.py TestForeachXPU.test_foreach_norm_empty_tensor_inf_error__foreach_norm_xpu_float64

[...]

PR #3070, opened 2026-03-17
Added check for infinity norm when one of the tensors is empty

Silv3S:
Fixes: https://github.com/intel/torch-xpu-ops/issues/2828

Align XPU backend to newly introduced `TORCH_CHECK`. It was added lately for CPU and CUDA in https://github.com/pytorch/pytorch/pull/173238.

Quote from original PR:
1. Checks if ord == infinity
2. If so, verifies all tensors have numel() > 0
3. Throws an error _foreach_norm cannot compute the infinity norm on an empty tensor because the operation does not have an identity


src/ATen/native/xpu/ForeachReduceOp.cpp
```diff
@@ -65,12 +65,29 @@ std::vector<Tensor> foreach_tensor_norm_xpu(
     }
   }();
   at::native::check_foreach_api_restrictions(tensors);
-  const bool has_int_or_complex =
-      std::any_of(tensors.begin(), tensors.end(), [](const auto& t) {
-        const auto scalar_type = t.scalar_type();
-        return at::isIntegralType(scalar_type, /*includeBool*/ true) ||
-            at::isComplexType(scalar_type);
-      });
+   // If the tensor is empty and norm == infty, we cannot compute the norm
```
Silv3S:
```suggestion
  // If the tensor is empty and norm == infty, we cannot compute the norm
```

Issue #2833, opened 2026-02-04
test_nn upsampling issue (4)


daisyden:
### 🐛 Describe the bug

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_upsamplingBiMode2d_consistency_memory_format1_mode_bicubic_antialias_False_align_corners_True_num_channels_5_output_size_32_check_as_unsqueezed_3d_tensor_True_non_contig_restrided_batch_size_1_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_upsamplingBiMode2d_consistency_memory_format1_mode_bicubic_antialias_False_align_corners_True_num_channels_5_output_size_32_check_as_unsqueezed_3d_tensor_True_non_contig_restrided_batch_size_5_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_upsamplingBiMode2d_consistency_memory_format1_mode_bicubic_antialias_False_align_corners_True_num_channels_5_output_size_32_check_as_unsqueezed_3d_tensor_True_non_contig_sliced_batch_size_1_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_upsamplingBiMode2d_consistency_memory_format1_mode_bicubic_antialias_False_align_corners_True_num_channels_5_output_size_32_check_as_unsqueezed_3d_tensor_True_non_contig_sliced_batch_size_5_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_upsamplingBiMode2d_consistency_memory_format1_mode_bicubic_antialias_False_align_corners_True_num_channels_5_output_size_600_check_as_unsqueezed_3d_tensor_False_non_contig_False_batch_size_1_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_upsamplingBiMode2d_consistency_memory_format1_mode_bicubic_antialias_False_align_corners_True_num_channels_5_output_size_600_check_as_unsqueezed_3d_tensor_False_non_contig_False_batch_size_5_xpu
[...]

PR #3078, opened 2026-03-19
Skip `test_upsamplingBiLinear2d_consistency_interp_size_bug` on XPU

chunhuanMeng:
This pull request makes a minor update to the test suite for PyTorch's neural network module on XPU devices. The change ensures that the test for bilinear upsampling skips both CUDA and XPU devices when uint8 is not supported, improving test consistency for device-specific limitations.
Fix https://github.com/intel/torch-xpu-ops/issues/2833

### DIFF
--- a/test/xpu/test_nn_xpu.py
+++ b/test/xpu/test_nn_xpu.py
@@ -13561,9 +13561,11 @@ class TestNNDeviceType(NNTestCase):
         self, device, memory_format, align_corners, input_size, output_size
     ):
         # Non-regression test for https://github.com/pytorch/pytorch/pull/101403
-
-        if torch.device(device).type in ("cuda", "xpu"):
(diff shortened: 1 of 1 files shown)


test/xpu/test_nn_xpu.py
```diff
@@ -13562,7 +13562,7 @@ def test_upsamplingBiLinear2d_consistency_interp_size_bug(
     ):
         # Non-regression test for https://github.com/pytorch/pytorch/pull/101403
 
-        if torch.device(device).type == "cuda":
+        if torch.device(device).type in ("cuda", "xpu"):
             raise SkipTest("CUDA implementation is not yet supporting uint8")
```
CuiYifeng: @chunhuanMeng Don't we need to update the error message?

CuiYifeng (approved):

PR #3092, opened 2026-03-19
[#3089] override test_max_pool2d_cudnn for xpu

jmamzax:
Pytorch test_max_pool2d_cudnn explicitly copy data to "cuda" device, so test was override in test_quantized_op_xpu.py and data is copied to "xpu:0" device. After this change test no longer fail with 'AssertionError: Torch not compiled with CUDA enabled' but also fail with 'NotImplementedError: Could not run 'aten::dequantize.self' with arguments from the 'QuantizedXPU' backend.' error. This should be fixed when https://github.com/pytorch/pytorch/pull/173923 will be merged.


test/xpu/quantization/core/test_quantized_op_xpu.py
```diff
@@ -75,7 +77,63 @@ def _test_max_pool2d_pt2e(self):
...
+    ),
+    kernel=st.sampled_from((3, 5, 7)),
+    stride=st.sampled_from((None, 1, 2)),
+    # currently there is no support for dilation for cudnn
+    # pooling
+    dilation=st.integers(1, 1),
+    padding=st.integers(0, 2),
+    ceil_mode=st.booleans(),
+)
+def _test_max_pool2d_cudnn(self, X, kernel, stride, dilation, padding, ceil_mode):
```
Silv3S, line 91:
```suggestion
def _test_max_pool2d_[?](self, X, kernel, stride, dilation, padding, ceil_mode):
```
Need other name for this test

test/xpu/quantization/core/test_quantized_op_xpu.py
```diff
@@ -75,7 +77,63 @@ def _test_max_pool2d_pt2e(self):
...
+    X=hu.tensor(
+        shapes=hu.array_shapes(min_dims=3, max_dims=4, min_side=1, max_side=10),
+        # cudnn's support for quantized pooling is limited to
+        # int8 currently
+        qparams=hu.qparams(dtypes=[torch.qint8]),
+    ),
+    kernel=st.sampled_from((3, 5, 7)),
+    stride=st.sampled_from((None, 1, 2)),
+    # currently there is no support for dilation for cudnn
+    # pooling
```
Silv3S:
```suggestion
```

test/xpu/quantization/core/test_quantized_op_xpu.py
```diff
@@ -75,7 +77,63 @@ def _test_max_pool2d_pt2e(self):
         self.assertEqual(a_pool, a_hat, msg="ops.quantized.max_pool2d results are off")
 
 
+@given(
+    X=hu.tensor(
+        shapes=hu.array_shapes(min_dims=3, max_dims=4, min_side=1, max_side=10),
+        # cudnn's support for quantized pooling is limited to
+        # int8 currently
```
Silv3S:
```suggestion
```

guangyey (commented):
We could change the code to catch the expected failure "NotImplementedError: Could not run 'aten::dequantize.self` explictly. And file a new PR to fix it after https://github.com/pytorch/pytorch/pull/173923 is landed.

jmamzax:
Pytorch change: https://github.com/pytorch/pytorch/pull/173923 was rejected so I add test to skip list as this test checks only quantization for max_pool2d.

jmamzax:
Skipped in https://github.com/intel/torch-xpu-ops/pull/3424 without fix as quantization in pytorch is obsolete and there will be no fixes for xpu.

Issue #2505, opened 2025-12-10
[upstream_ut]  AssertionError: 'tensor([ 100.0000, 0.0100])' != 'tensor([100.0000, 0.0100])'


libohao1201:
Cases:
op_ut,test_torch_xpu.TestTorch,test_print
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestTorch,test_print

pytest_command:
cd <pytorch> && PYTORCH_TEST_WITH_SLOW=1 pytest -v test_torch_xpu.py -k test_print

Error Message:
AssertionError: 'tensor([  100.0000,     0.0100])' != 'tensor([100.0000,   0.0100])'

Issue #2508, opened 2025-12-10
TypedStorage / TypedTensors deprecation


libohao1201:
Cases:
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_bool
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_uint8
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_int8
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_int32
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_int16
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_int64
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_float32
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_float64
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_complex64
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_complex128
op_ut,test_torch_xpu.TestTorch,test_storage_error
op_ut,test_torch_xpu.TestTorch,test_storage_error_no_attribute
op_ut,test_torch_xpu.TestTorch,test_typed_storage_deprecation_warning
op_ut,test_torch_xpu.TestTorch,test_typed_storage_internal_no_warning
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_grad_scaling_state_dict_xpu
op_ut,test_torch_xpu.TestTorch,test_print
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestTorch,test_print
op_ut,test_torch_xpu.TestTorchDeviceTypeXPU,test_serialization_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_complex128
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestTorchDeviceTypeXPU,test_storage_setitem_xpu_complex64
[...]

PR #3107, opened 2026-03-20
Remove extra whitespaces in expected print result

Silv3S:
Fixes https://github.com/intel/torch-xpu-ops/issues/2505

Currently test fails on extra white space, that will be fixed in this PR.
After fix, another test case will fail on missing torch.xpu.FloatStorage that is tracked in: https://github.com/intel/torch-xpu-ops/issues/2508/


kdrozd-dev (approved):

Silv3S:
This PR is trivial - removing whitespace in reference output in one testcase. CI is not stable and it fails on tests not related to introduced change. I will merge it, to not waste CI resources on reruns

Issue #2519, opened 2025-12-10
[upstream_ut]  TypeError: map2_ is only implemented on CPU tensors 



libohao1201:
Cases:
op_ut,test_torch.TestTorchDeviceTypeXPU,test_broadcast_fn_map2_xpu
op_ut,test_torch.TestTorchDeviceTypeXPU,test_broadcast_fn_map_xpu

pytest_command:
cd <pytorch> && PYTORCH_TEST_WITH_SLOW=1 pytest -v test_torch.py -k test_broadcast_fn_map2_xpu
cd <pytorch> && PYTORCH_TEST_WITH_SLOW=1 pytest -v test_torch.py -k test_broadcast_fn_map_xpu

Error Message:
 TypeError: map2_ is only implemented on CPU tensors 

Trace Example:

PR #3112, opened 2026-03-20
Skip test_broadcast_fn_map(2) on XPU

Silv3S:
Fixes https://github.com/intel/torch-xpu-ops/issues/2519
map and map2_ are only implemented on CPU tensors. CUDA has this tests skipped and XPU wasn't included on skiplist


test/xpu/test_torch_xpu.py
```diff
@@ -1399,8 +1399,8 @@ def test_broadcast(self, fn, device):
             small2 = torch.randn(*dims_small2, device=device).float()
             small2_expanded = small2.expand(*dims_full)
 
-        if small.is_cuda and fn in ["map", "map2"]:
-            # map and map2 are not implementd on CUDA tensors
+        if (small.is_cuda or small.is_xpu) and fn in ["map", "map2"]:
```
guangyey:
```suggestion
        if (small.device.type in ["cuda", "xpu"]) and fn in ["map", "map2"]:
```

guangyey (commented):
@daisyden, I think we should upstream the code change, what do you think.

chuanqi129:
✅ PR has been successfully merged by @Silv3S.

guangyey:
> @guangyey we discussed with @chuanqi129 that it would be best to incorporate this (an other similar) patches into single upstreaming PR, to not submit each small adjustment separately

Make sense.

PR #3119, opened 2026-03-23
Explicitly skip test_out_ for histc and mean ops

Silv3S:
Fixes https://github.com/intel/torch-xpu-ops/issues/2234
Both tests are expected to fail, same as on CUDA. We can utilize `skip_list_common.py` to skip them, before the official skip in `common_methods_invocations.py` is upstreamed


chuanqi129:

PR #3193, opened 2026-03-26
Fix PackedSequence.xpu binding and multi-XPU device handling in test

BartoszKokoszko:
- Bind PackedSequence.xpu directly to the helper instead of MethodType so Python creates a proper bound method; the previous MethodType call had swapped args, binding self to the class and triggering the PackedSequence constructor assertion.
- In the XPU test loop, create the PackedSequence on the specific target device (a.xpu(device=xpu)) so b.to(xpu) returns the same object and assertIs passes on multi-XPU setups.


CuiYifeng (approved):
LGTM.

test/xpu/nn/test_packed_sequence_xpu.py
```diff
@@ -67,7 +65,7 @@ def my_test_to(self):
                     "xpu",
                     "xpu:0" if torch.xpu.device_count() == 1 else "xpu:1",
                 ]:
-                    b = a.xpu()
+                    b = a.xpu(device=xpu)
```
Silv3S, line 68:
```suggestion
                    b = a.xpu()
```
Not sure if needed, but CUDA also invoked `a.cuda(device=cuda)`

Silv3S (approved):

test/xpu/nn/test_packed_sequence_xpu.py
BartoszKokoszko: If I remember correctly a.xpu() moves "a" to the default xpu device. We would like to move it to the specific one:

for xpu in [
                    "xpu",
                    "xpu:0" if torch.xpu.device_count() == 1 else "xpu:1",

test/xpu/nn/test_packed_sequence_xpu.py
```diff
@@ -67,7 +65,7 @@ def my_test_to(self):
...
                     self.assertIs(b, b.to(xpu))
```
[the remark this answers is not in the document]
BartoszKokoszko: For  kwargs {"device": "xpu:1"}  ex.device.type is equal to "xpu", so it will go through:
```
if ex.device.type == "xpu":
   return self.to(*args, **kwargs)
```
not through:
```
return self.to(*args, device="xpu", **kwargs)
```

chuanqi129:

Issue #2481, opened 2025-12-09
Two test_matmul_cuda.py cases failed in torch-xpu-ops CI but passed in local machine.


daisyden:
### 🐛 Describe the bug

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_matmul_cuda_xpu.TestMatmulCudaXPU,test_cublas_deterministic_shape_1024_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_matmul_cuda_xpu.TestMatmulCudaXPU,test_cublas_deterministic_shape_512_xpu_float32

pytest -v  test/xpu/test_matmul_cuda_xpu.py -k test_cublas_deterministic_shape_1024_xpu_float32
pytest -v  test/xpu/test_matmul_cuda_xpu.py -k test_cublas_deterministic_shape_512_xpu_float32

Traceback
```
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 3329, in wrapper
    method(*args, **kwargs)
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 428, in instantiated_test
    result = test(self, **param_kwargs)
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 1435, in only_fn
    return fn(slf, *args, **kwargs)
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 2009, in wrapper
    return fn(*args, **kwargs)
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/test_matmul_cuda_xpu.py", line 434, in test_cublas_deterministic
[...]

PR #3201, opened 2026-03-27
Stabilize XPU matmul determinism test coverage by separating CUDA default-path test from XPU deterministic-mode test.

BBBela:
Fix for: https://github.com/intel/torch-xpu-ops/issues/2481

## Summary

This change fixes determinism test coverage in `test_matmul_cuda_xpu.py` by excluding XPU from CUDA-specific test case.

The previous mixed-device test asserted strict bitwise equality on both CUDA and XPU without enabling deterministic mode. That assumption is valid for CUDA in this context, but not guaranteed for XPU default mode.

## Root Cause

XPU matmul uses oneDNN backend kernel that is by default non-deterministic according to tests and [DOCUMENTATION](https://www.intel.com/content/www/us/en/docs/onednn/developer-guide-reference/2024-1/primitive-attributes-deterministic.html). For some shapes and runner configurations, repeated matmul calls on identical inputs can produce small floating-point differences due to different reduction/order paths. The deterministic behavior of the matmul operator can be enforced, but that can damage the performance according to [DOCUMENTATION](https://www.intel.com/content/www/us/en/docs/onednn/developer-guide-reference/2024-1/primitive-attributes-deterministic.html), so the default setting is to keep the XPU matmul non-deterministic for performance purposes.

Because the old test used strict equality (`atol=0`, `rtol=0`) on both CUDA and XPU by default, CI could fail on XPU even when local runs passed.

Also, the failing test was a CUDA-specific test since the beginning. The XPU was added during porting tests from PyTorch to torch-xpu-ops, while execution of this test on XPU should not be added during porting.

## Proposed solution

The proposed solution is to remove the XPU from the so-far failing test case and mark it as `@onlyCUDA`.

### 1) Restrict existing strict default-path test to CUDA

- Changed `test_cublas_deterministic` decorator from:
  - `@onlyOn(["cuda", "xpu"])`
- To:
  - `@onlyCUDA`

This keeps existing strict bitwise default-path assertions for CUDA only.

### DIFF
--- a/.github/scripts/check-ut.py
+++ b/.github/scripts/check-ut.py
@@ -42,13 +42,20 @@ error_types = [
     "NotImplementedError",
 ]
 
+def _normalize_text(value):
+    if not value:
+        return ''
+    return ' '.join(str(value).split())
+
 def get_classname(case):
-    return ' '.join(case.classname.split()) if hasattr(case, 'classname') else case.get('classname', '')
+    if isinstance(case, dict):
+        return _normalize_text(case.get('classname', ''))
+    return _normalize_text(getattr(case, 'classname', ''))
 
 def get_name(case):
     if isinstance(case, dict):
-        return case.get('name', '')
-    return ' '.join(case.name.split())
+        return _normalize_text(case.get('name', ''))
+    return _normalize_text(getattr(case, 'name', ''))
 
 def get_category_from_case(case):
     if isinstance(case, dict):
@@ -105,7 +112,13 @@ def get_message(case):
                 indent_level = 0
                 break
 
-    return " ; ".join(error_messages) if error_messages else f"{case.result[0].message.splitlines()[0]}"
+    if error_messages:
+        return " ; ".join(error_messages)
+
+    message = getattr(case.result[0], 'message', '') or ''
+    if message:
+        return message.splitlines()[0]
+    return ""
 
 def print_md_row(row, print_header=False, failure_list=None):
     if print_header:
--- a/cmake/SYCLTLA.cmake
+++ b/cmake/SYCLTLA.cmake
@@ -26,7 +26,7 @@ if(NOT __SYCLTLA_INCLUDED)
   FetchContent_Declare(
       repo-sycl-tla
       GIT_REPOSITORY https://github.com/intel/sycl-tla.git
-      GIT_TAG        v0.6
+      GIT_TAG        v0.7
       GIT_SHALLOW    OFF
   )
   FetchContent_GetProperties(repo-sycl-tla)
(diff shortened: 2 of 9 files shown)


test/xpu/test_matmul_cuda_xpu.py
```diff
@@ -441,6 +442,19 @@ def test_cublas_deterministic(self, device, shape, dtype):
...
+    @onlyXPU
+    @parametrize("shape", [513, 767])
+    def test_matmul_deterministic_mode(self, device, shape):
+        original_deterministic = torch.are_deterministic_algorithms_enabled()
+        try:
+            torch.use_deterministic_algorithms(True)
+            input = torch.randn(shape, shape, device=device, dtype=torch.float32)
+            first = torch.matmul(input, input)
+            for _ in range(10):
+                self.assertEqual(first, torch.matmul(input, input), atol=0.0, rtol=0.0)
```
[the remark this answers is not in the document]
BBBela: Done

test/xpu/test_matmul_cuda_xpu.py
```diff
@@ -441,6 +442,19 @@ def test_cublas_deterministic(self, device, shape, dtype):
...
             self.assertEqual(first, torch.matmul(inp, inp), atol=0.0, rtol=0.0)
 
+    @onlyXPU
+    @parametrize("shape", [513, 767])
+    def test_matmul_deterministic_mode(self, device, shape):
+        original_deterministic = torch.are_deterministic_algorithms_enabled()
+        try:
+            torch.use_deterministic_algorithms(True)
+            input = torch.randn(shape, shape, device=device, dtype=torch.float32)
+            first = torch.matmul(input, input)
```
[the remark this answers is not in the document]
BBBela: Changed.

test/xpu/test_matmul_cuda_xpu.py
```diff
@@ -441,6 +443,15 @@ def test_cublas_deterministic(self, device, shape, dtype):
         for _ in range(10):
             self.assertEqual(first, torch.matmul(inp, inp), atol=0.0, rtol=0.0)
 
+    @onlyXPU
+    @parametrize("shape", [513, 767])
```
guangyey: This is a good idea, I think we should put this test case in PyTorch under `test/xpu/test_gemm.py`, what do you think.

test/xpu/test_matmul_cuda_xpu.py
```diff
@@ -431,7 +433,7 @@ def _expand_to_batch(t: torch.Tensor):
         # cross comparison
         self.assertEqual(out1_gpu, out2_gpu[0])
 
-    @onlyOn(["cuda", "xpu"])
```
guangyey, line 434: make sense.

test/xpu/test_matmul_cuda_xpu.py
BBBela: Sure, that seems a better place for that. I removed this `onlyXPU` test case from this PR and will prepare a PR in the `test/xpu/test_gemm.py` in PyTorch directly.

CuiYifeng: Agree with @guangyey's suggestion.

PR #3213, opened 2026-03-30
Fix compiler warnings to prepare for -Werror

chunhuanMeng:
Fix warnings that would fail under `-Werror`:

- `SpectralOps.cpp`: remove deprecated `CONJUGATE_EVEN_STORAGE`; fix `% 2` → `% 2 != 0`
- `BinaryShiftOpsKernels.cpp`: use `if constexpr (std::is_signed_v<scalar_t>)` to avoid `-Wtype-limits` on unsigned types
- `LinearInt4.cpp`: fix narrowing cast `int8_t` → `scalar_t`
- `SortingRadixSort.h`: fix signed overflow UB `1ll` → `uint64_t{1}`; handle `bool` key type explicitly
- `Macros.h`: add `#undef RESTRICT` guard before `#define`
- Remove unused `#include <comm/SYCLContext.h>` from 4 files

PR #3212, opened 2026-03-30
Suppress warnings from SYCL TLA flash attention headers

chunhuanMeng:
Two changes to fix compiler warnings in mha_fwd.h and mha_bwd.h:

1. **Header warnings**: Add clang/GCC diagnostic push/pop around SYCL TLA includes to suppress warnings (`-Wchanges-meaning`, `-Wsycl-strict`, `-Wunused-local-typedefs`, `-Wunused-variable`) that originate from the third-party SYCL TLA library, not our own code.

2. **Struct field reorder**: Move `pb_ptr` after `dv_ptr` in `Param` to match the designated initializer order and silence the out-of-order initialization warning.

The underlying issues are tracked in the SYCL TLA project and suppressions will be removed once upstream fixes land.

PR #3211, opened 2026-03-30
Suppress -Wdeprecated-declarations from oneAPI SYCL headers

chunhuanMeng:
This pull request introduces a consistent approach to suppressing deprecation warnings originating from oneAPI SYCL headers across the codebase. These warnings, which are not caused by our own code, would otherwise cause build failures when compiling with `-Werror`. The solution uses GCC diagnostic pragmas and defines `SYCL_DISABLE_FSYCL_SYCLHPP_WARNING` before and after including SYCL-related headers.

Two suppression mechanisms are applied around the affected includes:

* pragma GCC diagnostic ignored `-Wdeprecated-declarations` — suppresses the GCC warning at the compiler level.
* define `SYCL_DISABLE_FSYCL_SYCLHPP_WARNING` — the suppression macro officially provided by Intel SYCL headers for exactly this scenario (host-only compilation without -fsycl).

No functional or behavioral changes to the codebase were made—these modifications are solely to improve the build process and developer experience. Some versions of the DPC++ compiler expose internal oneAPI SYCL headers through non-system include paths, causing the host GCC compiler to emit deprecated-declaration warnings from third-party SYCL code. 
This suppression is scoped tightly around the offending includes and does not affect project code.

PR #3210, opened 2026-03-30
Suppress MSVC C4715 warning in SYCL kernel files

chunhuanMeng:
This pull request refactors the way return type warnings are suppressed in various SYCL kernel source files. Instead of using compiler-specific pragma directives directly in each file, the code now includes a common macro from `comm/Macros.h` to handle disabling and enabling return type warnings in a standardized way. This change improves code maintainability and consistency across the codebase.

**Refactoring and code consistency:**
* MSVC warning suppression additions
* Replaced manual `#pragma` directives for disabling return type warnings with the standardized macros `DISABLE_RETURN_TYPE_WARNING_BEGIN` and `DISABLE_RETURN_TYPE_WARNING_END` from `comm/Macros.h` in all affected SYCL kernel source files

[...]

PR #3214, opened 2026-03-30
Enable -Werror/-WX to treat warnings as errors

chunhuanMeng:
### 1. `mha_bwd_launch.h` — 3 fixes
- **`-Wdeprecated-declarations`**: `sg.barrier()` is deprecated in newer SYCL specs. Replaced with `sycl::group_barrier(sg)`.
- **`-Wdangling-pointer`**: The return-value version of `convert_type<DType>(rdQaccum)` creates a local `frag` variable and returns a tensor holding `&frag`, which dangles after the function returns. Removed this function and switched the `Is_even_M == true` branch to use the existing safe out-parameter version (`make_tensor_like` + `convert_type(src, dst)`), which is already used by the `Is_even_M == false` branch.
- **`-Warray-bounds`** (suppressed via `#pragma GCC diagnostic`): False positive from external `cutlass/array_subbyte.h:133` where `reinterpret_cast<Storage const&>(x)` reads a 4-byte `unsigned int` through a 2-byte `half_t` reference. This is intentional subbyte packing in the sycl-tla/cutlass library and cannot be fixed on our side.

### 2. `mha_common.h` — `static` → `inline`
- **`-Wunused-function`**: `is_64_bytes_aligned()` and `ensure_alignment_for_sdpa()` are `static` functions defined in a header, causing unused-function warnings in translation units that don't call them. Changed to `inline` to match standard header practice.

### 3. `Reduce.h` — `[=]` → `[=, this]`
- **`-Wdeprecated`**: Implicit capture of `this` via `[=]` is deprecated in C++20. Two lambdas in `ReduceOp` (lines 570, 996) capture `ops.combine()` which accesses `this`. Fixed by explicitly capturing `this`.

### 4. `IGammaKernel.cpp` — cross-compiler `optnone`
- **`-Wattributes`**: `[[clang::optnone]]` is a Clang-specific attribute ignored by GCC. Replaced with `#if defined(__clang__)` / `#elif defined(__GNUC__)` guards to use `[[clang::optnone]]` on Clang and `__attribute__((optimize("O0")))` on GCC, so optimization is actually disabled on both compilers.

### 5. `Attention.cpp` — suppress external SYCL header warnings
- **`-Wdeprecated-declarations`** and **`-Werror` from `sycl.hpp`**: This is a host-only `.cpp` file compiled by GCC (not the SYCL compiler). SYCL headers are pulled in transitively via `c10/xpu/XPUDeviceProp.h` using `-I` (not `-isystem`), exposing deprecated API warnings from the Intel oneAPI 2025.3 runtime headers. Suppressed with `#pragma GCC diagnostic ignored "-Wdeprecated-declarations"` around includes, and defined `SYCL_DISABLE_FSYCL_SYCLHPP_WARNING` as recommended by the SYCL header itself.

### 6.Add -Werror (GCC) SYCL host compilation flags so that all compiler warnings fail the build. This is the final step after:
  - stage1: fixing genuine warning-inducing code issues(https://github.com/intel/torch-xpu-ops/pull/3213)
  - stage2: suppressing MSVC C4715 in SYCL kernel files(https://github.com/intel/torch-xpu-ops/pull/3210)
  - stage3: suppressing -Wdeprecated-declarations from oneAPI SYCL headers(https://github.com/intel/torch-xpu-ops/pull/3211)
  - stage4: suppressing warnings from SYCL TLA flash attention headers(https://github.com/intel/torch-xpu-ops/pull/3212)


CuiYifeng (approved):
LGTM.

LuFinch (approved):
LGTM

chuanqi129:
✅ PR has been successfully merged by @chunhuanMeng.

Issue #2109, opened 2025-09-25
"worker 'gw0' crashed while running sparce_coo cases on PVC in CI 


daisyden:
### 🐛 Describe the bug

When the two cases removed from skiplist it will crash in torch-xpu-ops CI.
```
2025-09-24T11:35:09.5568323Z [gw0] linux -- Python 3.10.18 /tmp/xpu-tool/Python/3.10.18/x64/bin/python
2025-09-24T11:35:09.5568844Z worker 'gw0' crashed while running 'third_party/torch-xpu-ops/test/xpu/test_masked_xpu.py::TestMaskedXPU::test_mask_layout_sparse_coo_masked_amax_xpu_float64'
2025-09-24T11:35:09.5569356Z ____________ third_party/torch-xpu-ops/test/xpu/test_masked_xpu.py _____________
2025-09-24T11:35:09.5569678Z [gw5] linux -- Python 3.10.18 /tmp/xpu-tool/Python/3.10.18/x64/bin/python
2025-09-24T11:35:09.5570162Z worker 'gw5' crashed while running 'third_party/torch-xpu-ops/test/xpu/test_masked_xpu.py::TestMaskedXPU::test_mask_layout_sparse_coo_masked_amin_xpu_float64'
```

Reproduce command:
```
cd pytorch; pip install -r .ci/docker/requirements-ci.txt; cd third_party/torch-xpu-ops/xpu; PYTORCH_TEST_WITH_SLOW=1 pytest -v test_masked_xpu.py::TestMaskedXPU -k test_mask_layout_sparse_coo_masked_amax_xpu_float64
```
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_masked_xpu.TestMaskedXPU,test_mask_layout_sparse_coo_masked_amax_xpu_float64
op_ut,third_party.torch-xpu-ops.test.xpu.test_masked_xpu.TestMaskedXPU,test_mask_layout_sparse_coo_masked_amin_xpu_float64
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_masked_xpu.TestMaskedXPU,test_mask_layout_sparse_coo_masked_sum_xpu_float64~~

### Versions

PyTorch main
torch-xpu-ops: https://github.com/intel/torch-xpu-ops/pull/1825

PR #3223, opened 2026-03-30
Fix OOB read in PSTL merge for incomplete merge-sort pairs.

BBBela:
Fix for: #2109 

## Summary
This PR fixes an out-of-bounds read in XPU PSTL merge-sort that was causing segmentation fault in the test cases from issue #2109 

## Root Cause
In `third_party/torch-xpu-ops/src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h`, `merge()` could be called with an empty second sequence:
- `sq2_start == sq2_end == sort_sz`

Before this change, the code still evaluated:

```cpp
if (!comp_t(in_key[sq2_start], in_key[sq1_end - 1])) {
```

That dereferenced `in_key[sq2_start]` even when `sq2_start` was one past the valid range.
`in_key` is allocated for exactly `sort_sz` elements, so valid indices are `[0, sort_sz - 1]`; accessing `in_key[sort_sz]` is therefore a one-past-end read.

## What Changed
Added an early return in `merge()` to handle the empty-second-sequence case safely:

- If `sq2_start >= sq2_end`, copy only the first sequence chunk to output and return.
- This prevents evaluating `in_key[sq2_start]` when `sq2` is empty.

## Why This Is Correct
When one side of merge is empty, the correct merge result is the non-empty side unchanged.
So copying the first chunk and returning preserves merge semantics while removing invalid memory access.

## Behavior Before vs After
Before:
- Incomplete trailing pair could trigger `in_key[sq2_start]` with `sq2_start == sort_sz`.
- OOB read propagated invalid values and crash on XPU.

After:
- Empty `sq2` is detected up front.
- No OOB read.
- Trailing incomplete pair is handled as a simple copy of `sq1`.

### DIFF
--- a/cmake/SYCLTLA.cmake
+++ b/cmake/SYCLTLA.cmake
@@ -26,7 +26,7 @@ if(NOT __SYCLTLA_INCLUDED)
   FetchContent_Declare(
       repo-sycl-tla
       GIT_REPOSITORY https://github.com/intel/sycl-tla.git
-      GIT_TAG        v0.6
+      GIT_TAG        v0.7
       GIT_SHALLOW    OFF
   )
   FetchContent_GetProperties(repo-sycl-tla)
--- a/test/xpu/test_torch_xpu.py
+++ b/test/xpu/test_torch_xpu.py
@@ -6678,12 +6678,6 @@ else:
                     if torch.cuda.amp.common.amp_definitely_not_available()
                     else a.is_enabled()
                 )
-            elif device.type == "xpu":
-                self.assertTrue(
-                    not a.is_enabled()
-                    if torch.xpu.amp.common.amp_definitely_not_available()
-                    else a.is_enabled()
-                )
             else:
                 self.assertTrue(a.is_enabled())
             if lazy_init_scale:


src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h
```diff
@@ -1300,6 +1300,16 @@ inline void merge(
...
+  // Copy only the first sequence if the second one is empty.
+  if (sq2_start >= sq2_end) {
+    for (unsigned int i = 0; i < chunk1_size; ++i) {
+      out_key[chunk1_start + i] = in_key[chunk1_start + i];
+      out_val[chunk1_start + i] = in_val[chunk1_start + i];
+    }
+    return;
+  }
+
   if (!comp_t(in_key[sq2_start], in_key[sq1_end - 1])) {
```
BBBela, line 1313: **Additional info:**
On this line, the call
```cpp
in_key[sq2_start]
```
was causing out of bound memory access without checking if sequence 2 is not empty. The code was assuming explicitly that the second sequence has at least one element which is not always true.

Issue #3157, opened 2026-03-24
XPU out of memory. Tried to allocate 32.00 GiB. / 16.00GiB


kdrozd-dev:
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_tensor_creation_ops_xpu.TestLikeTensorCreationXPU,test_ones_large_xpu_complex128
op_ut,third_party.torch-xpu-ops.test.xpu.test_tensor_creation_ops_xpu.TestLikeTensorCreationXPU,test_ones_large_xpu_complex64
op_ut,third_party.torch-xpu-ops.test.xpu.test_tensor_creation_ops_xpu.TestLikeTensorCreationXPU,test_ones_large_xpu_float64
op_ut,third_party.torch-xpu-ops.test.xpu.test_tensor_creation_ops_xpu.TestLikeTensorCreationXPU,test_ones_large_xpu_int64
op_ut,third_party.torch-xpu-ops.test.xpu.test_tensor_creation_ops_xpu.TestLikeTensorCreationXPU,test_zeros_large_xpu_complex64
op_ut,third_party.torch-xpu-ops.test.xpu.test_tensor_creation_ops_xpu.TestLikeTensorCreationXPU,test_zeros_large_xpu_float64
op_ut,third_party.torch-xpu-ops.test.xpu.test_tensor_creation_ops_xpu.TestLikeTensorCreationXPU,test_zeros_large_xpu_int64
Error:
```
torch.OutOfMemoryError: XPU out of memory. Tried to allocate 32.00 GiB. GPU 0 has a total capacity of 11.59 GiB of which 9.01 GiB is free. Of the allocated memory 0 bytes is allocated by PyTorch, and 0 bytes is reserved by PyTorch but unallocated. Please use empty_cache to release all unoccupied cached memory. To execute this test, run the following from the base repo dir:     PYTORCH_TEST_WITH_SLOW=1 python test\xpu\test_tensor_creation_ops_xpu.py TestLikeTensorCreationXPU.test_ones_large_xpu_complex128 This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0
```

PR #3228, opened 2026-03-31
Add large tensor test decorator

kdrozd-dev:
Fixes https://github.com/intel/torch-xpu-ops/issues/3157

This pull request adds `@largeTensorTest` for the `test_zeros_large` and `test_ones_large` tests in `test/xpu/test_tensor_creation_ops_xpu.py` this ensures that they will only be run on machines with memory that matches the requirements. It is also parametrized by the dtype like in `test_nn.py` in upstream so that smaller dtype tests can be run on more devices.


Silv3S (approved):

chuanqi129:
✅ PR has been successfully merged by @kdrozd-dev.

Issue #2669, opened 2025-12-30
[upstream_ut]  AssertionError: Tensor-likes are not close! in functorch/test_vmap.py


daisyden:
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestVmapOperatorsOpInfoXPU,test_vmap_exhaustive_addmv_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_False_randomness_different_batched_input_first_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_False_randomness_different_batched_input_last_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_False_randomness_different_batched_input_none_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_False_randomness_same_batched_input_none_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_True_randomness_different_batched_input_first_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_True_randomness_different_batched_input_last_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_True_randomness_different_batched_input_none_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_True_randomness_same_batched_input_none_xpu
op_ut,test.functorch.test_vmap.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_True_randomness_different_batched_input_first_xpu
op_ut,test.functorch.test_vmap.TestRandomnessXPU,test_random_unary_out_of_place_use_generator_True_randomness_different_batched_input_none_xpu
[...]

PR #3239, opened 2026-03-31
Initialize Box-Muller state in rand_init function

tszulist-hbn:
Resolves https://github.com/intel/torch-xpu-ops/issues/2669

### Summary
Reset the Box-Muller transform state variables during Philox RNG initialization to prevent stale cached values from producing incorrect results.

### Root Cause
The `rand_init` function in `Philox4x32.h` initializes the Philox counter and key but does not reset the Box-Muller transform cache (`boxmuller_flag`, `boxmuller_extra`, and their double-precision counterparts). The Box-Muller transform produces normal variates in pairs, caching one for the next call. When the RNG state is reinitialized with a new seed, subsequence, or offset — as happens inside `vmap` with `randomness="different"` — the stale cached value from a previous configuration can be returned instead of a freshly generated one.

This caused a ~0.1% element mismatch (`1 / 1764`) in `TestRandomnessXPU` tests in `test/functorch/test_vmap.py`, where the greatest absolute difference reached `1.195` (tolerance: `1e-05`).

### Fix
Clear the four Box-Muller state fields at the end of `rand_init`. This ensures that after every `(re)initialization`, the first normal-variate request performs a full Box-Muller computation from the new Philox stream rather than returning a leftover value.

### DIFF
--- a/src/ATen/native/xpu/sycl/Philox4x32.h
+++ b/src/ATen/native/xpu/sycl/Philox4x32.h
@@ -261,6 +261,14 @@ static inline void rand_init(
   state->index = 0;
   skipahead_sequence(subsequence, state);
   skipahead(offset, state);
+  // Reset Box-Muller transform state. These fields are not part of the core
+  // Philox counter/key, but cache an extra normal variate generated from a
+  // previous Box-Muller call. When (re)initializing the Philox state, this
+  // cache must be cleared to avoid reusing a value derived from an old
+  // seed/subsequence/offset configuration.
+  state->boxmuller_flag = 0;
+  state->boxmuller_extra = 0.0f;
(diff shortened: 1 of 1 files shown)


astachowiczhabana:
Pre-existing issue spotted while reviewing:

`rand_normal_double` (line 443) stores `v.y` (a `double`) into `state->boxmuller_extra` (a `float`), silently truncating precision. The `PhiloxState` struct needs a `double boxmuller_extra_double` field, and `rand_normal_double` should use it instead. If fixed here, `rand_init` should also reset the new field.

chuanqi129:
✅ PR has been successfully merged by @tszulist-hbn.

Issue #2918, opened 2026-02-24
[XPU][upstream_ut][COW] Skip non-supported ops (jiterator + histogramdd)


gplutop7:
### 🐛 Describe the bug

Split from: https://github.com/intel/torch-xpu-ops/issues/2248

The following TestCompositeComplianceXPU failures are NOT related to COW
materialization problems. They fail because required operators are not
implemented on XPU backend.

These tests fail before reaching any COW validation logic.

# 1. Jiterator tests (not supported on XPU)

XPU backend does not support jiterator.

The following tests fail due to missing jiterator support:

- TestCompositeComplianceXPU.test_cow_input_jiterator_unary_xpu_float32
- TestCompositeComplianceXPU.test_cow_input_jiterator_binary_xpu_float32
- TestCompositeComplianceXPU.test_cow_input_jiterator_binary_return_by_ref_xpu_float32
- TestCompositeComplianceXPU.test_cow_input_jiterator_2inputs_2outputs_xpu_float32
- TestCompositeComplianceXPU.test_cow_input_jiterator_4inputs_with_extra_args_xpu_float32

Failure reason:
Jiterator-based ops are not implemented for XPU.
Tests fail before entering COW compliance checks.

# 2. histogramdd (operator not implemented)

Missing operator:

- aten::_histogramdd_bin_edges

Affected test:

- TestCompositeComplianceXPU.test_cow_input_histogramdd_xpu_float32

Failure reason:
aten::_histogramdd_bin_edges is not implemented on XPU.
The failure happens before any COW-related logic.

## Proposal

Skip the above tests on XPU until:

- jiterator support is implemented
- aten::_histogramdd_bin_edges is implemented

PR #3240, opened 2026-03-31
Keep jiterator CUDA-only in XPU test patching

gplutop7:
Fixes: 
https://github.com/intel/torch-xpu-ops/issues/2918

This PR fixes incorrect XPU execution of jiterator COW composite compliance tests.

What was done:

- Updated XPU test adapter logic to preserve CUDA-only jiterator decorators/wrappers instead of translating them to XPU.
- Added explicit affected test names to the extended XPU skip list as an extra CI guard.

Why:

- Jiterator is CUDA/ROCm-only.
- XPU adapter translation was making these tests run on XPU, causing failures instead of skips.
- The skip list alone is not sufficient: direct test runs (e.g. python test_ops_xpu.py TestName) bypass the skip list entirely, so the adapter-level fix is required as the primary guard.


chuanqi129:
✅ PR has been successfully merged by @gplutop7.

PR #3241, opened 2026-04-01
[windowsCI] Remove hardcoded MKL pin from CI requirements to avoid conflicts with Torch XPU wheel

jafraustro:
Remove the hardcoded MKL installation line from the Windows CI workflow so that the CI does not force-install MKL versions that can conflict with the locally-built Torch XPU wheel.

The change updates the Windows workflow file to remove lines mentioning mkl from the requirements before installing them.

Changed file:
`torch-xpu-ops/.github/workflows/_windows_ut.yml`

**Related PR:**
#3118 
#3245


CuiYifeng:
@chuanqi129 Any comments about Windows failure of this PR? Low risk?

PR #3245, opened 2026-04-01
Removed the BUILD_SEPARATE_OPS build option

chunhuanMeng:
This pull request simplifies the `src/BuildOnWindows.cmake` script by removing the conditional logic and macro related to building separate operator libraries on Windows. The changes consolidate the build process, making it more straightforward.

Build system simplification:

* Removed the `setup_common_libraries` macro and all logic related to the `BUILD_SEPARATE_OPS` option, including the creation and installation of separate SYCL operator libraries.
* Flattened the build flow by eliminating the `if(BUILD_SEPARATE_OPS)`/`else()` block, resulting in a single unified approach for building operator libraries on Windows. [[1]](diffhunk://#diff-7c496d77cdbeccfbcfc49fafab387712dbd0b2690543ba540bec243dee7a44b4L14-L43) [[2]](diffhunk://#diff-7c496d77cdbeccfbcfc49fafab387712dbd0b2690543ba540bec243dee7a44b4L125-R95)


guangyey (approved):
Sounds good! We should simplify this on Windows because nobody uses it, and it is never tested in CI.

draghan:
Hello there. What would be the reason behind removing this option, besides "nobody uses it"?
It will be a regression in functionality and step back in terms of parity with Linux. We will be doing more work on Windows, this feature will save the build time when doing dev/debug.

PR #3249, opened 2026-04-01
test: align XPU asarray alias test with requires_grad propagation

BartoszKokoszko:
torch.asarray (commit a97dcf9c584) started propagating requires_grad when copy=False. Update TestAsArrayXPU to inherit requires_grad from tensor inputs instead of always expecting False, fixing test_alias_from_tensor_xpu_* on XPU.


chuanqi129:

PR #3250, opened 2026-04-01
Adjust XPU unittest to use proper XPU module

Silv3S:
Fix for https://github.com/intel/torch-xpu-ops/issues/2920

Port forward_traceback_preserves_exception_with_checkpoint to use `torch.xpu` instead of `torch.cuda` from upstream. 

Without test in `test_autograd_xpu.py` we would try to execute stock upstream test that has hardcoded `cuda`


Silv3S:
CI logs: [linux-ut (op_ut)](https://github.com/intel/torch-xpu-ops/actions/runs/23904105874/job/69758414400?pr=3250#step:4:1839)

📊 New Passing Known Issues:
op_ut,third_party.torch-xpu-ops.test.xpu.test_autograd_xpu.TestAutograd,test_forward_traceback_preserves_exception_with_checkpoint

Issue #3266, opened 2026-04-06
[RFC] Migrate XPU kernel math functions from std::/:: to sycl::/sycl::native:: namespace


jianyizh:
## Summary

XPU SYCL kernels in `src/ATen/native/xpu/sycl/` currently use `std::`, `::`, or bare C math functions (e.g. `std::exp`, `::expf`, `sqrtf`) instead of the SYCL-native `sycl::` or `sycl::native::` namespace equivalents. This should be unified to use `sycl::` namespace functions for correctness, portability, and potential performance benefits on Intel GPU hardware.

### Mapping convention (mirroring CUDA → SYCL)

| Current (std/bare) | SYCL replacement | Notes |
|---|---|---|
| `std::exp`, `::expf`, `expf` | `sycl::exp` | |
| `std::log`, `::logf`, `logf` | `sycl::log` | |
| `std::sqrt`, `sqrtf` | `sycl::sqrt` | |
| `std::abs`, `std::fabs`, `::abs` | `sycl::fabs` (float/double), `sycl::abs` (int) | complex → keep `std::abs` |
| `std::pow` | `sycl::pow` | |
| `std::ceil`, `ceilf` | `sycl::ceil` | |
| `std::floor`, `floorf` | `sycl::floor` | |
| `std::tanh`, `::tanhf` | `sycl::tanh` | complex → keep `std::` |
| `std::fmod` | `sycl::fmod` | |
| `std::lgamma` | `sycl::lgamma` | |
| `std::log10` | `sycl::log10` | complex → keep `std::` |
| `std::log1p`, `::log1p` | `sycl::log1p` | complex → keep `std::` |
| `std::log2` | `sycl::log2` | complex → keep `std::` |
| `std::isinf` | `sycl::isinf` | device code only |
| `std::isnan` | `sycl::isnan` | device code only |
| `std::isfinite` | `sycl::isfinite` | device code only |
| `std::copysign` | `sycl::copysign` | |
| `std::erf` | `sycl::erf` | |
| `std::erfc` | `sycl::erfc` | |
| `std::exp2` | `sycl::exp2` | |
| `std::expm1` | `sycl::expm1` | complex → keep `std::` |
| `std::frexp` | `sycl::frexp` | |
| `std::sin` | `sycl::sin` | complex → keep `std::` |
| `std::cos` | `sycl::cos` | complex → keep `std::` |
| `std::tan` | `sycl::tan` | complex → keep `std::` |
[...]

PR #3268, opened 2026-04-06
Migrate trigonometric and related math functions from std:: to sycl:: namespace

jianyizh:
> **Note**: This PR was authored with the assistance of AI (Claude).

## Summary

Migrate 17 kernel files to use `sycl::` math functions instead of `std::` for trigonometric and related operations. This is part of the broader effort to align XPU kernels with SYCL-native math builtins (#3266).

### Functions migrated
- **Unary trig**: sin, cos, tan, asin, acos, atan, sinh, cosh, tanh, asinh, acosh, atanh
- **Binary geometric**: atan2, hypot
- **Other**: sinc, polar, mish (tanh inside), RNN tanh

### Key implementation details

1. **`opmath_type` cast to avoid Half/BFloat16 ambiguity**: `sycl::func(static_cast<opmath_t>(a))` where `opmath_t = at::opmath_type<scalar_t>`. This resolves ambiguity when `c10::Half`/`c10::BFloat16` can implicitly convert to both `float` and `sycl::half`.

2. **`if constexpr` for complex types**: SYCL math functions do not support `c10::complex<T>`. For functors dispatched with both real and complex types, we use:
   ```cpp
   if constexpr (c10::is_complex<scalar_t>::value) {
     return std::func(a);  // complex path unchanged
   } else {
     using opmath_t = at::opmath_type<scalar_t>;
     return sycl::func(static_cast<opmath_t>(a));
   }
   ```

3. **Files with separate ComplexFunctor**: Where complex dispatch uses a separate functor (e.g., AsinComplexFunctor), the plain functor is real-only and is directly migrated to `sycl::` without `if constexpr`.

### Not changed (intentionally)
- All `ComplexFunctor` variants (AsinComplex, AtanComplex, etc.) -- remain `std::`
- `ForeachUnaryKernels.cpp` STD_FUNCTOR macro -- separate PR scope
- Files where arguments are already safe types (mish uses `opmath_t`, RNN uses `H2F()` macro, sinc uses `opmath_t product`) -- only changed `std::` to `sycl::` prefix

### Testing
- **Accuracy**: All 17 ops pass across float16, bfloat16, float32, float64, complex64, complex128
- **Performance** (N=1M, PVC, before to after):
  - No regressions (>5% threshold)
  - Overall: +0.6% (within measurement noise, ~26us per op dominated by launch overhead)
  - One improvement: cosh bfloat16 -28% (baseline had a one-off spike)

```
        Op       dtype      before       after        diff      change
       sin     float16      26.7 us      27.0 us      +0.3 us     +1.1%
       sin     float32      26.4 us      26.7 us      +0.3 us     +1.1%
       cos     float16      26.1 us      26.6 us      +0.5 us     +1.9%
       cos     float32      26.1 us      26.6 us      +0.5 us     +1.9%
      tanh     float32      27.1 us      26.7 us      -0.4 us     -1.5%
      tanh     float64      31.0 us      31.0 us      +0.0 us     +0.0%
      cosh    bfloat16      37.1 us      26.7 us     -10.4 us    -28.0%
      mish     float64      39.6 us      39.6 us      +0.0 us     +0.0%
     polar     float32      27.2 us      27.7 us      +0.5 us     +1.8%
```
(Selected rows shown; all ops within +/-2.3% except cosh bf16 improvement)

Part of #3266

### DIFF
--- a/src/ATen/native/xpu/sycl/UnaryGeometricAcoshKernel.cpp
+++ b/src/ATen/native/xpu/sycl/UnaryGeometricAcoshKernel.cpp
@@ -9,6 +9,7 @@
  */
 
 #include <ATen/Dispatch.h>
+#include <ATen/OpMathType.h>
 #include <ATen/native/TensorIterator.h>
 #include <ATen/native/xpu/sycl/Loops.h>
 
@@ -16,10 +17,14 @@
 
 namespace at::native::xpu {
 
-template <typename scalar_t, typename acc_t = scalar_t>
+template <typename scalar_t, typename acc_t = at::opmath_type<scalar_t>>
 struct AcoshFunctor {
   scalar_t operator()(scalar_t a) const {
-    return std::acosh(static_cast<acc_t>(a));
+    if constexpr (c10::is_complex<acc_t>::value) {
+      return std::acosh(static_cast<acc_t>(a));
+    } else {
+      return sycl::acosh(static_cast<acc_t>(a));
+    }
   }
 };
 
--- a/src/ATen/native/xpu/sycl/ActivationMishKernels.cpp
+++ b/src/ATen/native/xpu/sycl/ActivationMishKernels.cpp
@@ -27,7 +27,7 @@ struct MishFunctor {
   scalar_t operator()(scalar_t x) const {
     using opmath_t = at::opmath_type<scalar_t>;
     const opmath_t x_acc = static_cast<opmath_t>(x);
-    return x_acc * std::tanh(std::log1p(std::exp(x_acc)));
+    return x_acc * sycl::tanh(std::log1p(std::exp(x_acc)));
   }
 };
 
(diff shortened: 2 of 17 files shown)


src/ATen/native/xpu/sycl/UnaryGeometricAcoshKernel.cpp
```diff
@@ -9,17 +9,22 @@
...
 
 namespace at::native::xpu {
 
-template <typename scalar_t, typename acc_t = scalar_t>
+template <typename scalar_t, typename acc_t = at::opmath_type<scalar_t>>
 struct AcoshFunctor {
   scalar_t operator()(scalar_t a) const {
-    return std::acosh(static_cast<acc_t>(a));
+    if constexpr (c10::is_complex<acc_t>::value) {
+      return std::acosh(static_cast<acc_t>(a));
```
CuiYifeng, line 24: Does sycl math function not support Complex dtype?

CuiYifeng (commented):
There is the same question as #3269. The main part of this PR looks good to me.

src/ATen/native/xpu/sycl/UnaryGeometricAcoshKernel.cpp
CuiYifeng: Got the answer from https://github.com/intel/torch-xpu-ops/pull/3269#discussion_r3144746936.

CuiYifeng (approved):

LuFinch (approved):

chuanqi129:

Issue #3177, opened 2026-03-25
Accuracy gap of BF16/FP16 test_block_addmm


CuiYifeng:
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_2_int32_noncontiguous_False_xpu_bfloat16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_2_int32_noncontiguous_False_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_2_int32_noncontiguous_True_xpu_bfloat16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_2_int32_noncontiguous_True_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_2_int64_noncontiguous_False_xpu_bfloat16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_2_int64_noncontiguous_False_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_2_int64_noncontiguous_True_xpu_bfloat16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_2_int64_noncontiguous_True_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_3_int32_noncontiguous_False_xpu_bfloat16
op_ut,third_party.torch-xpu-ops.test.xpu.test_sparse_csr_xpu.TestSparseCSRXPU,test_block_addmm_block_size_3_int32_noncontiguous_False_xpu_float16
[...]

PR #3273, opened 2026-04-06
Fix bf16/fp16 accuracy issue in sparsecsr addmm

jenniew:
Fix bf16/fp16 accuracy issue in sparsecsr addmm.
Related issue: https://github.com/intel/torch-xpu-ops/issues/3177


src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
```diff
@@ -79,11 +79,10 @@ Tensor addmm_calculation(
     const Scalar& alpha) {
   Tensor mat1_dense = mat1.layout() != kStrided ? mat1.to_dense() : mat1;
   Tensor mat2_dense = mat2.layout() != kStrided ? mat2.to_dense() : mat2;
-
   Tensor result_dense = mat1_dense.mm(mat2_dense) * alpha;
   if (beta.toComplexDouble() != 0.) {
     Tensor input_dense = input.layout() != kStrided ? input.to_dense() : input;
-    result_dense.add_(input_dense, beta);
+    result_dense.add_(input_dense * beta);
```
CuiYifeng, line 85: Could you explain the difference between these two calling methods?

CuiYifeng: The difference has been explained in #3476 by @jafraustro.

CuiYifeng (approved):

src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
etaf: LGTM, I'll be better if we can add some comments here.

etaf (approved):

src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
CuiYifeng: Sure, we can add comments in #3476 to avoid repeat CI tests.

PR #3287, opened 2026-04-08
Enable dpclang sycl compiler

dvrogozh:
`dpclang` compiler is an open source version of the closed source Intel DPC++ compiler. `dpclang` is being supported by Intel in the https://github.com/intel/llvm public repository and got recently packaged into upcoming Ubuntu 26.04 LTS.

This commit enables minimal support of `dpclang` compiler in PyTorch building towards fully open source PyTorch stack on Linux. Limitations:

* Kineto, XCCL and oneMKL paths were not enabled and might require additional work
* SYCL-TLA, oneDNN and Pytorch itself might require additional changes to properly handle dpclang compiler as it does not support `__INTEL_LLVM_COMPILER` macro

We will address above limitations in the follow up patches. At the moment the following procedure can be used to build PyTorch for XPU with `dpclang` compiler on Ubuntu 26.04:

* Do NOT activate oneAPI environment
* Install open source DPC++ compiler, OpenCL and few other tools with:
```
apt-get install dpclang-6 ocl-icd-opencl-dev pkg-config libze-dev
```
* Build PyTorch as follows:

```
XPU_SYCL_COMPILER=dpclang USE_KINETO=0 USE_ONEMKL_XPU=0 \
  python3 setup.py develop
```

The dependencies from https://github.com/pytorch/pytorch/pull/179763 and https://github.com/pytorch/pytorch/pull/185571 are weak in a sense that the change is transparent to the current production build of the PyTorch for XPU using oneAPI DL Essentials. The change in torch-xpu-ops is only required for the build with the dpclang.

Needs: https://github.com/pytorch/pytorch/pull/179763
Needs: https://github.com/pytorch/pytorch/pull/185571

CC: @guangyey @EikanWang @chuanqi129 @frenchwr

### DIFF
--- a/cmake/BuildFlags.cmake
+++ b/cmake/BuildFlags.cmake
@@ -111,8 +111,16 @@ macro(set_build_flags)
     elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
       set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -Wno-absolute-value)
       set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -fno-fast-math)
-      set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -fma)
-      set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -no-ftz)
+      # -fma which we used before is an alias used for -ffp-contract=fast for compatibility reasons
+      # with very old version of the ICX compiler. The -ffp-contract=fast is supported by both closed
+      # source and open source DPC++ compiler versions.
+      set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -ffp-contract=fast)
+      # -no-ftz is supported only by ICX compiler shipped with oneAPI Toolkits. For the
+      # DPCLANG open source compiler that's the default mode and no option is needed.
+      CHECK_SYCL_FLAG("-no-ftz" SUPPORTS_NO_FTZ)
+      if(SUPPORTS_NO_FTZ)
+        set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -no-ftz)
+      endif()
     endif()
 
     if(CMAKE_BUILD_TYPE MATCHES Debug)
--- a/cmake/Modules/FindSYCLToolkit.cmake
+++ b/cmake/Modules/FindSYCLToolkit.cmake
@@ -178,34 +178,5 @@ if(${has_werror} EQUAL -1)
 
 endif()
 
-# Create a clean working directory.
-set(SYCL_CMPLR_TEST_DIR "${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/TESTSYCLCMPLR")
-file(REMOVE_RECURSE ${SYCL_CMPLR_TEST_DIR})
-file(MAKE_DIRECTORY ${SYCL_CMPLR_TEST_DIR})
-# Create the test source file
-set(TEST_SRC_FILE "${SYCL_CMPLR_TEST_DIR}/llvm_features.cpp")
-set(TEST_EXE "${TEST_SRC_FILE}.exe")
-SYCL_CMPLR_TEST_WRITE(${TEST_SRC_FILE} "__INTEL_LLVM_COMPILER")
-# Build the test and create test executable
-SYCL_CMPLR_TEST_BUILD(error ${TEST_SRC_FILE} ${TEST_EXE})
-if(error)
-  message(FATAL_ERROR "Can not build SYCL_CMPLR_TEST")
-endif()
-# Execute the test to extract information
-SYCL_CMPLR_TEST_RUN(error ${TEST_EXE})
-if(error)
-  message(FATAL_ERROR "Can not run SYCL_CMPLR_TEST")
-endif()
-# Extract test output for information
-SYCL_CMPLR_TEST_EXTRACT(${test_output} "__INTEL_LLVM_COMPILER")
-
-# Check whether the value of __INTEL_LLVM_COMPILER macro was successfully extracted
-string(COMPARE EQUAL "${__INTEL_LLVM_COMPILER}" "" nosycllang)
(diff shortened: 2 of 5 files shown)


tsocha:
@dvrogozh
`USE_KINETO` is enabled by default in pytorch.
Could you recompile it with it enabled?
It's better to know issues as fast as possible.

dvrogozh:
@tsocha, `USE_KINETO` for XPU requires PTI library to be available, but it was not yet enabled with the `dpclang` compiler. We are working on that. If someone will attempt to build pytorch with the `USE_KINETO=1` he will get the straightforward configuration error that PTI was not found. Thus, at the moment there are no issues we can identify on pytorch side related to profiling path when we build with `dpclang` compiler. We will start see these errors once PTI will get enabled and we will have a chance to start enabling profiling path in pytorch when it is compiled with `dpclang`.

I will file an issue today to start tracking tasks required for enabling pytorch build against Linux provided dependencies instead of oneAPI and break down the planned steps in there.

```
  CMake Error at third_party/kineto/libkineto/src/plugin/xpupti/CMakeLists.txt:23 (find_package):
    By not providing "FindPti.cmake" in CMAKE_MODULE_PATH this project has
    asked CMake to find a package configuration file provided by "Pti", but
    CMake did not find one.

    Could not find a package configuration file provided by "Pti" with any of
    the following names:

      Pti.cps
      pti.cps
      PtiConfig.cmake
      pti-config.cmake

    Add the installation prefix of "Pti" to CMAKE_PREFIX_PATH or set "Pti_DIR"
    to a directory containing one of the above files.  If "Pti" provides a
    separate development package or SDK, be sure it has been installed.
```

guangyey:
@EikanWang WDYT

cmake/BuildFlags.cmake
```diff
@@ -109,8 +109,16 @@ macro(set_build_flags)
     elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
       set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -Wno-absolute-value)
       set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -fno-fast-math)
-      set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -fma)
-      set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -no-ftz)
+      # -fma which we used before is an alias used for -ffp-contract=fast for compatibility reasons
+      # with very old version of the ICX compiler. The -ffp-contract=fast is supported by both closed
+      # source and open source DPC++ compiler versions.
+      set(SYCL_KERNEL_OPTIONS ${SYCL_KERNEL_OPTIONS} -ffp-contract=fast)
```
EikanWang, line 117: Should it be `fast` or `on`? @chunhuanMeng

cmake/Modules/FindSYCLToolkit.cmake
```diff
@@ -178,34 +178,5 @@ if(${has_werror} EQUAL -1)
...
-# Extract test output for information
-SYCL_CMPLR_TEST_EXTRACT(${test_output} "__INTEL_LLVM_COMPILER")
-
-# Check whether the value of __INTEL_LLVM_COMPILER macro was successfully extracted
-string(COMPARE EQUAL "${__INTEL_LLVM_COMPILER}" "" nosycllang)
-if(nosycllang)
-  set(SYCLTOOLKIT_FOUND False)
-  set(SYCL_REASON_FAILURE "Can not find __INTEL_LLVM_COMPILER}")
-  set(SYCL_NOT_FOUND_MESSAGE "${SYCL_REASON_FAILURE}")
-endif()
```
EikanWang, line 208: What's the expected outcome now if the system installs neither `dpclang` nor `icpx`?

EikanWang (commented):
By the way, it does not belong to this PR. @guangyey , I think we should review all the  `public` link and `public` include in CMake and replace them with `private` to control the scope conservatively.

cmake/BuildFlags.cmake
dvrogozh: References:
* https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-2/ffp-contract.html
* https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-2/fma-qfma.html

From `-ffp-contract`:
* `fast` Fuses floating-point operations across statements.
* `on` Fuses floating-point operations within the same statement.

Per my understanding we want fusing across statements, i.e. `fast` which is analogue of `-fma`. @mdtoguchi :  can you, please, comment?

cmake/Modules/FindSYCLToolkit.cmake
dvrogozh: Behavior will be the same as before. The check for the available SYCL compiler is actually located in the upstream pytorch https://github.com/pytorch/pytorch/blob/main/cmake/Modules/FindSYCLToolkit.cmake. The check here is a redundant check which verifies availability of `__INTEL_LLVM_COMPILER` on top of already detected SYCL compiler. Further `__INTEL_LLVM_COMPILER` is `icpx` specific macro, `dpclang` does does not define it. Thus, this macros should not be checked on cmake level, but instead checked in source in the `#if defined(__INTEL_LLVM_COMPILER)` to distinguish `icpx` from `dpclang`.

We talked with @guangyey about that and I believe @guangyey was planning to drop this check here even before we started work on `dpclang`.

cmake/BuildFlags.cmake
mdtoguchi: For icx, `-fma` is the equivalent of `-ffp-contract=fast`. If you want to continue with the same behavior, `-ffp-contract=fast` is what to use here.

dvrogozh:
@EikanWang , @guangyey , can you, please, review and merge this PR?

EikanWang (approved):

guangyey (approved):
Thanks!
I will prepare a separate PR to change `PUBLIC` to `PRIVATE` properly.

chuanqi129:

chuanqi129:
✅ PR has been successfully merged by @dvrogozh (force merge).

PR #3298, opened 2026-04-10
Add static for kernel_bundle and sycl::kernel to enforce a single globally accessible instance.

minmingzhu:
By declaring kernel_bundle and sycl::kernel as static, they are created and initialized only once during the first invocation of syclMaxWorkGroupSize<KernelClass>(). Subsequent calls reuse the existing instances directly, avoiding repeated initialization.


tye1:
It is a WA in framework level not a real performance optimization. Close it

Issue #3106, opened 2026-03-20
Worker crashes when running TestDecompXPU,test_quick_core_backward_baddbmm_xpu_float64 in CI.


BBBela:
### 🐛 Describe the bug

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_decomp_xpu.TestDecompXPU,test_quick_core_backward_baddbmm_xpu_float64

## History

Initially the problem with this test case was that oneDNN matmul kernel called by the `baddbmm` operator wasn't handling the `float64` data type correctly by internally casting it to `float32`. That was causing the results to differ when compared to reference and test case to fail early, after first assert, not going through all the inputs and combinations.
This issue is already resolved by commit  https://github.com/pytorch/pytorch/commit/d26e57653ee745ad0231acfa7ee740bb09b699a2

Now the `float64` data type is being handled properly, but the test still fails due to worker crash, not the numerical mismatches or assertions.

## Investigation

I tried to launch this test case locally and it is passing, but it takes a lot of time.
On PVC platform it took 22 minutes:
```text
test_decomp_xpu.py::TestDecompXPU::test_quick_core_backward_baddbmm_xpu_float64 PASSED
1 passed, 9134 deselected in 1338.58s (0:22:18)
```
On BMG it was much worse, because it took more than 2 hours (9234 seconds exactly).

I checked the logs and it looks like the worker crashes after about 10 minutes, which correlates with the [current timeout](https://github.com/intel/torch-xpu-ops/blob/1a4d83524c70c05ae575132cc4721917fcafdef1/.github/workflows/_linux_ut.yml#L71) of 600 seconds.

It is even directly mentioned that this test case is slow even for CUDA A100: [HERE](https://github.com/pytorch/pytorch/blob/db9c26baad305c07f76307a1abf4a5de7bd36ccc/test/test_decomp.py#L538)

### Summary of investigation

Taking all the above into consideration, it looks like the problem is not with test logic, but with the currently set timeout, while the test takes longer to success.

### How it looks on CUDA

It looks like for CUDA the timeout is set to 6 times the standard timeout (600 seconds) for slow mode here [TIMEOUT SET](https://github.com/pytorch/pytorch/blob/db9c26baad305c07f76307a1abf4a5de7bd36ccc/test/run_test.py#L624-L628)
So basically on CUDA the timeout is 60 minutes when testing in SLOW mode.

## Potential solutions

[...]

PR #3299, opened 2026-04-10
[#2811] Increase timeout for test_quick_core_backward

jmamzax:
Part of the https://github.com/intel/torch-xpu-ops/issues/2811. The test_decomp_xpu.py::TestDecompXPU::test_quick_core_backward__unsafe_masked_index_put_accumulate_xpu_float64 test fails because of the timeout. By increasing timeout the test can finished with success. The same test on CUDA is listed in slow_test.json:
`"test_quick_core_backward__unsafe_masked_index_put_accumulate_cuda_float64 (__main__.TestDecompCUDA)": 1446.0608520507812,` and is running in separate workflow `slow.yml`

### DIFF
--- a/test/xpu/test_decomp_xpu.py
+++ b/test/xpu/test_decomp_xpu.py
@@ -16,6 +16,7 @@ import unittest
 from collections import defaultdict
 from functools import partial
 
+import pytest
 import torch._inductor.decomposition
 import torch.autograd
 from torch import Tensor
@@ -591,6 +592,7 @@ class TestDecomp(TestCase):
     @skipIfCrossRef
     @suppress_warnings
     @ops(_decomp_test_ops_core_autograd, allowed_dtypes=(torch.float64,))
+    @pytest.mark.timeout(2160)
     def test_quick_core_backward(self, device, dtype, op):
         test_keys = [
             (torch.device(device).type, dtype, op.name),


BBBela:
Hello @jmamzax,
please look at this issue I was working on: https://github.com/intel/torch-xpu-ops/issues/3106

I also had the test case that takes about 24 minutes to complete so it was killed by timeout before it succeeded. As you can see in the issue conversation we agreed with @chuanqi129 to remove those very long tests from standard pull workflows that run in every PR and to launch them only in Nightly tests with increased timeout. 

I prepared some proposal of such solution in https://github.com/intel/torch-xpu-ops/pull/3252 - waiting for review.
If those changes would be accepted, you could simply add this test case to `slow_dict` and it would be launched in Nightly tests only.

And if this test case is the last one in your issue, you can add it to https://github.com/intel/torch-xpu-ops/issues/3106 and close yours. And I will add it to `skip_dict` if PR is approved.

BBBela:
Hi @jmamzax,
the PR #3252 has been merged so now the test cases that takes more than 10 minutes should be added to the `skip_dict` and launched only during Nightly `skipped_ut` job.

jmamzax:
Solution for timeout errors was presented in issue: https://github.com/intel/torch-xpu-ops/issues/3106 and implemented in: https://github.com/intel/torch-xpu-ops/pull/3252

PR #3304, opened 2026-04-10
Fix uninitialized variable issue in AdaptiveMaxPool2dKernelFunctor

### LINKED ISSUE #2676: Random failure in CI test
### 🐛 Describe the bug

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.nn.test_pooling_xpu.TestPoolingNNDeviceTypeXPU,test_max_pool_nan_inf_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.nn.test_pooling_xpu.TestPoolingNNDeviceTypeXPU,test_max_pool_nan_inf_xpu_float64

### Versions

https://github.com/intel/torch-xpu-ops/actions/runs/20689988432/job/59428743168

BBBela:
Fix for: #2676 

## Short Summary

This PR adds initialization of `argmax` variable in `AdaptiveMaxPool2dKernelFunctor` to avoid using it uninitialized.

## Problem descritpion

The root problem is using uninitialized variable as index value.
```cpp
index_t argmax;
```
is later assigned to the list of indices:
```cpp
indices_[o_off] = argmax;
```
In the backward kernel, those indices are used. If the `argmax` has some garbage, then in backward kernel, we try to access the memory from this garbage index which can cause a segmentation fault.

For most cases it works correctly and the `argmax` is correctly assigned during program flow. But there is one case when it is used uninitialized and this case is explored in the failing test cases from #2676 

When all the values are `-inf`, then program never enters the if statement that sets the `argmax` variable:
```cpp
scalar_t max = at::numeric_limits<scalar_t>::lower_bound();
index_t argmax;
int64_t i_bp_off = ob * istrideB_ + op * istrideP_;
for (int64_t ih = istartH; ih < iendH; ih++) {
  for (int64_t iw = istartW; iw < iendW; iw++) {
    int64_t i_hw_off = ih * istrideH_ + iw * istrideW_;
    int64_t i_hw_id = ih * isizeW_ + iw;
    scalar_t val = input_[i_bp_off + i_hw_off];
    if ((val > max) || at::_isnan(val)) {
      max = val;
      argmax = i_hw_id;
    }
  }
}
```
the `max` equals `-inf`, but if all the checked values are also `-inf`, then the if condition is always false as `-inf` is not bigger than `-inf`. I mean we effectively get the comparison `-inf > -inf` and that is false. So in this scenario, we never set the `argmax` and use it uninitialized.

## Proposed solution

The proposed solution is to simply set the `argmax` variable to the first element's value.
The same approach is already being used in the 3D version of `AdaptiveMaxPool`: [3D VERSION EXAMPLE](https://github.com/intel/torch-xpu-ops/blob/4ae570b3944e5cddc063adc475a6fae486cc0be1/src/ATen/native/xpu/sycl/AdaptiveMaxPooling3dKernels.cpp#L74-L75)


tszulist-hbn (approved):

guangyey (approved):
Sounds good!

BBBela:
I did rerun the build and whole `op_ut` job twice and the failing test cases are different in both runs. So they are not caused by changes from this PR, but some PyTorch changes probably.
Rerunning with `disable_all` to enable merge after two CI launches and failing test cases analysis - not related failing test cases.

chuanqi129:
✅ PR has been successfully merged by @BBBela.

PR #2992, opened 2026-03-05
Support xpu graph capture for XCCL

Chao1Han:
This PR add xpu graph status check in collective/pointtopoint op. 
Eager mode: c10::xpu::CaptureStatus::Executing(similar to c10::cuda::CaptureStatus::None), work enqueue and query event status.
Graph mode: c10::xpu::CaptureStatus::Recording(similar to c10::cuda::CaptureStatus::Active), skip work enqueue/watchdog thread.

oneCCL support sycl graph after 2021.17.2(maybe 2022.0), so add errorIfCapturingNonCapturableXCCL func check if it support.
Add allreduce/allgather/reduce_scatter test cases. broadcast/p2p not support yet.

disable_e2e
disable_ut

PR #3307, opened 2026-04-10
Fix extra XPU context creation by moving DeviceGuard early

songhappy:
When running multi-rank XCCL distributed workloads, non-zero ranks (e.g. rank 1 on xpu:1) silently create an extra context on xpu:0, consuming ~500MB+ of additional device memory per rank. This wastes significant GPU memory at scale and causes test_extra_xpu_context to fail.
Fix PYTORCHDGQ-8337

### DIFF
--- a/src/xccl/ProcessGroupXCCL.cpp
+++ b/src/xccl/ProcessGroupXCCL.cpp
@@ -600,6 +600,7 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::endCoalescing(OpType optype) {
 
   auto comm = coalescedComm_;
   auto device = coalescedDevice_;
+  c10::OptionalDeviceGuard gpuGuard(device);
 
   const auto key = std::to_string(device.index());
   auto stream = xcclStreamsMap_.at(key).xpuStream;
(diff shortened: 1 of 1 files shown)


src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -631,6 +600,7 @@ c10::intrusive_ptr<Work> ProcessGroupXCCL::endCoalescing(OpType optype) {
 
   auto comm = coalescedComm_;
   auto device = coalescedDevice_;
+  c10::OptionalDeviceGuard gpuGuard(device);
```
guangyey, line 603: Is this really necessary?

Chao1Han: From nccl comments https://github.com/pytorch/pytorch/blob/30884b8fa9c9c24f62ef30118c2aaeebe54b43de/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L3771-L3772
We need and gpuguard in p2p and collective before graph capture status; but no need add it within endCoalescing.
And here https://github.com/intel/torch-xpu-ops/pull/2992/changes I have moved collective gpuguard location.

songhappy:
since  https://github.com/intel/torch-xpu-ops/pull/2992/changes has similar approach and merged, close this one

PR #3310, opened 2026-04-13
Remove generic entries from gen_op

Silv3S:


Silv3S:
Whole infra for gen_op was upstreamed, so this PR doesn't bring anything now

Issue #3314, opened 2026-04-13
[release/2.13] Test_xpu.py: Fatal Python error: Aborted on windows


RUIJIEZHONG66166:
### 🐛 Describe the bug

The test file: https://github.com/pytorch/pytorch/blob/main/test/test_xpu.py has below issue in windows CI. 
Detail log: https://github.com/intel/torch-xpu-ops/actions/runs/24223088962/job/70724781107
````
00007FF80CBB520800007FF80CBABC10 torch_python.dll!torch::autograd::THPCppFunction_requires_grad [<unknown file> @ <unknown line number>]
00007FF8CB6619D700007FF8CB6619C0 ucrtbase.dll!terminate [<unknown file> @ <unknown line number>]
00007FF8AD1B1943 <unknown symbol address> VCRUNTIME140_1.dll!<unknown symbol> [<unknown file> @ <unknown line number>]
00007FF8AD1B21AF <unknown symbol address> VCRUNTIME140_1.dll!<unknown symbol> [<unknown file> @ <unknown line number>]
00007FF8AD1B220D <unknown symbol address> VCRUNTIME140_1.dll!<unknown symbol> [<unknown file> @ <unknown line number>]
00007FF8AD1B408200007FF8AD1B3FD0 VCRUNTIME140_1.dll!_CxxFrameHandler4 [<unknown file> @ <unknown line number>]
00007FF8CDEA61CF00007FF8CDEA6130 ntdll.dll!_chkstk [<unknown file> @ <unknown line number>]
00007FF8CDD523A700007FF8CDD51E10 ntdll.dll!RtlLocateExtendedFeature [<unknown file> @ <unknown line number>]
00007FF8CDD4A96100007FF8CDD4A740 ntdll.dll!RtlRaiseException [<unknown file> @ <unknown line number>]
00007FF8CB2673FA00007FF8CB267370 KERNELBASE.dll!RaiseException [<unknown file> @ <unknown line number>]
00007FF8B8F752C700007FF8B8F75230 VCRUNTIME140.dll!CxxThrowException [<unknown file> @ <unknown line number>]
00007FF8476890B600007FF8476878C0 sycl8.dll!sycl::_V1::queue::ext_codeplay_supports_fusion [<unknown file> @ <unknown line number>]
00007FF84784031100007FF847815570 sycl8.dll!_sycl_register_lib [<unknown file> @ <unknown line number>]
[...]

PR #3316, opened 2026-04-13
[CI] Skip test_xpu.py in windows CI and nightly

RUIJIEZHONG66166:
Skip this test case in windows CI and nightly due to issue: https://github.com/intel/torch-xpu-ops/issues/3314, which makes windows CI green. Will put it back when issue fixed.

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -140,7 +140,7 @@ jobs:
     secrets: inherit
     uses: ./.github/workflows/_windows_ut.yml
     with:
-      ut: op_extended
+      ut: op_extended,test_xpu
       runner: Windows_CI
       build_runner: Windows_CI_Build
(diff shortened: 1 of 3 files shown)


.github/workflows/pull.yml
```diff
@@ -140,7 +140,7 @@ jobs:
     secrets: inherit
     uses: ./.github/workflows/_windows_ut.yml
     with:
-      ut: op_extended,test_xpu
+      ut: op_extended
```
chuanqi129: Similar with copilot comment, let's skip the crash test in `_windows_ut.yml` directly, to keep the sycl extention test case run

chuanqi129 (changes requested):

.github/workflows/pull.yml
RUIJIEZHONG66166: Modified. Could you please double check?

chuanqi129 (approved):

PR #3319, opened 2026-04-13
Enable vectorized memory access for fp32/fp16/bf16 dtype cast copies

CuiYifeng:
The generic `copy_kernel` path uses `gpu_kernel` which, for cross-dtype copies, detects `needs_dynamic_casting=true` and falls into the `LoadWithCast/StoreWithCast` code path. This path performs per-element runtime dtype dispatch and disables vectorized (wide, aligned) memory loads/stores — resulting in significantly lower memory bandwidth utilization.

This PR intercepts the 6 common fp32/fp16/bf16 cast combinations before reaching the generic path and uses gpu_kernel_nocast with an explicit `CastScalarFunc<in_t, out_t>` whose functor signature matches the tensor dtypes exactly, so `needs_dynamic_casting` evaluates to false and the vectorized kernel launch path is taken.

Covered conversions:

- fp32 ↔ fp16
- fp32 ↔ bf16
- fp16 ↔ bf16

### DIFF
--- a/src/ATen/native/xpu/sycl/CopyKernel.cpp
+++ b/src/ATen/native/xpu/sycl/CopyKernel.cpp
@@ -146,48 +146,62 @@ void float4_copy_kernel_xpu(TensorIteratorBase& iter) {
   }
 }
 
-// Optimized cast kernel for common fp32/fp16/bf16 conversions.
-// Uses gpu_kernel_nocast with explicit CastScalarFunc to enable vectorized
-// memory access, bypassing the generic LoadWithCast/StoreWithCast path
-// which disables vectorization and uses per-element runtime dtype dispatch.
-bool try_fast_copy_kernel_xpu(TensorIteratorBase& iter) {
+bool isFastCastType(TensorIteratorBase& iter) {
   ScalarType dst_dtype = iter.dtype(0);
   ScalarType src_dtype = iter.dtype(1);
+
   if (dst_dtype == src_dtype) {
     return false;
   }
 
+  if (src_dtype != kFloat && src_dtype != kHalf && src_dtype != kBFloat16) {
+    return false;
+  }
+
+  if (dst_dtype != kFloat && dst_dtype != kHalf && dst_dtype != kBFloat16) {
+    return false;
+  }
+
+  return true;
+}
+
+// Optimized cast kernel for common fp32/fp16/bf16 conversions.
+// Uses gpu_kernel_nocast with explicit CastScalarFunc to enable vectorized
+// memory access, bypassing the generic LoadWithCast/StoreWithCast path
+// which disables vectorization and uses per-element runtime dtype dispatch.
+void fast_copy_kernel_xpu(TensorIteratorBase& iter) {
+  ScalarType dst_dtype = iter.dtype(0);
+  ScalarType src_dtype = iter.dtype(1);
+
   // fp32 <-> fp16
   if (src_dtype == kFloat && dst_dtype == kHalf) {
     gpu_kernel_nocast(iter, CastScalarFunc<float, Half>());
-    return true;
+    return;
   }
   if (src_dtype == kHalf && dst_dtype == kFloat) {
     gpu_kernel_nocast(iter, CastScalarFunc<Half, float>());
-    return true;
+    return;
   }
 
   // fp32 <-> bf16
   if (src_dtype == kFloat && dst_dtype == kBFloat16) {
     gpu_kernel_nocast(iter, CastScalarFunc<float, BFloat16>());
-    return true;
+    return;
   }
   if (src_dtype == kBFloat16 && dst_dtype == kFloat) {
     gpu_kernel_nocast(iter, CastScalarFunc<BFloat16, float>());
-    return true;
+    return;
   }
 
   // fp16 <-> bf16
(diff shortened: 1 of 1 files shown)


CuiYifeng:
Speedup data collected from B60:

## float32 → float16
| Shape (batch, dim) | Speedup |
|---|---|
| [8192, 1024] | 1.20x |
| [131072, 1024] | 1.21x |
## float16 → float32
| Shape (batch, dim) | Speedup |
|---|---|
| [8192, 1024] | 1.41x |
| [131072, 1024] | 1.45x |
## bfloat16 → float32
| Shape (batch, dim) | Speedup |
|---|---|
| [8192, 1024] | 1.39x |
| [131072, 1024] | 1.44x |

jianyizh (approved):

src/ATen/native/xpu/sycl/CopyKernel.cpp
```diff
@@ -156,6 +200,9 @@ void copy_kernel(TensorIteratorBase& iter) {
     float8_copy_kernel_xpu(iter);
   } else if (iter.dtype(0) == kFloat4_e2m1fn_x2) {
     float4_copy_kernel_xpu(iter);
+  } else if (try_fast_copy_kernel_xpu(iter)) {
```
guangyey:
```suggestion
  } else if (isFastCastType(iter)) {
      fast_cast_copy_kernel_xpu(iter)
```
do you like this style?

guangyey (approved):

src/ATen/native/xpu/sycl/CopyKernel.cpp
CuiYifeng: Good idea! I also like this style, but we need to avoid duplicate dtype checks in `isFastCastType` and `fast_cast_copy_kernel_xpu`. Let me try.

CuiYifeng: @guangyey Code style has been updated and there are two work-arounds for Half -0.0 (which caused accuracy in some cases). Please take another look, thanks.

src/ATen/native/xpu/sycl/CopyKernel.cpp
```diff
@@ -35,6 +35,25 @@ struct CastScalarFunc {
...
 
+// TODO: Remove this work around for Half -0.0
+template <>
+struct CastScalarFunc<Half, float> {
+  float operator()(Half src_val) const {
+    if (src_val.x == 0x8000)
+      return -0.0f;
+    return (float)src_val;
+  }
+};
```
guangyey, line 45: haha...

guangyey (approved):

src/ATen/native/xpu/sycl/CopyKernel.cpp
guangyey: LGTM.

src/ATen/native/xpu/sycl/CopyKernel.cpp
```diff
@@ -35,6 +35,22 @@ struct CastScalarFunc {
...
 
+// TODO: Remove this workaround for Half -0.0
+template <>
+struct CastScalarFunc<Half, float> {
+  float operator()(Half src_val) const {
+    return c10::convert<float>(src_val);
+  }
+};
+
+// TODO: Remove this workaround for Half -0.0
```
EikanWang, line 46: @CuiYifeng , what's the issue here? Have we submitted an issue to the intel/llvm community or the IGC community?

src/ATen/native/xpu/sycl/CopyKernel.cpp
```diff
@@ -146,6 +162,64 @@ void float4_copy_kernel_xpu(TensorIteratorBase& iter) {
...
+    return false;
+  }
+
+  if (src_dtype != kFloat && src_dtype != kHalf && src_dtype != kBFloat16) {
+    return false;
+  }
+
+  if (dst_dtype != kFloat && dst_dtype != kHalf && dst_dtype != kBFloat16) {
+    return false;
+  }
```
EikanWang, line 179:
```suggestion
  constexpr std::array supported = {kFloat, kHalf, kBFloat16};
  auto is_supported = [&](ScalarType dt) {
    return std::find(supported.begin(), supported.end(), dt) != supported.end();
  };
  return is_supported(src_dtype) || is_supported(dst_dtype);
```

src/ATen/native/xpu/sycl/CopyKernel.cpp
```diff
@@ -146,6 +162,64 @@ void float4_copy_kernel_xpu(TensorIteratorBase& iter) {
   }
 }
 
+bool isFastCastType(const TensorIteratorBase& iter) {
```
EikanWang, line 165: Regarding the function name, it does not mean "Fast", right? The convert function only includes fp32<->bf16 and fp32<->fp16. Is there any real case that requires bf16<->fp16?

src/ATen/native/xpu/sycl/CopyKernel.cpp
CuiYifeng: Half -0.0 was not converted correctly. This issue has not been submitted to intel/llvm or IGC yet. That's why I hold on this PR. Will prepare standalone cases later.

PR #3323, opened 2026-04-14
[xpu][feature] Add fused Dual GEMM SiLU-Mul kernel (sycl-tla)

yucai-intel:
## Summary

Implements `xpu_ops::dual_gemm_silu_mul`, a fused kernel computing:
```
out = silu(x @ w1.T) * (x @ w3.T)
```
using the sycl-tla (CUTLASS SYCL) DualGemm primitive on Intel XPU (PVC).

The kernel loads the shared A matrix (`x`) from HBM only once, saving one full matrix read vs two separate GEMM+epilogue launches.

## New files

- `dual_gemm_api.h / .cpp`: public C++ API — validates inputs, transposes `w1`/`w3` to `[K, N]` row-major before calling the sycl-tla kernel
- `sycltla/dual_gemm_common.h`: sycl-tla header includes and type aliases
- `sycltla/dual_gemm.cpp`: CUTLASS kernel instantiation (BF16 in/out, float accumulator, TileShape 128×128×64, PVC XMX16)
- `DualGemmOps.cpp`: `TORCH_LIBRARY` schema + `TORCH_LIBRARY_IMPL` for `XPU` dispatch key and `Meta` dispatch key (required for `torch.compile` / FakeTensor shape propagation)

## CMake changes

- `ATen/CMakeLists.txt`: glob `dual_gemm/*.cpp` and `dual_gemm/sycltla/*.cpp`
- `CMakeLists.txt`: add `ATen_XPU_DUAL_GEMM_DIRS` cmake variable
- `BuildOnLinux.cmake`: add `ATen_XPU_DUAL_GEMM_DIRS` to sycltla include paths

## Known issues / follow-up

The kernel currently underperforms oneDNN/oneMKL eager on LLaMA-style shapes (small M, large N=14336). The hardcoded `TileShape = Shape<_128, _128, _64>` is not well-suited for small-M rectangular problems. Follow-up work needed to tune tile sizes or add a shape guard in the Inductor FX pass.

## Testing

Tested with `test_dual_gemm.py` (correctness + Inductor fusion trigger) and `bench_dual_gemm.py` (performance) on Intel Data Center GPU Max 1550.

### DIFF
--- a/src/ATen/native/transformers/xpu/dual_gemm/dual_gemm_api.cpp
+++ b/src/ATen/native/transformers/xpu/dual_gemm/dual_gemm_api.cpp
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2020-2026 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include <ATen/native/transformers/xpu/dual_gemm/dual_gemm_api.h>
+
+#ifdef USE_SYCLTLA
+#include <ATen/xpu/XPUContext.h>
+
+namespace sycltla {
+// Forward declaration of the sycltla implementation
+// (dual_gemm/sycltla/dual_gemm.cpp)
+at::Tensor dual_gemm_silu_mul_impl(
+    sycl::queue& queue,
+    const at::Tensor& x,
+    const at::Tensor& w1,
+    const at::Tensor& w3);
+} // namespace sycltla
+#endif // USE_SYCLTLA
+
+namespace sycltla {
+
+bool is_dual_gemm_available() {
+#ifndef USE_SYCLTLA
+  return false;
+#else
+  return true;
+#endif
+}
+
+at::Tensor dual_gemm_silu_mul(
+    const at::Tensor& x,
+    const at::Tensor& w1,
+    const at::Tensor& w3) {
+#ifndef USE_SYCLTLA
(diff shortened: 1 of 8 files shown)


src/ATen/native/transformers/xpu/dual_gemm/dual_gemm_api.cpp
```diff
@@ -0,0 +1,86 @@
...
+      "dual_gemm_silu_mul: x, w1, w3 must share the same K dimension");
+  TORCH_CHECK(
+      w1.size(0) == w3.size(0),
+      "dual_gemm_silu_mul: w1 and w3 must have the same N dimension");
+
+  // The sycl-tla DualGemm kernel expects B matrices in [K, N] row-major
+  // (i.e. already transposed relative to the [N, K] weight layout).
+  // Transpose w1 and w3 so the kernel computes x @ w1.T correctly.
+  at::Tensor w1_t = w1.t().contiguous();
+  at::Tensor w3_t = w3.t().contiguous();
```
jerrychenhf, line 79: For a linear module, the weight tensor is usually created in (N, K) = (out_feature, in_feature) row major.  weight.t() will be (K, N) column major. w1.t().contiguous() and w3.t().contiguous() get (K, N) row-major.

Will this in many cases causes a memory copy for weight from linear which is  (K, N) column major for weight.t()?
If there is a memory copy happening, all the fusion benefits will go away I think.

jerrychenhf (commented):
@yucai-intel Question as to at::Tensor w3_t = w3.t().contiguous();
If this will cause a memory copy and make performance worse, a better approach may check this out side before fusion and avoid going the fusion path for this case?

PR #3328, opened 2026-04-14
Fix alignment gating for vectorized QKV transform using alignof

gplutop7:
Fix alignment gating in transform_bias_rescale_qkv_kernel by checking pointer alignment against alignof(VectorT) instead of % TRANSFORM_BIAS_RESCALE_VEC. This makes the condition type-accurate and easier to maintain, while preserving behavior for typical aligned allocations.

This fixes a pre-existing bug introduced in commit b1582e14.


tszulist-hbn (approved):

src/ATen/native/transformers/sycl/AttentionKernels.cpp
```diff
@@ -442,7 +443,7 @@ void _transform_bias_rescale_qkv_kernel(
         const bool aligned =
             ((dim_per_head % TRANSFORM_BIAS_RESCALE_VEC) == 0) &&
             ((reinterpret_cast<intptr_t>(qkv_bias.data_ptr()) %
-              TRANSFORM_BIAS_RESCALE_VEC) == 0);
+              alignof(VectorT)) == 0);
```
guangyey: `alignof` here relaxes the alignment requirement. I don't think this is correct. @jianyizh what do you think

jianyizh: TRANSFORM_BIAS_RESCALE_VEC = 4,  vectorT is array. alignof(vectorT) will be the same as alignof(scalar_t)? It can be 2 in the case of half? Here you use Array but call it vector, you should use at::native::memory::aligned_vector if you want to do vectorization later.

gplutop7: Good catch - switched to memory::aligned_vector so alignof(LoadT<scalar_t, VEC>) gives the correct vector alignment. Also found the non-nested path was missing the input pointer alignment check - fixed that too.

jianyizh (approved):

gplutop7:
Adding disable_all since the new failures appear in unrelated test groups (AOTAutograd/Functorch, profiler, custom ops, autograd, fake tensor, linalg, and core tensor behavior), which are outside this PR’s scope.
This PR only changes the XPU attention/QKV transform path, so these failures are likely independent.

PR #3333, opened 2026-04-15
Fix nondeterministic alert for _histc_xpu: only alert for floating point input

chunhuanMeng:
_histc_xpu uses atomicAdd which is only nondeterministic for floating point types. Integer atomicAdd is deterministic. Align with CUDA behavior (_histc_cuda) by adding an isFloatingType guard.

Also update the test to parametrize with float32 and int32 dtypes, matching the CUDA test (test_nondeterministic_alert_histc_cuda_float32 and test_nondeterministic_alert_histc_cuda_int32).


yucai-intel (approved):

CuiYifeng (approved):
LGTM.

chuanqi129:

PR #3374, opened 2026-04-17
Rewrite `nonzero` kernel to use bounded scratch memory

pbielak:
For small tensors (N ≤ 2^25), keep the original single-pass `pstl::copy_if`
path which is faster (~1.2x). For large tensors, switch to a two-pass chunked
algorithm that bounds scratch memory at ~512 MB regardless of input size:

Pass 1 – count: a work-group tree reduction counts nonzeros per 2^25-element
  chunk and writes one partial sum per WG.

Pass 2 – scatter: for each chunk, a 0/1 mask and its inclusive prefix sum are
  computed in scratch buffers capped at 2^25 elements. ScatterToOutKernelFunctor
  converts flat indices to per-dimension coordinates and writes them directly
  into the output tensor.

The previous implementation allocated 3×N×8 bytes of scratch (mask + prefix-sum
+ flat-index output), exceeding 100 GB for a 4B-element tensor. The chunked
path makes `test_nonzero_large` (2^31-1 × 2, int8) pass on low-VRAM hardware.

Additional cleanups:
- Merge bool template specializations using `if constexpr`
- Replace manual ceiling division with `at::ceil_div()`
- Remove INT_MAX element-count limitation from Nonzero.cpp

### DIFF
--- a/src/ATen/native/xpu/sycl/NonzeroKernel.cpp
+++ b/src/ATen/native/xpu/sycl/NonzeroKernel.cpp
@@ -9,9 +9,11 @@
  */
 
 #include <ATen/Dispatch.h>
+#include <ATen/ceil_div.h>
 #include <ATen/core/Tensor.h>
 
 #include <ATen/native/xpu/sycl/pstl/PSTLFunctions.h>
+#include <comm/Memory.h>
 #include <comm/SYCLHelpers.h>
 #include <comm/TensorInfo.h>
 
@@ -20,6 +22,148 @@
 
 namespace at::native::xpu {
 
+// 0/1 int64 mask: global_mask[i] = 1 iff data[i] != 0.
+// For bool, use volatile int to prevent the compiler from eliminating the load.
+template <typename scalar_t>
+struct IsNonzeroKernelFunctor {
+  void operator()(sycl::item<1> item_id) const {
+    if constexpr (std::is_same_v<scalar_t, bool>) {
+      volatile int in = static_cast<int>(data_ptr_[item_id]);
+      global_mask_ptr_[item_id] = static_cast<int64_t>(in != 0);
+    } else {
+      global_mask_ptr_[item_id] =
+          static_cast<int64_t>(data_ptr_[item_id] != scalar_t(0));
+    }
+  }
+  IsNonzeroKernelFunctor(const scalar_t* data_ptr, int64_t* global_mask_ptr)
+      : data_ptr_(data_ptr), global_mask_ptr_(global_mask_ptr) {}
+
+ private:
+  const scalar_t* data_ptr_;
+  int64_t* global_mask_ptr_;
+};
+
+// Work-group-level reduction: counts nonzeros in [data_, data_+N_).
+// Each work-group writes its partial count to partial_sums_[group_id].
+template <typename scalar_t>
+struct CountNonzerosKernelFunctor : public __SYCL_KER_CONFIG_CONVENTION__ {
+  void operator()(sycl::nd_item<1> item) const {
+    const auto local_id = item.get_local_linear_id();
+    const auto global_id = item.get_global_linear_id();
+
+    if constexpr (std::is_same_v<scalar_t, bool>) {
+      int64_t val = 0;
+      if (global_id < static_cast<size_t>(N_)) {
+        volatile int in = static_cast<int>(data_[global_id]);
+        val = static_cast<int64_t>(in != 0);
+      }
+      local_buf_[local_id] = val;
+    } else {
+      local_buf_[local_id] = static_cast<int64_t>(
+          global_id < static_cast<size_t>(N_) &&
+          data_[global_id] != scalar_t(0));
+    }
+    sycl::group_barrier(item.get_group());
+
+    for (int64_t stride = wg_size_ / 2; stride > 0; stride >>= 1) {
+      if (local_id < static_cast<size_t>(stride))
+        local_buf_[local_id] += local_buf_[local_id + stride];
+      sycl::group_barrier(item.get_group());
+    }
+
+    if (local_id == 0)
+      partial_sums_[item.get_group_linear_id()] = local_buf_[0];
+  }
+
+  void sycl_ker_config_convention(sycl::handler& cgh) {
+    local_buf_ = sycl_local_acc_t<int64_t>(wg_size_, cgh);
+  }
+
+  CountNonzerosKernelFunctor(
+      const scalar_t* data,
+      int64_t N,
+      int64_t* partial_sums,
+      int64_t wg_size)
+      : data_(data), N_(N), partial_sums_(partial_sums), wg_size_(wg_size) {}
+
+ private:
+  const scalar_t* data_;
+  int64_t N_;
+  int64_t* partial_sums_;
+  int64_t wg_size_;
+  sycl_local_acc_t<int64_t> local_buf_;
+};
+
+// For each nonzero element, converts its flat index to per-dimension indices
+// and writes them directly into the output buffer (layout: dim-major, i.e.
+// out_ptr[d * num_nonzeros + slot]).
+struct ScatterToOutKernelFunctor {
+  void operator()(sycl::item<1> item_id) const {
+    if (global_mask_ptr_[item_id] != 0) {
+      // target_pos is the inclusive prefix sum of global_mask, so
+      // target_pos[i]-1 is this element's rank among nonzeros in the chunk.
+      // global_offset shifts it to the correct position in the full output.
+      const int64_t slot = global_offset_ + target_pos_ptr_[item_id] - 1;
+      const int64_t flat_idx =
+          chunk_start_ + static_cast<int64_t>(item_id.get_linear_id());
+      for (int64_t d = 0; d < num_dim_; d++) {
+        out_ptr_[d * num_nonzeros_ + slot] = flat_idx / divisor_[d] % sizes_[d];
+      }
+    }
+  }
+  ScatterToOutKernelFunctor(
+      const int64_t* global_mask_ptr,
+      const int64_t* target_pos_ptr,
+      int64_t* out_ptr,
+      int64_t chunk_start,
+      int64_t global_offset,
+      int64_t num_nonzeros,
+      int64_t num_dim,
+      int64_t* divisor,
+      int64_t* sizes)
+      : global_mask_ptr_(global_mask_ptr),
+        target_pos_ptr_(target_pos_ptr),
+        out_ptr_(out_ptr),
+        chunk_start_(chunk_start),
+        global_offset_(global_offset),
+        num_nonzeros_(num_nonzeros),
+        num_dim_(num_dim) {
+    for (int64_t d = 0; d < num_dim; d++) {
+      divisor_[d] = divisor[d];
+      sizes_[d] = sizes[d];
+    }
+  }
+
+ private:
+  const int64_t* global_mask_ptr_;
+  const int64_t* target_pos_ptr_;
+  int64_t* out_ptr_;
+  int64_t chunk_start_;
+  int64_t global_offset_;
+  int64_t num_nonzeros_;
+  int64_t num_dim_;
+  int64_t divisor_[XPU_MAX_TENSORINFO_DIMS];
+  int64_t sizes_[XPU_MAX_TENSORINFO_DIMS];
+};
+
+// Predicate for pstl::copy_if: returns true if self_begin_[x] != 0.
+template <typename scalar_t>
+struct CopyIfFunc {
+  bool operator()(int64_t x) const {
+    if constexpr (std::is_same_v<scalar_t, bool>) {
+      volatile int in = static_cast<int>(self_begin_[x]);
+      return in != 0;
+    } else {
+      return self_begin_[x] != scalar_t(0);
+    }
+  }
+  CopyIfFunc(const scalar_t* self_begin) : self_begin_(self_begin) {}
+
+ private:
+  const scalar_t* self_begin_;
+};
+
+// Converts flat nonzero indices to per-dimension coordinates.
 struct FlattenIdxtoRealIdxKernelFunctor {
   void operator()(sycl::nd_item<1> item_id) const {
     auto global_id = item_id.get_global_linear_id();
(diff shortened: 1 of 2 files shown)


pbielak:
Comparing the old and new implementations:

| Shape | dtype | NNZ | Time (new/old) | Scratch (new/old) |
| --- | --- | --- | --- | --- |
| (1.05M,) | float32 | 100% | 0.89x | 0.99x |
| (1.05M,) | float32 | 1% | 0.68x | 1.47x |
| (2.05K, 2.05K) | float32 | 100% | 0.75x | 0.75x |
| (2.05K, 2.05K) | float32 | 1% | 0.48x | 1.48x |
| (2.05K, 2.05K) | int8 | 50% | 0.54x | 1.06x |
| (2.05K, 2.05K) | bool | 50% | 0.54x | 1.06x |
| (8.19K, 8.19K) | float32 | 100% | 1.19x | 1.00x |
| (8.19K, 8.19K) | float32 | 1% | 0.82x | 2.94x |
| (8.19K, 8.19K) | int8 | 100% | 1.18x | 1.00x |
| (537M,) | float32 | 100% | (OOM before) | OOM → 4.5 GB |
| (537M,) | float32 | 0.1% | (OOM before) | OOM → 520 MB |
| (1.07B, 2) | int8 | 0.01% | (OOM before) | OOM → 532 MB |
| (2.15B, 2) | int8 | 2.328e-07% | (OOM before) | OOM → 544 MB |

Speedup >1x = new kernel is faster / uses less scratch.
Speedup <1x = new kernel is slower / uses more scratch.

astachowiczhabana:
Is there any issue linked? Or is this improvement PR?

pbielak:
> Is there any issue linked? Or is this improvement PR?

@astachowiczhabana If accepted, this PR will be a replacement for https://github.com/intel/torch-xpu-ops/pull/2884 fixing the `test_unary_ufuncs_xpu.TestUnaryUfuncsXPU,test_nonzero_large_xpu_int8` test

astachowiczhabana:
Great work! I really like how this trades extra computation for reduced memory usage. Specifically fusing position lookup and coordinate conversion into a single kernel is very clever. Only a couple of minor comments below.

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -94,55 +194,116 @@ void nonzero_template(const Tensor& self_, Tensor& out) {
...
+      LEGACY_CONTIGUOUS_MEMORY_FORMAT);
+
+  // scatter_chunk_size caps the global_mask and target_pos scratch buffers
+  // to scatter_chunk_size * sizeof(int64) = 256 MB each (at 2^25).
+  const int64_t scatter_chunk_size = int64_t(1) << 25;
+  const int64_t num_chunks = (N + scatter_chunk_size - 1) / scatter_chunk_size;
 
-  Tensor idx_flat = at::empty(
-      {N}, out.options().memory_format(LEGACY_CONTIGUOUS_MEMORY_FORMAT));
+  // ---- Pass 1: count nonzeros per chunk via work-group reduction ----
```
astachowiczhabana, line 211: Could you add a short comment explaining why `1 << 25` was chosen as the chunk size?

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -94,55 +194,116 @@ void nonzero_template(const Tensor& self_, Tensor& out) {
...
+    CountFunctor count_kfn(
+        self_data + start,
+        this_chunk,
+        partial_sums.data_ptr<int64_t>(),
+        count_wg_size);
+    sycl_kernel_submit(
+        num_wgs * count_wg_size, count_wg_size, queue, count_kfn);
+
+    chunk_offsets[ci] = num_nonzeros;
+    chunk_counts[ci] = partial_sums.sum().item<int64_t>();
```
astachowiczhabana: Could you enqueue all count kernels into a single pre-allocated `int64_t[num_chunks]` device buffer without blocking, then do a single device→host sync to retrieve all partial counts at once before allocating the output?

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: Done

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: Thanks for the idea - applied

astachowiczhabana (approved):

CuiYifeng (commented):
A good idea for functionality of large N!
However, my concern is that 2-pass implementation will cause performance regression for N which could have been executed originally. How about using new implementation only for large N (such as N>=INT_MAX)?

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -20,72 +20,172 @@
...
+  int64_t* global_mask_ptr_;
+};
+
+// Specialization for bool: use volatile int to prevent the compiler from
+// eliminating the load when scalar_t == bool.
+template <>
+struct IsNonzeroKernelFunctor<bool> {
+  void operator()(sycl::item<1> item_id) const {
+    volatile int in = (int)data_ptr_[item_id];
+    global_mask_ptr_[item_id] = static_cast<int64_t>(in != 0);
```
guangyey: I can't believe that we still need this workaround to handle bool logic. @pbielak Could you help verify if this workaround is still needed.

pbielak: Unfortunately, we need this specialized case - for normal bool tensors it would work without it, but if we create a bool view from "non-normalized" values (like 2 or 255), the nonzero kernel fails without this specialized class.

I confirmed this by removing the specialized classes and running some tests against the recompiled code: standard bool tests passed, non-normalized ones produced garbage.

pbielak:
> A good idea for functionality of large N!
> However, my concern is that 2-pass implementation will cause performance regression for N which could have been executed originally. How about using new implementation only for large N (such as N>=INT_MAX)?

@CuiYifeng Good point. However, I don't think `N >= INT_MAX` is the right threshold - the old kernel OOMs well before that. Even for a ~500M element tensor, the old approach allocates ~12 GB of scratch memory (3*N-sized int64 buffers), while the new approach stays bounded at ~512 MB regardless of N.

Looking at the benchmarks, the performance crossover is around 67M elements (8K x 8K), where the new kernel is already 1.18–1.19x faster. Below ~4M elements, the old kernel is ~1.5–2x faster, but the absolute times are sub-millisecond either way.

In my opinion, we should keep a single code path to avoid the maintenance cost of two separate kernel implementations (including duplicated bool specializations). 

However, if you'd like the dual path approach, I would suggest to set the threshold to 2^25 (33M, the scatter_chunk_size) instead of INT_MAX.

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: However, I think we could slightly improve the code by using a data type comparison instead of keeping separate specialization classes, e.g.,

```cpp
template <typename scalar_t>
inline int64_t load_as_nonzero(const scalar_t* data, size_t idx) {
  if constexpr (std::is_same_v<scalar_t, bool>) {
    volatile int in = static_cast<int>(data[idx]);
    return static_cast<int64_t>(in != 0);
  } else {
    return static_cast<int64_t>(data[idx] != scalar_t(0));
  }
}
```

and then use it in the `IsNonzeroKernelFunctor`:

```cpp
template <typename scalar_t>
struct IsNonzeroKernelFunctor {
  void operator()(sycl::item<1> item_id) const {
    global_mask_ptr_[item_id] = load_as_nonzero(data_ptr_, item_id);
  }
  // ...
};
```

WDYT @guangyey ?

guangyey: Let's keep it simple. e.g.
```cpp
template <typename scalar_t>
struct IsNonzeroKernelFunctor {
  void operator()(sycl::item<1> item_id) const {
    if constexpr (std::is_same_v<scalar_t, bool>) {
      volatile int in = static_cast<int>(data[item_id]);
      global_mask_ptr_[item_id] = static_cast<int64_t>(in != 0);
    } else {
       global_mask_ptr_[item_id] = static_cast<int64_t>(data[item_id] != scalar_t(0));
    }
  }
  // ...
};
```

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -20,72 +20,172 @@
...
-struct CopyIfFunc<bool> {
-  bool operator()(int64_t x) const {
-    // Using data type conversion to break deduce of execution chain in bool.
-    // Bool operations will be removed in the compiler optimization.
-    // The function returns a bool variable with one byte value stored in
-    // self_begin_ not 1 specified here.
-    volatile int in = (int)self_begin_[x];
-    bool res = in != int(0) ? 1 : 0;
-    return res;
+struct CountNonzerosKernelFunctor<bool>
```
guangyey: Let's absorb this specialization via `if constexpr`.

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -94,55 +194,140 @@ void nonzero_template(const Tensor& self_, Tensor& out) {
...
+      LEGACY_CONTIGUOUS_MEMORY_FORMAT);
+
+  // scatter_chunk_size caps the global_mask and target_pos scratch buffers to
+  // scatter_chunk_size * sizeof(int64_t) = 256 MB each. 2^25 (≈33 M elements)
+  // was chosen as a practical trade-off: large enough that the number of chunks
+  // stays small for typical workloads (e.g. a 4 B-element tensor needs only
+  // ~128 chunks), yet small enough that each pair of scratch buffers fits
+  // comfortably within the memory budget of low-VRAM devices.
+  const int64_t scatter_chunk_size = int64_t(1) << 25;
+  const int64_t num_chunks = (N + scatter_chunk_size - 1) / scatter_chunk_size;
```
guangyey:
```suggestion
  const int64_t num_chunks = at::ceil_div(N, scatter_chunk_size);
```

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -94,55 +194,140 @@ void nonzero_template(const Tensor& self_, Tensor& out) {
...
+
+  // ---- Pass 1: count nonzeros per chunk via work-group reduction ----
+  using CountFunctor = CountNonzerosKernelFunctor<scalar_t>;
+  const auto count_wg_size = syclMaxWorkGroupSize<CountFunctor>();
+
+  // Pre-allocate a single device buffer wide enough to hold every WG's partial
+  // sum for every chunk. All count kernels are enqueued without blocking so
+  // only one device→host transfer is needed at the end.
+  const int64_t max_wgs_per_chunk =
+      (scatter_chunk_size + count_wg_size - 1) / count_wg_size;
```
guangyey:
```suggestion
  const int64_t max_wgs_per_chunk = at::ceil_div(scatter_chunk_size, count_wg_size);
```

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -94,55 +194,140 @@ void nonzero_template(const Tensor& self_, Tensor& out) {
...
-      {N}, out.options().memory_format(LEGACY_CONTIGUOUS_MEMORY_FORMAT));
+  std::vector<int64_t> chunk_wgs(num_chunks);
 
-  const scalar_t* self_begin = self.const_data_ptr<scalar_t>();
-  int64_t* idx_flat_begin = idx_flat.data_ptr<int64_t>();
-  int64_t* range_begin = nullptr;
+  for (int64_t ci = 0; ci < num_chunks; ci++) {
+    const int64_t start = ci * scatter_chunk_size;
+    const int64_t this_chunk = std::min(scatter_chunk_size, N - start);
+    const int64_t num_wgs = (this_chunk + count_wg_size - 1) / count_wg_size;
```
guangyey:
```suggestion
    const int64_t num_wgs = at::ceil_div(this_chunk, count_wg_size);
```

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: Done

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: Done

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: Done

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: Done

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: Done

guangyey (approved):
Hi @pbielak Overall LGTM.
I think @CuiYifeng’s comment makes sense. The new kernel appears to have worse performance than the old one, and the gap becomes larger when sparsity is low. However, it does reduce memory consumption. It's definitely a valuable implement.

I think you could find a threshold to detect which path is currently preferred. What do you think @jianyizh

jianyizh:
What's the implementation of CUDA?

CuiYifeng:
> Hi @pbielak Overall LGTM. I think @CuiYifeng’s comment makes sense. The new kernel appears to have worse performance than the old one, and the gap becomes larger when sparsity is low. However, it does reduce memory consumption. It's definitely a valuable implement.
> 
> I think you could find a threshold to detect which path is currently preferred. What do you think @jianyizh

In my opinion, using `2^25` as the threshold (only 1 chunk) is acceptable.

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
```diff
@@ -94,55 +151,140 @@ void nonzero_template(const Tensor& self_, Tensor& out) {
...
+        this_chunk,
+        all_partial_sums_ptr + ci * max_wgs_per_chunk,
+        count_wg_size);
+    sycl_kernel_submit(
+        num_wgs * count_wg_size, count_wg_size, queue, count_kfn);
+  }
 
-  auto num_nonzeros = std::distance(idx_flat_begin, idx_flat_end);
+  // Single device→host sync: retrieve all partial sums at once.
+  Tensor all_partial_sums_cpu = all_partial_sums.to(at::kCPU);
```
guangyey: Use `memcpyDeviceToHost` to reduce the dispatch overhead.

pbielak:
> What's the implementation of CUDA?

@jianyizh CUDA's `nonzero` is also two-pass:

1. **Pass 1 – Count:** `cub::DeviceReduce::Sum` with a `NonZeroOp<T>` transform iterator. For tensors exceeding `INT_MAX`, the input is split into chunks of ~2^30 elements, each reduced separately. Per-chunk counts are copied to pinned host memory and summed.

2. **D2H sync** to get `num_nonzeros`, allocate output `{ndim, num_nonzeros}`.

3. **Pass 2 – Compact:** `cub::DeviceSelect::Flagged` writes flat nonzero indices, then a `write_indices` kernel converts flat indices to multi-dimensional coordinates via div/mod.

No dual-path / small-tensor fast path — CUDA uses the same 2-pass CUB approach for all sizes. The key difference is that CUB's device-level primitives are highly optimized, so the two-pass overhead is minimal even for small inputs.

src/ATen/native/xpu/sycl/NonzeroKernel.cpp
pbielak: Done

pbielak:
> > Hi @pbielak Overall LGTM. I think @CuiYifeng’s comment makes sense. The new kernel appears to have worse performance than the old one, and the gap becomes larger when sparsity is low. However, it does reduce memory consumption. It's definitely a valuable implement.
> > I think you could find a threshold to detect which path is currently preferred. What do you think @jianyizh
> 
> In my opinion, using `2^25` as the threshold (only 1 chunk) is acceptable.

@CuiYifeng @guangyey @jianyizh I implemented the dual path approach with `scatter_chunk_size` (2^25) as the threshold, as suggested. For `N ≤ 2^25`, the original single-pass `pstl::copy_if` path is used; for larger tensors, the new chunked two-pass algorithm is used.

Speedups of the dual-path vs. the chunked-only implementation (for small tensors):

| Config | dtype | nonzero% | Speedup |
|---|---|---|---|
| (1.05M,) | float32 | 100% | 1.21x |
| (1.05M,) | float32 | 1% | 1.25x |
| (2K, 2K) | float32 | 100% | 1.16x |
| (2K, 2K) | float32 | 1% | 1.27x |
| (2K, 2K) | int8 | 50% | 1.22x |
| (2K, 2K) | bool | 50% | 1.19x |

Large tensors (67M+) are unaffected — they take the chunked path as before.

CuiYifeng (approved):

chuanqi129:
✅ PR has been successfully merged by @pbielak (force merge).

PR #3397, opened 2026-04-20
[#2811] Add long test to skip list to run it only on nightly

jmamzax:
Part of https://github.com/intel/torch-xpu-ops/issues/2811. The TestDecompXPU,test_quick_core_backward__unsafe_masked_index_put_accumulate_xpu_float64 executes more than 10 minutes because of that it's timeout and fails as 10 minutes timeout is set for all tests. The test is skipped to run it only on nightly with longer timeout as described in this issue: https://github.com/intel/torch-xpu-ops/issues/3106

### DIFF
--- a/test/xpu/skip_list_common.py
+++ b/test/xpu/skip_list_common.py
@@ -56,6 +56,7 @@ skip_dict = {
     "test_decomp_xpu.py": (
         # Slow test case: it takes more than 10 minutes to run on XPU.
         "test_quick_core_backward_baddbmm_xpu_float64",
+        "test_quick_core_backward__unsafe_masked_index_put_accumulate_xpu_float64",
     ),
     "test_distributions_xpu.py": None,
     "test_dynamic_shapes_xpu.py": None,
--- a/test/xpu/skip_list_common.py
+++ b/test/xpu/skip_list_common.py
@@ -56,6 +56,7 @@ skip_dict = {
     "test_decomp_xpu.py": (
         # Slow test case: it takes more than 10 minutes to run on XPU.
         "test_quick_core_backward_baddbmm_xpu_float64",
+        # Slow test case: it takes more than 10 minutes to run on XPU.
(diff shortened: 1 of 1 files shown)


Silv3S (approved):

BBBela (approved):
I am just thinking about maybe explicitly adding the `# Slow test case` comment before every slow test we add to avoid any confusion in the future. Because it is not obvious if just the first test case is slow or all the listed test cases from `test_decomp_xpu.py` are slow.

I mean, the approach to add the slow tests to `skipped_ut` is not perfect, but simplest solution. We mix the tests that are skipped because they are failing with tests that pass, but need more time to pass. I was thinking about somehow logically separating them in the future, so we have for example the `skip_failing_dict` and `skip_slow_dict` and to change the summary workflow a little to clearly print what tests are expected to fail and failing and what are expected to pass after longer time and passing. But for now maybe we could explicitly mark every slow test case with comment just before it - what do you think?

test/xpu/skip_list_common.py
```diff
@@ -56,6 +56,7 @@
     "test_decomp_xpu.py": (
         # Slow test case: it takes more than 10 minutes to run on XPU.
         "test_quick_core_backward_baddbmm_xpu_float64",
+        "test_quick_core_backward__unsafe_masked_index_put_accumulate_xpu_float64",
```
BBBela, line 60:
```suggestion
        # Slow test case: it takes more than 10 minutes to run on XPU.
        "test_quick_core_backward__unsafe_masked_index_put_accumulate_xpu_float64",
```
Just a suggestion ;)

guangyey (approved):

guangyey:
Refer to https://github.com/pytorch/pytorch/blob/95e6cf8d58df44a5c92cd96a6d087cd1081c9d17/test/slow_tests.json#L184-L185

chuanqi129:
✅ PR has been successfully merged by @jmamzax.

PR #3400, opened 2026-04-20
[Perf] Add vectorized scatter kernel for bf16/fp16

yanqiuz:
Add a vectorized scatter fast path mirroring the existing vectorized
gather kernel. For scatter_(dim=0) with TensorAssign, contiguous slices
are copied using 16-byte wide vector load/store operations instead of
element-by-element stores.

Results on Intel B60 (BMG, 0xE211):
- bf16 scatter bandwidth: 224 GB/s -> 377 GB/s (1.68x speedup)

### Summary

Add a vectorized scatter kernel to eliminate the **4× L3 write amplification** caused by LSC `d16u32` partial-write stores on Intel Xe2 GPUs (BMG/B580).

### Root Cause

On BMG (Xe2), a SIMD-32 scatter store with `d16u32` (2-byte data, 4-byte address) writes 32×2B = 64B per instruction. However, each 16B-aligned partial write touches only 2B within that 16B block, forcing the L3 cache to perform **4 separate 16B read-modify-write** cycles per cache line. This results in 4× write amplification compared to fp32 (`d32` store, 1 write per cache line).

**unitrace -g ComputeBasic evidence (B580, shape=[4096, 4096], scatter dim=0):**

| Metric | fp32 baseline | bf16 baseline | bf16 vectorized |
|--------|--------------|---------------|-----------------|
| L3_WRITE | 2.10M | **8.39M** | **2.10M** |
| SEND_ALL | 4.20M | 4.20M | 2.23M |
| XVE_STALL | 67.9% | 79.1% | 62.3% |
| Latency (µs) | 415 | 754 | **387** |
| BW (GB/s) | 323 | 178 | **347** |

- bf16 baseline L3_WRITE = 8.39M ≈ 4 × fp32's 2.10M → confirms **4× write amplification**
- bf16 vectorized L3_WRITE = 2.10M = fp32 level → amplification eliminated
- bf16 vectorized is **~1.95× faster** than bf16 baseline, and even slightly faster than fp32

### Approach

Pack 8 consecutive `bf16`/`fp16` elements into a `uint4` (16 bytes) and issue a single `d128` (128-bit) vector store per work-item, avoiding partial writes entirely.

**Eligibility conditions** (`fast_scatter_kernel_eligible`):
1. All tensors (dst, src, index) are contiguous with matching sizes
2. Inner dimension is 16-byte aligned (≥ 8 elements for bf16/fp16)
3. Scatter dim = 0 (gather dimension matches scatter dimension)

When eligible, the kernel dispatches to `VectorizedScatterKernel`; otherwise falls back to the existing elementwise path.

### DIFF
--- a/src/ATen/native/xpu/sycl/IndexKernelUtils.h
+++ b/src/ATen/native/xpu/sycl/IndexKernelUtils.h
@@ -48,6 +48,37 @@ inline bool fast_gather_kernel_eligible(
       get_alignment(static_cast<size_t>(iter.strides(0)[1])) == alignment;
 }
 
+// Eligibility check for vectorized scatter (is_scatter_like=true).
+//
+// For scatter with TensorAssign:
+//   dst[index[i], j] = src[i, j]
+// TensorIterator layout (fastest-to-slowest):
+//   strides(0) = dst: [element_size, 0]  ← dim restrided to 0
+//   strides(1) = src: [element_size, dim_size*element_size]  ← original
+//   strides(2) = idx: [0, idx_elem_size]  ← expanded along dim
+template <int alignment>
+inline bool fast_scatter_kernel_eligible(
+    const TensorIterator& iter,
+    char* const dst_ptr,
+    char* const src_ptr,
+    const size_t index_stride_bytes,
+    const size_t element_size) {
+  using at::native::memory::get_alignment;
+  const auto index_element_size = iter.element_size(2);
+
+  return iter.ndim() == 2 && iter.strides(2)[0] == 0 &&
+      iter.strides(2)[1] == index_element_size &&
+      static_cast<size_t>(iter.strides(0)[0]) == element_size &&
+      static_cast<size_t>(iter.strides(1)[0]) == element_size &&
+      static_cast<size_t>(iter.strides(0)[1]) == 0 &&
+      get_alignment(dst_ptr) == alignment &&
+      get_alignment(src_ptr) == alignment &&
+      get_alignment(static_cast<size_t>(iter.shape()[0] * element_size)) ==
+      alignment &&
+      get_alignment(static_cast<size_t>(index_stride_bytes)) == alignment &&
+      get_alignment(static_cast<size_t>(iter.strides(1)[1])) == alignment;
+}
+
 #define SIMD 32
 
 template <int Alignment, typename index_t>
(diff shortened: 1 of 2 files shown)


src/ATen/native/xpu/sycl/IndexKernelUtils.h
```diff
@@ -48,6 +48,37 @@ inline bool fast_gather_kernel_eligible(
...
+template <int alignment>
+inline bool fast_scatter_kernel_eligible(
+    const TensorIterator& iter,
+    char* const dst_ptr,
+    char* const src_ptr,
+    const size_t index_stride_bytes,
+    const size_t element_size) {
+  using at::native::memory::get_alignment;
+  const auto index_element_size = iter.element_size(2);
+
```
jianyizh, line 68: Be careful with the condition here. The condition I think is 1. contiguous source and dst 2. aligned pointers 3. size and stride can be divided by 4 or 8 4. We read row wise

yanqiuz: guarantee the memory layout is a "contiguous row-copy" pattern with all addresses and strides 16-byte aligned, violating to access misaligned memory, read/write out of bounds, or produce incorrect results

jianyizh (approved):

jianyizh:
@yanqiuz I ask AI to root cause the reason why vectorization brings larger improvement on 16bit datatype. You can update your pr description. For fp32, if you see improvement, you can do ablation study, if it's because less send instruction or we skipped OffsetCalculator .

## Op-Level Performance Analysis: Vectorized Scatter (B580, BMG)

Measured on Intel B580 (BMG), GPU freq pinned at 2850 MHz, using `unitrace -g ComputeBasic` for hardware counters. Benchmark: `scatter_(dim=0)` with **expanded index** (stride(dim)=0, which triggers the vectorized path).

### Speedup Summary

| dtype | Shape | Baseline GpuTime | Vec GpuTime | **Speedup** |
|:---:|:---:|:---:|:---:|:---:|
| **bf16** | [1024, 65536] | 1,070 μs | 579 μs | **1.85×** |
| **fp32** | [1024, 32768] | 576 μs | 579 μs | **1×** |

Both shapes have the same total data volume: 128 MB output (+ 128 MB read).

### Root Cause of bf16 Speedup: Eliminating d16u32 4× L3 Write Amplification

| Metric | Baseline bf16 | Vec bf16 | Baseline fp32 | Vec fp32 |
|--------|:---:|:---:|:---:|:---:|
| L3_WRITE | **8,388,651** | **2,097,195** | 2,097,195 | 2,097,193 |
| L3_WRITE reduction | — | **4.0×** | — | 1.0× (none) |
| SEND_ALL | 29,365,248 | 5,242,880 | 14,685,184 | 5,242,880 |
| XVE_STALL | 30.5% | 90.0% | 44.3% | 90.0% |

**bf16 baseline**: The scalar kernel uses `d16u32` (16-bit data, unpacked to 32-bit register) store instructions. With SIMD32, each store covers 32×2B = 64B = 1 CL, but the hardware performs 4 partial writes per CL (each writing only 16B). This causes **4× L3 write amplification**: `L3_WRITE = 8.39M = total_bytes / 16` instead of the theoretical minimum `total_bytes / 64 = 2.10M`.

**bf16 vectorized**: `st_vec<16>` stores 16 bytes per WI as a packed d32x4, eliminating the d16u32 partial-write issue. L3_WRITE drops to the theoretical minimum 2.10M.

**fp32 baseline**: Scalar `d32` stores with expanded index already coalesce perfectly (32 lanes × 4B = 128B = 2 CLs). L3_WRITE is already at minimum — **no coalescing improvement possible**.

### Why fp32 Speedup is Modest (1.17×)

The fp32 scalar kernel was already well-optimized with expanded index: L3_WRITE at theoretical minimum, only 44% XVE_STALL. The vectorized kernel reduces SEND count by 2.8× and eliminates OffsetCalculator overhead, but becomes memory-latency bound (90% XVE_STALL) due to less instruction-level parallelism to hide latency. The SEND reduction doesn't translate proportionally into time reduction.

### Key Takeaway

**This PR primarily benefits bf16/fp16** by eliminating the d16u32 4× L3 write amplification inherent in scalar 16-bit stores. For fp32, the benefit is secondary (reduced instruction count). The optimization only applies to the expanded-index pattern (stride(dim)=0); random-index scatter remains inherently uncoalesced regardless of kernel implementation.

Issue #2630, opened 2025-12-23
Title: [upstream_ut]  AssertionError: Scalars are not equal!


daisyden:
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNN,test_cudnn_weight_format
op_ut,third_party.torch-xpu-ops.test.xpu.test_binary_ufuncs_xpu.TestBinaryUfuncsXPU,test_fmod_remainder_by_zero_integral_xpu_int64

pytest_command:
cd <pytorch>/third_party/torch-xpu-ops/test/xpu  && PYTORCH_TEST_WITH_SLOW=1 pytest -v third_party/torch-xpu-ops/test/xpu/test_binary_ufuncs_xpu.py -k test_fmod_remainder_by_zero_integral_xpu_int64
cd <pytorch>/third_party/torch-xpu-ops/test/xpu  && PYTORCH_TEST_WITH_SLOW=1 pytest -v third_party/torch-xpu-ops/test/xpu/test_nn_xpu.py -k test_cudnn_weight_format

Error Message:
 AssertionError: Scalars are not equal! 

Trace Example:

```
Command: cd <pytorch>/third_party/torch-xpu-ops/test/xpu  && PYTORCH_TEST_WITH_SLOW=1 pytest -v third_party/torch-xpu-ops/test/xpu/test_binary_ufuncs_xpu.py -k test_fmod_remainder_by_zero_integral_xpu_int64
Traceback (most recent call last):
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 3329, in wrapper
    method(*args, **kwargs)
  File "/tmp/xpu-tool/Python/3.10.19/x64/lib/python3.10/site-packages/torch/testing/_internal/common_device_type.py", line 428, in instantiated_test
    result = test(self, **param_kwargs)
[...]

PR #3401, opened 2026-04-20
[#2630] Override test_fmod_remainder_by_zero_integral to work with int64 on xpu

jmamzax:
Part of https://github.com/intel/torch-xpu-ops/issues/2630. Behavior of torch.fmod(input, other) for other equals int64(0) is undefined and is differently handled by xpu and cuda devices. XPU returns -1 for all input values, but cuda for positive values int64 returns 4294967295 and -1 for negative.


test/xpu/test_binary_ufuncs_xpu.py
```diff
@@ -12,16 +12,36 @@
...
+
+
+@dtypes(*integral_types())
+def _test_fmod_remainder_by_zero_integral(self, device, dtype):
+    fn_list = (torch.fmod, torch.remainder)
+    for fn in fn_list:
+        x = make_tensor((10, 10), device=device, dtype=dtype, low=-9, high=9)
+        zero = torch.zeros_like(x)
+        value = 255 if dtype == torch.uint8 else -1
+        self.assertTrue(torch.all(fn(x, zero) == value))
```
Silv3S:
```suggestion
    @dtypes(*integral_types())
    def test_fmod_remainder_by_zero_integral(self, device, dtype):
        fn_list = (torch.fmod, torch.remainder)
        for fn in fn_list:
            # check integral tensor fmod/remainder to zero
            x = make_tensor((10, 10), device=device, dtype=dtype, low=-9, high=9)
            zero = torch.zeros_like(x)
            # RuntimeError on CPU
            if self.device_type == "cpu":
                with self.assertRaisesRegex(RuntimeError, "ZeroDivisionError"):
                    fn(x, zero)
            elif torch.version.hip is not None:
                # ROCm behavior: x % 0 is a no-op; x is returned
                self.assertEqual(fn(x, zero), x)
            else:
                # CUDA behavior: Different value for different dtype
                # Due to it's an undefined behavior, CUDA returns a pattern of all 1s
                # for integral dividend (other than int64) divided by zero. For int64,
                # CUDA returns all 1s for negative dividend, half 1s for positive dividend.
                # uint8: 0xff -> 255
                # int32: 0xffffffff -> -1
                if torch.device(device).type == "cuda" and dtype == torch.int64:
                    self.assertEqual(fn(x, zero) == 4294967295, x >= 0)
                    self.assertEqual(fn(x, zero) == -1, x < 0)
                else:
                    value = 255 if dtype == torch.uint8 else -1
                    self.assertTrue(torch.all(fn(x, zero) == value))
```

I think the better fix is to add CUDA check for their custom behavior on running int64. This change is oneliner that will be easier to upstream

test/xpu/test_binary_ufuncs_xpu.py
```diff
@@ -12,16 +12,36 @@
...
+from torch.testing._internal.common_utils import make_tensor, run_tests
 
 try:
     from xpu_test_utils import XPUPatchForImport
 except Exception as e:
     from .xpu_test_utils import XPUPatchForImport
 
 with XPUPatchForImport(False):
-    from test_binary_ufuncs import TestBinaryUfuncs
+    from test_binary_ufuncs import integral_types, TestBinaryUfuncs
```
tszulist-hbn: Indirect import. Import `integral_types` directly from `torch.testing._internal.common_device_type` alongside `dtypes` and `instantiate_device_type_tests`.

tszulist-hbn (changes requested):

test/xpu/test_binary_ufuncs_xpu.py
```diff
@@ -23,6 +28,41 @@
...
+            # ROCm behavior: x % 0 is a no-op; x is returned
+            self.assertEqual(fn(x, zero), x)
+        elif self.device_type == "cuda" and dtype == torch.int64:
+            # CUDA behavior: Different value for different dtype
+            # Due to it's an undefined behavior, CUDA returns a pattern of all 1s
+            # for integral dividend (other than int64) divided by zero. For int64,
+            # CUDA returns all 1s for negative dividend, half 1s for positive dividend.
+            # uint8: 0xff -> 255
+            # int32: 0xffffffff -> -1
+            if dtype == torch.int64:
```
pbielak: The same condition is included in the `elif` block (i.e., `elif self.device_type == "cuda" and dtype == torch.int64:`)

Please check if the `else` block below is intended correctly

test/xpu/test_binary_ufuncs_xpu.py
```diff
@@ -23,6 +28,41 @@
...
+                self.assertEqual(fn(x, zero) == 4294967295, x >= 0)
+                self.assertEqual(fn(x, zero) == -1, x < 0)
+        else:
+            value = 255 if dtype == torch.uint8 else -1
+            self.assertTrue(torch.all(fn(x, zero) == value))
+
```
pbielak, line 59: The PR description tells that XPU returns always `-1`, but I don't see a separate `elif` block for XPU - is that correct?

pbielak (changes requested):

test/xpu/test_binary_ufuncs_xpu.py
Silv3S: Right, it's not needed

test/xpu/test_binary_ufuncs_xpu.py
Silv3S: Yes it's correct.  XPU goes through last `else` block which in upstream is:
```Python
else:
    # CUDA behavior: Different value for different dtype
    # Due to it's an undefined behavior, CUDA returns a pattern of all 1s
    # for integral dividend (other than int64) divided by zero. For int64,
    # CUDA returns all 1s for negative dividend, half 1s for positive dividend.
    # uint8: 0xff -> 255
    # int32: 0xffffffff -> -1
    if dtype == torch.int64:
        self.assertEqual(fn(x, zero) == 4294967295, x >= 0)
        self.assertEqual(fn(x, zero) == -1, x < 0)
    else:
        value = 255 if dtype == torch.uint8 else -1
        self.assertTrue(torch.all(fn(x, zero) == value))
```
CUDA has some special handling of `int64` which we don't, so the proper fix is to extract that special CUDA case into separate if branch and so XPU is checked by the default assert
```Python
elif self.device_type == "cuda" and dtype == torch.int64:
    # CUDA behavior: Different value for different dtype
    # Due to it's an undefined behavior, CUDA returns a pattern of all 1s
    # for integral dividend (other than int64) divided by zero. For int64,
    # CUDA returns all 1s for negative dividend, half 1s for positive dividend.
    # uint8: 0xff -> 255
    # int32: 0xffffffff -> -1
    self.assertEqual(fn(x, zero) == 4294967295, x >= 0)
    self.assertEqual(fn(x, zero) == -1, x < 0)
else:
    value = 255 if dtype == torch.uint8 else -1
    self.assertTrue(torch.all(fn(x, zero) == value))
```

pbielak: Makes sense, thanks!

pbielak (approved):

test/xpu/test_binary_ufuncs_xpu.py
```diff
@@ -12,8 +12,13 @@
...
 
-from torch.testing._internal.common_device_type import instantiate_device_type_tests
-from torch.testing._internal.common_utils import run_tests
+import torch
+from torch.testing._internal.common_device_type import (
+    dtypes,
+    instantiate_device_type_tests,
+)
+from torch.testing._internal.common_dtype import integral_types
+from torch.testing._internal.common_utils import make_tensor, run_tests
```
BBBela:
```suggestion
from torch.testing import make_tensor
from torch.testing._internal.common_utils import run_tests
```
In most upstream tests the `make_tensor` is imported from the public `torch.testing` API instead of internal `common_utils`. Importing from `common_utils` is indirect import as `torch.testing._internal.common_utils` imports the `make_tensor` exactly from `torch.testing`. Also in the upstream `test_binary_ufuncs.py` the `make_tensor` is imported from `torch.testing`. I think we should stay aligned to upstream as much as possible to make the future test cases porting to PyTorch easier and simpler.

BBBela (commented):
Beside this one import comment below, everything looks good 😉

BBBela (approved):
Thanks!

tszulist-hbn (approved):

PR #3413, opened 2026-04-21
fused Adam/AdamW: in-kernel mixed-precision support & remove unused tests

### LINKED ISSUE #3267: New failed test cases 2026-04-06
### 🐛 Describe the bug

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_hook_skips_existing_state_amsgrad_False_Adam_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_hook_skips_existing_state_amsgrad_True_AdamW_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_hook_skips_existing_state_amsgrad_True_Adam_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_numerics_AdamW_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_numerics_Adam_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_state_init_amsgrad_False_AdamW_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_state_init_amsgrad_False_Adam_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_state_init_amsgrad_True_AdamW_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_state_init_amsgrad_True_Adam_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_fused_mixed_precision_hook_skips_existing_state_amsgrad_False_AdamW_xpu_float32
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestBasicVitalSigns,test_basic_vitals
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestBasicVitalSigns,test_basic_vitals_read_write
op_ut,third_party.torch-xpu-ops.test.xpu.test_torch_xpu.TestBasicVitalSigns,test_dataloader_vitals

### Versions

2.12

PatrykWilczewski:
Add mixed-precision support for fused Adam and AdamW on XPU, matching CUDA's behavior. Previously, fused Adam/AdamW on XPU required strict fast-path dtype alignment across params, grads, and states, so mixed-state scenarios (e.g. fp32 params with bf16 optimizer states) failed with runtime errors.
This PR handles the bf16 <-> fp32 conversion entirely inside the SYCL kernel (FusedAdamMathFunctorMP), mirroring CUDA's 'FusedAdamMathFunctorMP'. There is no host-side tensor allocation or copy per step, so the mixed-precision path is safe under graph capture. The supported configuration is fp32 params/grads + bf16 states, validated by 'validate_fused_mixed_precision_dtypes'; other dtype combinations are rejected with a clear error.
Routing is keyed on dtype (params vs states), and 'check_fast_path_restrictions' is enforced with "skip-cross_list_dtype_check' so device/layout/contiguity mismatches surface their own diagnostic instead of misleading dtype error.

Fixes: https://github.com/intel/torch-xpu-ops/issues/3267


src/ATen/native/xpu/FusedAdam.cpp
```diff
@@ -23,14 +27,48 @@
...
+    ScalarType exp_avg_sq_dtype,
+    std::optional<ScalarType> max_exp_avg_sq_dtype) {
+  auto math_dtype = promoteTypes(param_dtype, grad_dtype);
+  math_dtype = promoteTypes(math_dtype, exp_avg_dtype);
+  math_dtype = promoteTypes(math_dtype, exp_avg_sq_dtype);
+  if (max_exp_avg_sq_dtype.has_value()) {
+    math_dtype = promoteTypes(math_dtype, *max_exp_avg_sq_dtype);
+  }
+  return isFloatingType(math_dtype) ? toOpMathType(math_dtype) : math_dtype;
+}
```
kdrozd-dev: This batch of functions is duplicated in both .cpp files couldn't it be moved to a common header to avoid duplication?

kdrozd-dev (changes requested):

src/ATen/native/xpu/FusedAdam.cpp
PatrykWilczewski: Done

kdrozd-dev (approved):

src/ATen/native/xpu/FusedAdamMixedPrecisionUtils.h
```diff
@@ -0,0 +1,53 @@
...
+  return found_inf.has_value() && found_inf->item<double>() != 0.0;
+}
+
+inline double grad_scale_value(const std::optional<at::Tensor>& grad_scale) {
+  return grad_scale.has_value() && grad_scale->is_cpu()
+      ? grad_scale->item<double>()
+      : 1.0;
+}
+
+inline bool can_cast_without_narrowing(ScalarType from, ScalarType to) {
```
etaf: CUDA’s mixed-precision fused Adam/AdamW path only supports `float32` params/grads with `bfloat16` optimizer states and explicitly rejects other mixed configurations in `aten/src/ATen/native/cuda/fused_adam_utils.cuh:29-70`. This PR therefore silently enables unvalidated combinations such as FP32 params with FP16 states by casting through temporary FP32 tensors and copying back. The XPU path should enforce the same dtype contract as CUDA, or the PR needs a documented and tested reason to intentionally support additional mixed-state configurations.

src/ATen/native/xpu/FusedAdamMixedPrecisionUtils.h
```diff
@@ -0,0 +1,53 @@
...
+#include <ATen/OpMathType.h>
+#include <ATen/core/Tensor.h>
+
+#include <c10/core/ScalarType.h>
+
+#include <optional>
+
+namespace at::native {
+
+inline bool found_inf_nonzero(const std::optional<at::Tensor>& found_inf) {
```
etaf: Here reads `found_inf` with `.item<double>()` before the mixed/fallback paths run, and the fallback implementations read device `state_steps` and device tensor LR on the host at `src/ATen/native/xpu/FusedAdam.cpp:253`, `src/ATen/native/xpu/FusedAdam.cpp:501`, `src/ATen/native/xpu/FusedAdamW.cpp:254`, and `src/ATen/native/xpu/FusedAdamW.cpp:501`. For fused optimizers, `found_inf`, `state_steps`, and capturable/tensor LR are device-side scalars specifically to avoid host synchronization; CUDA keeps these as device pointers and reads them in-kernel. These host reads defeat the fused AMP path’s async behavior and can break graph/capturable usage when the fallback is selected. The mixed/fallback paths should keep these scalars device-side or reject unsupported capturable/device-scalar cases instead of synchronizing.

etaf (changes requested):
The dtype contract and host-synchronization issues affect semantic parity and XPU fused optimizer behavior, and please enable or created the corresponding UT in this PR if they are skipped.

src/ATen/native/xpu/FusedAdam.cpp
```diff
@@ -152,28 +374,140 @@ void _fused_adam_kernel_xpu_(
...
+            state_steps,
+            lr,
+            beta1,
+            beta2,
+            weight_decay,
+            eps,
+            amsgrad,
+            maximize,
+            grad_scale,
+            found_inf);
```
CuiYifeng: Some copies will be wasted if `try_mixed_precision_fused_adam_xpu_` returns false, right? Can we divide inspection and execution into two parts？

src/ATen/native/xpu/FusedAdamW.cpp
```diff
@@ -98,35 +191,179 @@ void _fused_adamw_kernel_xpu_(
...
+          state_steps,
+          lr,
+          beta1,
+          beta2,
+          weight_decay,
+          eps,
+          amsgrad,
+          maximize,
+          grad_scale,
+          found_inf);
```
CuiYifeng: Ditto.

src/ATen/native/xpu/FusedAdam.cpp
```diff
@@ -8,7 +8,12 @@
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
+#include <ATen/OpMathType.h>
```
CuiYifeng: `OpMathType.h` has been included in `FusedAdamMixedPrecisionUtils.h`.

src/ATen/native/xpu/FusedAdamW.cpp
```diff
@@ -8,7 +8,12 @@
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
+#include <ATen/OpMathType.h>
```
CuiYifeng: Included in `FusedAdamMixedPrecisionUtils.h`.

CuiYifeng (changes requested):

src/ATen/native/xpu/FusedAdam.cpp
PatrykWilczewski: Reworked in the latest push. The old try_mixed_precision_fused_adam_xpu_ is gone. Inspection and execution are now separate: checK-fast_path_restrictions picks the path, and inside the mixed helper validate_fused_mixed_precision_dtypes rejects unsupported dtype combos via TORCH_CHECK before any copy is made. The helper returns void and only ever runs when it will complete, so there are no wasted copies. The call site is now a simple if (fast path) { ... } else { run_mixed_precision_...(); }.

src/ATen/native/xpu/FusedAdamW.cpp
PatrykWilczewski: Same rework as in FusedAdam.cpp

src/ATen/native/xpu/FusedAdam.cpp
PatrykWilczewski: Removed

src/ATen/native/xpu/FusedAdamW.cpp
PatrykWilczewski: Removed

src/ATen/native/xpu/FusedAdamMixedPrecisionUtils.h
PatrykWilczewski: Fixed. The can_cast_without_narrowing / promoteTypes logic that silently enabled unvalidated combos (e.g. fp32 params + fp16 states) is gone. The header now exposes validate_fused_mixed_precision_dtypes(), a direct port of CUDA's validate_mixed_precision_dtypes (fused_adam_utils.cuh): it TORCH_CHECKs params/grads == float32 and exp / exp_avg_sq / max_exp_avg_sq == bfloat16, rejecting anything else.
Verified: fp16 states now raise "_fused_adam with mixed dtype requires bfloat16 optimizer states, got Half".

src/ATen/native/xpu/FusedAdamMixedPrecisionUtils.h
PatrykWilczewski: Fixed by removing the eager fallback that caused these host reads. The non-fast-path now only handles the CUDA-supported fp32/bf16 mixed case. The only remaining .item() is lr.item() strictly inside the pre-existing 'if (lr.is_cpu())' branch, which reads a CPU-resident scalar (same as CUDA) and is not a device sync.

PatrykWilczewski:
> The dtype contract and host-synchronization issues affect semantic parity and XPU fused optimizer behavior, and please enable or created the corresponding UT in this PR if they are skipped.

The dtype contract now matches CUDA exactly (fp32 params/grads + bf16 states, everything else rejected via TORCH_CHECK), and the host-synchronizing eager fallback has been removed.
I also re-enabled the three previously @onlyCUDA test_fused_mixed_precision_* tests for XPU in test_optim_xpu.py. All 10 tests pass on Arc B580; 144 existing fused/state_dict tests still pass with no regression

src/ATen/native/xpu/FusedAdam.cpp
```diff
@@ -39,17 +45,58 @@ void _fused_adam_kernel_xpu_(
...
+  std::vector<at::Tensor> exp_avg_sqs_fp32;
+  std::vector<at::Tensor> max_exp_avg_sqs_fp32;
+  exp_avgs_fp32.reserve(exp_avgs.size());
+  exp_avg_sqs_fp32.reserve(exp_avg_sqs.size());
+  if (amsgrad) {
+    max_exp_avg_sqs_fp32.reserve(max_exp_avg_sqs.size());
+  }
+
+  for (const auto i : c10::irange(params.size())) {
+    exp_avgs_fp32.emplace_back(exp_avgs[i].to(param_dtype));
```
etaf: run_mixed_precision_fused_adam_xpu_ allocates new tensors via exp_avgs[i].to(param_dtype) and copies results back via exp_avgs[i].copy_(exp_avgs_fp32[i]) for every step. CUDA and XPU graph capture require a fixed allocation graph — dynamic at::Tensor allocation during a captured region causes capture failure or silent correctness bugs. When fused=True is combined with a tensor lr (which implies capturable usage), this path will silently produce incorrect updates.

  CUDA resolves this by handling the bf16↔fp32 casting entirely inside the GPU kernel (FusedAdamMathFunctorMP) without any host-side allocation. The XPU path would need a proper mixed-precision SYCL kernel like CUDA's FusedAdamMathFunctorMP.

src/ATen/native/xpu/FusedAdam.cpp
```diff
@@ -135,42 +261,59 @@ void _fused_adam_kernel_xpu_(
...
-        weight_decay,
-        eps,
-        maximize,
-        grad_scale,
-        found_inf);
+  const bool fast_path = amsgrad
+      ? at::native::check_fast_path_restrictions(
+            {params, grads, exp_avgs, exp_avg_sqs, max_exp_avg_sqs})
+      : at::native::check_fast_path_restrictions(
+            {params, grads, exp_avgs, exp_avg_sqs});
```
etaf: In both FusedAdam.cpp and FusedAdamW.cpp, !fast_path routes to run_mixed_precision_fused_adam_xpu_. But check_fast_path_restrictions returns false for layout
  mismatches, device mismatches, and non-contiguous tensors — not only dtype mismatches. Any such tensor combination will silently fall into the mixed-precision path
  and then fail with a misleading "requires bfloat16 optimizer states" error rather than the real diagnosis.

etaf (changes requested):

src/ATen/native/xpu/FusedAdam.cpp
PatrykWilczewski: Addressed. Ported CUDA's in-kernel approach to SYCL: new FusedAdamMathFunctorMp casts bf16 <-> fp32 entirely inside the kenrel via load_store_cast / load_args_mixed_prec / store_args_cast. There is no host-side at::Tensor allocation or .copy_() per step, so the fixed-allocation requirement for graph capture is satisfied. Found_inf, state_steps, grad_scale, and tensor lr are all read device-side, matching CUDA's 'FusedAdamMathFunctorMP'. The host side upcast/copy-back wrapper is removed. Verified: fused=True with a tensor lr across Adam/AdamW x amsgrad produces correct fp32 params + bf16 states

src/ATen/native/xpu/FusedAdam.cpp
PatrykWilczewski: Fixed. Routing now keys on dtype (params[0].scalar_type(0 != exp_avgs[0].scalar_type()) rather than !check_fast_path_restrictions, and wraps check_fast_path_restrictions in a TORCH_CHECK with skip_cross_list_dtype_check=is_mixed_precision (CUDA parity, FusedAdamkernel.cu:32-64). A non-contiguous/non-dense or device/layout-mismatched fp32 input now raises "must have same dtype, devie and layout" instead of the misleading "requires bfloat16" message. Verified: strided fp32 -> layout error; dense fp32 -> runs.

test/xpu/test_torch_xpu.py
```diff
@@ -240,61 +239,6 @@ def my_should_stop_test_suite(self):
 )
 
 
-@contextlib.contextmanager
```
etaf, line 243: Please exclude these non PR related change.

etaf (changes requested):
Seems new changes doesn't be pushed.

PatrykWilczewski:
> Seems new changes doesn't be pushed.

You are right. Now it is pushed.

src/ATen/native/xpu/sycl/FusedAdamUtils.h
```diff
@@ -208,4 +255,210 @@ struct FusedAdamMathFunctor {
...
+    typename scalar_type,
+    typename param_type,
+    typename grad_type,
+    typename exp_avg_type,
+    typename exp_avg_sq_type,
+    typename max_exp_avg_sq_type,
+    int depth,
+    ADAM_MODE adam_mode,
+    bool amsgrad>
+struct FusedAdamMathFunctorMP {
```
etaf, line 268: The aligned fast path in `FusedAdamMathFunctorMP` is entirely scalar, defeating its own purpose. `load_store_cast`  is an element-wise loop, whereas `load_store` does a single aligned_vector`<T, kILP>` (128-bit) access. The MP functor calls load_store_cast for all five tensors — but param_type and grad_type are float, i.e. identical to scalar_type, so those four loads and two stores could use the vectorized path unchanged. CUDA does exactly this: fused_adam_utils.cuh guards each with if constexpr (!std::is_same_v<scalar_type, param_type>) and falls through to plain load_store when the types match.

src/ATen/native/xpu/sycl/FusedAdamUtils.h
```diff
@@ -208,4 +255,210 @@ struct FusedAdamMathFunctor {
...
+          store_args_cast(
+              max_exp_avg_sq_args,
+              r_args[kMaxExpAvgSqIdx],
+              i_start,
+              chunk_size,
+              n,
+              item_id,
+              local_range);
+        }
+        if (grad_scale_ptr) {
```
etaf, line 449: `optim_inputs_func` never sets `grad_scale`, so the branch never runs. fp32 params + bf16 states is exactly the AMP scenario — add a case that passes `grad_scale`/`found_inf` and checks grads are written back unscaled.

etaf (changes requested):
Please also fix the two issues copilot found.

src/ATen/native/xpu/sycl/FusedAdamUtils.h
```diff
@@ -22,6 +25,50 @@ namespace at::native::xpu {
...
+    at::TensorList params,
+    at::TensorList grads,
+    at::TensorList exp_avgs,
+    at::TensorList exp_avg_sqs,
+    const char* op_name) {
+  TORCH_CHECK(
+      params[0].scalar_type() == at::kFloat,
+      op_name,
+      " with mixed dtypes requires float32 params, got ",
+      params[0].scalar_type());
```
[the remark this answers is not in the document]
PatrykWilczewski: Added `test_fused_mixed_precision_rejects_unsupported_dtypes`, parametrized over six named dtype
cases (`_MP_REJECT_CASES`) and over `Adam`/`AdamW` — 12 instantiations, all passing.
Each case asserts the **whole** diagnostic rather than a fragment:

```python
msg = f"{op_name} with mixed dtypes {expected}"
with self.assertRaisesRegex(RuntimeError, msg):
    op(...)
```

so the op-name prefix, the `with mixed dtypes` qualifier and the per-tensor-list wording are all
pinned, which is what makes the *diagnostics* regression-proof and not just the guards:

| case | param | grad | exp_avg | exp_avg_sq | max_exp_avg_sq | asserted message |
|---|---|---|---|---|---|---|
| `param_f16` | f16 | f32 | bf16 | bf16 | -| `<op> with mixed dtypes requires float32 params` |
| `param_bf16` | bf16 | f32 | f32 | f32 | -| `… requires float32 params` |
| `grad_bf16` | f32 | bf16 | bf16 | bf16 | -| `… requires float32 grads` |
| `exp_avg_f16` | f32 | f32 | f16 | bf16 | -| `… requires bfloat16 optimizer states` |
| `exp_avg_sq_f16` | f32 | f32 | bf16 | f16 | -| `… requires bfloat16 optimizer states` |
| `max_exp_avg_sq_f16` | f32 | f32 | bf16 | bf16 | f16 | `… requires bfloat16 max_exp_avg_sqs` |

That covers all four categories, plus `exp_avg` for symmetry.

test/xpu/test_optim_xpu.py
```diff
@@ -191,6 +191,167 @@ def _test_peak_memory_foreach(self, device, dtype, optim_info):
 TestOptimRenewed.test_peak_memory_foreach = _test_peak_memory_foreach
 
 
+@parametrize("amsgrad", [False, True])
+@optims(
+    [o for o in optim_db if o.optim_cls.__name__ in ["Adam", "AdamW"]],
+    dtypes=[torch.float32],
+)
+def _test_fused_mixed_precision_state_init(self, device, dtype, optim_info, amsgrad):
```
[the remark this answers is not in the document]
PatrykWilczewski: Removed all three (`_test_fused_mixed_precision_state_init`, `_test_fused_mixed_precision_hook_skips_existing_state`, `_test_fused_mixed_precision_numerics`) together with their `TestOptimRenewed` assignments, and dropped the now - unused `TensorTracker` import. `_bf16_state_init_hook` and `parametrize` are kept - both are still used by the new `grad_scale` test added for the other review comment.

One correction to the mechanism, because it decides whether the coverage really is preserved: upstream decorates these three with `@onlyAccelerator` + `@skipMPS`, **not** `@onlyCUDA` (`test/test_optim.py:2519-2526`, `:2559-2566`, `:2617-2623`), so no `XPUPatchForImport` decorator remapping is involved. `onlyAccelerator` (`torch/testing/_internal/common_device_type.py:2173`) skips only `cpu` and `meta`, so the upstream methods already execute under `instantiate_device_type_tests(TestOptimRenewed, …, only_for="xpu", allow_xpu=True)`.

So the conclusion holds and is stronger than stated: the local copies were shadowing tests that were already running on XPU. Confirmed by re-running the three upstream test names after the removal - `state_init` (4 cases), `hook_skips_existing_state` (4) and `numerics` (2) all **pass on XPU**, none skipped. Supporting detail: those three tests and `_bf16_state_init_hook` all arrived upstream in pytorch #175230 ("Initial version of mixed precision optimizers", 2026-04-12), and torch-xpu-ops has no skip or expected-failure entry matching `fused_mixed_precision` - so they were never skipped here, they were simply failing until this PR supplies the XPU mixed-precision path.

src/ATen/native/xpu/sycl/FusedAdamUtils.h
PatrykWilczewski: Fixed

src/ATen/native/xpu/sycl/FusedAdamUtils.h
PatrykWilczewski: Added `test_fused_mixed_precision_grad_scale`, parametrized over `amsgrad`. It uses fp32 params with bf16 states via `_bf16_state_init_hook`, pre-scales the grads by 128.0, and sets `grad_scale` / `found_inf` as optimizer **attributes**

src/ATen/native/xpu/sycl/FusedAdamUtils.h
```diff
@@ -22,6 +25,50 @@ namespace at::native::xpu {
...
+inline void validate_fused_mixed_precision_dtypes(
+    at::TensorList params,
+    at::TensorList grads,
+    at::TensorList exp_avgs,
+    at::TensorList exp_avg_sqs,
+    const char* op_name) {
+  TORCH_CHECK(
+      params[0].scalar_type() == at::kFloat,
+      op_name,
+      " with mixed dtypes requires float32 params, got ",
```
etaf, line 37: These messages no longer match CUDA. CUDA's `validate_mixed_precision_dtypes` produces `Mixed-precision fused Adam requires float32 params, got ...` (and `Mixed-precision fused AdamW ...`), while XPU produces `_fused_adam with mixed dtypes requires float32 params, got ...`. The point of this PR is CUDA parity, and anything that matches on the text (the new tests here, or a shared upstream test later) would then need a separate XPU branch.

I'd keep the helper as it is and pass the same strings CUDA uses from the call sites in the four `Fused*Kernels.cpp` files: `"Mixed-precision fused Adam"` / `"Mixed-precision fused AdamW"`, and drop the ` with mixed dtypes` infix.

test/xpu/test_optim_xpu.py
```diff
@@ -191,6 +191,183 @@ def _test_peak_memory_foreach(self, device, dtype, optim_info):
...
+    ),
+}
+
+
+@parametrize("case", list(_MP_REJECT_CASES))
+@optims(
+    [o for o in optim_db if o.optim_cls.__name__ in ["Adam", "AdamW"]],
+    dtypes=[torch.float32],
+)
+def _test_fused_mixed_precision_rejects_unsupported_dtypes(
```
etaf: Neither of the two new tests (this one and `_test_fused_mixed_precision_grad_scale`) touches anything XPU-specific. Once the error strings match CUDA, they would pass as-is on CUDA, and upstream `test/test_optim.py` has no rejection test and no grad_scale test for the mixed-precision path. Their natural home is next to `test_fused_mixed_precision_numerics` in `test_optim.py` under `@onlyAccelerator`, which is where the three existing mixed-precision tests already run for XPU. If they stay here only, they will drift from upstream and CUDA won't get the coverage. Can you send them upstream, or at least open a follow-up to do so?

Smaller point on this test: `@optims(..., dtypes=[torch.float32])` is only used to choose between `_fused_adam_` and `_fused_adamw_`. `dtype` is never used and `optim_info` is only read for its class name. `@parametrize("op", [torch._fused_adam_, torch._fused_adamw_])` would say the same thing without the unused parameters.

src/ATen/native/xpu/FusedAdam.cpp
```diff
@@ -118,8 +132,10 @@ void _fused_adam_kernel_xpu_(
     return;
   }
 
-  // Manually check devices since we specify no device check in
-  // native_functions.yaml
+  if (params.empty()) {
```
etaf: This deletes the `// Manually check devices since we specify no device check in native_functions.yaml` comment above the device checks. It still applies (CUDA's `FusedAdamKernel.cu` keeps it), and removing it isn't needed for this change. Please keep it above `Device param_device = ...`.

src/ATen/native/xpu/sycl/ForeachFunctors.h
```diff
@@ -83,6 +83,101 @@ void store_args(
...
+    SrcT* src,
+    int64_t dst_offset,
+    int64_t src_offset) {
+  if constexpr (std::is_same_v<DstT, SrcT>) {
+    load_store(dst, src, dst_offset, src_offset);
+  } else {
+#pragma unroll
+    for (int ii = 0; ii < kILP; ++ii) {
+      dst[dst_offset * kILP + ii] =
+          static_cast<DstT>(src[src_offset * kILP + ii]);
```
etaf, line 99: When the types differ, this does element-wise global loads and stores. CUDA's `FusedAdamMathFunctorMP` casts element-wise into a local `kILP` array too, but then goes through `load_store` for the global side, so at least the bf16 stores (`exp_avg`, `exp_avg_sq`, `max_exp_avg_sq`) are vectorized aligned writes. Something like this would do the same here:

```cpp
} else {
  DstT tmp[kILP];
#pragma unroll
  for (int ii = 0; ii < kILP; ++ii) {
    tmp[ii] = static_cast<DstT>(src[src_offset * kILP + ii]);
  }
  load_store(dst, tmp, dst_offset, 0);
}
```

I haven't measured this on XPU, and IGC may already combine the scalar stores. A quick before/after on a large bf16-state step would show whether it matters.

etaf (changes requested):
Thanks, we're almost there.

PR #3414, opened 2026-04-21
test_transformers_xpu: skip test_disable_fastpath in transformer suite

BartoszKokoszko:
TestTransformersXPU::test_disable_fastpath_xpu currently fails because it asserts MHA fastpath invocation. MHA fast would be deprecated. Add an explicit XPU skip on test_disable_fastpath in test_transformers_xpu.py.

### DIFF
--- a/test/xpu/test_transformers_xpu.py
+++ b/test/xpu/test_transformers_xpu.py
@@ -1900,6 +1900,7 @@ class TestTransformers(NNTestCase):
         torch.jit.script(mha)
 
     @unittest.skipIf(TEST_WITH_CROSSREF, "Fastpath not available with crossref")
+    @skipIfXpu(msg="MHA fastpath is not available on XPU")
     @torch.no_grad()
     def test_disable_fastpath(self, device):
         def _test_te_fastpath_called(


test/xpu/test_transformers_xpu.py
```diff
@@ -1900,6 +1900,7 @@ def test_script_mha_in_proj_weight_none(self):
         torch.jit.script(mha)
 
     @unittest.skipIf(TEST_WITH_CROSSREF, "Fastpath not available with crossref")
+    @unittest.skipIf(TEST_XPU, "MHA fastpath would be deprecated on XPU")
```
Silv3S:
```suggestion
    @skipIfXpu(msg="MHA fastpath is not available on XPU")
```
I think it was never enabled on XPU, so current comment is misleading

Silv3S:
It doesn't make sense to run all CI tests for single test skip. Please use `disable_*` labels to reduce CI overload

test/xpu/test_transformers_xpu.py
BartoszKokoszko: done

Silv3S (approved):

kdrozd-dev (approved):

guangyey (approved):

guangyey (commented):
Refer to https://github.com/pytorch/pytorch/pull/180829#issuecomment-4286320190

PR #3429, opened 2026-04-22
[CI] Add snapshotting skipped issues before build start

### LINKED ISSUE #3470: [CI] False UT failures when skipped issue is closed mid-run after fix is merged into main.
## Summary

Linux CI UT summary can produce false failures because the list of skipped test cases from issues with the `skipped` label is read only at summary time, while build + UT can run for hours.

I already prepared PR with proposal of fix: #3429 
## Problem

The UT result check currently depends on live GitHub issue state at the moment the summary job runs.
Because issues with the `skipped` label can be closed during a long workflow run, the effective allowlist can drift from the state that existed when the build started.

This creates a race between:
- build start time
- UT summary time
- external issue state changes on GitHub

## Why this is incorrect

UT failures should be evaluated against a stable baseline for that run.
When baseline data changes mid-run, the same test result can be interpreted differently depending on when summary executes.

## False regression when issue closes mid-run

The most common trigger is fix merging to `torch-xpu-ops` main:

1. An issue with the `skipped` label is open when the CI build starts — the build does not include the fix.
2. During the multi-hour build+UT run, the fix is merged into main and the issue is closed immediately on GitHub.
3. The UT fails for exactly that known upstream reason (the fix is not in the build, it started too early).
4. At summary time, the issue is already closed, so summary no longer treats the failure as expected and marks the run as failed.

**The key point:** the build was never supposed to contain the fix — it started before fix for the issue was added — but the issue closure outraces the summary, producing a false failure.

## Impact

- Flaky CI outcomes for long-running Linux pipelines.
- Reduced trust in UT summary signal.
- Extra reruns/manual triage for non-code regressions.
- Need to manually search and find if some `skipped`-labeled issue was closed mid-run, reopening it for summary rerun and closing after.

BBBela:
Fix for: #3470 

## Summary

This PR adjusts Linux CI skipped-issues handling to use a snapshot+live merge model for UT summary allowlist construction.

Detailed problem description in the issue: #3470 

## What This PR Changes

1. Add a pre-build snapshot step to save the state of the `skipped` issues just before the build starts.
2. Download and use the snapshot during UT summary.
3. Merge snapshot data with current live open `skipped` issues when building the list of skipped test cases in the summary.

## Reasoning for This Design

- Snapshot preserves build-start context for the running build.
- Live fetch at summary time includes newly opened valid `skipped` entries.
- Union of both sources provides stable and up-to-date allowlist semantics for long-running workflows.

## Resulting Behavior

UT summary now evaluates results against:

- build-start snapshot state
- current live open issues with label `skipped`

This means:

- entries that existed at build start remain represented even if later closed
- entries opened during the run are also represented

## Testing

I have validated and confirmed the correct behavior in the On-demand [WORKFLOW](https://github.com/intel/torch-xpu-ops/actions/runs/24987002166/job/73208894011)
The behavior is as expected.


.github/workflows/_linux_build.yml
```diff
@@ -53,9 +53,35 @@ jobs:
...
+          name: Skipped-Issues-Snapshot-${{ github.event.pull_request.number || github.sha }}-linux-build-start
+          path: |
+            known_issue_static.log
+            issues_snapshot.log
+
   build:
     name: ${{ inputs.pytorch }}
-    needs: runner
+    needs: [runner, snapshot-skipped-issues]
     if: ${{ ! endsWith(inputs.pytorch, '_wheel') }}
```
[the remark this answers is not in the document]
BBBela: I added change that creates empty files first and tolerates the gh failures.

.github/workflows/_linux_build.yml
```diff
@@ -53,9 +53,35 @@ jobs:
...
+      issues: read
+    env:
+      GH_TOKEN: ${{ github.token }}
+      UT_SKIP_ISSUE: 1624
+    steps:
+      - name: Snapshot open skipped issues at build start
+        run: |
+          # Capture static distributed known issues baseline.
+          gh --repo intel/torch-xpu-ops issue view $UT_SKIP_ISSUE --json body -q .body \
+            | sed -E '/^(#|$)/d' > known_issue_static.log || touch known_issue_static.log
```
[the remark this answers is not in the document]
BBBela: I added change that creates empty files first and tolerates the gh failures.

.github/workflows/_linux_ut.yml
```diff
@@ -165,15 +171,20 @@ jobs:
...
+          #   2. Current live state (covers new issues opened after build start)
+          snapshot_dir="${{ github.workspace }}/issues_snapshot"
+          touch issues.log Known_issue.log.tmp
+          echo "Restoring snapshotted skipped issues if available."
+          cp "${snapshot_dir}/known_issue_static.log" Known_issue.log.tmp 2>/dev/null || true
+          cp "${snapshot_dir}/issues_snapshot.log" issues.log 2>/dev/null || true
+          echo "Fetching current UT_SKIP_ISSUE baseline from GitHub."
+          gh --repo intel/torch-xpu-ops issue view $UT_SKIP_ISSUE --json body -q .body \
+            | sed -E '/^(#|$)/d' >> Known_issue.log.tmp
+          echo "Fetching current open skipped issues from GitHub."
```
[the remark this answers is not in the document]
BBBela: That was the previous behavior, not my changes. I am just reusing what already existed.

.github/workflows/_linux_build.yml
```diff
@@ -53,9 +53,36 @@ jobs:
         id: runner-info
         uses: ./.github/actions/get-runner
 
+  snapshot-skipped-issues:
+    runs-on: ubuntu-24.04
```
pbielak: add `container: image: ubuntu:24.04` as suggested in https://github.com/intel/torch-xpu-ops/pull/3438

.github/workflows/_linux_build.yml
```diff
@@ -53,9 +53,36 @@ jobs:
...
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      issues: read
+    env:
+      GH_TOKEN: ${{ github.token }}
+      UT_SKIP_ISSUE: 1624
+    steps:
+      - name: Snapshot open skipped issues at build start
```
pbielak: The code is duplicated also in `_linux_ut.yml` - consider extracting it into an action

.github/workflows/_linux_ut.yml
```diff
@@ -152,6 +152,11 @@ jobs:
         with:
           pattern: Inductor-XPU-UT-Data-${{ github.event.pull_request.number || github.sha }}-${{ inputs.ut }}-*
           path: ${{ github.workspace }}/ut_log
+      - name: Download skipped issues snapshot
+        uses: actions/download-artifact@v4
+        with:
+          name: Skipped-Issues-Snapshot-${{ github.event.pull_request.number || github.sha }}-linux-build-start
+          path: ${{ github.workspace }}/issues_snapshot
```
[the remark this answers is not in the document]
BBBela: Added `continue-on-error: true`

.github/workflows/_linux_ut.yml
```diff
@@ -165,15 +170,17 @@ jobs:
...
-              --jq '.[] | select(.pull_request == null) | "Issue #\(.number): \(.title)\n\(.body)\n"' > issues.log
-          fi
+          # Build the known-issues list by merging two sources:
+          #   1. Snapshot captured at build start (covers issues closed mid-run)
+          #   2. Current live state (covers new issues opened after build start)
+          snapshot_dir="${{ github.workspace }}/issues_snapshot"
+          touch issues.log Known_issue.log.tmp
+          echo "Restoring snapshotted skipped issues if available."
+          cp "${snapshot_dir}/known_issue_static.log" Known_issue.log.tmp
+          cp "${snapshot_dir}/issues_snapshot.log" issues.log
```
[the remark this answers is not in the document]
BBBela: Added `|| true` not to fail if the files do not exist.

.github/scripts/fetch_issues.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
+UT_SKIP_ISSUE=1624
```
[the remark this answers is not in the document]
BBBela: Removed the `env: UT_SKIP_ISSUE` as it is no longer used in the .yml file. Only in the script file.

BBBela:
Thank you @pbielak for the review!
Your feedback made me think about the design and I refactored the PR a little. Here are the refactoring changes:
- Moved the duplicated code into separate reusable bash script with improved error logging.
- Refactored the separate snapshot job into two steps in the build job.

.github/workflows/_linux_build.yml
BBBela: Inspired by this suggestion, I did a little refactor and extracted duplicated code but into separate bash script instead of separate action. Let me know what you think about this new approach?

.github/workflows/_linux_build.yml
BBBela: After refactor there is no new job `snapshot-skipped-issues`. Its functional behavior is now moved into build job, so it will be executed on the same runner as build.

.github/workflows/_linux_build.yml
pbielak: The bash script seems to be a good fit here

.github/scripts/fetch_issues.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
```
pbielak, line 5: I would add a check that makes sure that `GITHUB_REPOSITORY` is set when calling this script

.github/scripts/fetch_issues.sh
```diff
@@ -0,0 +1,31 @@
...
+if [[ "$1" == "static" ]]; then
+  fetch_function=fetch_static
+elif [[ "$1" == "open" ]]; then
+  fetch_function=fetch_open
+else
+  printf 'Error: invalid mode "%s". Expected "static" or "open".\n' "$1" >&2
+  exit 1
+fi
+
+if ! output=$("$fetch_function"); then
```
pbielak: The error message suggest that only `gh` can fail, but `sed` and `jq` might be also potential failure points. Wondering if `set -e` combined with `set -x` wouldn't be a better option rather than an explicit error message (we could see at which point it failed, but on the other hand we would "pollute" the job logs) - WDYT @BBBela ?

.github/workflows/_linux_build.yml
```diff
@@ -93,6 +93,20 @@ jobs:
...
+          bash torch-xpu-ops/.github/scripts/fetch_issues.sh open > issues_snapshot.log || true
+      - name: Upload skipped issues snapshot
+        if: ${{ ! cancelled() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: Skipped-Issues-Snapshot-${{ github.event.pull_request.number || github.sha }}-linux-build-start
+          path: |
+            known_issue_static.log
+            issues_snapshot.log
+        continue-on-error: true
```
pbielak: The `continue-on-error` combined with the `|| true` syntax mean that we accept scenarios where the know issues cannot be collected before build time. In such a case the whole execution "degrades" back to the current situation, where some tests might have been closed in the meantime.

I would vote for keeping this as an integral step, which should fail the whole pipeline if something goes wrong.

.github/scripts/fetch_issues.sh
BBBela: That makes sense. I adjusted the script to use `set -xe` and removed the explicit error log. I left just the one error log related to wrong input mode.

.github/scripts/fetch_issues.sh
BBBela: As for the `GITHUB_REPOSITORY`, from what I know it is always present in the actions workflow and the script is rather specific, not general-purpose. I think that adding additional check is a little over-protective coding in this specific situation and I wanted to keep this script as simple as possible. But it is not that I am strongly against - if you still want that change I can add it, no problem. Just let me know.

.github/workflows/_linux_build.yml
BBBela: Yes, to be honest I wasn't sure what is the best way. I decided to make it non-blocking for the start, so if anything fails, the behavior is like it was so far. That was intentional - not to stop the build if there is any issue with the gh call. But it would still fail if the final live check is not successful during the summary.
I suppose we could discuss it and decide during Tuesday meeting and I will adjust the code to what we agree on. I am neutral on this.

.github/scripts/fetch_issues.sh
pbielak: Ok, convinced - resolving the comment

.github/workflows/_linux_build.yml
BBBela: Removed the `continue-on-error` and `|| true`.

BBBela:
Hi @pbielak, as per today's team conversation, I introduced changes to make the snapshotting feature fail the CI if the snapshotting itself have any issues.

AKloniecki (approved):
LGTM

pbielak (approved):
LGTM!

@BBBela The following is out-of-scope for this PR, but maybe let's create a follow-up issue: Instead of creating two separate files, we could update the `fetch_issues.sh` script to already produce one file which (1) merges both known issues sources, (2) prepares the final test description format (instead of transforming it in the summary stage).

BBBela:
> @BBBela The following is out-of-scope for this PR, but maybe let's create a follow-up issue: Instead of creating two separate files, we could update the `fetch_issues.sh` script to already produce one file which (1) merges both known issues sources, (2) prepares the final test description format (instead of transforming it in the summary stage).

That is a very good follow up task idea. I was also thinking about similar changes during this PR development, but I wanted to keep the PR changes modular and simple and go step by step. I will add it to my TODO list!

.github/workflows/_linux_ut.yml
```diff
@@ -35,7 +35,6 @@ env:
   HF_TOKEN: ${{ secrets.HUGGING_FACE_HUB_TOKEN }}
   HUGGING_FACE_HUB_TOKEN: ${{ secrets.HUGGING_FACE_HUB_TOKEN }}
   DOCKER_REGISTRY_AUTH_TOKEN: ${{ secrets.DOCKER_HUB_TOKEN }}
-  UT_SKIP_ISSUE: 1624
   NEW_PASSED_ISSUE: 2333
```
chuanqi129: I think we don't need this issue for now, can delete it

chuanqi129 (approved):
LGTM, thanks for the enhancement

.github/workflows/_linux_ut.yml
BBBela: Removed NEW_PASSED_ISSUE

.github/workflows/_linux_build.yml
```diff
@@ -93,6 +93,19 @@ jobs:
         uses: actions/checkout@v4
         with:
           path: torch-xpu-ops
+      - name: Snapshot open skipped issues at build start
+        run: |
+          touch known_issue_static.log issues_snapshot.log
+          bash torch-xpu-ops/.github/scripts/fetch_issues.sh static > known_issue_static.log
+          bash torch-xpu-ops/.github/scripts/fetch_issues.sh open > issues_snapshot.log
```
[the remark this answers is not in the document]
BBBela: That is desired behavior for the build stage.

.github/workflows/_linux_ut.yml
```diff
@@ -165,15 +168,17 @@ jobs:
...
+          #   1. Snapshot captured at build start (covers issues closed mid-run)
+          #   2. Current live state (covers new issues opened after build start)
+          snapshot_dir="${{ github.workspace }}/issues_snapshot"
+          touch issues.log Known_issue.log.tmp
+          echo "Restoring snapshotted skipped issues if available."
+          cp "${snapshot_dir}/known_issue_static.log" Known_issue.log.tmp
+          cp "${snapshot_dir}/issues_snapshot.log" issues.log
+          echo "Fetching current skipped-issues data from GitHub."
+          bash ${{ github.workspace }}/.github/scripts/fetch_issues.sh static >> Known_issue.log.tmp
+          bash ${{ github.workspace }}/.github/scripts/fetch_issues.sh open >> issues.log
```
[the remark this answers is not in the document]
BBBela: This is the desired behavior.

.github/workflows/_linux_ut.yml
```diff
@@ -152,6 +150,11 @@ jobs:
         with:
           pattern: Inductor-XPU-UT-Data-${{ github.event.pull_request.number || github.sha }}-${{ inputs.ut }}-*
           path: ${{ github.workspace }}/ut_log
+      - name: Download skipped issues snapshot
+        uses: actions/download-artifact@v4
+        with:
+          name: Skipped-Issues-Snapshot-${{ github.event.pull_request.number || github.sha }}-linux-build-start
+          path: ${{ github.workspace }}/issues_snapshot
       - name: Check UT Results
```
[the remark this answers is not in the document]
BBBela: Done

.github/workflows/_linux_ut.yml
```diff
@@ -165,15 +168,17 @@ jobs:
...
-              --jq '.[] | select(.pull_request == null) | "Issue #\(.number): \(.title)\n\(.body)\n"' > issues.log
-          fi
+          # Build the known-issues list by merging two sources:
+          #   1. Snapshot captured at build start (covers issues closed mid-run)
+          #   2. Current live state (covers new issues opened after build start)
+          snapshot_dir="${{ github.workspace }}/issues_snapshot"
+          touch issues.log Known_issue.log.tmp
+          echo "Restoring snapshotted skipped issues if available."
+          cp "${snapshot_dir}/known_issue_static.log" Known_issue.log.tmp
+          cp "${snapshot_dir}/issues_snapshot.log" issues.log
```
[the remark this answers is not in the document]
BBBela: Addressed and fixed

BBBela:
I checked the CI logs and everything looks correct - the mechanism introduced by this PR works as expected. New stages are visible and uploaded artifacts are downloadable and contain what expected.

BBBela:
> ❌ CI checks have not passed. Failed checks: linux-build (build) / main: cancelled
> 
> Use `/merge -f` to force merge (maintainers only).

That is a weird message from github-actions bot as all the checks passed or were skipped and any of them was cancelled in latest CI run:
https://github.com/intel/torch-xpu-ops/pull/3429/checks?sha=813ec4c8e5e51664ec1095a9f44098c4dd960777

Using force merge only because it looks like some auto-merge problem while all the CI jobs passed or were skipped in the latest run with latest changes.

chuanqi129:
✅ PR has been successfully merged by @BBBela (force merge).

PR #3437, opened 2026-04-23
[copilot] Init skills for PR creation

Stonepia:
Add init instructions for pr creation

### DIFF
--- a/.github/skills/xpu-ops-pr-creation/SKILL.md
+++ b/.github/skills/xpu-ops-pr-creation/SKILL.md
@@ -31,12 +31,12 @@ Branch must follow: `agent/<slug>` (lowercase, hyphens, max 50 chars).
 
 ## Step 2: Write a reproducer test (mandatory)
 
-Every PR must include a pytest-compatible test under `test/regressions/`:
+Every PR must include a pytest-compatible test under `test/regressions/` if it is newly introduced. If the test is in PyTorch or torch-xpu-ops tests, you should explicitly write the run command in the PR body (see Step 3).:
 
 - File name: `test_<description>.py`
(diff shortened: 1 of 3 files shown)


Stonepia:
@copilot Fix all the comments

.github/skills/xpu-ops-pr-creation/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+git diff main --stat              # sanity-check the changed files
+```
+
+Branch must follow: `agent/<slug>` (lowercase, hyphens, max 50 chars).
+
+---
+
+## Step 2: Write a reproducer test (mandatory)
+
+Every PR must include a pytest-compatible test under `test/regressions/` if it is newly introduced. If the test is in PyTorch or torch-xpu-ops tests, you should explicitly write the run command in the PR body (see Step 3).:
```
chuanqi129:
```suggestion
Every PR must include a pytest-compatible test under `test/repro/` if it is newly introduced. If the test is in PyTorch or torch-xpu-ops tests, you should explicitly write the run command in the PR body (see Step 3).:
```

chuanqi129: let's start with the issue has reproducer, next step to cover the issue for pytorch ut

Stonepia: Thanks! I updated the skill to explicitly points out if there is any pytorch issue. Right now, I think we didn't submit any issues without the reproducer, so from the PR creation skill perspective, I just leave it as it is.

chuanqi129 (approved):

PR #3440, opened 2026-04-23
optimize rms_norm kernel perfomance.

nc-BobLee:


jianyizh:
@nc-BobLee What's the motivation of this PR? Please add some PR description. You ported RMS norm from IPEX. Does this provide same accuracy? what's the difference between current code and your PR?

PR #3466, opened 2026-04-24
Add pytorch-cuda-fix-xpu-alignment agent skill

laifenxiawucha:
## Summary

Add a new agent skill (`pytorch-cuda-fix-xpu-alignment`) that discovers CUDA and other backend bug fixes in `pytorch/pytorch` and validates whether the same bugs exist on XPU.

## Input / Output

**Input**: A time window (e.g., "last 1 day") and target repository (`pytorch/pytorch`).

**Output**: A local `triage_scan_<date>.md` file containing:
- Candidate list with URLs, titles, and dates
- Filter decisions (reject/pass) with reasons for each candidate
- Adapted XPU reproducer scripts for passed candidates
- Local XPU nightly validation results (confirmed / not-reproduced / unverified)
- Filter and validation statistics summary

## Workflow example
**Use**r: "Scan pytorch/pytorch for CUDA bug fixes in the last 1 day and check if they affect XPU."

The skill then:
1. **Collect** — Search `pytorch/pytorch` via GitHub MCP for recent backend bug-fix signals (issues, PRs, commits). Save a lightweight candidate list to `triage_scan_<date>.md`.
2. **Process one at a time** — For each candidate:
   - **Filter**: Read details, reject if infra-only / build-CI-only / docs-typo-only; otherwise pass
   - **Reproduce**: Extract or adapt the upstream regression test for `torch.xpu`
   - **Validate**: Run on local XPU nightly, record confirmed / not-reproduced / unverified
3. **Summarize** — Append filter statistics and validation statistics to the scan file.

## What is included

| File | Purpose |
|------|---------|
| `.github/skills/pytorch-cuda-fix-xpu-alignment/SKILL.md` | Skill definition and end-to-end workflow |
| `.github/copilot-instructions.md` | Registered in the skills index (1 row added) |

## What is missing

I also have two scripts:

- One script is used to set up a daily scheduled run: `daily_scan.sh`.
- Another script, `batch_scan.sh`, is used for batch scanning at scale, such as large-volume scans covering around 90 days.

In addition, there will be a README that explains how to configure `cron` and provides a quick start guide.

## Output Structure

Each run produces:
```
runs/<date-or-range>/
├── artifacts/
│   ├── raw_candidates.json       — deduplicated search results
│   ├── candidate_ledger.jsonl    — state-machine progress tracker
│   ├── details/<id>.json         — fetched issue/PR bodies and commit diffs
│   ├── output_<id>.log           — repro execution output
[...]

PR #3467, opened 2026-04-24
Add xpu-ops-issue-creation agent skill

laifenxiawucha:
## Summary

Add a new skill for turning a locally reproduced XPU bug into a structured issue for `intel/torch-xpu-ops`.

This is the issue-filing half of the CUDA-fix-to-XPU triage flow and is intended to run only after local validation confirms a real bug.

## What is included

- the skill definition and filing workflow instructions
- duplicate-check guidance
- issue structure and guardrails
- an issue template asset
- a short validation reference for confirming whether a repro is worth filing


.github/skills/xpu-bug-issue-filer/scripts/find_xpu_python.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
Stonepia: We don't need this script. If the python can't be found, then there should be an issue

Stonepia:
Hi @ZhaoqiongZ , as you are also creating the issue template, do you have any comments on this PR?

chuanqi129 (approved):

.github/skills/xpu-ops-issue-creation/references/local-xpu-validation-reference.md
```diff
@@ -0,0 +1,35 @@
+# Local XPU validation reference
```
ZhaoqiongZ: This file shares the same name (local-xpu-validation-reference.md) as the one in #3466  under pytorch-cuda-xpu-triage/references/, but they live at different paths and carry different content. This will cause confusion and maintenance drift.

Please either:

Merge them into a single shared file that both skills reference, or
Rename them to clearly indicate their distinct purpose (e.g., validation-criteria-reference.md for the shorter one focused on confirmation criteria, vs local-xpu-repro-reference.md for the fuller one with environment setup and repro guidance).

.github/skills/xpu-ops-issue-creation/assets/xpu_issue_template.md
```diff
@@ -0,0 +1,35 @@
+### 🐛 Describe the bug
+
+[sentence 1: what fails on XPU and on which build channel]
+[sentence 2: the concrete observed mismatch, crash, or wrong result]
+
+Reproducer:
```
ZhaoqiongZ, line 47: The current issue template is tailored for one scenario — manually validated bugs with custom reproducer scripts. However, there is another equally common scenario: CI UT failures caused by upstream PyTorch commits, where the failing test itself is the reproducer and the CI job link replaces collect_env.

I'd suggest unifying both scenarios into a single template rather than maintaining two separate ones, since the core information overlaps heavily (upstream commit/PR, error output, affected op). The difference is only in how the bug was discovered.

Here's a rough sketch of what a unified template could look like:

### 🐛 Describe the bug

[What fails on XPU and on which build/channel]
[Concrete mismatch, crash, or wrong result]

**Failure source:** CI UT failure / Manual validation

**Affected op/module:** [e.g. `aten::native_batch_norm`]

### Upstream reference

- Upstream commit: [commit SHA or URL that triggered the failure]
- Upstream PR: [PR URL]
- Upstream issue: [issue URL, if applicable]
- What the upstream change did (one line): [e.g. "added stricter dtype validation in native_batch_norm"]

### Failure details

**Failure type:** new test added / existing test broken / test expectation changed / build break


- CI job link: [URL]
- Failing test: `test/test_ops.py::TestCommonXPU::test_foo_xpu_float32`
- Run command: `python test/test_ops.py -k "test_foo_xpu_float32"`


- Reproducer:
  ```python
  [minimal script]

Error output
[traceback or mismatch output]

Versions
<!-- Only needed for manual validation; CI failures can reference the job link above --> <details><summary>collect_env</summary>
[collect_env output]

</details> ```
Key additions compared to the current template:

Failing test path & function name — the test IS the reproducer for CI failures
CI job link — currently missing entirely
Failure type classification — helps the fixer decide the approach (kernel fix vs test skip vs upstream fix)
One-line summary of the upstream change — saves the fixer from reading the upstream diff
Affected op/module — enables routing to the right owner
And fields that become conditional rather than always-required:

Standalone reproducer script → only for manual validation
collect_env → only for manual validation (CI env is a known BKC)

ZhaoqiongZ (approved):

.github/skills/xpu-ops-issue-creation/assets/xpu_issue_template.md
```diff
@@ -0,0 +1,75 @@
+### 🐛 Describe the bug
+
+[sentence 1: what fails on XPU and on which build channel]
```
Stonepia, line 3: @laifenxiawucha  This should align with the template formerly defined.

https://github.com/intel/torch-xpu-ops/pull/3507

PR #3476, opened 2026-04-25
 Add comment about addmm calculation for precision on XPU (bf16/fp16)

jafraustro:
Add comment about addmm calculation for precision on XPU (bf16/fp16)


src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
```diff
@@ -80,15 +80,25 @@ Tensor addmm_calculation(
...
+      (mat2.layout() != kStrided ? mat2.to_dense() : mat2).to(compute_dtype);
+
+  // Convert mm result back to output dtype before applying alpha/beta,
+  // to match the reference: alpha * (mm.to(out_dtype)) + beta * c
+  Tensor result = mat1_dense.mm(mat2_dense).to(out_dtype) * alpha;
   if (beta.toComplexDouble() != 0.) {
     Tensor input_dense = input.layout() != kStrided ? input.to_dense() : input;
-    result_dense.add_(input_dense, beta);
+    // Two-step (not fused) to match reference precision: beta*c first, then add
+    result = result + input_dense * beta;
```
CuiYifeng: @jafraustro Could check if the accuracy can be fixed with using non-fused add & mul ?
https://github.com/intel/torch-xpu-ops/pull/3273 also encountered similar problem.

src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
```diff
@@ -80,15 +80,25 @@ Tensor addmm_calculation(
...
     const Scalar& beta,
     const Scalar& alpha) {
-  Tensor mat1_dense = mat1.layout() != kStrided ? mat1.to_dense() : mat1;
-  Tensor mat2_dense = mat2.layout() != kStrided ? mat2.to_dense() : mat2;
-
-  Tensor result_dense = mat1_dense.mm(mat2_dense) * alpha;
+  auto compute_dtype =
+      (mat1.scalar_type() == at::kHalf || mat1.scalar_type() == at::kBFloat16)
+      ? at::kFloat
+      : mat1.scalar_type();
```
CuiYifeng: Just a reminder: to promote dtype, we can use `using opmath_t = at::opmath_type<scalar_t>;`, but I guess we won't need dtype promotion this time.

CuiYifeng (changes requested):

src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
jafraustro: Yes, it can be fixed similar to #3273, thanks

CuiYifeng: Good. Could you further investigate the difference between the fused and non-fused path?

jafraustro: A) `result_dense.add_(input_dense * beta)` executes two rounding:
1.  `input_dense (bf16) * beta (f32)` -> `intermediate_value (bf16)`
2.  `result_dense (bf16) + intermediate_value (bf16)`

B) `result_dense.add_(input_dense , beta)`, the kernel internally promotes both bf16 operands to f32:
`read result_dense(bf16) -> f32 register`
`read input_dense(bf16) -> f32 register`,
computes `result_dense (f32) + input_dense (f32) * beta (f32)` ,
then rounds to bf16 `result_dense (bf16)`

CuiYifeng: @jafraustro Thanks for the elaboration. In my view, using dtype promotion is more reasonable for accuracy. Seems the accuracy gap we encountered is related to the calculation of ground truth.
This fix is acceptable for UT. I noticed that the change in PR is the same as #3273. My suggestion is to land #3273 since CI is already passed.

CuiYifeng: #3273 is merged. @jafraustro Please add comments for the change. For comments changed only, we can add label `disable_all` to avoid redundant tests.

jafraustro: @CuiYifeng, Done

CuiYifeng (approved):

chunhuanMeng (approved):

chuanqi129:

PR #3487, opened 2026-04-27
Add input dimension check in sparse CSR addmm to match dense path

CuiYifeng:
The dense addmm in Blas.cpp validates that the self tensor's dimensionality does not exceed the result shape before expanding. The sparse CSR addmm_out_sparse_compressed_xpu was missing this check, causing test_addmm_errors to fail because the sparse path produced a different error message than the dense path when self is 3D.

Add the same dimension check (result_shape.size() >= self.dim()) before the expand_size call so both paths raise identical errors.

To solve #2245


LuFinch (approved):

chunhuanMeng (approved):

chuanqi129:

PR #3495, opened 2026-04-27
Fix randint uniformity for large ranges on XPU

### LINKED ISSUE #3473: [Bug Skip] test_randint_distribution on XPU: non-uniform distribution for large n_max
## Bug Description

`test_randint_distribution` fails on XPU with a histogram bin-spread assertion. The test draws `1_000_000` samples from `torch.randint(n_max=0.75 * 2**32, ...)`, bins them into 8 equal-width buckets, and asserts that each bucket is within `~3·sqrt(N/8)/(N/8)` relative error of the expected count. On XPU, the maximum relative error exceeds the threshold, indicating a non-uniform distribution for very large `n_max`.

## Affected Tests

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_tensor_creation_ops_xpu.TestRandomTensorCreationXPU,test_randint_distribution_xpu

## Error Message

```
AssertionError: tensor(False) is not true
```

## Test Code Snippet

```python
# third_party/torch-xpu-ops/test/xpu/test_tensor_creation_ops_xpu.py:4310-4327
@unittest.skipIf(
    IS_FBCODE or IS_SANDCASTLE,
    "For fb compatibility random not changed in fbcode",
)
def test_randint_distribution(self, device):
    size = 1_000_000
    n_max = int(0.75 * 2**32)
    n_bins = 8

    def bin(index, max_size):
        return index // (max_size // n_bins)

    res = torch.randint(n_max, (size,), device=device)
    bins = bin(res, n_max).float().cpu()
    hist, _ = bins.histogram(8, range=(0, n_bins))
    expected_bin = res.shape[0] / 8
    expected_error = math.sqrt(expected_bin) / expected_bin * 3
    error = (hist - expected_bin).abs().max() / expected_bin
    self.assertTrue(error < expected_error)   # <-- fails on XPU
```

## Traceback

```
pytest_command:
cd ~/daisy_pytorch/third_party/torch-xpu-ops/test/xpu
pytest -v test_tensor_creation_ops_xpu.py -k test_randint_distribution_xpu

E   AssertionError: tensor(False) is not true
test_tensor_creation_ops_xpu.py:4327: AssertionError
```

## Root Cause Analysis

The failure points to a uniformity gap in XPU's `torch.randint` for large `n_max` close to `2**32`. The CPU and CUDA implementations satisfy the binning bound; the XPU implementation does not. Likely contributors:

- 32-bit philox / mt19937 path producing uneven coverage when `n_max` is not a power of two near `2**32` (modulo bias).
- XPU `randint` may sample raw 32-bit values then take `%n_max`, which introduces visible bias for `n_max ~ 0.75·2**32`.

This is a numeric/RNG bug in the XPU `randint` kernel rather than a test infrastructure issue.

## Related Intel/torch-xpu-ops Issues

- (no exact match for `randint_distribution` in 2026-04 issue list)

## Versions

- pytorch: source build (HEAD)
- torch-xpu-ops: HEAD
- env: pytorch_opencode_env on PVC

Silv3S:
Fixes https://github.com/intel/torch-xpu-ops/issues/3473

XPU `random_from_to_kernel` was using threshold condition from `fbcode`  path (switch to 64-bit RNG only when  `range >= 2³²` for specific dtypes), instead of `range >= 2²⁸`. This patch changes it, to use the default threshold that matches upstream CUDA non-fbcode path. 
 
Upstream covers both paths, but `fbcode` is skipped from ut: https://github.com/pytorch/pytorch/pull/144819


kdrozd-dev (approved):

PawelSwider2000 (approved):

guangyey (approved):
Nice!

Silv3S:
/merge -f

Test linux-distributed was skipped as it already passed in previous CI run (only rebase diff)

chuanqi129:

PR #3502, opened 2026-04-28
[CI] Restart pytest worker when XPU memory utilization is high

mengfei25:
## Summary

Add a pytest-xdist watchdog that restarts UT workers when an XPU device approaches OOM, plus the supporting CI plumbing on Linux and Windows. Net effect across the PR: 7 files changed, +653 / −59.

## Motivation

Long UT runs occasionally leak GPU memory across tests, so later tests in the same xdist worker fail with OOM / UR errors and the whole job becomes flaky. We want each worker to exit cleanly and respawn whenever its device is too full or after a known-fatal failure pattern, so the rest of the suite can continue.

## Layout

A small three-file package under `.github/scripts/pytest_configs/`, loaded by the workflows via `PYTHONPATH = .../pytest_configs` and `-p xpu_worker_restart`:

### `xpu_smi_install.py` — locate or install `xpu-smi`
- Linux: `apt-get update && apt-get install -y xpu-smi`, fallback `dnf install -y xpu-smi`. Uses `sudo -n` when not root.
- Windows: download a pinned `xpumanager` release zip.
  - HTTPS-only.
  - SHA-256 pinned (`d1b44b58…f9de`).
  - 50 MiB streaming size cap; partial file removed on any failure.
  - Zip extraction guards against zip-slip (rejects absolute paths, drive letters, symlinks, `..` traversal).
- Public API: `ensure_xpu_smi()`, `which_xpu_smi()`. Never raises; always returns a `bool` / `Optional[str]`.

### `xpu_mem_monitor.py` — background `xpu-smi` poller (writer / reader split)
- **Writer** (xdist controller, or non-xdist run): owns one daemon thread that runs `xpu-smi dump … -n 1` every 3 s and atomically rewrites a small (~100 B) snapshot file (`mkstemp` + `os.replace`). Probes tiles-per-device once via `xpu-smi stats -d 0` and publishes the count via the `XPU_SMI_TILES` env var.
- **Reader** (xdist workers): only call `attach()` then `get_max_mem_util()`. Pure file read + regex; never spawns subprocesses, never imports the installer module (lazy import in `start()`).
- When `tiles > 1` the writer issues `xpu-smi dump -t 0,1,…,N-1 -m 5 -n 1` to capture every tile in one shot.
- `get_max_mem_util(cards=None)` returns `(card_id, util)` for the busiest visible device/tile, where `card_id = dev * tiles + tile` (or just `dev` in non-tile mode). With `cards=None` the watchdog considers every row in the snapshot.
- All hooks/threads are guarded so the poller can never crash the test session.

### `xpu_worker_restart.py` — pytest plugin
- Controller publishes `XPU_SMI_LOG_PATH` / `XPU_SMI_LOG_OWNER_PID`; workers attach to the same snapshot file. Cleanup of the shared log is gated on the owner pid.
- Per worker, target cards come from `ZE_AFFINITY_MASK` (supports `0`, `0,1`, `0,2-3`); when the env var is unset the watchdog considers **all** visible cards/tiles.
- On test teardown, if the **max** utilization across the worker's cards is ≥ **80%**, the worker calls `gc.collect` + `torch.xpu.synchronize/empty_cache`, prints `!RESTART <worker> (...)` via a single `_log()` helper, and exits with code **101** so pytest-xdist respawns it under `--max-worker-restart`.
- A single combined regex covers fatal failure markers in test output: `ur_result_error`, `out of memory`, `segmentation fault`, `bus error`, `kernel died`, `illegal memory`, `failed on setup with … crashed while running` (case-insensitive, DOTALL).
- A re-entrancy guard prevents double-restart on cascading failures. The controller never restarts itself.
- Every pytest hook is wrapped in `try/except` so the watchdog cannot break the test session.

## CI integration

- **Linux UT** (`.github/workflows/_linux_ut.yml`, both UT jobs):
  - `PYTHONPATH: ${{ github.workspace }}/.github/scripts/pytest_configs`
  - `PYTEST_ADDOPTS: … --max-worker-restart 1000000 -p xpu_worker_restart …`
- **Windows UT** (`.github/workflows/_windows_ut.yml`): same `PYTHONPATH` / `-p xpu_worker_restart` wiring.
- Removed the legacy `cp …/conftest.py …` step from `.github/actions/linux-uttest/action.yml` and deleted the old `.github/scripts/conftest.py`; the plugin is now loaded explicitly by name to avoid pytest's auto-discovery of files named `conftest.py`.

## Behavior

- **Healthy run**: snapshot file stays ~100 B, rewritten every 3 s; no test impact.
- **Worker hits ≥80% on its busiest visible card/tile** (or matches a fatal pattern): worker prints `!RESTART <worker> (...)` and exits 101; xdist respawns it.
- **Controller exit / non-xdist run**: poller stopped via `atexit` / `pytest_unconfigure`; only the owner pid does cleanup.
- **`xpu-smi` unavailable**: install attempt is best-effort; on failure the watchdog logs, disables itself, and the test session continues normally.

## Tunables

| Constant | Value |
| --- | --- |
| `_GPU_MEM_THRESHOLD` | `80.0` (%) |
| `_XPU_SMI_INTERVAL` | `3` s |
| `_XPU_SMI_TIMEOUT` | `60` s |
| `_WORKER_RESTART_CODE` | `101` |

## Code quality

- All three modules use `from __future__ import annotations`, native PEP 604 union syntax, `pathlib.Path`, `contextlib.suppress`, walrus operators, `subprocess.run(..., text=True, capture_output=True)`, and `max(..., key=..., default=None)`.
- `Iterable` / `Iterator` imported from `collections.abc` (ruff `UP035`).
- Strict separation between the three responsibilities (install / poll / pytest hooks); workers don't pull in the installer's `urllib`/`zipfile`/`hashlib` etc.

## Validation

- Local smoke tests with stubbed `_run_xpu_smi` cover both single-tile and multi-tile snapshots: correct command shape (`dump -m 5 -n 1` vs `dump -t 0,1,… -m 5 -n 1`), correct max-utilization across all devices/tiles, correct subset filtering by `ZE_AFFINITY_MASK`, and `attach()` from a separate "worker" sees the same data.
- Crafted zip with a `../evil.txt` member is rejected by `_safe_extract_zip`; benign zip extracts.
- `_run_xpu_smi` raising mid-loop does not kill the poller thread (caught + logged).
- `ZE_AFFINITY_MASK` parsing handles single ids, ranges (`2-3`), commas, blank tokens, and bad tokens (skipped).
- Fatal-pattern regex matches `UR_RESULT_ERROR`, `Out Of Memory`, `Segmentation fault`, etc.; does not match generic assertion errors.
- All three modules pass `python3 -c "import ast; ast.parse(...)"` and import cleanly under stubbed `pytest` / `xpu_smi_install` modules.


chuanqi129 (approved):

.github/scripts/xpu_worker_restart.py
```diff
@@ -0,0 +1,406 @@
...
+        _log("xpu-smi.exe not found after extraction")
+        return False
+    os.environ['PATH'] = os.path.dirname(cached) + os.pathsep + os.environ.get('PATH', '')
+    return True
+
+
+def _ensure_xpu_smi():
+    if _which_xpu_smi():
+        return True
+    return _install_windows() if _is_windows() else _install_linux()
```
chuanqi129: Can we split the xpu-smi installation into another dedicated scripts?

.github/scripts/xpu_worker_restart.py
```diff
@@ -0,0 +1,406 @@
...
+    if not _ensure_xpu_smi():
+        _log("xpu-smi unavailable; memory monitoring disabled")
+        return
+
+    fd = None
+    path = None
+    try:
+        fd, path = tempfile.mkstemp(prefix='xpu_smi_', suffix='.log')
+        env = os.environ.copy()
+        env['ZE_AFFINITY_MASK'] = ''
```
chuanqi129: What's the behavior on multi-devices system?

mengfei25: Get all GPUs memory to file, then get the memory used from file filter by each GPU in each worker

.github/scripts/xpu_worker_restart.py
mengfei25: Done

chuanqi129 (approved):

jmamzax:
Is this going to work correctly on GPU Max 1550, as this is two tile device? ZE_AFFINITY_MASK isn't set per device but 0,1 correspond with device 0. Also calling `xpu-smi dump -m 5 -n 1` will give memory utilization per device no per tile. Example 
test_mem_eff_attention_large_seq_len_uniform_attention_xpu test run:

```
Timestamp, DeviceId, GPU Memory Utilization (%)
11:57:10.891,    0, 49.24
11:57:10.891,    1, 0.07
```

But running `xpu-smi dump -m 5 -n 1 -t 0,1` on the same test will return:

```
Timestamp, DeviceId, TileId, GPU Memory Utilization (%)
11:56:14.962,    0,    0, 98.71
11:56:14.962,    0,    1, 0.07
11:56:14.962,    1,    0, 0.07
11:56:14.962,    1,    1, 0.07
```

But `-t` option shouldn't be use on devices that don't support tiles. Arc B570 return error or wrong value. Don't know how behave GPU Max 1100.

chuanqi129:
@jmamzax good capture! @mengfei25 let's check the device first and use different parameters. BTW, can this pytest config work for distributed UT test? Do we plan to support it?

mengfei25:
> Is this going to work correctly on GPU Max 1550, as this is two tile device? ZE_AFFINITY_MASK isn't set per device but 0,1 correspond with device 0. Also calling `xpu-smi dump -m 5 -n 1` will give memory utilization per device no per tile. Example test_mem_eff_attention_large_seq_len_uniform_attention_xpu test run:
> 
> ```
> Timestamp, DeviceId, GPU Memory Utilization (%)
> 11:57:10.891,    0, 49.24
> 11:57:10.891,    1, 0.07
> ```
> 
> But running `xpu-smi dump -m 5 -n 1 -t 0,1` on the same test will return:
> 
> ```
> Timestamp, DeviceId, TileId, GPU Memory Utilization (%)
> 11:56:14.962,    0,    0, 98.71
> 11:56:14.962,    0,    1, 0.07
> 11:56:14.962,    1,    0, 0.07
> 11:56:14.962,    1,    1, 0.07
> ```
> 
> But `-t` option shouldn't be use on devices that don't support tiles. Arc B570 return error or wrong value. Don't know how behave GPU Max 1100.

@jmamzax  Thanks for your point out, I just focused on client, will add support on multiple tiles device

mengfei25:
> @jmamzax good capture! @mengfei25 let's check the device first and use different parameters. BTW, can this pytest config work for distributed UT test? Do we plan to support it?

@chuanqi129  This is for all pytest test cases, if distributed, will use max utilization for all used GPUs.

.github/scripts/pytest_configs/xpu_worker_restart.py
```diff
@@ -0,0 +1,193 @@
...
+        if report.failed and report.longrepr is not None:
+            if _FATAL_RE.search(str(report.longrepr)):
+                _restart_worker("fatal pattern matched")
+                return  # _restart_worker returns only when disarmed
+
+        if report.when == "teardown" and _target_cards != []:
+            hot = xpu_mem_monitor.get_max_mem_util(_target_cards)
+            if hot is not None and hot[1] >= _GPU_MEM_THRESHOLD:
+                card, mem = hot
+                _restart_worker(
```
jmamzax, line 189: Why `_restart_worker` here, not call `_cleanup_torch_xpu` only. If there will be a test case that uses 82% of memory and finish correctly with **pass** the worker will restart again to run test. Calling `_cleanup_torch_xpu` should be enough to clean gpu memory for the next test.

mengfei25: This PR aims to mitigate issues involving machine reboot caused by out of memory, especially on client GPU. We don't know how much memory will be used exactly for a case, so hard to handle edge-case scenarios and used a conservative threshold. And after worker restart, will rerun the case too (rerun <= 2), if memory enough it will be passed.

jmamzax: But if the single test case uses from start more then 80% of the memory the test will fail if the memory will be checked. It will be rerun but every time it will fail at check. In line 180 you check for errors so this is not the out of memory case, just high memory usage. But if this should works like that then ok.

chuanqi129:
@mengfei25 do we still need this PR?

PR #3508, opened 2026-04-28
docs: clarify issue skill capture rules

laifenxiawucha:
This updates the xpu-ops issue creation skill package so CI-filed issues retain build and version context in plain text instead of relying on a job link alone, and so upstream issue/PR/commit anchors are required for CI-triggered cases but only included for manual validation when applicable. The SKILL guidance, issue template, and validation reference now align on shared environment fields and conditional upstream capture for the two filing paths.

Test: none (docs-only skill/template update; validated with git diff --check)


laifenxiawucha:
Follow-up clarification after review: this skill now targets exactly two upstream-anchored entry points.

1. Manual validation path: start from an upstream CUDA or PyTorch issue, PR, or commit, run the XPU reproducer locally, and file an intel/torch-xpu-ops issue if the XPU bug is confirmed.
2. CI path: start from a PyTorch CI failure tied to an upstream change and file an intel/torch-xpu-ops issue from that failure context.

The latest commit tightens the wording so upstream anchors are required for both paths, while keeping the earlier fix that requires build/version context to be captured in plain text for CI-filed issues.

laifenxiawucha:
Superseded: the requested changes have now been pushed directly to the original PR branch for #3467 (`add-xpu-bug-issue-filer-skill`). Please review and merge through #3467 instead of this follow-up PR.

Issue #3503, opened 2026-04-28
Issue Creation for CI Failure Tracking — PR Split and Validation Plan


ZhaoqiongZ:
# Issue Creation for CI Failure - PR Split & Validation Plan

**Last Updated:** 2026-04-29
**Target Repo:** intel/torch-xpu-ops
**Tracking Issue:** #3503

---

## Design Principles

1. **Label:** `pytorch-ci-failure` - distinguishes PyTorch mainline CI failures from torch-xpu-ops internal issues
2. **Title Prefix:** `[PyTorch CI]`
3. **Classification Strategy:**
   - **Summary issue:** factual - NEW_FAILURE / EXISTING_FAILURE (pure data comparison between runs)
   - **Sub-issues:**  Creating sub-issues would result in a multi-issue list and could cause the number of issues to explode, so this will not be automated for now.
4. **Schedule Control:** repo variable `CI_MONITOR_ENABLED` (true/false) gates cron triggers; `workflow_dispatch` always available

---

## Pipeline Overview

```
Step 1: Scan CI (PR2 - MERGED)
  Input:  GitHub API (PyTorch XPU workflow runs)
  Output: ci_results.json, all_runs.json (uploaded as artifact)

Step 2: Create Summary Issue (PR3)
  Input:  ci_results.json, all_runs.json
  Output: GitHub issue with commit scope, NEW/EXISTING classification

Step 3: Workflow Integration (PR3)
  Input:  All scripts from Step 1-3
  Output: Updated ci_break_monitor.yml with all steps wired together

Create Sub-Issues (will not be automated for now and manually trigger with local script)
  Input:  ci_results.json, summary issue number
  Output: One sub-issue per test-file group
```

---

### PR 1: Issue Template - MERGED (#3507 )

**Files:**
- `.github/ISSUE_TEMPLATE/ci-failure-tracking.yml`

### PR 2: CI Scanner Script + Workflow - MERGED (#3525)

**Files:**
- `.github/scripts/check_nightly_status.py`
- `.github/workflows/ci_break_monitor.yml`

**Key features:**
- Scan PyTorch XPU CI workflow runs via GitHub API
- Deep-scan EXISTING failures to find first-fail commit scope
- File-path based workflow ID lookup with name-based fallback
- Auth header leak prevention on log download redirects
- Non-success conclusion coverage (not just failure/timed_out)
- `pull_request` trigger with path filters for CI validation

[...]

PR #3507, opened 2026-04-28
Add PyTorch CI failure tracking issue template

ZhaoqiongZ:
PR1 for plan #3503 

Add a structured issue template for tracking PyTorch CI failures detected on the XPU CI workflow.

**What this adds:**
- `.github/ISSUE_TEMPLATE/ci-failure-tracking.yml`

**Template fields:**
- Failure Type (NEW_FAILURE / EXISTING_FAILURE)
- Commit Scope (last pass → first fail range)
- Failed Tests list
- Error Log
- CI Job URL
- Repro Commands
- Action Items checklist

**Label:** Auto-applies `pytorch-ci-failure` to distinguish PyTorch mainline CI failures from torch-xpu-ops internal issues.

This template will be used by upcoming CI monitoring scripts (PRs 2–5 in #3503 ) to auto-create issues when XPU CI failures are detected.

PR #3532, opened 2026-04-29
Add issue creation scripts and workflow Steps 2-3

ZhaoqiongZ:
## PR3: Issue Creation for CI Failures

Implements Steps 2-3 of CI Break Monitor (tracking issue #3503).

### Changes:
1. **create_tracking_issue.py** - Summary issue with commit scope, suspects, NEW/EXISTING/FIXED breakdown
2. **classify_failures.py** - Sub-issues per test-file group, links DISABLED tests from pytorch/pytorch, auto-close fixed
3. **ci_break_monitor.yml** - Steps 2-3 integration (dry-run on PR, real on schedule)

### Self-review applied:
- Null safety (dict.get everywhere)
- Dedup before creating issues
- [ai_generated] prefix on all titles
- Labels: pytorch-ci-failure, agent:blocked, ai_generated
- Local dry-run tested


Stonepia (commented):
About the file structure, I am having a concern if we should put the scripts into separate agent folders. For example,

```
.github/scripts/xxx_agent
```

I assume that in the future, there are many agents. If we all put under the `scripts/` folder, then the reading content for AI coding might be complicated and easily break other code

.github/scripts/ci_monitor_agent/create_tracking_issue.py
```diff
@@ -0,0 +1,616 @@
...
+        })
+
+    return suspects[:10]
+
+
+# ============================================================================
+# Issue Body Formatting
+# ============================================================================
+
+def format_issue_body(data, all_runs_data=None):
```
Stonepia, line 172: About this, maybe we should let LLM to create the issue from formerly defined issue template?

https://github.com/intel/torch-xpu-ops/pull/3507

ZhaoqiongZ: Currently the generated issues partially align with the CI failure template, but the script does not directly render from the template file. Some fields are covered structurally (commit scope, failed tests, CI job links, repro commands/action items in sub-issues), while fields like Platform, OS, and Error Log are not explicitly populated yet.

ZhaoqiongZ:
> About the file structure, I am having a concern if we should put the scripts into separate agent folders. For example,
> 
> ```
> .github/scripts/xxx_agent
> ```
> 
> I assume that in the future, there are many agents. If we all put under the `scripts/` folder, then the reading content for AI coding might be complicated and easily break other code

updated and put those ci monitor scripts in scripts/ci_monitor_agents

.github/scripts/ci_monitor_agent/create_tracking_issue.py
ZhaoqiongZ: I agree we should align this with the formal template to avoid drift. Let’s iterate a few times to see whether the generated issue needs more information or less information, and then fully align it with the issue template later.

PR #3469, opened 2026-04-24
Update lintrunner to changes from upstream

Silv3S:
We have few linters that were removed/replaced in upstream. This patch is collection of cherry-picks, aligning our backend to upstream config
- use PEP 723 inline script dependencies
- replace ufmt with pyfmt
- replace black with ruff (partially)
- remove custom constexpr lint
- remove bazel lint
- update ruff/flake8 config

PR #3533, opened 2026-04-29
Update lintrunner to changes from upstream (part 2)

Silv3S:
Follow up to https://github.com/intel/torch-xpu-ops/pull/3469

Upstream migrated formatting lint from `black` to `ruff`. To align our codebase format, we should do the same


test/xpu/higher_order_ops/test_invoke_subgraph_xpu.py
```diff
@@ -328,7 +328,9 @@ def fn(mod, x, y):
         x_clone = x.detach().clone().requires_grad_(True)
         y_clone = y.detach().clone().requires_grad_(True)
         backend = EagerAndRecordGraphs()
-        with (torch.no_grad(),):
+        with (
+            torch.no_grad(),
+        ):
```
PawelSwider2000, line 333: why not:

`with torch.no_grad():`

test/xpu/higher_order_ops/test_invoke_subgraph_xpu.py
```diff
@@ -1335,7 +1337,9 @@ def fn(x, y):
 
         opt_fn = torch.compile(fn, backend="inductor", fullgraph=True)
 
-        with (torch.no_grad(),):
+        with (
+            torch.no_grad(),
+        ):
```
PawelSwider2000, line 1342: ditto.

test/xpu/higher_order_ops/test_invoke_subgraph_xpu.py
Silv3S: We aim to have 1:1 alignment with upstream. While I agree that oneliner is much cleaner I think we should keep exact copy, so it's easier to migrate

PawelSwider2000: So if there is also in upstream that maybe in fact will be better to keep this similar

PawelSwider2000 (approved):

chuanqi129 (approved):

chuanqi129:

PR #3547, opened 2026-04-30
#2618 Skip linalg tests that fails on numpy 1.23.2

jmamzax:
Part of https://github.com/intel/torch-xpu-ops/issues/2618. The tests fails only on python 3.10 because then pytorch for CI uses NumPy 1.23.2 with Open BLAS 0.3.20 which have a bug that can be observed only on Intel(R) Sapphire Rapids CPU which is used on every XPU CI runners.
NumPy issue: https://github.com/numpy/numpy/issues/22502


test/xpu/skip_list_common.py
```diff
@@ -99,6 +94,18 @@
         "test_compile_dyn_quant_matmul_4bit_m_32_k_64_n_4096_xpu",
         "_tunableop_",
         "_tuning_tunableop_",
+        # skipped due to a bug in OpenBLAS 0.3.20 used in numpy 1.23.* related with Intel(R) Sapphire Rapids CPU
```
Silv3S, line 102: Could you please add some TODO, to mark that these tests are expected to work once we switch to Python 3.11? It will be easier to unskip them

Silv3S (approved):

Issue #3426, opened 2026-04-22
[ai_generated] torch.topk raises RuntimeError for dimensions larger than INT_MAX on XPU


laifenxiawucha:
### 🐛 Describe the bug

torch.topk on XPU rejects an input dimension with more than INT_MAX elements on the PyTorch 2.11.0+xpu build used here.
For a shape based on upstream CUDA issue #173622, XPU fails earlier with `RuntimeError: The dimension being select can not have more than INT_MAX elements` instead of executing the 64-bit topk path.

Reproducer:

```python
import time
import torch

total_len = 4294967296 + 1000
print(f"Allocating {total_len} float16 elements on XPU (~{total_len * 2 / 1024**3:.1f} GB)...")
torch.manual_seed(0)
data_xpu = torch.randn(1, total_len, device="xpu", dtype=torch.float16)

k = 4294967296 + 530
print(f"Running topk k={k}...")
start = time.time()
values_xpu, indices_xpu = torch.topk(data_xpu, k, dim=1, sorted=False)
print("topk completed in", time.time() - start, "seconds")
print(values_xpu.shape, indices_xpu.shape)
```

Observed output:

```text
Allocating 4294968296 float16 elements on XPU (~8.0 GB)...
Allocation done in 0.1s
Running topk k=4294967826...
Traceback (most recent call last):
  File "<string>", line 15, in <module>
RuntimeError: The dimension being select can not have more than INT_MAX elements.
```

Additional context:

- Date: 2026-04-22
- Build: torch 2.11.0+xpu
- **Upstream reference:** https://github.com/pytorch/pytorch/issues/173622
- Related upstream PR: https://github.com/pytorch/pytorch/pull/176095
- Note: upstream CUDA issue was a silent wrong-result bug caused by 64-bit indexing overflow; XPU instead errors out before running topk when the selected dimension exceeds INT_MAX.
- Assisted-by: open code: GPT-5.4 [GitHub-API] [collect_env]

### Versions

<details>
<summary>Collected with torch/utils/collect_env.py</summary>

```text
Collecting environment information...
PyTorch version: 2.11.0+xpu
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: Ubuntu 24.04.2 LTS (x86_64)
GCC version: (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
Clang version: 18.1.3 (1ubuntu1)
CMake version: Could not collect
[...]

PR #3555, opened 2026-05-05
SortingKernels: fixes num_tiles and tile-count overflows

BartoszKokoszko:
This PR fixes overflows in SortingKernels:
- fixes num_tiles computation i.e. uses 64-bit arithmetic for num_tiles computation - earlier num_tiles was computed using 32-bit arithmetics. For large num_elements, the ceil-division expression could overflow before kernel launch, leading to incorrect behavior;
- fixes integer overflow in SortingRadixSort tile-count computation;
- adds a regression test (test_topk_num_tiles_no_overflow) to validate overflow-safe TopK behavior.

Co-authored-by: Benedykt Bela <benedykt.bela@intel.com>


src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -348,8 +348,9 @@ void segmented_radix_sort_pairs_kernel(
     int num_segments,
     int num_elements) {
   constexpr int TILE_PROCESSING_LENGTH = GROUP_SIZE * KEYS_PER_ITEM;
-  int num_tiles =
-      (num_elements + TILE_PROCESSING_LENGTH - 1) / TILE_PROCESSING_LENGTH;
+  int64_t num_tiles =
+      (static_cast<int64_t>(num_elements) + TILE_PROCESSING_LENGTH - 1) /
+      TILE_PROCESSING_LENGTH;
```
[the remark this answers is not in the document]
BartoszKokoszko: done

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -348,8 +350,14 @@ void segmented_radix_sort_pairs_kernel(
...
-  int num_tiles =
-      (num_elements + TILE_PROCESSING_LENGTH - 1) / TILE_PROCESSING_LENGTH;
+  int64_t num_tiles_64 =
+      (static_cast<int64_t>(num_elements) + TILE_PROCESSING_LENGTH - 1) /
+      TILE_PROCESSING_LENGTH;
+  TORCH_CHECK(
+      num_tiles_64 <= std::numeric_limits<int>::max(),
+      "num_tiles overflow: ",
+      num_tiles_64);
+  int num_tiles = static_cast<int>(num_tiles_64);
```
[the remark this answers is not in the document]
BartoszKokoszko: added cast for the first operand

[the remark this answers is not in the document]
BartoszKokoszko: done

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -359,11 +379,12 @@ void segmented_radix_sort_pairs_kernel(
...
   at::DataPtr counts_data = c10::GetAllocator(kXPU)->allocate(
-      num_segments * RADIX_BUCKETS * num_tiles * sizeof(int));
+      static_cast<size_t>(num_segments) * RADIX_BUCKETS * num_tiles *
+      sizeof(int));
   at::DataPtr keys_temp_data = c10::GetAllocator(kXPU)->allocate(
-      num_segments * num_elements * sizeof(key_t));
+      static_cast<size_t>(num_segments) * num_elements * sizeof(key_t));
   at::DataPtr values_temp_data = c10::GetAllocator(kXPU)->allocate(
-      num_segments * num_elements * sizeof(value_t));
+      static_cast<size_t>(num_segments) * num_elements * sizeof(value_t));
```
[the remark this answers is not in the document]
BartoszKokoszko: done

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -173,8 +175,14 @@ void segmented_radix_sort_pairs_upsweep_kernel(
...
+  int64_t num_tiles_64 =
+      (static_cast<int64_t>(num_elements) + method_t::PROCESSING_LENGTH - 1) /
       method_t::PROCESSING_LENGTH;
+  TORCH_CHECK(
+      num_tiles_64 <= std::numeric_limits<int>::max(),
+      "num_tiles overflow: ",
+      num_tiles_64);
+  int num_tiles = static_cast<int>(num_tiles_64);
   auto caller = SegmentedRadixSortPairsUpsweepFunctor<method_t, key_t, value_t>(
       keys_in, counts, num_elements, begin_bit, end_bit);
```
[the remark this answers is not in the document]
BartoszKokoszko: done

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -108,8 +110,10 @@ struct SegmentedRadixSortPairsUpsweepFunctor
     : public __SYCL_KER_CONFIG_CONVENTION__ {
   SYCL_REQD_SUB_GROUP_SIZE(method_t::SUBGROUP_SIZE)
   void operator()(sycl::nd_item<1> item) const {
-    int num_tiles = (num_elements_ + method_t::PROCESSING_LENGTH - 1) /
+    int64_t num_tiles_64 = (static_cast<int64_t>(num_elements_) +
+                            method_t::PROCESSING_LENGTH - 1) /
         method_t::PROCESSING_LENGTH;
```
BBBela: There is a `ceil_div()` method that can be included with: `#include <ATen/ceil_div.h>`
It does exactly the same and I personally find it more verbose and clear.
We could change the current "manual" ceil_div with:
```cpp
ceil_div<int64_t>(num_elements, method_t::PROCESSING_LENGTH)
```
When the template parameter is explicitly given as `int64_t`, then the arguments will be automatically promoted to the `int64_t`.

I won't repeat this comment, but it applies to all the similar occurrences in this file.

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -305,8 +317,14 @@ void segmented_radix_sort_pairs_downsweep_kernel(
...
       value_t>;
-  int num_tiles = (num_elements + method_t::PROCESSING_LENGTH - 1) /
+  int64_t num_tiles_64 =
+      (static_cast<int64_t>(num_elements) + method_t::PROCESSING_LENGTH - 1) /
       method_t::PROCESSING_LENGTH;
+  TORCH_CHECK(
+      num_tiles_64 <= std::numeric_limits<int>::max(),
+      "num_tiles overflow: ",
+      num_tiles_64);
+  int num_tiles = static_cast<int>(num_tiles_64);
```
BBBela: The same TORCH_CHECK is repeating 3 times. Could we extract it to separate function?
The function could not only run the TORCH_CHECK, but also return the value casted back to `int`.
I mean something that:
```cpp
inline int checked_num_tiles(int64_t n) {
  TORCH_CHECK(
      n <= std::numeric_limits<int>::max(), "num_tiles overflow: ", n);
  return static_cast<int>(n);
}
```

then is all three places we can simply do that:
```cpp
 int num_tiles = checked_num_tiles(num_tiles_64);
```

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -358,12 +382,38 @@ void segmented_radix_sort_pairs_kernel(
...
+      checked_mul_size(keys_temp_bytes, sizeof(key_t), "keys_temp");
+  at::DataPtr keys_temp_data =
+      c10::GetAllocator(kXPU)->allocate(keys_temp_bytes);
+
+  size_t values_temp_bytes =
+      checked_mul_size(num_segments_sz, num_elements_sz, "values_temp");
+  values_temp_bytes =
+      checked_mul_size(values_temp_bytes, sizeof(value_t), "values_temp");
+  at::DataPtr values_temp_data =
+      c10::GetAllocator(kXPU)->allocate(values_temp_bytes);
```
BBBela:
```suggestion
  at::DataPtr counts_data = c10::GetAllocator(kXPU)->allocate(
      static_cast<size_t>(num_segments) * RADIX_BUCKETS * num_tiles * sizeof(int));
  at::DataPtr keys_temp_data = c10::GetAllocator(kXPU)->allocate(
      static_cast<size_t>(num_segments) * num_elements * sizeof(key_t));
  at::DataPtr values_temp_data = c10::GetAllocator(kXPU)->allocate(
      static_cast<size_t>(num_segments) * num_elements * sizeof(value_t));
```
I think that adding `static_cast<size_t>(num_segments)` would be sufficient here instead of this new approach. But I may be wrong and I may miss some important detail, so please let me know if the `static_cast()` approach is not sufficient.

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -230,8 +240,10 @@ struct SegmentedRadixSortPairsDownsweepFunctor
     : public __SYCL_KER_CONFIG_CONVENTION__ {
   SYCL_REQD_SUB_GROUP_SIZE(method_t::SUBGROUP_SIZE)
   void operator()(sycl::nd_item<1> item) const {
-    int num_tiles = (num_elements_ + method_t::PROCESSING_LENGTH - 1) /
+    int64_t num_tiles_64 = (static_cast<int64_t>(num_elements_) +
+                            method_t::PROCESSING_LENGTH - 1) /
         method_t::PROCESSING_LENGTH;
+    int num_tiles = static_cast<int>(num_tiles_64);
```
BBBela: We first calculate the `num_tiles` in the `segmented_radix_sort_pairs_downsweep_kernel` function, then we recalculate the same value in the `SegmentedRadixSortPairsDownsweepFunctor`. Could we simply pass the `num_tiles` value calculated in the `segmented_radix_sort_pairs_downsweep_kernel` function as a `SegmentedRadixSortPairsDownsweepFunctor` parameter? The same is for `SegmentedRadixSortPairsUpsweepFunctor`

src/ATen/native/xpu/sycl/SortingKernels.h
BartoszKokoszko: Updated

src/ATen/native/xpu/sycl/SortingKernels.h
BartoszKokoszko: updated

src/ATen/native/xpu/sycl/SortingKernels.h
BartoszKokoszko: updated

src/ATen/native/xpu/sycl/SortingKernels.h
BartoszKokoszko: I have added checked_mul_size(), because theoretically there is possibility of size_t input argument overflow here.

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -348,8 +361,9 @@ void segmented_radix_sort_pairs_kernel(
     int num_segments,
     int num_elements) {
   constexpr int TILE_PROCESSING_LENGTH = GROUP_SIZE * KEYS_PER_ITEM;
-  int num_tiles =
-      (num_elements + TILE_PROCESSING_LENGTH - 1) / TILE_PROCESSING_LENGTH;
+  int64_t num_tiles_64 =
+      ceil_div<int64_t>(num_elements, method_t::PROCESSING_LENGTH);
```
BBBela:
```suggestion
      ceil_div<int64_t>(num_elements, TILE_PROCESSING_LENGTH);
```
That makes the build to fail. In this place there is no the `method_t` template parameter yet.

src/ATen/native/xpu/sycl/SortingKernels.h
BBBela: I agree that there is such a theoretical possibility, but in practice it is impossible.
`size_t` on 64-bit systems is an unsigned 64-bit integer with range `[0, 2^64 - 1]`, so the maximum addressable size is 16 exabytes — equivalent to about 16 million people each holding a 1 TB drive.

I don't think the code should be complicated with overflow checks that are impossible to trigger in practice.

BartoszKokoszko: @BBBela  I agree. I have changed it into:
```
 at::DataPtr counts_data = c10::GetAllocator(kXPU)->allocate(
      static_cast<size_t>(num_segments) * RADIX_BUCKETS * num_tiles * sizeof(int));
  at::DataPtr keys_temp_data = c10::GetAllocator(kXPU)->allocate(
      static_cast<size_t>(num_segments) * num_elements * sizeof(key_t));
  at::DataPtr values_temp_data = c10::GetAllocator(kXPU)->allocate(
      static_cast<size_t>(num_segments) * num_elements * sizeof(value_t));
```

BBBela (dismissed):
Looks good to me.
Thank you! 😉

pponikox (approved):

pawel-olejniczak (commented):
LGTM

Silv3S (approved):

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -10,16 +10,24 @@
...
+#include <limits>
 
 namespace at {
 namespace native {
 namespace xpu {
 
+inline int checked_num_tiles(int64_t n) {
+  TORCH_CHECK(n <= std::numeric_limits<int>::max(), "num_tiles overflow: ", n);
+  return static_cast<int>(n);
+}
```
CuiYifeng: For `int32 a >= 0` and `int32 b > 0`, using `int64_t` for the intermediate computation makes sense to avoid overflow in `(a + b - 1)`. However, the result of `ceil_div<int64_t>(a, b)` cannot overflow int32. So `checked_num_tiles()` seems redundant here.

CuiYifeng (commented):
Do we have any test cases for such overflow? The remaining part of this PR LGTM.

src/ATen/native/xpu/sycl/SortingKernels.h
BartoszKokoszko: done

BartoszKokoszko:
> Do we have any test cases for such overflow? The remaining part of this PR LGTM.

@CuiYifeng 

Directly no. It was taken from https://github.com/intel/torch-xpu-ops/issues/3426

CuiYifeng:
> > Do we have any test cases for such overflow? The remaining part of this PR LGTM.
> 
> @CuiYifeng
> 
> Directly no. It was taken from #3426

Please add a test case in torch-xpu-ops referring to `test_topk_large_k` in `test/test_sort_and_select.py`.

BartoszKokoszko:
@CuiYifeng  I have added _test_topk_large_k test case, which checks OutOfMemoryError

test/xpu/test_sort_and_select_xpu.py
```diff
@@ -126,6 +127,25 @@ def _test_sort_large_slice(self, device):
 
     TestSortAndSelect.test_sort_large_slice = _test_sort_large_slice
 
+    # XPU variant of upstream test_sort_and_select.py:test_topk_large_k.
+    # Based on /home/bkokoszx/scripts/topk.py and targets the large-k path
```
CuiYifeng: Please remove "Based on /home/bkokoszx/scripts/topk.py" from the comment.

BartoszKokoszko: done

CuiYifeng (approved):

BBBela (changes requested):
I checked this newly added test case `_test_topk_large_k` and I have some doubts about correctness. Below some of my considerations:

1. CUDA supports num elements bigger than `INT_MAX`, while XPU does not support it. XPU implementation has `TORCH_CHECK` that throws when the number of elements exceeds the `INT_MAX` value. Was it considered if we should enable number of elements bigger than `INT_MAX` as CUDA does? It is okay if we support number of elements up to `INT_MAX`, but just wanted to ask if it was considered.
2. Regarding the test case - it seems not correct for me. We gate it with `@largeTensorTest("20GB", "xpu")` so it is skipped if GPU has less than 20 GB of memory, but then we check for the Out Of Memory exception when trying to allocate 16 GB. So if the GPU has more than 20 GB of memory, then why can't we allocate 16 GB of data? Even if there are some previous allocations, and the available free memory is less than 16 GB, then we would get the Out Of Memory exception, but only if this allocation exceeds the available memory. On a GPU with enough memory this test would fail. So the test checks the GPU memory rather than topk algorithm and PyTorch mechanisms and checks.
3. I believe that in current code shape we should check if the `TORCH_CHECK` throws the exception when we try to allocate more than `INT_MAX` elements. This would check the PyTorch itself. Current test doesn't check the topk, but it checks if GPU throws the Out Of Memory exception when we want to allocate too big tensor.

Maybe I am missing something - please let me know what do you think?

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -115,7 +115,13 @@ struct SegmentedRadixSortPairsUpsweepFunctor
...
     int tile_offset = tile_idx * method_t::PROCESSING_LENGTH;
-    int tile_end = tile_offset + method_t::PROCESSING_LENGTH;
+    // Compute tile_end without int32 overflow: tile_offset + PROCESSING_LENGTH
+    // can wrap to INT_MIN when tile_offset is near INT_MAX (e.g. n = INT_MAX),
+    // which would cause process_full_tile() to be called without bounds checks,
+    // reading one element past the end of the input buffer.
+    int tile_end = (num_elements_ - tile_offset < method_t::PROCESSING_LENGTH)
+        ? num_elements_
+        : tile_offset + method_t::PROCESSING_LENGTH;
     tile_end = tile_end > num_elements_ ? num_elements_ : tile_end;
```
BBBela: The line
```cpp
tile_end = tile_end > num_elements_ ? num_elements_ : tile_end;
```
can be now removed as it always takes the false branch after the above refactor.

src/ATen/native/xpu/sycl/SortingKernels.h
```diff
@@ -115,7 +115,13 @@ struct SegmentedRadixSortPairsUpsweepFunctor
     auto keys_in_seg = keys_in_ + seg_idx * num_elements_;
     auto counts_seg = counts_ + seg_idx * method_t::RADIX_BUCKETS * num_tiles_;
     int tile_offset = tile_idx * method_t::PROCESSING_LENGTH;
-    int tile_end = tile_offset + method_t::PROCESSING_LENGTH;
+    // Compute tile_end without int32 overflow: tile_offset + PROCESSING_LENGTH
+    // can wrap to INT_MIN when tile_offset is near INT_MAX (e.g. n = INT_MAX),
+    // which would cause process_full_tile() to be called without bounds checks,
+    // reading one element past the end of the input buffer.
```
BBBela: I would personally remove this whole comment. The test case already checks for correctness and potential regressions and the newly added code fragment is rather clear. Or at least shorten it to one-line. I am a fan of self-explaining code that doesn't need additional comments.

test/regressions/test_sort.py
```diff
@@ -39,3 +40,20 @@ def test_sort_large_bool(self):
...
             self.assertEqual(sorted.cpu(), sorted_cpu)
+
+    @largeTensorTest("48GB", device="xpu")
+    def test_topk_num_tiles_no_overflow(self):
+        # INT_MAX:
+        # triggers overflow in the unfixed int32 computation
+        n = 2**31 - 1
+        # k > 256 routes through
+        # topk_out_with_sort -> segmented_radix_sort
+        k = 300
```
BBBela:
```suggestion
        n = 2**31 - 1
        # k > 256 routes through topk_out_with_sort -> segmented_radix_sort
        k = 300
```
I think we could simplify and shorten this.

src/ATen/native/xpu/sycl/SortingRadixSort.h
```diff
@@ -700,8 +700,12 @@ class RadixSortUpsweep {
   }
 
   inline void process_partial_tile(int group_offset, int group_end) {
-    for (int offset = group_offset + lid_; offset < group_end;
-         offset += GROUP_THREADS) {
+    // Use a relative loop variable to avoid int32 overflow: when group_offset
+    // is near INT_MAX, `group_offset + lid_ + k*GROUP_THREADS` wraps to INT_MIN
+    // on the next iteration, making the loop condition false-positive.
```
BBBela: As above - I would personally remove this additional explanatory comment. I think it is not necessary here.

src/ATen/native/xpu/sycl/SortingRadixSort.h
```diff
@@ -784,9 +788,11 @@ class RadixSortUpsweep {
     reset_unpacked_counters();
 
     // Unroll batches of full tiles
+    // Use subtraction form (a <= c - b) instead of addition (a + b <= c) to
+    // avoid int32 overflow when group_offset is close to INT_MAX.
```
BBBela: As above - unnecessary comment in my opinion.

test/xpu/test_sort_and_select_xpu.py
BBBela, line 1: I suppose we can now revert all changes in this file as the new test case is added to the `regressions/test_sort.py`, right?

BBBela (changes requested):
In general it looks good to me know, but below are some comments regarding recent changes.

src/ATen/native/xpu/sycl/SortingKernels.h
BartoszKokoszko: updated

src/ATen/native/xpu/sycl/SortingKernels.h
BartoszKokoszko: updated

test/regressions/test_sort.py
BartoszKokoszko: updated

src/ATen/native/xpu/sycl/SortingRadixSort.h
BartoszKokoszko: updated

test/xpu/test_sort_and_select_xpu.py
BartoszKokoszko: updated

src/ATen/native/xpu/sycl/SortingRadixSort.h
BartoszKokoszko: updated

BBBela (approved):
For me it looks good now. Thank you for your effort and this contribution! 😉

CuiYifeng (approved):
Since the changes are not only for `num_tile` overflow, please also update PR title and description.

BartoszKokoszko:
@CuiYifeng @BBBela 
PR description has been updated.

chuanqi129:
✅ PR has been successfully merged by @BartoszKokoszko (force merge).

PR #3558, opened 2026-05-05
Fix `int32` overflow in `max_pool3d` backward kernel

pbielak:
Widen parameters and variables in `max_pool3d_with_indices_backward_template` and `MaxPool3dBackwardKernelFunctor` from `int` to `int64_t` to prevent overflow when computing offsets for large tensors (e.g. shape `[70, 32, 100, 100, 100]`).

Fixes test case: `test_pool3d_large_size_int64`

### DIFF
--- a/src/ATen/native/xpu/sycl/DilatedMaxPool3d.cpp
+++ b/src/ATen/native/xpu/sycl/DilatedMaxPool3d.cpp
@@ -20,6 +20,7 @@ DISABLE_RETURN_TYPE_WARNING_BEGIN
 #include <ATen/ATen.h>
 #include <ATen/AccumulateType.h>
 #include <ATen/ceil_div.h>
+#include <ATen/native/CanUse32BitIndexMath.h>
 #include <ATen/native/Pool.h>
 #include <ATen/native/utils/ParamUtils.h>
 
@@ -316,24 +317,24 @@ void max_pool3d_with_indices_out_template(
   sycl_kernel_submit(global_range, work_group_size, queue, kfn);
 }
 
-template <typename scalar_t, bool channels_last>
+template <typename scalar_t, bool channels_last, typename index_t>
 struct MaxPool3dBackwardKernelFunctor {
   void operator()(sycl::nd_item<1> item) const {
-    auto outputIndex = item.get_global_id(0);
+    index_t outputIndex = item.get_global_id(0);
     if (outputIndex < gradOutputSize_) {
-      int batch = outputIndex / out_nbatch_stride_;
+      index_t batch = outputIndex / out_nbatch_stride_;
       if constexpr (channels_last) {
-        int channel = outputIndex % features_;
+        index_t channel = outputIndex % features_;
         int64_t index = indicesData_[outputIndex];
-        int64_t gradIn_offset =
+        index_t gradIn_offset =
             batch * in_nbatch_stride_ + channel + index * features_;
         atomicAdd(
             (sycl_global_ptr<scalar_t>)&gradInputData_[gradIn_offset],
             gradOutputData_[outputIndex]);
       } else {
-        int channel = outputIndex / out_cf_channel_stride_ % features_;
+        index_t channel = outputIndex / out_cf_channel_stride_ % features_;
         int64_t index = indicesData_[outputIndex];
-        int64_t gradIn_offset =
+        index_t gradIn_offset =
             batch * in_nbatch_stride_ + channel * in_cf_channel_stride_ + index;
         atomicAdd(
             (sycl_global_ptr<scalar_t>)&gradInputData_[gradIn_offset],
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/sycl/DilatedMaxPool3d.cpp
```diff
@@ -386,12 +386,13 @@ void max_pool3d_with_indices_backward_template(
...
+      (int64_t)obatch * features * otime * oheight * owidth;
 
-  auto out_cf_channel_stride = otime * oheight * owidth;
-  auto in_cf_channel_stride = itime * iheight * iwidth;
-  auto out_nbatch_stride = features * out_cf_channel_stride;
-  auto in_nbatch_stride = features * in_cf_channel_stride;
+  int64_t out_cf_channel_stride = (int64_t)otime * oheight * owidth;
+  int64_t in_cf_channel_stride = (int64_t)itime * iheight * iwidth;
+  int64_t out_nbatch_stride = (int64_t)features * out_cf_channel_stride;
+  int64_t in_nbatch_stride = (int64_t)features * in_cf_channel_stride;
```
BBBela: Regarding casting - what do you think about using the `static_cast<>()` instead of C-style casts? That follows the [CPP CORE GUIDELINE](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast) and [GOOGLE C++ STYLE GUIDE](https://google.github.io/styleguide/cppguide.html#Casting)?

I am also wondering if maybe we could just change the function parameters from `int` to `int64_t`? Then we would not need explicit casting - what do you think?

pbielak: Thanks for catching this - I updated the function parameters, as in fact there were multiple casting steps (even in the old code), int64 -> int32 -> int64. Now, all arguments are passed as int64_t. I also updated some local variables that were still computed as int32.

BBBela (approved):
Looks good to me, thanks for addressing review comments!

guangyey (approved):
Thanks!

guangyey:
@pbielak Could you help check if the same issue persisted in the forward kernel? If yes, better to file another PR to fix it.

pbielak:
@guangyey Thanks for the idea - I checked the forward kernel and it was also susceptible to int32 overflow, but there was no existing UT for it; I fixed the forward kernel and added the new unit test in PR: https://github.com/intel/torch-xpu-ops/pull/3632

src/ATen/native/xpu/sycl/DilatedMaxPool3d.cpp
```diff
@@ -345,12 +345,12 @@ struct MaxPool3dBackwardKernelFunctor {
...
       const scalar_t* gradOutputData,
       const int64_t* indicesData,
-      int features,
+      int64_t features,
       int64_t gradOutputSize,
-      int out_cf_channel_stride,
-      int in_cf_channel_stride,
-      int out_nbatch_stride,
-      int in_nbatch_stride)
+      int64_t out_cf_channel_stride,
```
guangyey: cc @jianyizh

jianyizh (changes requested):
same as #3558, I prefer keep two instance of int32 and int64. We can dispatch according to shape

guangyey:
> same as #3558, I prefer keep two instance of int32 and int64. We can dispatch according to shape

Good catch!

jianyizh:
> same as #3558, I prefer keep two instance of int32 and int64. We can dispatch according to shape

@Stonepia @liangan1 I see copilot will read some skills. Can we update the skill so that copilot can be aware of such cases that may hurt performance? Maybe just ask copilot will this PR hurt perf?

Stonepia:
> > same as #3558, I prefer keep two instance of int32 and int64. We can dispatch according to shape
> 
> @Stonepia @liangan1 I see copilot will read some skills. Can we update the skill so that copilot can be aware of such cases that may hurt performance? Maybe just ask copilot will this PR hurt perf?

I think currently, the copilot does not have the skill for performance related issues. We need to manually give it skills. So when reviewing, it will aware such issues. As you could see there is a check list in `.github/skills/xpu-ops-pr-review/references/review-checklist.md`, we could enrich it. Currently it basically all written by AI, so only workflow exists.

jianyizh:
> > > same as #3558, I prefer keep two instance of int32 and int64. We can dispatch according to shape
> > 
> > 
> > @Stonepia @liangan1 I see copilot will read some skills. Can we update the skill so that copilot can be aware of such cases that may hurt performance? Maybe just ask copilot will this PR hurt perf?
> 
> I think currently, the copilot does not have the skill for performance related issues. We need to manually give it skills. So when reviewing, it will aware such issues. As you could see there is a check list in `.github/skills/xpu-ops-pr-review/references/review-checklist.md`, we could enrich it. Currently it basically all written by AI, so only workflow exists.

@Stonepia  I mean just add one line to ask copilot if this pr hurt perf， no extra skill needed

Stonepia:
> > > > same as #3558, I prefer keep two instance of int32 and int64. We can dispatch according to shape
> > > 
> > > 
> > > @Stonepia @liangan1 I see copilot will read some skills. Can we update the skill so that copilot can be aware of such cases that may hurt performance? Maybe just ask copilot will this PR hurt perf?
> > 
> > 
> > I think currently, the copilot does not have the skill for performance related issues. We need to manually give it skills. So when reviewing, it will aware such issues. As you could see there is a check list in `.github/skills/xpu-ops-pr-review/references/review-checklist.md`, we could enrich it. Currently it basically all written by AI, so only workflow exists.
> 
> @Stonepia I mean just add one line to ask copilot if this pr hurt perf， no extra skill needed

Oh that's possible. Will add that.

pbielak:
@jianyizh Added the template parameter of the index dtype - waiting for CI signal

jianyizh (approved):

chuanqi129:
✅ PR has been successfully merged by @pbielak (force merge).

PR #3567, opened 2026-05-06
Update numpy for python 3.10 and cooperlake CPU

jmamzax:
Part of the https://github.com/intel/torch-xpu-ops/issues/2618 issue. There is a bug in OpenBLAS 0.3.20 used in numpy 1.23.* that appear on CI configuration in linalg tests. Pytorch for CI uses python 3.10 and numpy 1.23.2. By adding extra step to setup Linux test environment the numpy is updated to 1.24.* version which also uses new OpenBLAS release 0.3.21 whit fix for the problem on cooperlake.


.github/actions/linux-testenv/action.yml
```diff
@@ -192,6 +192,28 @@ runs:
           # yolov requires numpy>=1.23
           pip install -U numpy==1.26.4
         fi
+    - name: Update numpy
+      shell: bash -xe {0}
+      run: |
+        # TODO: Remove this step once Python 3.10 is no longer supported.
+        if [[ "${{ inputs.python }}" == "3.10" ]]; then
+          OPENBLAS_DETECT_OUTPUT=$(OPENBLAS_VERBOSE=2 python -c "import numpy as np; np.dot(np.ones(1), np.ones(1))" 2>&1 > /dev/null)
+          if echo "${OPENBLAS_DETECT_OUTPUT}" | grep -qi "cooperlake"; then
```
Silv3S: Let's just update numpy globally, without CPU arch dependency. It might be problematic if we start comparing results from two different runners that will have different numpy version.

This is temporary w/a anyway, as Python 3.10 will be gone soon (officially Oct, but if upstream switch faster, we will probably do the same)

.github/actions/linux-testenv/action.yml
```diff
@@ -192,6 +192,28 @@ runs:
...
+            echo "OpenBLAS detected CPU as Cooperlake; checking numpy for update"
+            CURRENT_NUMPY=$(python -c "import numpy as np; print(np.__version__)")
+            if [[ "${CURRENT_NUMPY}" == 1.23.* ]]; then
+              echo "Detected numpy ${CURRENT_NUMPY}; upgrading to 1.24.*"
+              pip install -U "numpy>=1.24,<1.25"
```
Silv3S:
```suggestion
              pip install -U "numpy==1.24.?"
```
Pick one patch version

.github/actions/linux-testenv/action.yml
```diff
@@ -192,6 +192,22 @@ runs:
...
+          if [[ "${CURRENT_NUMPY}" == 1.23.* ]]; then
+            echo "Detected numpy ${CURRENT_NUMPY}; upgrading to 1.24.4"
+            pip install -U "numpy==1.24.4"
+            python -c "import numpy as np; print(f'numpy version after conditional upgrade: {np.__version__}')"
+          else
+            echo "numpy is ${CURRENT_NUMPY}; skipping upgrade because version is not 1.23.*"
+          fi
+        else
+          echo "Skipping Update numpy because python is not 3.10 (python=${{ inputs.python }})"
+        fi
```
Silv3S, line 214:
```suggestion
        if [[ "${{ inputs.python }}" == "3.10" ]]; then
            pip install -U "numpy==1.24.4"
        fi
```

Silv3S (approved):

.github/actions/linux-testenv/action.yml
Silv3S: Consider just reinstalling numpy when Py3.10 is detected without other checks

chuanqi129 (approved):

.github/actions/linux-testenv/action.yml
jmamzax: In previous step there is:
```
if [ "${{ inputs.suite }}" != "None" ];then
          # To install numpy 1.x for benchmarks as CUDA
          # yolov requires numpy>=1.23
          pip install -U numpy==1.26.4
        fi
```
So I don't want to update NumPy if not 1.23.*

kdrozd-dev (approved):

chuanqi129:
✅ PR has been successfully merged by @Silv3S.

PR #3577, opened 2026-05-07
Relax XPU Muon state_dict parity tolerance

gplutop7:
Fixes:

Slightly relax the XPU Muon state_dict test tolerance to account for small float32 CPU-XPU numerical

### DIFF
--- a/test/xpu/test_optim_xpu.py
+++ b/test/xpu/test_optim_xpu.py
@@ -254,13 +254,17 @@ def _test_state_dict_with_cuda_params(self, device, dtype, optim_info):
         for _ in range(5):
             optimizer.step(closure)
             optimizer_cuda.step(closure)
-            self.assertEqual(params, params_cuda, atol=2e-5, rtol=1.3e-6)
-            self.assertEqual(
-                optimizer.state_dict(),
-                optimizer_cuda.state_dict(),
-                atol=2e-5,
-                rtol=1.3e-6,
-            )
+            if optim_cls.__name__ == "Muon":
+                self.assertEqual(params, params_cuda, atol=2e-5, rtol=1.3e-6)
+                self.assertEqual(
+                    optimizer.state_dict(),
+                    optimizer_cuda.state_dict(),
+                    atol=2e-5,
+                    rtol=1.3e-6,
+                )
+            else:
+                self.assertEqual(params, params_cuda)
+                self.assertEqual(optimizer.state_dict(), optimizer_cuda.state_dict())
 
 
 TestOptimRenewed.test_state_dict_with_cuda_params = _test_state_dict_with_cuda_params
--- a/test/xpu/test_optim_xpu.py
+++ b/test/xpu/test_optim_xpu.py
@@ -258,7 +258,7 @@ def _test_state_dict_with_cuda_params(self, device, dtype, optim_info):
             self.assertEqual(
                 optimizer.state_dict(),
                 optimizer_cuda.state_dict(),
-                atol=1.9e-5,
+                atol=2e-5,
                 rtol=1.3e-6,
             )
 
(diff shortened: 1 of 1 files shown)


BBBela (commented):
This change increases tolerance for all devices, although we could theoretically add additional condition to apply override only for xpu.
Have you thought about adding tolerance override here?: [common_optimizers.py](https://github.com/pytorch/pytorch/blob/9fad81aaeed8d11bf2f4b0420797a0daa30be29b/torch/testing/_internal/common_optimizers.py#L1922-L1942)
I mean this is similar mechanism to `common_methods_invocations.py` but for optimizers.
We could add something that to the mentioned code fragment:
```python
  decorators=(
      DecorateInfo(
          toleranceOverride({torch.float32: tol(atol=2e-5, rtol=1.3e-6)}),
          "TestOptimRenewed",
          "test_state_dict_with_cuda_params",
          device_type="xpu",
      ),
  ),
```

It would be xpu-only change, not affecting other backends and not changing the test body.
I know that this would affect all `self.assertEqual` calls in that functions, but the tolerance increase is really small from `1e-5` to `2e-5` atol and I don't suppose that would be the problem.

Let me know what you think! 😄

gplutop7:
> This change increases tolerance for all devices, although we could theoretically add additional condition to apply override only for xpu. Have you thought about adding tolerance override here?: [common_optimizers.py](https://github.com/pytorch/pytorch/blob/9fad81aaeed8d11bf2f4b0420797a0daa30be29b/torch/testing/_internal/common_optimizers.py#L1922-L1942) I mean this is similar mechanism to `common_methods_invocations.py` but for optimizers. We could add something that to the mentioned code fragment:
> 
> ```python
>   decorators=(
>       DecorateInfo(
>           toleranceOverride({torch.float32: tol(atol=2e-5, rtol=1.3e-6)}),
>           "TestOptimRenewed",
>           "test_state_dict_with_cuda_params",
>           device_type="xpu",
>       ),
>   ),
> ```
> 
> It would be xpu-only change, not affecting other backends and not changing the test body. I know that this would affect all `self.assertEqual` calls in that functions, but the tolerance increase is really small from `1e-5` to `2e-5` atol and I don't suppose that would be the problem.
> 
> Let me know what you think! 😄

Thanks, but I’d prefer to keep this local to the Muon test case. Moving it to common_optimizers.py would broaden the change and affect all assertEqual calls in that test body, which seems unnecessary here.

BBBela:
> Thanks, but I’d prefer to keep this local to the Muon test case. Moving it to common_optimizers.py would broaden the change and affect all assertEqual calls in that test body, which seems unnecessary here.

I would like to consult this approach with @guangyey before moving forward with review.
I still see two main disadvantages of current approach when compared to putting the override to `common_optimizers.py`:
- The solution is not scalable. Let's consider more optimizers would require tolerance adjustment for XPU. Then we would end up with series of `if-elif` statements in the test function body. And I believe that test function body should contain only the tested logic to make it clear what is being tested.
- Current approach changes tolerance for all devices that are being tested with this test function. I think we should not change the tolerance for other devices and CPU when it is not necessary.

In most cases to override the tolerance we use the decorator just above the test function or we add tolerance override for specific operator in `common_methods_invocations.py` and I think we should follow this approach for optimizers also.

@guangyey - what do you think?

guangyey (commented):
I think we should do:
1. root cause why `Muon` has accuracy lose on XPU?
2. if it is reasonable (such as deterministic, bfloat16, etc), a better place to skip it is https://github.com/pytorch/pytorch/blob/7ff1adf1d1ba890dc2e4e4eb02cb74e61fe4aa1b/torch/testing/_internal/common_optimizers.py#L1923 using  `DecorateInfo` and `toleranceOverride`

gplutop7:
I closed this PR. After discussion, I created a new PR: https://github.com/pytorch/pytorch/pull/183903

Issue #3529, opened 2026-04-29
[Bug Skip] test_multiprocessing_xpu — 5 cases fail on XPU (is_shared, cuda_bad_call, meta_simple, autograd_fine_with_spawn, integer_parameter_serialization_cpu)


daisyden:
## Bug Description

Five upstream `test_multiprocessing.py::TestMultiprocessing` cases fail on XPU
when run via the XPU-port wrapper `test/xpu/test_multiprocessing_xpu.py`.
Failures cover three distinct underlying gaps but are tracked as one issue
because they all live in the same ported test class and the dynamic-skip
mechanism filters by `Cases:` lines. Filing them together keeps `Cases:`
line drift to a minimum.

## Affected Tests

Cases:
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_multiprocessing_xpu.TestMultiprocessing,test_is_shared_cuda~~
op_ut,third_party.torch-xpu-ops.test.xpu.test_multiprocessing_xpu.TestMultiprocessing,test_integer_parameter_serialization_cpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_multiprocessing_xpu.TestMultiprocessing,test_meta_simple
op_ut,third_party.torch-xpu-ops.test.xpu.test_multiprocessing_xpu.TestMultiprocessing,test_autograd_fine_with_spawn
op_ut,third_party.torch-xpu-ops.test.xpu.test_multiprocessing_xpu.TestMultiprocessing,test_cuda_bad_call

### 1. `test_is_shared_cuda`

The PR overrides this test with `_test_is_shared_xpu` which asserts that
`torch.randn(5, 5).xpu().is_shared()` returns `True`. On CUDA, GPU tensors
are auto-shared (CUDA IPC semantics); XPU does not auto-share fresh tensors.

```
AssertionError: False is not true
File "test_multiprocessing_xpu.py", line 82
    self.assertTrue(t.is_shared())
```

XPU semantics may legitimately differ from CUDA here (no IPC by default).
Either the assertion needs adjusting to XPU behaviour or XPU needs to mark
device tensors shared after `.xpu()` (matching CUDA).

### 2. `test_integer_parameter_serialization_cpu`

Upstream test, runs `_test_integer_parameter_serialization(device="cpu")`.
The CPU-device variant fails:

```
AssertionError: Scalars are not equal!
Expected 1 but got 0.
Failed to serialize successfully for "cpu" device!
[...]

PR #3592, opened 2026-05-08
Fix XPU multiprocessing spawn import and fork exception helper

gplutop7:
Fixes:
https://github.com/intel/torch-xpu-ops/issues/3529

Add explicit test path setup for spawn subprocess imports and add an XPU-specific queue_get_exception helper so the forked subprocess path validates XPU behavior instead of CUDA.

Fixed tests:
cd third_party/torch-xpu-ops/test/xpu
pytest -v test_multiprocessing_xpu.py::TestMultiprocessing::test_integer_parameter_serialization_cpu
pytest -v test_multiprocessing_xpu.py::TestMultiprocessing::test_meta_simple
pytest -v test_multiprocessing_xpu.py::TestMultiprocessing::test_autograd_fine_with_spawn
pytest -v test_multiprocessing_xpu.py::TestMultiprocessing::test_cuda_bad_call


test/xpu/test_multiprocessing_xpu.py
```diff
@@ -20,12 +21,25 @@
...
-    from .xpu_test_utils import XPUPatchForImport
+    from .xpu_test_utils import ensure_pytorch_test_path, XPUPatchForImport
 
-with XPUPatchForImport(False):
-    from test_multiprocessing import queue_get_exception, TestMultiprocessing
+with XPUPatchForImport(False) as patcher:
+    from test_multiprocessing import TestMultiprocessing
+
+
+ensure_pytorch_test_path(patcher)
```
tszulist-hbn: Non-blocking: The pattern of using a context manager, then immediately adding paths back that were just removed is confusing. The context manager adds test paths in `__enter__`, removes them in `__exit__`, then `ensure_pytorch_test_path` adds them back. While functionally correct, this is unnecessarily indirect. Consider inlining:

```
with XPUPatchForImport(False) as patcher:
    from test_multiprocessing import TestMultiprocessing

test_dir = os.path.abspath(patcher.test_package[0])
if test_dir not in sys.path:
    sys.path.append(test_dir)
```

Alternatively, if you anticipate `ensure_pytorch_test_path` useful in the future, consider a cleaner contract by adding a small property on the patcher (e.g. patcher.root_test_dir) or to pass the path explicitly. Minor maintainability nit.

test/xpu/test_multiprocessing_xpu.py
```diff
@@ -20,12 +21,25 @@
...
+
+
+def queue_get_exception(inqueue, outqueue):
```
tszulist-hbn, line 36: The `queue_get_exception` function parameter `inqueue` is unused. Is it for signature compatibility with other test targets?

test/xpu/xpu_test_utils.py
```diff
@@ -1266,3 +1266,9 @@ def launch_test(test_case, skip_list=None, exe_list=None):
             f"pytest --junit-xml=./op_ut_with_all.{module_name}.xml " + test_case
         )
     return os.system(test_command)
+
+
+def ensure_pytorch_test_path(patcher):
+    test_dir = os.path.abspath(patcher.test_package[0])
+    if test_dir not in sys.path:
+        sys.path.append(test_dir)
```
tszulist-hbn: The function name `ensure_pytorch_test_path` uses the `patcher` object but doesn't indicate it requires the patcher to have been initialized. A more descriptive name like `append_test_path_from_patcher` would clarify the dependency.

tszulist-hbn (approved):

test/xpu/test_multiprocessing_xpu.py
gplutop7: Yes - it’s intentional for signature compatibility with upstream.

test/xpu/test_multiprocessing_xpu.py
gplutop7: Thanks for the suggestion - I applied the cleaner contract and now pass the path explicitly

test/xpu/xpu_test_utils.py
gplutop7: Addressed by changing the helper contract to accept an explicit path.

pbielak (approved):

gplutop7:
linux-ut (op_ut) tests are currently failing with the following environment/library-loading error:
OSError: Could not load this library: /tmp/xpu-tool/Python/3.10.20/x64/lib/python3.10/site-packages/torch/build/lib/libtorchbind_test.so

This failure is unrelated to the changes introduced in this PR. To avoid blocking merge due to an external CI issue, the disable_all label has been added.

chuanqi129:
✅ PR has been successfully merged by @gplutop7.

PR #3627, opened 2026-05-11
[CI] Automatically label PRs created by AI with `ai_generated` label

Stonepia:
We found that when we assigned copilot to fix, it doesn't label the PR, so it won't trigger the reproducer running CI. This PR is to ensure the label be marked, even if it doesn't have the reproducer.

### DIFF
--- a/.github/workflows/label-ai-prs.yml
+++ b/.github/workflows/label-ai-prs.yml
@@ -9,6 +9,7 @@ jobs:
     if: github.event.pull_request.user.login == 'copilot[bot]'
     runs-on: ubuntu-latest
     permissions:
+      issues: write
       pull-requests: write
     steps:
       - uses: actions/github-script@v7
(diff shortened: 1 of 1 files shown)


.github/workflows/label-ai-prs.yml
```diff
@@ -0,0 +1,23 @@
+name: Label AI-generated PRs
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  add-label:
```
chuanqi129, line 8: Does it mean all copilot created PR will be labeled with `ai_generated`? Based on current flow, all `ai_generated` label PR need include reproducer case or UT case. Do we align with it?

Stonepia: I think we should anyway trigger the reproducer CI for the copilot. The reproducer generation, in theory should be restricted in skills to add the reproducer if needed. I checked random 8 issues, only 3 are not with the reproducer because it skip the tests.

1. If no reproducer: then running that ci stage should automatically pass. So it won't affect later CIs.
2. If have reproducer, then this loop is our desired loop.

So I think we could always label the `ai_generated` for all copilot PRs, this won't affect us.

Issue #3628, opened 2026-05-11
Improve random label behavior


jmamzax:
### 🚀 The feature, motivation and pitch

As it is now the random label prevent from closing issue when it pass. But also preventing from inserting comment with pass message so after adding random to issue there is no update from CI when issue pass or fail. Which in some cases is problematic as issue reproduce in specific condition.

### Alternatives

Add new check to ut_result_check.sh script to insert comment with information if test cases from issue marked as random fail or pass. This will give us clear history of nightly runs and also we will be able quickly check how often the problem occur.

### Additional context

I had issue: https://github.com/intel/torch-xpu-ops/issues/3195 marked as random and I was not able to find single nightly run that fail but I also don't check every run. Only fail is a pull request run. So with such nightly history I will be able to quickly tell if the test fail again.

jmamzax:
After proposing this I also create a script to download artifacts from github and see if test fail or pass so this can be solved in other way. But this solution give as a clear run history in github issue without additional steps. The disadvantage is that each run will insert new comment into issue.

jmamzax:
Closing as the same can be done with github cli.

PR #3640, opened 2026-05-12
Check if random issue pass or fail in a run and add comment

jmamzax:
Address problem described in: https://github.com/intel/torch-xpu-ops/issues/3628


jmamzax:
Closing as the same can be done with github cli.

Issue #3647, opened 2026-05-12
Improve errors logging when run_test_with_skip.py is executed


jmamzax:
### 🚀 The feature, motivation and pitch

When tests are executed stderr is redirected to a file *_test_error.log. But it is not shown in console so GitHub will not saved them. Also because two separate files are created *_test_error.log and *_test.log without any timestamps there is no possibility to see what test cause what error or what happened during each test execution. This is not a problem when the same error is captured by python and show in test summary but that's no the case for some tests. Also if test is random sometime is hard to reproduce a problem but if occur on CI, it will be a great addition to debugging if there is a log error with additional information.

```
python run_test_with_skip.py \
          2> ${{ github.workspace }}/ut_log/op_ut/op_ut_with_skip_test_error.log | \
          tee ${{ github.workspace }}/ut_log/op_ut/op_ut_with_skip_test.log
```

### Alternatives

1) First fix is to just save all the data in one file. So redirect 2>&1. Main log file for Linux op_ut is ~83MB so adding an error log ~180KB is not a big problem in terms of log size.
2) Second idea is to leave two separate logs but use for both of them tee and redirect stderr to console to have all needed data in one GitHub log.

```
python run_test_with_skip.py \
          > >(tee ${{ github.workspace }}/ut_log/op_ut/op_ut_with_skip_test.log) \
          2> >(tee ${{ github.workspace }}/ut_log/op_ut/op_ut_with_skip_test_error.log >&2)
```

PR #3648, opened 2026-05-12
Save stderr in separate file and show in console

jmamzax:
One of the proposed solution from https://github.com/intel/torch-xpu-ops/issues/3647


kdrozd-dev (approved):
Looks good to me. Seems like a useful improvement as it makes debugging obscure CI issues easier.

kdrozd-dev:
Could this also be done to windows ut runs?

PR #3660, opened 2026-05-13
Add runtime env to use copy kernel on PVC

frost-intel:
As a follow-up to https://github.com/intel/torch-xpu-ops/pull/2095, this PR adds a runtime variable, `TORCH_XPU_USE_COPY_ENGINE`, which allows the use of the SYCL queue. This is relevant on PVC where copy kernel is faster on small benchmarks, but a flexible mechanism is desired for experimentation.

From https://github.com/intel/torch-xpu-ops/pull/2095#issuecomment-4427769280

> Addressing Level Zero (L0) performance issues on PVC is very difficult. In some workload scenarios, it is important to evaluate whether an overall end-to-end (E2E) performance benefit can still be achieved, even when there is a drop in raw D2D copy performance.
> Our main PVC customer deploys a single PyTorch build as part of the system image across compute nodes in a large-scale supercomputer, validated to scale to thousands of nodes. Maintaining multiple PyTorch builds to experiment with different behaviors is not a viable option for production releases.
> A well-documented runtime flag provides a practical and flexible mechanism to enable such experimentation without requiring changes to the deployed software stack. This allows users to explore potential E2E performance gains in real workloads while preserving deployment simplicity and consistency.

Regarding documentation: Ideally we would have a doc similar to https://github.com/pytorch/pytorch/blob/main/docs/source/mps_environment_variables.md which would outline the existing runtime variables (such as those in ProcessGroupXCCL). Since there is no current document, the only documentation is in a comment, and a full documentation is marked as a TODO.

@pkourdis


frost-intel:
@CuiYifeng @jianyizh Can you review this related to #2095

jianyizh (approved):

CuiYifeng (approved):

CuiYifeng:
@frost-intel Please check the terms used in the title and description. "forces the use of **the copy engine instead of the SYCL queue**" is inconsistent with that `TORCH_XPU_USE_COPY_ENGINE` is false in default. How about replacing `copy engine` with `copy kernel` in the title and description?

frost-intel:
Confirmed that UT failures and build failure are not related, due to other issues in the stack.

chuanqi129:
✅ PR has been successfully merged by @dvrogozh (force merge).

PR #3690, opened 2026-05-15
Fix DecorateInfo test-name matching for XPU overrides via register_test

gplutop7:
Fixes:

This PR fixes a mismatch between overridden XPU test function names and DecorateInfo name matching.

What changed:
- Added a shared register_test() helper in XPU test utilities.
- Replaced direct Class.test_* = _test_* assignments with register_test(...) in affected XPU tests.


chuanqi129:
@copilot New UT failures detected in **op_ut** tests (4 total). Please check these new failures and analyze whether they are caused by the PR changes.

**Important: Do NOT update the plan in the PR description directly. Use reply comments to update the status.**

Job log: https://github.com/intel/torch-xpu-ops/actions/runs/25928824349
Full new failure report: https://github.com/intel/torch-xpu-ops/actions/runs/25928824349/artifacts/7032504143

**New failures:**

| Category | Class name | Test name | Status | Message |
| --- | --- | --- | --- | --- |
| op_ut | third_party.torch-xpu-ops.test.xpu.dynamo.test_misc_xpu.MiscTests | test_const_dict_variable_python_type | failed | AttributeError: '_thread._local' object has no attribute 'current_tx'
| op_ut | third_party.torch-xpu-ops.test.xpu.profiler.test_profiler_xpu.TestProfiler | test_disable_external_correlation | failed | AssertionError: False is not true
| op_ut | third_party.torch-xpu-ops.test.xpu.profiler.test_profiler_xpu.TestProfiler | test_dynamic_toggle | failed | AssertionError: False is not true
| op_ut | third_party.torch-xpu-ops.test.xpu.test_custom_ops_xpu.TestCustomOpAPI | test_no_grad_skips_autograd | failed | AssertionError: Scalars are not equal!

Instructions:
1. Download the full new failure report artifact for complete error messages
2. Analyze whether the failures are caused by PR changes
3. If caused by PR changes, fix those failures on this PR branch directly
4. Add failing test cases with pytest commands in the PR description for the reproducer job to use
5. Use reply comments to report analysis status - do NOT modify the PR description plan directly

test/xpu/xpu_test_utils.py
```diff
@@ -626,7 +627,30 @@ def ModuleTest_test_xpu(self, test_case):
         self.test_noncontig(test_case, xpu_module, xpu_input_tuple)
 
 
-ModuleTest.test_cuda = ModuleTest_test_xpu
+def register_test(cls, func, *, name=None):
```
BBBela: The current signature hides its contract — you have to read the body to know that the function name must start with `test_` and that the target method name is derived from it. A three-argument form like that:
```python
def register_test(original_test_class, original_test_func, overriden_func):
```
would make the relationship explicit at the call site and removes all the name-derivation logic entirely.

What do you think?

BBBela (commented):
The PR converts many test registrations that don't appear to have confirmed
`DecorateInfo` breakage.
Given the stated plan to port all XPU tests to PyTorch directly, it might be worth
scoping this PR to only the confirmed broken cases to keep the PR focused.
It looks like most of those changes are preventive rather than fixing existing problems - is that correct?

test/xpu/xpu_test_utils.py
gplutop7: Thanks, I had the same thought initially.

A fully explicit 3-argument form makes the contract clearer, but it adds cost to the common case. Most overrides follow _test_* -> test_*, so requiring the target name everywhere would duplicate that repeatedly and make renames more fragile.

I also considered introducing the helper only for new/broken cases and leaving existing direct assignments as-is. But that would create two inconsistent patterns in the codebase: some tests using the helper, others the old direct assignment. That invites confusion for future contributors — which pattern should they copy? — and risks repeating the same fix in different ways.

So I went with a hybrid approach:

`def register_test(cls, override_func, *, target_name=None):`

This keeps the common case concise while requiring target_name for non-standard mappings. And I updated all existing registrations to use it consistently, so the pattern is now uniform across the test suite.

Does that resolve the scope concern? Do you see any risk in standardizing on this approach repo-wide?

chuanqi129:
@copilot New UT failures detected in **op_ut** tests (1 total). Please check these new failures and analyze whether they are caused by the PR changes.

**Important: Do NOT update the plan in the PR description directly. Use reply comments to update the status.**

Job log: https://github.com/intel/torch-xpu-ops/actions/runs/26033399176
Full new failure report: https://github.com/intel/torch-xpu-ops/actions/runs/26033399176/artifacts/7066563163

**New failures:**

| Category | Class name | Test name | Status | Message |
| --- | --- | --- | --- | --- |
| op_ut | third_party.torch-xpu-ops.test.xpu.functorch.test_vmap_xpu.TestVmapOperatorsOpInfoXPU | test_op_has_batch_rule_count_nonzero_xpu_float32 | failed | Unexpected success

Instructions:
1. Download the full new failure report artifact for complete error messages
2. Analyze whether the failures are caused by PR changes
3. If caused by PR changes, fix those failures on this PR branch directly
4. Add failing test cases with pytest commands in the PR description for the reproducer job to use
5. Use reply comments to report analysis status - do NOT modify the PR description plan directly

test/xpu/xpu_test_utils.py
BBBela: Thank you for response and clarification! 😃

I understand your point of view and it makes sense. I may even agree on the repo-wide scope to be consistent. However, regarding the function signature I am still not convinced, because the main argument for current approach is that it will make each registration several lines shorter, while making the contract less clear. So we trade the number of code lines at the cost of simplicity and clarity - at least that is my point of view.

But to be clear - those are my personal, subjective design preferences, so I am not going to block anything on this PR and I am not going to request any changes, I am just not approving something I do not fully agree with. So feel free to request others for review and proceed with this PR 😉

gplutop7: Thanks a lot for the thoughtful feedback and for the constructive discussion.

You were absolutely right about the clarity tradeoff, and I decided to follow your suggestion. I updated the helper to the fully explicit 3-argument form so the contract is now clear at each call site.

I also migrated all registrations in this PR to match that explicit style (including non-standard mappings), so the pattern stays consistent repo-wide while keeping the interface unambiguous

BBBela: Here is some analysis from my point of view:
1. The new `register_test` function looks good. It performs detailed preventive checks and registers overridden test case only when everything is okay while raising an exception when anything is not ok. That is valid and preventive enhancement. Function's signature is clear and simple. Good change overall.
2. This PR changes all current "_simple_" registrations to registrations with `register_test` function. That makes all the registrations aligned, leaving no ambiguity about how to handle future changes. All future registrations should use `register_test` function. That approach is good in general when developing evolving software. I have some doubts only because those changes affect tests that are going to be migrated to PyTorch and currently affected files will be removed in near future. But overall, I like the idea to make everything aligned, leaving no ambiguity or several possible options.

Additionally, all those mechanisms, porting tests and the architecture was designed mostly by the PRC team and this change is potentially affecting many developers and is changing many files.

Taking all the above into consideration, I would prefer the PRC team to decide if we want to introduce the comprehensive enhancement from this PR or keep the change the simplest (the fix is really required only for one test case).

In short, I think that current shape of this PR is functionally correct, the idea behind it is in my opinion valid and I treat it as a positive enhancement. But I would prefer to leave the decision for someone responsible for the whole testing part of `torch-xpu-ops`.

If someone from the PRC team approves, you have my second approval! 😉

gplutop7:
I analyzed all test overrides and split them into three groups:
- Group 1 (32 cases): overrides where the target name can be derived by stripping a leading underscore (e.g., test_x <- _test_x).
- Group 2 (10 cases): overrides that require explicit name mapping (non-trivial renames, e.g., different suffix/prefix or double-underscore naming).
- Group 3 (65 cases): overrides without a leading underscore, but still with a function-name mismatch (test_* method name differs from the assigned function’s __name__.

Key point: the issue is not test discovery (tests still run), but decorator matching by test name.
In PyTorch, DecorateInfo / skip / xfail logic matches against test_name. If an override leaves function.__name__ inconsistent with the method name, decorators may not apply as intended.

gplutop7:
We decided not to roll out this change broadly in this PR. Instead, we will apply a targeted fix only to the tests where the issue actually occurs.

The main issue is not test discovery, since the tests are still being executed. The problem is how decorators are matched by test name. In PyTorch, DecorateInfo, skip, and xfail rules are resolved against test_name. If an override keeps function.name out of sync with the actual method name, those decorators can be matched incorrectly and may not be applied as expected.

PR #3706, opened 2026-05-19
[CI] Add auto-label workflow for PR CI optimization

chuanqi129:
## Summary

Add an auto-labeling system that determines `disable_*` CI labels based on PR file paths, reducing unnecessary CI load.

## Changes

- **`.github/skills/xpu-pr-auto-labeling/SKILL.md`** — Maintainable rule set mapping file paths to labels. Scoped to labels used by `pull.yml` only (`disable_all/ut/e2e/distributed/win/build`). Handles `src/xccl/` correctly (no `disable_distributed` since it's the distributed backend).
- **`.github/workflows/auto_label.yml`** — Workflow triggered on PR open/sync/reopen/ready_for_review that:
  1. Gets changed file list
  2. Calls GitHub Models API with skill rules as context for label determination
  3. Falls back to deterministic shell logic if API fails (covers all major path patterns)
  4. Merges new labels with existing (union, never removes)
  5. Cancels in-flight `pull` workflow, then toggles PR draft->ready to trigger fresh event with updated labels

### DIFF
--- a/.github/skills/xpu-pr-auto-labeling/SKILL.md
+++ b/.github/skills/xpu-pr-auto-labeling/SKILL.md
@@ -79,8 +79,7 @@ Rules are evaluated top-to-bottom; use the FIRST matching rule set.
 
 - `src/ATen/` only (kernel/operator code, no xccl):
   - Base: `disable_e2e`, `disable_distributed`
-  - If changes are only in `src/ATen/native/xpu/sycl/` (pure kernel logic, no new ops): add `disable_win`
-  - If changes touch `src/ATen/native/sparse/`: add `disable_win`
+  - Do NOT add `disable_win` — kernel fixes may be platform-specific (e.g. a fix specifically for Windows)
 
 - `src/xccl/` involved (distributed communication backend):
   - Base: `disable_e2e` only — do NOT add `disable_distributed` (xccl changes must run distributed tests)
--- a/.github/skills/xpu-pr-auto-labeling/SKILL.md
+++ b/.github/skills/xpu-pr-auto-labeling/SKILL.md
@@ -1,0 +1,144 @@
+---
+name: xpu-pr-auto-labeling
+description: >
+  Rules for automatically determining which disable_* labels to apply to a PR
+  based on the file paths changed. Used by the auto-label workflow.
+---
+
+# PR Auto-Labeling Rules
+
+## Purpose
+
+Determine which `disable_*` CI labels to apply to a PR based on the files changed.
+These labels control which CI jobs run, reducing unnecessary CI load.
+
+## Available Labels
+
+| Label | Effect |
+|-------|--------|
+| `disable_all` | Skip all CI test jobs, only run lint |
+| `disable_ut` | Skip non-distributed UT jobs |
+| `disable_e2e` | Skip e2e (inductor benchmark) jobs |
+| `disable_distributed` | Skip distributed UT jobs |
+| `disable_win` | Skip Windows CI jobs |
+| `disable_build` | Skip source build, use nightly wheel |
+| `disable_accelerate` | Skip accelerate test job |
+| `disable_transformers` | Skip transformers UT job |
+
+## Decision Rules
+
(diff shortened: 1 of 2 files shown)


.github/skills/xpu-pr-auto-labeling/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Rule 5: Kernel/operator source changes (src/)
+
+**Condition:** Changed files include `src/ATen/` or `src/xccl/`.
+
+**Labels:** Depends on which subdirectory:
+
+- `src/ATen/` only (kernel/operator code, no xccl):
+  - Base: `disable_e2e`, `disable_distributed`
+  - If changes are only in `src/ATen/native/xpu/sycl/` (pure kernel logic, no new ops): add `disable_win`
```
Stonepia: This line seems not safe. If there is one kernel change that fixed specifically for win, we will skip the win test.

.github/workflows/auto_label.yml
```diff
@@ -0,0 +1,309 @@
...
+            HAS_WORKFLOW=false
+            HAS_GITHUB_META=false
+            HAS_YAML=false
+            ONLY_SKIP_EXPECT=true
+            HAS_WIN_TEST=false
+            HAS_SYCL_ONLY=true
+            HAS_SPARSE=false
+
+            while IFS= read -r file; do
+              [ -z "$file" ] && continue
```
Stonepia, line 143: So we have both the LLM for labelling and the rule labelling, right?

.github/workflows/auto_label.yml
```diff
@@ -0,0 +1,309 @@
+name: auto-label
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, ready_for_review]
```
Stonepia, line 5: This `synchronize` trigger might trigger the workflow of auto_labeling on every push. I wonder if we only trigger  on the `opened`, `ready_for_review` , `reopened` for smaller trigger frequency?

Stonepia (approved):
Overall LGTM. The only thing I wonder is about label remove, currently we never automatically remove the label, seems fine by now, but not sure it would fit the future need.

.github/workflows/auto_label.yml
chuanqi129: Yes

.github/workflows/auto_label.yml
```diff
@@ -0,0 +1,309 @@
...
+          echo "Changed files:"
+          echo "$FILES"
+
+      - name: Determine labels via GitHub Models
+        id: determine-labels
+        env:
+          GH_TOKEN: ${{ secrets.MERGE_TOKEN }}
+        run: |
+          SKILL_CONTENT=$(cat .github/skills/xpu-pr-auto-labeling/SKILL.md)
+          FILES="${{ steps.changed-files.outputs.files }}"
```
ZhaoqiongZ: Direct ${{ }} interpolation inside a run block allows command injection via malicious filenames. Since this workflow uses pull_request_target, this is a high-severity risk.

Please pass the data safely via env variables instead:

```
env:
  CHANGED_FILES: ${{ steps.changed-files.outputs.files }}
  PR_IS_DRAFT: ${{ github.event.pull_request.draft }}
run: |
  FILES="$CHANGED_FILES"
  IS_DRAFT="$PR_IS_DRAFT"
```

.github/skills/xpu-pr-auto-labeling/SKILL.md
chuanqi129: addressed

ZhaoqiongZ (approved):

.github/workflows/auto_label.yml
chuanqi129: keep sync in case PR update change the modify scope

.github/workflows/auto_label.yml
chuanqi129: addressed

chuanqi129:

Issue #3477, opened 2026-04-26
[upstream_ut] export/test_torchbind.py NotImplementedError: The operator '_TorchScriptTesting::queue_push' is not currently implemented for the XPU device


zxd1997066:
### 🐛 Describe the bug

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.export.test_torchbind_xpu.TestCompileTorchbind,test_export_obj_torchbind_op_with_autocast_device_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.export.test_torchbind_xpu.TestCompileTorchbind,test_compile_obj_torchbind_op_with_autocast_device_xpu_backend_inductor
op_ut,third_party.torch-xpu-ops.test.xpu.export.test_torchbind_xpu.TestCompileTorchbind,test_compile_obj_torchbind_op_with_autocast_device_xpu_backend_aot_eager
op_ut,third_party.torch-xpu-ops.test.xpu.export.test_torchbind_xpu.TestCompileTorchbind,test_compile_obj_torchbind_op_with_autocast_device_xpu_backend_eager

```
_ TestCompileTorchbind.test_compile_obj_torchbind_op_with_autocast_device_xpu_backend_inductor _
[gw3] linux -- Python 3.10.20 /tmp/xpu-tool/Python/3.10.20/x64/bin/python
Traceback (most recent call last):
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/export/test_torchbind_xpu.py", line 1591, in test_compile_obj_torchbind_op_with_autocast
    self, f(_empty_tensor_queue(), x), opt_f(_empty_tensor_queue(), x)
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/export/test_torchbind_xpu.py", line 1582, in f
    torch.ops._TorchScriptTesting.queue_push(tq, x.cos())
  File "/tmp/xpu-tool/Python/3.10.20/x64/lib/python3.10/site-packages/torch/_ops.py", line 1275, in __call__
    return self._op(*args, **kwargs)
[...]

PR #3732, opened 2026-05-21
Implement XPU queue_push torchbind kernel

jkosnox:
Add a native boxed implementation for _TorchScriptTesting::queue_push on XPU
and register it for both XPU and AutocastXPU dispatch keys.

The new kernel mirrors CPU behavior by validating the TorchBind object,
calling its push method, and cloning the tensor before enqueueing to avoid
aliasing issues.

Fix  for: [#3477](https://github.com/intel/torch-xpu-ops/issues/3477)

### DIFF
--- a/src/ATen/native/xpu/TorchScriptTesting.cpp
+++ b/src/ATen/native/xpu/TorchScriptTesting.cpp
@@ -1,0 +1,48 @@
+#include <ATen/core/class_type.h>
+#include <ATen/core/ivalue.h>
+#include <torch/library.h>
+
+namespace at::native::xpu {
+namespace {
+
+void queue_push_boxed(
+    const c10::OperatorHandle& op,
+    c10::DispatchKeySet ks,
+    c10::Stack* stack) {
+  (void)op;
+  (void)ks;
+
+  auto x = torch::jit::pop(*stack).toTensor();
+  auto tq = torch::jit::pop(*stack);
+
+  TORCH_CHECK(
+      tq.isObject(),
+      "Expected _TorchScriptTesting::queue_push first argument to be a TorchBind object");
+
+  auto class_type = tq.type()->cast<c10::ClassType>();
+  TORCH_CHECK(
+      class_type != nullptr,
+      "Expected _TorchScriptTesting::queue_push first argument to have ClassType");
+  TORCH_CHECK(
+      class_type->hasMethod("push"),
+      "Expected _TorchScriptTesting::queue_push object to implement method 'push'");
+
+  // Match CPU implementation in test_custom_class_registrations.cpp.
+  class_type->getMethod("push")({std::move(tq), x.clone()});
+}
+
+} // namespace
+
+TORCH_LIBRARY_IMPL(_TorchScriptTesting, XPU, m) {
+  m.impl(
+      "queue_push",
+      torch::CppFunction::makeFromBoxedFunction<&queue_push_boxed>());
+}
+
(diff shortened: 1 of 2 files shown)


src/ATen/native/xpu/TorchScriptTesting.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace at::native::xpu {
+namespace {
+
+void queue_push_boxed(
+    const c10::OperatorHandle& op,
+    c10::DispatchKeySet ks,
+    c10::Stack* stack) {
+  (void)op;
+  (void)ks;
```
tszulist-hbn, line 13: Seems like unused. Is it to meet API compatibility? If so, consider adding comment or using kind of `[maybe_unused]` marker.

BBBela (commented):
Have you considered fixing this issue in the same way that CUDA and CPU do?
I tried myself if that would work, and with changes from [THIS COMMIT](https://github.com/BBBela/pytorch/commit/0ffc3113ee6fb3a276a1e9a834315e1671506fd2), the tests pass and it is basically aligning with CUDA and CPU.

Have you checked that option? Do you see any potential downsides of [THIS APPROACH](https://github.com/BBBela/pytorch/commit/0ffc3113ee6fb3a276a1e9a834315e1671506fd2) versus implementation from this PR? Please let me know if there are any additional aspects that I may have missed 😉

src/ATen/native/xpu/TorchScriptTesting.cpp
jkosnox: Will do

jkosnox:
> Have you considered fixing this issue in the same way that CUDA and CPU do? I tried myself if that would work, and with changes from [THIS COMMIT](https://github.com/BBBela/pytorch/commit/0ffc3113ee6fb3a276a1e9a834315e1671506fd2), the tests pass and it is basically aligning with CUDA and CPU.
> 
> Have you checked that option? Do you see any potential downsides of [THIS APPROACH](https://github.com/BBBela/pytorch/commit/0ffc3113ee6fb3a276a1e9a834315e1671506fd2) versus implementation from this PR? Please let me know if there are any additional aspects that I may have missed 😉

My idea was to keep all changes in torch-xpu-ops, but I agree  [THIS COMMIT](https://github.com/BBBela/pytorch/commit/0ffc3113ee6fb3a276a1e9a834315e1671506fd2) looks cleaner. Do you want to create a PR based on your commit? Or I can do that myself if you'd rather not get additional notifications from another PR.

BBBela:
> My idea was to keep all changes in torch-xpu-ops, but I agree [THIS COMMIT](https://github.com/BBBela/pytorch/commit/0ffc3113ee6fb3a276a1e9a834315e1671506fd2) looks cleaner. Do you want to create a PR based on your commit? Or I can do that myself if you'd rather not get additional notifications from another PR.

I totally understand your idea to keep the changes in torch-xpu-ops as this path is mostly much faster than PyTorch contributions, but in this case I think we should go with the desired final solution in the first place to make the change simpler and adjust with already existing solutions for CUDA and CPU.

> Do you want to create a PR based on your commit?

Please create a PR directly in PyTorch yourself, as you already handle the issue and the whole topic.
I don't mind additional notifications, but you own the topic, so let's keep it that way 😉

jkosnox:
> Please create a PR directly in PyTorch yourself, as you already handle the issue and the whole topic. I don't mind additional notifications, but you own the topic, so let's keep it that way 😉

Will do, thank you

PR #3734, opened 2026-05-21
Remove _flash_attention_forward from XPU CPU fallback list

gplutop7:
Fixes:
https://github.com/intel/torch-xpu-ops/issues/3698

This PR removes aten::_flash_attention_forward from the XPU CPU fallback list, so XPU can use the native XPU implementation path instead of being forced to CPU fallback.

This change is required together with PyTorch PR https://github.com/pytorch/pytorch/pull/184724, which adds/enables the XPU-side flash attention forward implementation.


LuFinch:
Already remove in https://github.com/intel/torch-xpu-ops/pull/3404/changes#diff-3659788f094e19517d8f731e18e4f317e573793bc1f1ec5eca2fe7359bc7b093

PR #3738, opened 2026-05-22
Refine layer_norm_wg_size_select to reduce tree-reduce barrier cost

CuiYifeng:
This PR is to set `wg_size` to `SIMD * 4` for the vectorized LayerNorm kernel when `N / vec_size <= max_wg_size`.

Motivation:
The vectorized LayerNorm kernel uses inter-subgroup tree-reduce in `compute_stats()`. For `wg_size = 1024` and `SIMD32`, the inter-subgroup tree-reduce requires 5 rounds of reduction, where average subgroup utilization during reduce is only ~19.4%. Stall sampling shows barrier synchronization as the biggest stall source (~31.5% of all stalls). Reducing `wg_size / SIMD` to 4 cuts tree-reduce to 2 rounds, eliminates ~92% of sync stalls.

Why gate on `N / vec_size <= max_wg_size`:
Smaller `wg_size / SIMD` means more concurrent workgroups per Xe-core, increasing L1 pressure. The Pass2 of vectorized LayerNorm kernel has a working set that scales as: 
`W_L1 = concurrent_WGs × N × element_size`
when `W_L1` exceeds the LSC cache capacity, Pass2 degrades from L1 hit to L3 hit, offsetting the barrier reduction gains. Since LSC cache sizes vary across platforms and cannot be queried at runtime, we need a platform-independent guard.

### DIFF
--- a/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
@@ -589,11 +589,34 @@ struct VectorizedLayerNormKernelFunctor
   sycl_local_acc_t<T_ACC> buf_;
 };
 
-int64_t layer_norm_wg_size_select(int64_t max_wg_size, int n) {
-  while (max_wg_size > n && max_wg_size > SIMD) {
-    max_wg_size >>= 1;
+int64_t layer_norm_wg_size_select(
+    const int64_t max_wg_size,
+    const int64_t M,
+    const int n) {
+  if (n > max_wg_size)
+    return max_wg_size;
+
+  int64_t wg_size = max_wg_size;
+  while (wg_size > n && wg_size > SIMD) {
+    wg_size >>= 1;
   }
-  return max_wg_size;
+
+  // To reduce the barrier overhead during tree-reduce
+  // with 4 subgroups per workgroup
+  constexpr int64_t threads_per_wg = 4;
+  constexpr int64_t preferred_wg_size = SIMD * threads_per_wg;
+
+  // keep wg_size when n is not large enough to utilize preferred_wg_size
+  if (wg_size <= preferred_wg_size)
+    return wg_size;
+
+  // (XeCore count * EUs per XeCore) * HW threads per EU
+  int64_t total_hw_threads = syclGpuEuCount() * syclGpuHWThreadsPerEU();
+  // Only use preferred_wg_size when less than 50% HW threads would be left idle
+  if (M * threads_per_wg > total_hw_threads / 2)
+    return preferred_wg_size;
+
+  return wg_size;
 }
 
 template <typename T, typename T_ACC, bool rms_norm>
@@ -609,7 +632,7 @@ void launch_vectorized_layer_norm_kernel(
     T_ACC* rstd_data) {
   using KernelClass = VectorizedLayerNormKernelFunctor<T, T_ACC, rms_norm>;
   auto wg_size = layer_norm_wg_size_select(
-      syclMaxWorkGroupSize<KernelClass>(), N / vec_size);
+      syclMaxWorkGroupSize<KernelClass>(), M, N / vec_size);
   KernelClass kfn(
       N,
       eps,


CuiYifeng:
LayerNorm Speedup Ratio:
Shape (batch, dim) | Speedup (fp32) | Speedup (fp16) | Speedup (bf16)
-- | -- | -- | --
(1024, 1024) | 1.44x | 1.34x | 1.34x
(1024, 2048) | 1.29x | 1.79x | 1.65x
(1024, 4096) | 1.02x | 2.20x | 1.83x

RMSNorm Speedup Ratio:
Shape (batch, dim) | Speedup (fp32) | Speedup (fp16) | Speedup (bf16)
-- | -- | -- | --
(1024, 1024) | 1.20x | 1.19x | 1.20x
(1024, 2048) | 1.02x | 1.65x | 1.45x
(1024, 4096) | 0.97x | 1.87x | 1.58x

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -590,8 +590,11 @@ struct VectorizedLayerNormKernelFunctor
 };
 
 int64_t layer_norm_wg_size_select(int64_t max_wg_size, int n) {
-  while (max_wg_size > n && max_wg_size > SIMD) {
-    max_wg_size >>= 1;
+  // To reduce the barrier overhead during tree-reduce
+  // with 4 subgroups per workgroup
+  constexpr int64_t preferred_wg_size = SIMD * 4;
+  if (n <= max_wg_size && preferred_wg_size <= max_wg_size) {
+    return preferred_wg_size;
```
jianyizh, line 617: @CuiYifeng 
1. `preferred_wg_size <= max_wg_size` Why we need this? Max_wg_size for every device should > 128
2. Copilot comments makes sense when n < 128?
3. You also need consider occupancy. Consider batchsize=64 on B70, n=1024. if you have 32 xe core. wg_size = 1024 gives occupancy 100%, but may suffer sync stall. wg_size 512 only 50% occupancy, but less cost on barrier. Does wg_size 128 gives the best performance? You need have a test.

jianyizh (changes requested):

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
CuiYifeng: @jianyizh 

1. Because not sure if max_wg_size >= SIMD * 4 always holds true. However, agree with your opinion about `max_wg_size`. `preferred_wg_size <= max_wg_size` has been removed.
2. Do you mean `when max_wg_size < SIMD*4, the function now returns max_wg_size without shrinking it relative to n`? Make sense. That's another reason to remove `preferred_wg_size <= max_wg_size`.
3. Good point. Based on the ratio ratio for batch_size=40 and dim_size=4096 (n=1024), no obvious regression has been observed. Below is the data collected from B60:

|   dtype | Speedup |
|---------|-----------|
|   fp32   |      0.98x |
|   fp16   |      1.00x |
|   bf16   |      1.01x |

jianyizh: For 2, If dim_size/4==64, should you set wg size 64? 
For 3, if dim_size==4096, batch_size=40, you don't see improvement to decrease wg size from 1024 to 128. Does it mean your sync stall is lower but the improvement is covered by low occupancy?

CuiYifeng: @jianyizh `wg_size` for the following scenarios is retained as-is now.

- `dim_size / 4 < SIMD * 4` 
- `occupancy < 100%`

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -589,11 +589,34 @@ struct VectorizedLayerNormKernelFunctor
...
+  constexpr int64_t preferred_wg_size = SIMD * threads_per_wg;
+
+  // keep wg_size when n is not large enough to utilize preferred_wg_size
+  if (wg_size <= preferred_wg_size)
+    return wg_size;
+
+  // (XeCore count * EUs per XeCore) * HW threads per EU
+  int64_t total_hw_threads = syclGpuEuCount() * syclGpuHWThreadsPerEU();
+  // Only use preferred_wg_size when no HW threads would be left idle
+  if (M * threads_per_wg >= total_hw_threads)
```
jianyizh: you can guard occupancy<=50%. i.e. if occupancy<=50%, use larger wg size.

jianyizh (changes requested):
Another question is why we choose to decrease wg size instead of decrease the number of barriers? Do you know why decrease the number of barriers not perform good?

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
CuiYifeng:
```suggestion
  // Only use preferred_wg_size when no more than 50% HW threads would be left idle
  if (M * threads_per_wg > total_hw_threads / 2)
```

CuiYifeng: Done.

CuiYifeng:
> Another question is why we choose to decrease wg size instead of decrease the number of barriers? Do you know why decrease the number of barriers not perform good?

Good question. Will be processed as Step 2.

jianyizh (approved):

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -589,11 +589,34 @@ struct VectorizedLayerNormKernelFunctor
...
+  constexpr int64_t threads_per_wg = 4;
+  constexpr int64_t preferred_wg_size = SIMD * threads_per_wg;
+
+  // keep wg_size when n is not large enough to utilize preferred_wg_size
+  if (wg_size <= preferred_wg_size)
+    return wg_size;
+
+  // (XeCore count * EUs per XeCore) * HW threads per EU
+  int64_t total_hw_threads = syclGpuEuCount() * syclGpuHWThreadsPerEU();
+  // Only use preferred_wg_size when less than 50% HW threads would be left idle
```
LuFinch, line 615: Why 50%? Do you have data to demonstrate performance benefit for this PR when the input size can't have max occupancy?

CuiYifeng: @LuFinch `preferred_wg_size` only takes effect when `dim_size` belongs to `[128 * vec_size, 1024 * vec_size]`. On this basis, op time is dominated by the host bound when `batch_size` reduces the occupancy to below 100%. Taking `[256, 768]` as an example on B60, where the occupancy under new policy is 80%, the speed ratio of fp32/fp16/bf16 is 1.05/1.04/0.98. The threshold 50% is to prevent the new policy from causing too low occupancy.

LuFinch (approved):

CuiYifeng:
@copilot Please check if "UT check failed with new failures" are relevant to this PR.

Issue #3209, opened 2026-03-30
[Win][Build] There is Cyclic dependencies error when build with BUILD_SEPARATE_OPS=true


NeoZhangJianyu:
### 🐛 Describe the bug

When build PyTorch on windows, there two modes: BUILD_SEPARATE_OPS or not.
In release build case, BUILD_SEPARATE_OPS is false. It's passed in building on windows.
But when set BUILD_SEPARATE_OPS=true or CMAKE_BUILD_TYPE is "Debug", the build mode BUILD_SEPARATE_OPS will lead to error:
```
CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
    "torch_xpu" of type SHARED_LIBRARY
      depends on "torch_xpu_ops" (weak)
    "torch_xpu_ops" of type STATIC_LIBRARY
      depends on "torch_xpu" (weak)
  At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
  CMake Generate step failed.  Build files cannot be regenerated correctly.
```

Reproduce:
Apply following diff in torch_xpu_ops, and build again.
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8f6fa8d6..033a47d5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,7 @@ endif()

 # Only for debugging. Save building time by shrinking translation unit scope.
 set(BUILD_SEPARATE_OPS $ENV{BUILD_SEPARATE_OPS})
+set(BUILD_SEPARATE_OPS TRUE)
 if(CMAKE_BUILD_TYPE MATCHES "(Debug|RelWithDebInfo)")
   set(BUILD_SEPARATE_OPS TRUE)
 endif()
```

### Versions

https://github.com/intel/torch-xpu-ops 

commit 1c25bb5ddf4f5cfcf84d2614864c9ca5d445426a (HEAD, main)
Author: Piotr Bielak <pbielak@users.noreply.github.com>
Date:   Thu Mar 26 20:09:23 2026 +0100

xuhancn:
I remember we are not guarantee `BUILD_SEPARATE_OPS` works well on Windows, as it is a corner case.

PR #3755, opened 2026-05-26
Fix Debug/RelWithDebInfo PyTorch+XPU on Windows

draghan:
Fix for #3209.

Building and using PyTorch with Windows `Debug` and `RelWithDebInfo` configuration with XPU support, exposed several Windows-specific issues:
- cyclic dependency patterns in separate-ops configure/link flow,
- SYCL CRT/debug-info causing linker/compiler pressure,
- incremental codegen fragility from configure-time template copy/delete handling.

This PR contains three commits, fixing above issues. They are gathered together in one PR, as they are complementary and addressing one functional issue: building non-release packages with XPU on Windows. Those fixes are not full solution: they must be complemented with another set of fixes in the PyTorch repository: https://github.com/pytorch/pytorch/pull/185217

Commits in this PR cover:
1. Build graph/cyclic deps: break cyclic dependency in Windows separate-ops configure/link stages.
2. Debug compile settings: align SYCL host CRT for debug and reduce debug-info pressure in large SYCL translation units.
3. Incremental codegen stability: replace fragile configure-time template copy/delete flow with build-time template sync + stamp dependency.


cmake/Codegen.cmake
```diff
@@ -24,16 +24,34 @@ set(XPU_AOTI_INSTALL_DIR ${TORCH_ROOT}/torch/csrc/inductor/aoti_torch/generated/
...
-  execute_process(COMMAND cmd /c xcopy ${SrcPATH} ${DestPATH} /E /H /C /I /Y > nul)
+  # on Windows, sync pytorch templates on every build (including incremental builds)
+  add_custom_command(
+    OUTPUT ${XPU_TEMPLATE_STAMP}
+    COMMAND ${CMAKE_COMMAND} -E rm -rf "${XPU_TEMPLATE_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy_directory "${XPU_TEMPLATE_SRC_DIR}" "${XPU_TEMPLATE_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E touch "${XPU_TEMPLATE_STAMP}"
+    DEPENDS
+      ${TORCH_ROOT}/aten/src/ATen/templates/RegisterDispatchDefinitions.ini
+      ${TORCH_ROOT}/aten/src/ATen/templates/RegisterDispatchKey.cpp
```
BBBela:
```suggestion
      ${XPU_TEMPLATE_SRC_DIR}/RegisterDispatchDefinitions.ini
      ${XPU_TEMPLATE_SRC_DIR}/RegisterDispatchKey.cpp
```

BBBela (approved):
Thanks for those changes. Approving! 😉 

Short summary of my reasoning behind the approval:

As for my understanding, before this PR, Windows Debug/RelWithDebInfo builds with XPU were completely broken. This PR along the one from PyTorch fixes that.

**Change 1 (cyclic deps):** Affects `BUILD_SEPARATE_OPS` only. From what I know, the cyclic dependency (`torch_xpu` → `torch_xpu_ops` → per-op `.dll` → `torch_xpu`) worked on Linux because Linux `.so` allows unresolved symbols at link time, resolved lazily at runtime. Windows `.dll` apparently requires every symbol to be explicitly resolved at link time, so MSVC rejected the cycle. Switching to static per-op libs defers all symbol resolution to the single `torch_xpu.dll` link, which should eliminate the cycle. The downside I can see is that the `torch_xpu` link step gets heavier, but that seems like an acceptable cost given the alternative is a broken build.

**Change 2 (debug compile settings):** Debug builds are now enabled. The `gline-tables-only` downside only affects SYCL device code — host code should still retain full debug info. The alternative (full `-g`) apparently exhausts Clang source locations and fails to compile entirely, so this seems like the right tradeoff.

**Change 3 (incremental codegen):** Incremental builds on Windows were silently broken — as far as I can tell, templates were deleted after every build, so the next incremental run used a missing templates directory. This fixes that.

Just one minor suggestion below.

cmake/Codegen.cmake
draghan: Thank you, applied!

draghan:
Thanks for your time @BBBela!

draghan:
Awaiting for landing of PR #5316

PR #3766, opened 2026-05-27
[SYCLTLA] Enable dropout for FlashAttention

LuFinch:
# Flash Attention Dropout Implementation Summary

### Mathematical Definition

$$S = QK^T, \quad P = \text{softmax}(S), \quad \tilde{P} = \frac{1}{1-p} P \odot M, \quad O = \tilde{P} V$$

where $M$ is a 0/1 dropout mask, $p$ is the drop probability, and $\frac{1}{1-p}$ is the rescale factor.

### Implementation Key Points

1. **No mask storage**: Only the RNG state (two integers) is saved; the backward pass regenerates the exact same mask using the same RNG state.
2. **Dropout applied after softmax**: Applied to $P$ (softmax output), not to $S$ (raw scores).
3. **Rescale**: Multiply by $\frac{1}{1-p}$ to keep the expected value unchanged.

### Philox RNG Implementation (philox.cuh)

FlashAttention uses the Philox4x32-7 (7 rounds) counter-based random number generator.

#### Core Primitive

```
output(4×uint32) = philox(key(2×uint32), counter(4×uint32))
```

#### Parameter Mapping

| User Level | Philox Internal | Bit Width |
|-----------|----------------|-----------|
| seed | key | 64 bit (2×uint32) |
| offset | counter[0:1] | 64 bit (low 64 bits) |
| subsequence | counter[2:3] | 64 bit (high 64 bits) |

#### Key Functions

- **`mulhilo32(a, b)`**: Uses PTX `mul.wide.u32` to compute the full-width multiplication of two uint32 values, returning (lo, hi).
- **`philox_single_round(ctr, key)`**: Single-round permutation, performing multiplication + XOR mixing on both halves of the counter.
- **`philox(seed, subsequence, offset)`**: Main function, executes 7 rounds (6 loop iterations + 1 tail round). After each round, the key is incremented by Weyl sequence constants `(0x9E3779B9, 0xBB67AE85)`, producing 4 uint32 random numbers.

##### Core Design Principle

A single Philox invocation produces 4×uint32 random numbers (reinterpretable as 8×uint16 or 16×uint8). To fully amortize the computational cost of each Philox call, each thread/lane should process an equal number of S matrix elements per call.

To ensure **fwd/bwd generate exactly the same dropout mask regardless of different TiledMMA tile sizes and thread layouts**, the S matrix is partitioned into fixed-size **MMA atomic blocks** as the fundamental dropout unit. For each atomic block:

```
Philox(seed,
       offset      = f(batch_id, head_id, lane_id),   
       subsequence = f(block_row_id, block_col_id))      
```

offset and subsequence occupy different dimensions of the 128-bit counter (low/high 64 bits), **orthogonal and non-overlapping**. Given (seed, batch, head, lane, block position), fwd/bwd will necessarily produce the same random numbers for the same set of elements, thus generating an identical mask.

##### CUDA FA vs Intel XPU FA Comparison

|  | CUDA FA (Tri Dao) | Intel XPU FA (SyclTLA) |
|---|---|---|
| MMA atom | m16×n8 | m8×n16 |
| Thread group | warp = 32 threads | subgroup = 16 lanes |
| Elements per thread/lane per atom | 4 | 8 |
| Dropout basic block | (16, 32) = 4 m16n8 atoms | (8, 16) = 1 m8×n16 atom |
| Elements per Philox call | 4 atoms × 4 elements = 16 elements | 1 atom × 8 elements = 8 elements |
| Random number reinterpretation | 4×uint32 → 16×uint8 | 4×uint32 → 8×uint16 |
| Comparison precision | uint8 (256 levels) | uint16 (65536 levels) |

**CUDA FA Details**: Each thread in an m16n8 atom holds only 4 elements, but Philox produces 16 uint8 values. Therefore, a (16, 32) block (= 4 m16n8 atoms, 1×M 4×N) is used as the basic unit, and one Philox call covers all 4×4 = 16 elements held by the current thread within that block:

```cpp
for (int j = 0; j < 2; j++) {          // 2 groups, each with 2 n-atoms
    for (int i = 0; i < 8; i++) {       // 8 elements per group
        tensor(i, m, n * 2 + j) = encode_dropout(rnd_8[j * 8 + i] <= threshold, ...);
    }
}
```

**Intel XPU FA Details**: Each lane in an m8×n16 atom holds 8 elements, which exactly matches 8×uint16. One Philox call covers a complete atomic block without needing to combine across atoms.

##### Offset Increment

After each forward kernel invocation, the global Philox offset increases by `batch × heads × lane_count`:

- CUDA: `batch × heads × 32` (warp size)
- XPU: `batch × heads × 16` (subgroup size)

---

### Dropout Flow in Forward Kernel

Mathematically $O = \frac{1}{1-p}(P \odot M) V$, but the kernel **defers rescaling to the epilogue**:

**Mainloop** (each K block iteration):
1. $S = Q K^T$ (GEMM1)
2. $P = \text{softmax}(S)$ (online softmax)
3. **Apply dropout mask**: Generate mask $M$ using Philox, **zero out** dropped elements: $\tilde{P} = P \odot M$
4. $O_{acc} += \tilde{P} \cdot V$ (GEMM2, accumulate to output registers)

**Epilogue**:
5. **Dropout rescale**: $O = \frac{1}{1-p} \cdot O_{acc}$

Deferring the $\frac{1}{1-p}$ multiplication to the epilogue avoids scaling $\tilde{P}$ before each GEMM2, reducing intermediate computation. The final result is equivalent to $O = \frac{1}{1-p}(P \odot M) V$.

### Mathematical Derivation

Need to compute $dV, dK, dQ$. Dropout propagates to all gradients through the chain rule.

**Key Equations:**

- $d\tilde{P} = dO \cdot V^T$ (gradient w.r.t. dropout output)
- $dP = \frac{M}{1-p} \odot d\tilde{P}$ (backprop through dropout)
- $D_i = \sum_j P_{ij} \cdot dP_{ij} = dO_i \cdot O_i$ (normalization term gradient inside softmax)
- $dS = P \odot (dP - D)$ (softmax backward)
- $dV = \tilde{P}^T \cdot dO, \quad dK = scale \cdot dS^T \cdot Q, \quad dQ = scale \cdot dS \cdot K$

### CUDA Code Implementation

The code employs a **sign-bit encoding + deferred scaling** strategy, avoiding explicit mask storage and reducing extra computation.

#### 1. RNG Reconstruction (No Mask Storage Needed)

```cpp
FLASH_NAMESPACE::Dropout dropout(params.rng_state[0], params.rng_state[1],
    params.p_dropout_in_uint8_t, bidb, bidh, tidx, params.h);
```

Uses the RNG state saved during forward to regenerate the exact same dropout pattern.

#### 2. Sign-Bit Encoding of Mask

```cpp
dropout.template apply_dropout</*encode_dropout_in_sign_bit=*/true>(acc_s, ...);
```

When applying dropout to the softmax output $P$, instead of zeroing out elements, it **flips the sign bit to make them negative**. A single float simultaneously encodes:
- **Absolute value** = original $P_{ij}$
- **Sign** = whether it was dropped (negative = dropped)

#### 3. Using Sign Bit to Obtain $P \odot M$

```cpp
Tensor rP = Is_dropout
    ? FLASH_NAMESPACE::convert_type_relu<Element>(acc_s)   // negative → 0
    : FLASH_NAMESPACE::convert_type<Element>(acc_s);
```

`convert_type_relu` zeros out negative values. The result `rP` = $P \odot M$, written to shared memory `sP` for use in the $dV$ gemm.

#### 4. Computing $D'_i$ (Pre-compensated Scaling)

```cpp
dot_do_o<...>(tdOrdO, tdOrO, gdPsum, ..., params.p_dropout);
```

Stores $D'_i = (1-p) \cdot dO_i \cdot O_i$, rather than the true $D_i = dO_i \cdot O_i$.

**Reason**: The code uses a deferred scaling strategy where the epilogue multiplies the entire $dS$ by $\frac{1}{1-p}$. If $D_i$ were stored directly, it would be incorrectly divided by $(1-p)$ an extra time. Pre-multiplying by $(1-p)$ exactly cancels this extra scaling:

$$\frac{1}{1-p} \cdot P(d\tilde{P} - \underbrace{(1-p) D_i}_{D'}) = P\left(\frac{d\tilde{P}}{1-p} - D_i\right) \quad \checkmark$$

#### 5. `pointwise_mult`: Implicit Mask Handling for $dS$

```cpp
auto pointwise_mult = [](float p, float dp, float d) {
    return p * (!Is_dropout || p >= 0 ? dp - d : d);
};
```

- `p` = sign-encoded $P_{ij}$, `dp` = $d\tilde{P}_{ij}$ = $(dO \cdot V^T)_{ij}$, `d` = $D'_i$
- **Not dropped** (`p >= 0`): $dS' = P \cdot (d\tilde{P} - D')$
- **Dropped** (`p < 0`): $dS' = (-|P|) \cdot D'$ (the negative sign participates in the correct subtraction)

**No explicit mask operation** — the mask effect is entirely achieved through the sign-bit branch.

#### 6. Epilogue: Deferred $\frac{1}{1-p}$ Scaling

```cpp
// dV multiplied by 1/(1-p) separately
if (Is_dropout) { acc_dv(i) *= params.rp_dropout; }
// dK combines softmax_scale and 1/(1-p)
acc_dk(i) *= params.scale_softmax_rp_dropout;  // = scale / (1-p)
// dQ similarly
acc_dq(i) *= params.scale_softmax_rp_dropout;
```

All preceding computations exclude $\frac{1}{1-p}$, which is uniformly applied in the epilogue.

### Parameter Reference

| Parameter | Meaning |
|-----------|---------|
| `p_dropout` | Keep probability $1-p$ |
| `rp_dropout` | $\frac{1}{1-p}$ (reciprocal of keep prob) |
| `scale_softmax_rp_dropout` | $\frac{scale\_softmax}{1-p}$ |
| `p_dropout_in_uint8_t` | uint8-encoded dropout probability (for RNG comparison) |

### Design Philosophy Summary

| Technique | Purpose |
|-----------|---------|
| RNG state reconstruction | Avoids storing $O(n^2)$ mask tensor |
| Sign-bit encoding | A single float represents both the $P$ value and the mask, saving extra storage and operations |
| `convert_type_relu` | Obtains $P \odot M$ directly using the sign bit, zero overhead |
| $D'_i$ pre-multiplied by $(1-p)$ | Works with deferred scaling to prevent $D_i$ from being incorrectly scaled |
| Deferred scaling | Merges $\frac{1}{1-p}$ into a single multiplication in the epilogue, reducing intermediate computation |

---

### Intel XPU (SyclTLA) Code Implementation

The XPU backward implementation is mathematically equivalent to CUDA but uses an **immediate scaling** strategy (as opposed to CUDA's deferred scaling).

### Overall Flow

The XPU backward consists of 3 kernels:

1. **`mha_dot_do_o`** (`compute_o_dot_do`): Computes $D_i = \sum_j dO_{ij} \cdot O_{ij}$, also zeros out `dQaccum`
2. **`mha_backward_seq`** (`dq_dk_dv_1colblock`): Main loop, computes dS, dV, dK, dQ
3. **`mhd_convert_dq`** (`convert_dq`): Converts float32 `dQaccum` to bf16/fp16 and writes to `dQ`

#### 1. $D_i$ Without Pre-scaling

CUDA stores $D'_i = (1-p) \cdot dO_i \cdot O_i$ (pre-multiplied by dropout keep probability), while XPU directly stores $D_i = dO_i \cdot O_i$ without any preprocessing.

#### 2. `softmax_backward`: All Scaling Done Immediately

```cpp
auto pointwise_mult = [is_dropout, scale, rp_dropout](float p, float dp, float d) {
    return scale * p * (!is_dropout || p >= 0 ? dp * rp_dropout - d : d);
};
```

- `p` = sign-encoded $P_{ij}$, `dp` = $d\tilde{P}_{ij}$, `d` = $D_i$ (not pre-scaled)
- `scale` = softmax_scale = $\frac{1}{\sqrt{d}}$
- `rp_dropout` = $\frac{1}{1-p}$

**Not dropped** (`p >= 0`):

$$dS_{ij} = \text{scale} \cdot P_{ij} \cdot \left(\frac{d\tilde{P}_{ij}}{1-p} - D_i\right)$$

**Dropped** (`p < 0`):

$$dS_{ij} = \text{scale} \cdot (-|P_{ij}|) \cdot D_i = -\text{scale} \cdot |P_{ij}| \cdot D_i$$

Unlike CUDA, here dS already includes `scale` and `rp_dropout`, requiring no epilogue post-processing.

#### 3. $\tilde{P}$ Processing for dV

```cpp
// Apply relu + rescale on sign-encoded P, obtaining P_tilde in one step
void apply_dropout_on_signed_P(Tensor& P, const float& rp_dropout) {
    P(i) = P(i) >= 0 ? P(i) * rp_dropout : 0;  // = P * M / (1-p)
}
```

CUDA does this in two steps (`convert_type_relu` to get $P \odot M$, then epilogue multiplies by $\frac{1}{1-p}$), while XPU does it in one step.

#### 4. No Epilogue Scaling

CUDA uniformly multiplies scaling factors in the epilogue; XPU's dK/dQ/dV require no additional scalar multiplication before being written out.

### XPU Main Loop Pseudocode (`dq_dk_dv_1colblock`)

```
rdV = 0, rdK = 0                         // register accumulators

for m_block in 0..max_m_block:
    // 1. Reconstruct P
    rS = Q[m_block] * K^T                 // GEMM: S = Q K^T
    scores = exp2(rS * scale_log2 - LSE * log2e)  // P = softmax(S)

    // 2. Sign-bit encode dropout mask (same as CUDA)
    if is_dropout:
        dropout.apply_dropout<sign_bit=true>(rS)

    // 3. Compute d_tilde_P
    rdP = dO[m_block] * V^T               // GEMM: dP_tilde = dO V^T

    // 4. softmax_backward: immediate scaling (differs from CUDA)
    dS = scale * P_signed * (dP_tilde * rp_dropout - D_i)  // or dropped branch

    // 5. Process P → P_tilde (for dV)
    if is_dropout:
        rS = relu(rS) * rp_dropout        // apply_dropout_on_signed_P

    // 6. Write to global memory intermediate buffer
    store(rS → mP),  store(dS → mdP)

    // 7. GEMM to compute dV, dK, dQ
    rdV += P_tilde^T * dO[m_block]         // dV accumulation
    rdK += dS^T * Q[m_block]               // dK accumulation (dS already includes scale)
    dQaccum[m_block] += dS * K             // dQ atomic accumulation

store rdV → dV,  store rdK → dK           // write out after loop ends
```

### CUDA vs XPU Strategy Comparison

| Aspect | CUDA (Deferred Scaling) | XPU (Immediate Scaling) |
|--------|------------------------|------------------------|
| $D_i$ storage | $D'_i = (1-p) \cdot D_i$ (pre-scaled) | $D_i$ (raw value) |
| `softmax_backward` | Excludes `scale`, excludes `rp_dropout` | Includes `scale` and `rp_dropout` |
| dS output | Unscaled $dS'$ | Fully scaled final $dS$ |
| $P \odot M$ for dV | `convert_type_relu` → epilogue $\times \frac{1}{1-p}$ | `apply_dropout_on_signed_P` in one step |
| Epilogue dK/dQ | $\times \frac{\text{scale}}{1-p}$ | None |
| Epilogue dV | $\times \frac{1}{1-p}$ | None |
| Intermediate buffer | Shared memory (sP, sdS) | Global memory (pb_ptr) |

### Mathematical Equivalence Verification

Expanding the final $dK_{ij}$ for non-dropped elements:

**CUDA**:

$$dK = \frac{\text{scale}}{1-p} \cdot \underbrace{P \cdot (d\tilde{P} - (1-p) D_i)}_{dS'} \cdot Q = \text{scale} \cdot P \cdot \left(\frac{d\tilde{P}}{1-p} - D_i\right) \cdot Q$$

**XPU**:

$$dK = \underbrace{\text{scale} \cdot P \cdot \left(\frac{d\tilde{P}}{1-p} - D_i\right)}_{dS} \cdot Q \quad \checkmark$$

The dropped elements follow the same logic; both yield identical final results.

### DIFF
--- a/src/ATen/native/transformers/xpu/flash_attn/sycltla/kernel/xe_fmha_fwd_kernel.h
+++ b/src/ATen/native/transformers/xpu/flash_attn/sycltla/kernel/xe_fmha_fwd_kernel.h
@@ -228,11 +228,21 @@ class XeFMHAFwdKernel {
 
     SharedStorage& shared_storage = *reinterpret_cast<SharedStorage*>(smem_buf);
 
+    int thr_id = int(ThreadIdxX());
+
+    bool is_dropout = params.mainloop.p_dropout < 1.0f;
+    auto seed_offset = at::xpu::philox::unpack(params.mainloop.philox_args);
+    // Save seed and offset for backward, before any early exiting. Otherwise
+    // the 0-th thread block might exit early and no one saves the rng states.
+    if (is_dropout && BlockIdxX() == 0 && BlockIdxY() == 0 &&
+        BlockIdxZ() == 0 && thr_id == 0) {
+      params.mainloop.rng_seed[0] = std::get<0>(seed_offset);
+      params.mainloop.rng_offset[0] = std::get<1>(seed_offset);
+    }
+
     auto& p = params.kernel;
     ProblemShape const& s = p.shape;
     int head_group_q = s.num_heads_q / s.num_heads_kv;
-
-    int thr_id = int(ThreadIdxX());
     int q_sg_tile = get<0>(shape_div(TileShapeQK{}, shape(SubgroupLayoutQK{})));
 
     auto cS = make_identity_tensor(take<0, 2>(TiledMMAQK{}.tile_mnk()));
@@ -249,6 +259,15 @@ class XeFMHAFwdKernel {
           tile_scheduler.get_block_coord(); // (Q,V,h,b)
       auto blk_qv = make_coord(blk_q, blk_v);
       int head = head_q / head_group_q;
+
+      FLASH_NAMESPACE::Dropout dropout(
+          std::get<0>(seed_offset),
+          std::get<1>(seed_offset),
+          params.mainloop.p_dropout_in_uint16_t,
+          idx_b,
+          head_q,
+          thr_id,
+          s.num_heads_q);
 
       auto [seq_len_qo, seq_len_kv] = get_sequence_length_shape(s, idx_b);
       if (blk_q * get<0>(TileShapeQK{}) >= seq_len_qo)
@@ -333,6 +352,9 @@ class XeFMHAFwdKernel {
       // Main loop
       int l_coord = is_var_len ? 0 : idx_b;
       CollectiveMainloop mainloop(params.mainloop, shared_storage.mainloop);
+      if (mainloop.params.p_ptr)
+        mainloop.params.p_ptr +=
+            (l_coord * s.num_heads_q + head_q) * seq_len_qo * seq_len_kv;
       mainloop(
           Q(_, _, head_q, l_coord),
           K(_, _, head, l_coord),
@@ -350,7 +372,9 @@ class XeFMHAFwdKernel {
           seq_len_kv,
           idx_b,
           tile_row_idx,
-          rows_of_maxima);
+          rows_of_maxima,
+          is_dropout,
+          dropout);
 
       if constexpr (
           !is_empty_v<MainloopSharedStorage> &&
@@ -376,7 +400,8 @@ class XeFMHAFwdKernel {
           blk_qv,
           thr_id,
           dpLSE,
-          metadata_for_lse);
+          metadata_for_lse,
+          is_dropout);
     }
   }
 };
--- a/src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd_launch.h
+++ b/src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd_launch.h
@@ -25,7 +25,7 @@ template <typename T, int Headdim, bool is_causal>
 void run_mha_bwd_(sycl::queue& queue, FLASH_BWD_params& params);
 
 template <bool Is_even_M, class T>
-void compute_o_dot_do(
+CUTLASS_DEVICE void compute_o_dot_do(
     T& trait,
     Param<typename T::DType>& param,
     const int m_block,
@@ -179,7 +179,7 @@ void mha_dot_do_o(T trait, Param<typename T::DType> param) {
 }
 
 template <typename Layout>
-auto convert_layout_2d_layout(Layout layout) {
+CUTLASS_DEVICE auto convert_layout_2d_layout(Layout layout) {
   auto l =
       make_layout(make_layout(get<0>(layout), get<1>(layout)), get<2>(layout));
   return l;
@@ -196,7 +196,7 @@ CUTLASS_DEVICE auto convert_layout_acc_layout(Layout acc_layout) {
 }
 
 template <typename T, class Trait, class MTensor, class TiledMMA>
-auto create_reg(
+CUTLASS_DEVICE auto create_reg(
     Trait const& trait,
     MTensor const& C,
     TiledMMA const& tiled_mma) {
@@ -229,7 +229,7 @@ template <
     class Layout2,
     class TVLayout2,
     class TiledMMA>
-void gemm_kernel(
+CUTLASS_DEVICE void gemm_kernel(
     Trait& trait,
     Tensor<Engine0, Layout0> const& A,
     Tensor<Engine1, Layout1> const& B,
@@ -315,7 +315,7 @@ template <
     class Layout2,
     class TVLayout2,
     class TiledMMA>
-void gemm_SdP(
+CUTLASS_DEVICE void gemm_SdP(
     Trait& trait,
     Tensor<Engine0, Layout0> const& A,
     Tensor<Engine1, Layout1> const& B,
@@ -334,7 +334,7 @@ template <
     class Layout2,
     class TVLayout2,
     class TiledMMA>
-void gemm_dKV(
+CUTLASS_DEVICE void gemm_dKV(
     Trait& trait,
     Tensor<Engine0, Layout0> const& A,
     Tensor<Engine1, Layout1> const& B,
@@ -352,7 +352,7 @@ template <
     class Engine2,
     class Layout2,
     class TiledMMA>
-void gemm_dQ(
+CUTLASS_DEVICE void gemm_dQ(
     Trait& trait,
     Tensor<Engine0, Layout0> const& A,
     Tensor<Engine1, Layout1> const& B,
@@ -502,7 +502,7 @@ CUTLASS_DEVICE void softmax_backward(
 }
 
 template <typename Engine, typename Layout>
-void apply_dropout_on_signed_P(
+CUTLASS_DEVICE void apply_dropout_on_signed_P(
     Tensor<Engine, Layout>& P,
     const float& rp_dropout) {
   CUTLASS_PRAGMA_UNROLL
@@ -519,7 +519,7 @@ template <
     class TVLayout0,
     class Engine1,
     class Layout1>
-void mha_copy(
+CUTLASS_DEVICE void mha_copy(
     Trait& trait,
     TiledMma& tiled_mma,
     SubgroupTensor<Engine0, Layout0, TVLayout0>& r,
@@ -545,7 +545,7 @@ template <
     class TVLayout0,
     class Engine1,
     class Layout1>
-void mha_reorder_copy(
+CUTLASS_DEVICE void mha_reorder_copy(
     Trait& trait,
     TiledMma& tiled_mma,
     SubgroupTensor<Engine0, Layout0, TVLayout0>& r,
@@ -556,7 +556,7 @@ void mha_reorder_copy(
 }
 
 template <bool Is_even_N, bool is_dropout, class Trait>
-void dq_dk_dv_1colblock(
+CUTLASS_DEVICE void dq_dk_dv_1colblock(
     Trait& trait,
     Param<typename Trait::DType>& param,
     const int bidb,
@@ -863,7 +863,7 @@ CUTLASS_DEVICE void convert_type(
 }
 
 template <bool Is_even_M, class T>
-void convert_dq(
+CUTLASS_DEVICE void convert_dq(
     T& trait,
     Param<typename T::DType>& param,
     int m_block,
--- a/src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd_launch.h
+++ b/src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd_launch.h
@@ -201,8 +201,8 @@ CUTLASS_DEVICE void apply_mask_causal(
   for (int n = 0; n < size<1>(tensor); ++n) {
     CUTLASS_PRAGMA_UNROLL
     for (int m = 0; m < size<0>(tensor); ++m) {
-      int y = m_offset + get<1>(rC_2d(m, n)) + diagonal_offset;
-      int x = n_offset + get<0>(rC_2d(m, n));
+      int y = m_offset + get<0>(rC_2d(m, n)) + diagonal_offset;
+      int x = n_offset + get<1>(rC_2d(m, n));
       if (x > y) {
         tensor(m, n) = -INFINITY;
       }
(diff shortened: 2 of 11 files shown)


src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_common.h
```diff
@@ -59,6 +60,31 @@
...
+    } else if (HEADDIM <= 128) {          \
+      constexpr static int Headdim = 128; \
+      return __VA_ARGS__();               \
+    } else if (HEADDIM <= 192) {          \
+      constexpr static int Headdim = 192; \
+      return __VA_ARGS__();               \
+    } else if (HEADDIM <= 256) {          \
+      constexpr static int Headdim = 256; \
+      return __VA_ARGS__();               \
+    }                                     \
```
jianyizh: else raise error?

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_common.h
```diff
@@ -59,6 +60,31 @@
...
+  }()
+
+#define DROPOUT_SWITCH BOOL_SWITCH
```
jianyizh: Not used?

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd_launch.h
```diff
@@ -522,35 +476,86 @@ CUTLASS_DEVICE void softmax_backward(
...
+      } else {
+        dP(mi, ni) = scale * P(mi, ni) * (dP(mi, ni) - dpsum);
       }
     }
   }
 }
 
-template <bool Is_even_N, bool Seq_parallel, class Trait>
+template <typename Engine, typename Layout>
+void apply_dropout_on_signed_P(
```
jianyizh: Do we need CUTLASS_DEVICE ?

src/ATen/native/transformers/xpu/flash_attn/sycltla/kernel/xe_fmha_fwd_kernel.h
```diff
@@ -250,6 +260,15 @@ class XeFMHAFwdKernel {
       auto blk_qv = make_coord(blk_q, blk_v);
       int head = head_q / head_group_q;
 
+      FLASH_NAMESPACE::Dropout dropout(
```
jianyizh, line 263: dropout(...) is constructed inside the head loop even when is_dropout is false. Consider guarding the construction (or hoisting to a default state) to avoid unnecessary computation per work-item.

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_common.h
```diff
@@ -196,6 +241,12 @@ inline void set_params_fprop(
   // Other params
   params.is_causal = is_causal;
   params.scale = scale;
+  // Set this to probability of keeping an element to simplify things.
+  TORCH_CHECK(p_dropout < 1.0f);
```
jianyizh: TORCH_CHECK(p_dropout >= 0.0f && p_dropout < 1.0f)

jianyizh (approved):

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_common.h
LuFinch: No. Here is aligned with Pytorch CUDA and TriDao. The headdim check is applied before launch.

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_common.h
LuFinch: Good catch! Removed.

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd_launch.h
LuFinch: Good point. We should add for all kernel function. Added.

src/ATen/native/transformers/xpu/flash_attn/sycltla/kernel/xe_fmha_fwd_kernel.h
LuFinch: This is also aligned with CUDA. I think it is OK because the computation is not huge.

src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_common.h
LuFinch: Done. Thanks for your review.

EikanWang:
@copilot+claude-opus-4.6 please review the changes in depth.

src/ATen/native/transformers/xpu/flash_attn/sycltla/philox.h
EikanWang, line 1: Please reuse https://github.com/intel/torch-xpu-ops/blob/main/src/ATen/native/xpu/sycl/Philox4x32.

LuFinch: Sure. Done.

LuFinch:
test case will be enabled in Pytorch side.

EikanWang (approved):

LuFinch:
@copilot Does the UT failures relevant to this PR change?

PR #3770, opened 2026-05-28
Fix LSTM fused cell bias handling: check each bias independently

jianyizh:
## Summary

Fix a bug in `LstmCellForwardFunctor` where bias presence is determined by a single check on `bias1_` pointer only. When `bias1` is nullptr but `bias2` is valid (e.g. when input gates are pre-computed without bias, but hidden bias is still passed), `bias2` is incorrectly skipped.

## Problem

The original code:
```cpp
bool has_bias = bias1_.data != nullptr;
```
This single flag controls both `bias1` and `bias2` loading. If `bias1` is nullptr, `bias2` is also zeroed out regardless of whether it contains valid data.

## Fix

Check each bias pointer independently:
```cpp
bool has_bias1 = bias1_.data != nullptr;
bool has_bias2 = bias2_.data != nullptr;
```

This allows the common case `(bias1=valid, bias2=valid)` and `(None, None)` to work as before, while also correctly handling `(None, bias2)` which occurs in LSTM pre-compute optimizations.

## Testing

Verified correctness across 8 LSTM configurations (bidirectional/unidirectional, with/without bias, packed/padded sequences) with max diff < 1e-6 vs CPU reference.

### DIFF
--- a/src/ATen/native/xpu/sycl/RNNKernels.cpp
+++ b/src/ATen/native/xpu/sycl/RNNKernels.cpp
@@ -111,7 +111,8 @@ template <
     int indexing_kind>
 struct LstmCellForwardFunctor {
   void operator()(sycl::nd_item<1> item) const {
-    bool has_bias = bias1_.data != nullptr;
+    bool has_bias1 = bias1_.data != nullptr;
+    bool has_bias2 = bias2_.data != nullptr;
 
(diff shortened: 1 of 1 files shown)


LuFinch (approved):

weishi-deng (approved):

chuanqi129:

PR #3771, opened 2026-05-28
Enable torch_xpu and inductor test suites in CI

chuanqi129:
## Summary

- Add `torch_xpu` (test/xpu/* + test_xpu.py) to the basic CI suite with full log processing and failure capture
- Add `inductor` (test/inductor/*) as a standalone test suite (not in basic scope) with its own log/failure pipeline
- Integrate both suites into `ut_result_check.sh`, `check-ut.py`, and the workflow failure collection steps
- Collect JUnit XML reports from `run_test.py` for structured failure analysis


chuanqi129:
@copilot /pr-review

PR #3815, opened 2026-06-01
Fix test_nan_assert in test_c10d_xccl suite

HSienkiewicz:
Fix test_nan_assert in test_c10d_xccl suite by adding skip if GPUs device count is lower than expected


libohao1201 (commented):
LGTM.

zxd1997066 (approved):

PR #3825, opened 2026-06-02
ci: remove unnecessary actions:write permission from auto_label workflow

chuanqi129:
## Summary

- Remove `actions: write` permission from `auto_label.yml` GITHUB_TOKEN since the workflow cancellation step uses `MERGE_TOKEN` (PAT), not GITHUB_TOKEN
- Follows principle of least privilege for workflow token permissions
- Addresses OpenSSF Scorecard warning about excessive token permissions

## Details

The `auto_label.yml` workflow declared `actions: write` at the top level, but the only step that interacts with the Actions API (cancelling workflow runs at line 272-278) uses `secrets.MERGE_TOKEN` (a PAT), not `GITHUB_TOKEN`. Therefore `actions: write` on the default token is unnecessary and should be removed.


mengfei25 (approved):

chuanqi129:

PR #3826, opened 2026-06-02
Support independent bias checking in GRU fused cell kernel

jianyizh:
## Summary
- Split `has_bias` into `has_bias1`/`has_bias2` in `GruCellForwardFunctor`, matching the pattern from #3770 (LSTM fix)
- Relax `checkSizes` to validate each bias tensor independently rather than requiring both to exist together

## Motivation
The GRU fused cell kernel currently uses a single `has_bias = Bias1_.data != nullptr` flag to control loading of both input and hidden biases. This prevents calling the kernel with only one bias defined (e.g., `input_bias=None, hidden_bias=defined`), which is needed for the pre-computed input gates optimization (pytorch/pytorch#185507) where the input bias is already folded into the matmul result.

This is the GRU counterpart to #3770 which applied the same fix to the LSTM kernel.

### DIFF
--- a/src/ATen/native/xpu/sycl/RNNKernels.cpp
+++ b/src/ATen/native/xpu/sycl/RNNKernels.cpp
@@ -59,7 +59,10 @@ void checkSizes(
   if (input_bias->defined()) {
     checkDim(c, input_bias, 1);
     checkNumel(c, input_bias, gates_size);
-    checkSameSize(c, input_bias, hidden_bias);
+  }
+  if (hidden_bias->defined()) {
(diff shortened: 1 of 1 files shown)


EikanWang:
@copilot please review the changes in depth

EikanWang:
@copilot /pr-review

jianyizh:
@copilot+claude-opus-4.8 /pr-review https://github.com/pytorch/pytorch/pull/185507 and https://github.com/intel/torch-xpu-ops/pull/3826

PR #3866, opened 2026-06-04
Fix intermittent CI failures due to Intel GPU repository mirror sync

mengfei25:
## Summary
This PR addresses intermittent CI build failures caused by Intel GPU repository mirror sync issues.
Found in https://github.com/intel/torch-xpu-ops/actions/runs/26892464152

## Solution
Remove the repo in test image, because driver already installed.

### DIFF
--- a/.github/actions/linux-testenv/action.yml
+++ b/.github/actions/linux-testenv/action.yml
@@ -37,8 +37,6 @@ runs:
         if [ ! -z ${AGENT_TOOLSDIRECTORY} ];then
           rm -rf ${AGENT_TOOLSDIRECTORY}
         fi
-        sudo apt update
-        sudo apt install -y jq numactl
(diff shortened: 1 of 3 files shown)


.github/actions/linux-testenv/action.yml
```diff
@@ -37,8 +37,6 @@ runs:
         if [ ! -z ${AGENT_TOOLSDIRECTORY} ];then
           rm -rf ${AGENT_TOOLSDIRECTORY}
         fi
-        sudo apt update
-        sudo apt install -y jq numactl
```
chuanqi129, line 41: Why we don't need those 2 packages?

mengfei25: already installed in the image and CI hosts

chuanqi129 (approved):

PR #3867, opened 2026-06-04
Fix and enhance auto merge workflow

chuanqi129:
## Summary

Fixes two reported bugs and adds several robustness enhancements to the `/merge` comment workflow.

### 1. Approval count incorrectly reports < 2 despite having 2 approvals

**Root cause:** The review deduplication logic (`latestByUser` Map) kept the *latest* review per user regardless of state. If a reviewer approved, then later submitted a `COMMENTED` review (e.g., just leaving inline comments), the Map overwrote their `APPROVED` entry with `COMMENTED`, effectively losing the approval.

**Fix:** Skip `COMMENTED` reviews during deduplication — they don't change a user's approval/rejection state. Also added `github.paginate()` to handle PRs with many reviews (previously limited to 30).

### 2. "Resource not accessible by personal access token" on merge

**Root cause:** Two issues:
- The workflow `permissions` block was missing `contents: write`, which is required for the merge API
- `secrets.MERGE_TOKEN` was used with no fallback — if the secret isn't configured, the API call fails with an opaque error

**Fix:** Added `contents: write` permission, use `secrets.MERGE_TOKEN || github.token` as fallback, and provide an actionable error message explaining what to configure.

## Additional Enhancements

- **Treat `neutral` conclusion as acceptable** — matches GitHub's own behavior for check runs
- **Simplify force-merge authorization** — use `author_association` (already available in the event payload) instead of a team API call that requires `org:read` and a specific team slug
- **Increase mergeability retry delay** from 3s to 5s for reliability
- **Add `per_page: 100`** to `checks.listForRef` to reduce missed checks

## Test Plan

Validated YAML syntax. Full functional testing requires deploying the workflow and triggering via `/merge` comment on a PR with:
- 2+ approvals (one reviewer having also left a comment review)
- Passing CI checks
- No merge conflicts

### DIFF
--- a/.github/workflows/auto_merge_by_comment.yml
+++ b/.github/workflows/auto_merge_by_comment.yml
@@ -6,6 +6,7 @@ on:
 
 permissions:
   checks: read
+  contents: write
   issues: write
   pull-requests: write
   statuses: read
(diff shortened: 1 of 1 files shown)


chuanqi129:
@copilot /pr-review

CuiYifeng (approved):
Good fixing!

RUIJIEZHONG66166 (approved):

chuanqi129:

PR #3893, opened 2026-06-08
Fix warning for Indexing kernel

CuiYifeng:
This PR is to fix `error: ‘*’ in boolean context, suggest ‘&&’ instead [-Werror=int-in-bool-context]`.

**Improvements to type handling:**

* Added `#include <type_traits>` to enable type trait checks for conditional logic based on data type.
* Updated both `IndexFuncSmallIndexFunctor` and `IndexFuncLargeIndexFunctor` to use `std::is_same_v<T, bool>`: now, if the data type is `bool`, the operation uses logical AND (`&&`) with `alpha_` instead of multiplication, ensuring correct behavior for boolean tensors.

### DIFF
--- a/src/ATen/native/xpu/sycl/Indexing.cpp
+++ b/src/ATen/native/xpu/sycl/Indexing.cpp
@@ -981,8 +981,7 @@ template <
     bool IndexIsMajor,
     typename func_t>
 struct IndexFuncLargeIndexFunctor : public __SYCL_KER_CONFIG_CONVENTION__ {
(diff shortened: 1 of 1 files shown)


LuFinch (approved):

BBBela (approved):
I just encountered the problem with `intel::reqd_sub_group_size(SIMD)` when I tried to rebuild PyTorch locally. I wanted to create PR with same fix, but I see that this PR already fixes that. Thank you for that change!

PR #3900, opened 2026-06-08
Exclude frexp operator from _xpu_computation_op_list.

BBBela:
WIP


BBBela:
After further and more in-depth investigation, it looks like changes from this PR are not correct. There is a real problem on Windows OS that needs to be checked and addressed. On Linux everything works correctly.

PR #3911, opened 2026-06-09
Refine reduction over subgroup

EikanWang:
Mark `shift_group_left+combine` loops as a must-fix performance issue: 

- The pattern generates excessive ALU-INT instructions, causing pipeline stalls.
- Reviewers should recommend `sycl::reduce_over_group` instead.

### DIFF
--- a/.claude/skills/pr-review/SKILL.md
+++ b/.claude/skills/pr-review/SKILL.md
@@ -307,6 +307,64 @@ int subgroup_size = 16;
 int xecore_count = zeDeviceGetSubsliceCount(device);  // legacy API name
 ```
 
+## Subgroup Reduction Anti-Pattern
+
+**Principle: Never use `sycl::shift_group_left` in a loop to perform subgroup reductions (sum, min, max, mean). Always use `sycl::reduce_over_group` instead.**
+
+This is a **🔴 Must Fix** category.
+
+### The Problem
+
+The following pattern manually implements subgroup reduction using shift operations:
+
+```cpp
+// 🔴 Bad — shift_group_left generates massive int-related instructions,
+// causing ALU-INT pipe stall
+for (int offset = 1; offset < sg_size; offset <<= 1) {
+  arg_t other = sycl::shift_group_left(sg, value[i], offset);
+  value[i] = combine(value[i], other);  // combine = add/min/max
+}
+```
+
+`sycl::shift_group_left` produces excessive integer ALU instructions for index manipulation, which stalls the ALU-INT pipeline and degrades performance. The SYCL runtime's built-in `reduce_over_group` can use hardware-optimized reduction paths.
+
+### The Fix
+
+Replace with `sycl::reduce_over_group` using the appropriate SYCL binary operation:
+
+```cpp
+// ✅ Good — uses hardware-optimized reduction
+value[i] = sycl::reduce_over_group(sg, value[i], sycl::plus<arg_t>());    // for sum/mean
+value[i] = sycl::reduce_over_group(sg, value[i], sycl::minimum<arg_t>()); // for min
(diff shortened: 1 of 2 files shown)


LuFinch (approved):
LGTM

.claude/skills/pr-review/SKILL.md
```diff
@@ -307,6 +307,64 @@ int subgroup_size = 16;
 int xecore_count = zeDeviceGetSubsliceCount(device);  // legacy API name
 ```
 
+## Subgroup Reduction Anti-Pattern
+
+**Principle: Never use `sycl::shift_group_left` in a loop to perform subgroup reductions (sum, min, max, mean). Always use `sycl::reduce_over_group` instead.**
```
CuiYifeng, line 312: Just a reminder: `T` should be a fundamental type.
https://github.khronos.org/SYCL_Reference/iface/group-algorithms-library.html#reduce-over-group

guangyey (approved):

.claude/skills/pr-review/SKILL.md
```diff
@@ -307,6 +307,64 @@ int subgroup_size = 16;
...
+    arg_t other = sycl::shift_group_left(sg, value[i], offset);
+    value[i] = combine(value[i], other);
+  }
+}
+```
+
+### Where to Check
+
+- `src/ATen/native/xpu/sycl/` — All SYCL kernel files
+- Any file using `sycl::shift_group_left` in combination with arithmetic/comparison operators
```
liangan1, line 365:
```suggestion
- Any file using `sycl::shift_group_left` in combination with arithmetic/comparison operators to do the subgroup level reduction.
```

CuiYifeng:
I will provide more suggestion after experiment.

.claude/skills/pr-review/SKILL.md
EikanWang: Why does the skill need to care about `T`?

.claude/skills/pr-review/SKILL.md
EikanWang: @liangan1 , it is a child-section of `## Subgroup Reduction Anti-Pattern`. So, I suppose the agent could understand the context well.

.claude/skills/pr-review/SKILL.md
CuiYifeng: Discussed offline. The skill for PR review doesn't need to care about `T`. Thanks for the correcting.

CuiYifeng (approved):

chuanqi129:

Please ensure the repository has a `MERGE_TOKEN` secret configured with `contents: write` permission, or that the default GITHUB_TOKEN has sufficient permissions.

chuanqi129:

.claude/skills/pr-review/SKILL.md
```diff
@@ -307,6 +307,64 @@ int subgroup_size = 16;
...
+```cpp
+// 🔴 Bad — shift_group_left generates massive int-related instructions,
+// causing ALU-INT pipe stall
+for (int offset = 1; offset < sg_size; offset <<= 1) {
+  arg_t other = sycl::shift_group_left(sg, value[i], offset);
+  value[i] = combine(value[i], other);  // combine = add/min/max
+}
+```
+
+`sycl::shift_group_left` produces excessive integer ALU instructions for index manipulation, which stalls the ALU-INT pipeline and degrades performance. The SYCL runtime's built-in `reduce_over_group` can use hardware-optimized reduction paths.
```
jianyizh, line 329: I think the stall is caused by indirect read, mov(16) r[a0.0]<1,0>:d, it will take 300+ cycles. So pipeline stall because this instruction is ready to be scheduled to int-pipeline but read of r[a0.0] is not finished. Anyway we should use reduce_over_group for subgroup reduction

PR #3914, opened 2026-06-09
Add compiler version block for SymmetricMemory

Chao1Han:
This PR introduces a conditional build step for the XPU SymmetricMemory feature to ensure compatibility with the required oneAPI compiler version. The main change is to skip compiling the `XPUSymmetricMemory.cpp` translation unit if the SYCL compiler version is older than 2026.0, preventing build errors on unsupported toolchains.

### DIFF
--- a/src/xccl/CMakeLists.txt
+++ b/src/xccl/CMakeLists.txt
@@ -13,6 +13,17 @@ file(GLOB xccl_cpp "*.cpp")
 list(REMOVE_ITEM xccl_cpp "${CMAKE_CURRENT_SOURCE_DIR}/NanCheck_XPU.cpp")
 list(REMOVE_ITEM xccl_cpp "${CMAKE_CURRENT_SOURCE_DIR}/Signal.cpp")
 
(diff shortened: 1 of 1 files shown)


CuiYifeng (approved):

chuanqi129 (approved):

zhangxiaoli73 (approved):

chuanqi129:

Please ensure the repository has a `MERGE_TOKEN` secret configured with `contents: write` permission, or that the default GITHUB_TOKEN has sufficient permissions.

chuanqi129:
✅ PR has been successfully merged by @chuanqi129 (force merge).

guangyey:
I don't think this is a good implementation. Instead, we should use an `#ifdef SYCL_COMPILER_VERSION` guard and provide a clear runtime error indicating that the feature is not supported on older compiler versions, along with guidance to upgrade to a newer compiler. Printing a message at build time is easy to miss, and users may not understand why the feature is unavailable at runtime.

PR #3918, opened 2026-06-09
Fix GCC 13 build failure in sycltla flash attention headers.

BBBela:
This PR fixes the build error when building with GCC 13.

Add `#pragma GCC diagnostic ignored "-Wpragmas"` before the existing GCC diagnostic suppression
blocks in `mha_bwd.h` and `mha_fwd.h`. This silences the `-Wpragmas` warning that GCC 13 emits
when it encounters the unrecognized `-Wtemplate-id-cdtor` flag (introduced in GCC 14), preventing
`-Werror=pragmas` from turning it into a hard error.

The problematic `-Wtemplate-id-cdtor` was introduced in [THIS RECENT COMMIT](https://github.com/intel/torch-xpu-ops/commit/0b3806ed41694c14d7f30851167204bdee41ccab)

The fix has no effect on GCC 14+ (which recognizes the flag natively) or on clang/icpx builds
(which use the `#pragma clang diagnostic` path and already suppress unknown warning options).


BBBela:
Closing as this is duplicated with #3916

PR #3920, opened 2026-06-09
Add skill to avoid using shift_group_left for reduction

EikanWang:
Refine the pr-review skill to address Liangang's comments and add more test cases to validate the newly added changes for `reduce over subgroup`

### DIFF
--- a/.claude/commands/validate-pr-review-test.md
+++ b/.claude/commands/validate-pr-review-test.md
@@ -1,0 +1,25 @@
+# Review Test Cases
+
+Review the files under `test/agentic/pr-review` to validate the pr-review skill.
+
+## Instructions
+
+1. Load the `pr-review` skill.
+2. For each file in `test/agentic/pr-review/`, perform a review as if the file were part of a PR submission.
+3. For each test case in the file:
+   - Identify whether it is marked as "Expected: FLAG" (bad) or "Expected: PASS" (good).
+   - Apply the pr-review skill's detection rules (e.g., subgroup reduction anti-pattern).
+   - Report whether the skill would correctly flag or pass each case.
+4. Output a summary table showing:
+   - Case name
+   - Expected result (FLAG or PASS)
+   - Actual result from review (FLAGGED or PASSED)
+   - Verdict (CORRECT or MISSED or FALSE POSITIVE)
+
+## Optional argument
+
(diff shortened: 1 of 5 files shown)


EikanWang:
@copilot+claude-opus-4.6 /validate-pr-review-test

EikanWang:
@chuanqi129 , @liangan1 , do you think this PR could be positioned as a reference to validate newly added SKILL documents? If so, @chuanqi129 , we need to update the label rules a little bit to avoid triggering the CI process if the PR only changes files under `test/agentic`.

guangyey (approved):

chuanqi129 (approved):

.claude/commands/validate-pr-review.md
chuanqi129, line 1: Suggest to name it as pr-review-test directly to simplify the cmd

EikanWang: Do you think `validate-pr-view` is a good cmd name?

.claude/commands/validate-pr-review-test.md
```diff
@@ -0,0 +1,25 @@
+# Review Test Cases
```
liangan1: Who will use this command? The skills maintainer? or add it in the CI in the future?

.claude/commands/validate-pr-review.md
liangan1: The test for skills is not included by  the https://github.com/intel/torch-xpu-ops/blob/b0d1cef75a94491bad061131f205a8718e5b317a/.claude/skills/skill-writer/SKILL.md, whether we need another pr to add this part in the skills writer?

.claude/commands/validate-pr-review-test.md
EikanWang: SKILL contributors. Any SKILL contributor needs to provide test cases. That's why we need this command.

.claude/commands/validate-pr-review.md
EikanWang: For PR review skill, I predict that we will continue adding items to the checklist. However, it is hard for reviewers to check whether the newly added items are valid. And the skill-writer SKILL is a static check while this command is a runtime check.

.claude/commands/validate-pr-review-test.md
EikanWang: In the future, it will be part of CI dedicated to SKILL changes.

PR #3934, opened 2026-06-09
ci: add workflow to validate build with dpclang

dvrogozh:
Current limitations for the pytorch build with dpclang:
1. Triton can not find dpclang and dpclang sycl (dynamo and inductor cases fail)
2. oneCCL not available: https://github.com/uxlfoundation/oneCCL/issues/191
3. PTI not available: https://github.com/intel/pti-gpu/issues/111
4. MKL not enabled (as it's closed source library)

### DIFF
--- a/.github/workflows/pull_dpclang.yml
+++ b/.github/workflows/pull_dpclang.yml
@@ -31,8 +31,8 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: intel/llvm
-          # WA for https://github.com/intel/llvm/issues/22308
-          ref: 8c9609e0723795c0c82b7d2d097eed406e71156b # sycl-rel-7_0
+          # TODO: change to versioned release tag when available
+          ref: eb36014ea34e57e534af8febd14f58dfb181f96e  # branch: sycl-rel-7_0
           path: intel-llvm
       - name: Install prerequisites
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/pull_dpclang.yml
```diff
@@ -0,0 +1,129 @@
...
+          # TODO: change to versioned release tag when available
+          ref: eb36014ea34e57e534af8febd14f58dfb181f96e  # branch: sycl-rel-7_0
+          path: intel-llvm
+      - name: Install prerequisites
+        run: |
+          rm -rf /tmp/xpu-tool/myvenv
+          /opt/python/cp314-cp314/bin/python3 -m venv /tmp/xpu-tool/myvenv
+          pip install cmake ninja
+          dnf install -y libtool libzstd-devel
+      - name: Build intel-llvm
```
chuanqi129, line 51: Is there any pre-built binary in public channel?

dvrogozh: There are pre-built binaries, but they are built for one of the later Ubuntu versions and don't run on almalinux 8 which we need for the build stage - this concerns pre-built binaries available at commit granularity which we currently need to setup CI on our side.

mengfei25 (approved):

chuanqi129 (approved):

PR #3942, opened 2026-06-10
Add missing input checks to quantized_max_pool2d

cyyever:
The XPU `quantized_max_pool2d` kernel omitted two `TORCH_CHECK`s the CPU reference (`q_maxpool_2d`) performs: `padding <= kernel_size/2` and non-zero input dims. With `padding > kernel/2`, an all-padding corner window left the output at the uint8 sentinel `lowest()=0` -- silently wrong data instead of an error. `quantized::max_pool2d` is a custom op (no structured meta), so the kernel is the only validation point.

Test: negative case added to `test/xpu/quantization/core/test_quantized_op_xpu.py`

```
pytest test/xpu/quantization/core/test_quantized_op_xpu.py -k invalid_padding
```


EikanWang (approved):

guangyey (commented):
I think this op `quantized_max_pool2d_xpu` will be removed after https://github.com/pytorch/pytorch/pull/181233. Let's pend this PR first and try to land after 181233.

guangyey (approved):

guangyey:
It's fine to keep this usage.

PR #3943, opened 2026-06-10
Use at::ceil_div instead of local CeilDiv helpers

cyyever:
Delete the duplicate `CeilDiv` templates in `ScanUtils.h` and `EmbeddingBackwardKernel.h` and call `at::ceil_div` (`ATen/ceil_div.h`), already used in 10+ files in this repo. Same formula, bit-identical results; all call sites pass matching integral types, and the upstream template is stricter (mixed types fail to compile instead of silently promoting).

Test: no behavior change; covered by existing scan/embedding/triu-tril UTs.


guangyey (approved):
Thanks!

EikanWang (changes requested):
`at::ceil_div` triggers the aten dispatch logic. So, please use `ceil_div` directly instead of `at::ceil_div`.

https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/ceil_div.h - FYI

EikanWang (approved):
My bad. I misunderstood `at::ceil_div` requires a dispatch.

PR #3944, opened 2026-06-10
Use at::ceil_div instead of local div_up/divup/ceil_div clones

cyyever:
Follow-up to #3943: replace four more local spellings of ceiling division (`ceil_div` in TensorModeKernel, `div_up` in TensorShapeKernels and BatchNormKernels, `divup` in WelfordNorm.h) with `at::ceil_div`. All call sites pass matching integral types, so results are bit-identical.

`Reduce.h` keeps its `div_up`: it mirrors CUDA's `Reduce.cuh`, which has the same local helper, and its call sites rely on implicit `int`->`int64_t` promotion that the deducing template would reject.

Test: no behavior change; covered by existing mode/cat/batch_norm UTs.


guangyey (approved):
Thanks!

EikanWang (approved):

PR #3945, opened 2026-06-10
Use margin loss shape checks from ATen/native/LossMulti.h

cyyever:
Delete the verbatim copies of `multi_margin_loss_shape_check` and `multilabel_margin_loss_shape_check`; upstream exports them inline in `<ATen/native/LossMulti.h>` (same signature and messages), and the CUDA kernels call them from there. Call sites resolve to `at::native::` through the enclosing namespace. Net -74 lines.

Test: no behavior change; covered by existing margin-loss UTs.


guangyey (approved):
Nice!

EikanWang (approved):

guangyey:
waits for CI build results.

PR #3950, opened 2026-06-10
Use C++20 bit_ceil/bit_floor instead of hand-rolled power-of-2 helpers

cyyever:
Replace `c10::llvm::PowerOf2Ceil`/`PowerOf2Floor` (BatchKernel.h, top-k kernels) and the local bit-twiddling helpers `next_highest_power_of_2` (TensorModeKernel) and `radix_select_last_power2` (SortingKernels, whose body is actually a bit_ceil despite the name) with `std::bit_ceil`/`std::bit_floor`.

Behavior is identical on all reachable inputs: for n >= 1 the helpers and `std::bit_ceil` agree bit-for-bit, and at n=0 both switch call sites land in the same `default`/assert branch. `std::bit_ceil` additionally fixes the MSVC path in TensorModeKernel that skipped the `n >> 32` smear. The `lastPow2`/`last_pow2` helpers are deliberately kept: they return `max(1, bit_floor(n))`, a different contract that guards work-group sizes against 0.

Test: no behavior change; covered by existing top-k/mode/sort UTs.

### DIFF
--- a/src/ATen/native/xpu/sycl/BatchKernel.h
+++ b/src/ATen/native/xpu/sycl/BatchKernel.h
@@ -9,10 +9,9 @@
  */
 
 #pragma once
-#include <c10/util/llvmMathExtras.h>
-
 #include <comm/SYCLContext.h>
 #include <algorithm>
+#include <bit>
 
 namespace at::native::xpu {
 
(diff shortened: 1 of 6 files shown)


guangyey (approved):

guangyey:
Need to resolve the conflicts.

cyyever:
@guangyey Fixed

guangyey (approved):
Thanks for your update!

PR #3935, opened 2026-06-10
Use C++20 STL utils and enable C++20 for Windows SYCL device code

cyyever:
Replace c10 utility usages with their C++20 standard-library equivalents:

- `c10::array_of` -> `std::to_array` in SDPUtils.cpp
- `c10::ln_2<T>` -> `std::numbers::ln2_v<T>` in LogAddExpKernels.cpp
- `c10::frac_1_sqrt_pi<T>` -> `std::numbers::inv_sqrtpi_v<T>` in MathExtensions.h

The math constant replacements are bit-for-bit identical for both `float` and `double`: c10 and `std::numbers` represent the same mathematical constants and round to the same IEEE-754 value. `calc_erfinv<T>` is only instantiated with `float`/`double` (the `BFloat16`/`Half` overloads cast to `float`), so `std::numbers` (floating-point only) covers all cases.

Also fixes C++20 support for Windows SYCL device compilation, which `std::numbers` (here and in upstream `c10/util/MathConstants.h` after pytorch/pytorch#186877) depends on:

- icx in clang-cl mode ignores the GCC-style `-std=c++20`, so Windows device code silently compiled as C++17. Pass `-Qstd=c++20` instead.
- Real C++20 exposed that MSVC's `std::gcd` dispatches on a runtime ISA global and x86 builtins, which SYCL kernels cannot use. `reduce_fraction` now uses Euclid's algorithm on MSVC (matching CUDA's `Reduce.cuh`).

Test: windows-build CI; constants verified bit-identical.

PR #3954, opened 2026-06-10
Use C++20 std::bit_cast instead of union type punning

cyyever:
Rewrite the `__float_as_int` family in `comm/Scalar.h` with `std::bit_cast`; reading the inactive union member was UB. Also drop the unused `__double_as_int`/`__int_as_double` (the latter did a value conversion, not a bit cast) and the six now-unused union types. The helper names are kept since they mirror the CUDA intrinsics at the kernel call sites (Atomics.h, SortingCommon.h, SortingRadixSelect.h). `std::bit_cast` is constexpr and SYCL-device-safe.

Test: no behavior change; covered by existing atomics/sort UTs.

### DIFF
--- a/src/comm/Scalar.h
+++ b/src/comm/Scalar.h
@@ -12,24 +12,22 @@
 
 #include <sycl/sycl.hpp>
 
-#include <bit>
-
 static inline uint32_t __float_as_int(float val) {
-  return std::bit_cast<uint32_t>(val);
+  return sycl::bit_cast<uint32_t>(val);
 }
 
 static inline float __int_as_float(uint32_t val) {
-  return std::bit_cast<float>(val);
+  return sycl::bit_cast<float>(val);
 }
 
 static inline unsigned long long __double_as_long_long(double val) {
-  return std::bit_cast<unsigned long long>(val);
+  return sycl::bit_cast<unsigned long long>(val);
 }
 
 static inline double __long_long_as_double(unsigned long long val) {
-  return std::bit_cast<double>(val);
+  return sycl::bit_cast<double>(val);
 }
 
 static inline sycl::half __ushort_as_half(unsigned short int val) {
-  return std::bit_cast<sycl::half>(val);
+  return sycl::bit_cast<sycl::half>(val);
 }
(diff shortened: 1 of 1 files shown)


src/comm/Scalar.h
```diff
@@ -12,74 +12,24 @@
...
+  return std::bit_cast<unsigned long long>(val);
 }
 
 static inline double __long_long_as_double(unsigned long long val) {
-  ull_to_double cn;
-  cn.in = val;
-  return cn.out;
-}
-
-static inline uint32_t __double_as_int(double val) {
```
guangyey, line 71: `__double_as_int` seems missing.

src/comm/Scalar.h
```diff
@@ -12,74 +12,24 @@
...
-  double_to_u32 cn;
-  cn.in = val;
-  return cn.low;
-}
-
-static inline float __int_as_double(uint32_t val) {
```
guangyey, line 77: `__int_as_double` as well

src/comm/Scalar.h
cyyever: It is not used and removed.

src/comm/Scalar.h
cyyever: Also not used.

guangyey (dismissed):

guangyey:
@copilot please help check whether the functions are used in device code only.

EikanWang (approved):

src/comm/Scalar.h
```diff
@@ -12,74 +12,24 @@
...
-    uint32_t high, low;
-  };
-};
+#include <bit>
 
 static inline uint32_t __float_as_int(float val) {
-  f32_to_u32 cn;
-  cn.in = val;
-  return cn.out;
+  return std::bit_cast<uint32_t>(val);
```
guangyey:
```suggestion
  return sycl::bit_cast<uint32_t>(val);
```
Hi @cyyever I recall that the functions will be used in device code and we recommend using sycl:: one if it exists.

guangyey (approved):
Thanks!

jianyizh (approved):

EikanWang:
@chuanqi129 , do you know why the CI signal is not green?

guangyey:
Because the c++20 is not enabled on Windows. We have to land https://github.com/intel/torch-xpu-ops/pull/3935 first and then the others.

PR #3958, opened 2026-06-10
Check if CUDA is available before querying capabilities.

### LINKED ISSUE #3846: [release/2.13] Hardcoded torch.cuda API calls in upstream tests — porting bug
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.nn.test_convolution_xpu.TestConvolutionNN,test_slow_conv_dilated3d_unbatched
op_ut,third_party.torch-xpu-ops.test.xpu.test_indexing_xpu.TestIndexingXPU,test_index_add_smem_stage_alignment_regression_xpu_bfloat16
op_ut,third_party.torch-xpu-ops.test.xpu.test_indexing_xpu.TestIndexingXPU,test_index_add_smem_stage_alignment_regression_xpu_float32

## Root Cause

These upstream PyTorch tests contain hardcoded `torch.cuda` API calls that fail on XPU-only builds:

- **index_add tests** (`test_indexing.py:2152`): call `torch.cuda.get_device_properties(0).multi_processor_count` to determine shared memory staging parameters. This triggers `AssertionError: Torch not compiled with CUDA enabled` on XPU builds.
- **slow_conv test** (`test_convolution.py:613`): uses `torch.randn(..., device=device)` which internally triggers CUDA lazy init on this code path.

This is a classic CUDA→XPU porting issue. The XPU test wraps the upstream test but the upstream code still uses CUDA-specific APIs.

**Fix:** Either:
1. Override the test method in the XPU test file to replace `torch.cuda.get_device_properties(0).multi_processor_count` with `torch.xpu.get_device_properties(0).max_compute_units` (or equivalent).
2. Submit an upstream PyTorch PR to make the test device-agnostic.
3. Skip these tests on XPU with an appropriate comment.

## Reproducer

```bash
source ~/.bashrc && cd third_party/torch-xpu-ops/test/xpu && python -m pytest -sxv test_indexing_xpu.py -k "test_index_add_smem_stage_alignment_regression_xpu_bfloat16"
```

```bash
source ~/.bashrc && cd third_party/torch-xpu-ops/test/xpu && python -m pytest -sxv nn/test_convolution_xpu.py -k "test_slow_conv_dilated3d_unbatched"
```

## Representative Error

```
  File "test/test_indexing.py", line 2152, in test_index_add_smem_stage_alignment_regression
    sm = torch.cuda.get_device_properties(0).multi_processor_count
  File "torch/cuda/__init__.py", line 733, in get_device_properties
    _lazy_init()
  File "torch/cuda/__init__.py", line 518, in _lazy_init
    raise AssertionError("Torch not compiled with CUDA enabled")
AssertionError: Torch not compiled with CUDA enabled
```

AKloniecki:
Partially fixes: https://github.com/intel/torch-xpu-ops/issues/3846


pbielak (approved):

BBBela (commented):
1. How does this PR relates to the issue mentioned in the description
 #3846?
This PR changes the `conv3d_depthwise_naive_groups` test case, while it is not mentioned in the above issue at all.
2. Is this change necessary? From what I know, the `instantiate_device_type()` only instantiates the cuda device if `torch.cuda.is_available()` evaluates to `True`. So if `device` parameter is `"cuda"`, then it should be guaranteed that `torch.cuda.is_available()` is `True`. Unless I am missing something - please let me know then 😉 
3. Is there any failing test case that can be fixed with changes from this PR or is it rather some preventive change?

test/xpu/nn/test_convolution_xpu.py
```diff
@@ -179,9 +179,11 @@ def conv3d_depthwise_naive_groups(self, device, dtype):
             i2 = i.detach()[:, 1:].clone().requires_grad_()
             output2 = m2(i2)
             output2.backward(grad_output[:, offset:].contiguous())
-            is_cuda_sm86 = device.startswith(
-                "cuda"
-            ) and torch.cuda.get_device_capability(0) == (8, 6)
+            is_cuda_sm86 = (
+                device.startswith("cuda")
```
guangyey, line 183: Doesn't `device` be xpu:0`?

AKloniecki: no. this test was copied directly from pytorch, and this is the original execution path, having special flow depending on capabilities.
Since we don't care about these capabilities, we could completely remove this deduction, but I've decided to only fix the faulty `torch.cuda.get_device_capability()` call, when cuda is not available, to keep test easier to maintain.

BBBela: @AKloniecki - but how is it possible that this test case is instantiated with `device` parameter equal to `"cuda"`? This test case should never be instantiated with the `device` parameter equal to `"cuda"` on the platform that doesn't have CUDA from what I know.
And by the way - this test case is passing in latest Nightly CI - can you share the CI workflow where this test case failed?

BBBela:
Hello @AKloniecki - could you please answer my review comments?

PR #3962, opened 2026-06-10
Add SYCL based FFT kernel

dbollapr:
Adds a SYCL based FFT kernel with limited size support for 2D Complex-to-Complex batched problems.

Details:
- A new dispatching check in `xpu/SpectralOps.cpp` that checks if the given size can be executed by the SYCL kernel.
- The kernel itself is added to the `xpu/sycl/FFTKernelFunctor.cpp`. It re-uses the same dispatch structure in `xpu/mkl/SpectralOps.cpp`.
- The kernel uses the SYCL online compilation feature to JIT compile the kernel from source string during `commit` time.
- Moved utility functions like `_sort_dims(..)`, `_dft_scale(..)`, `fft_apply_normalization(..)` to `xpu/sycl/FFTKernelFunctor.cpp` to re-use them. 
- Added a test `test_spectral_ops_xpu.py::TestFFTXPU::test_fft_c2c_sycl_kernel_xpu` to selectively verify functionality of this kernel.

@dvrogozh @riverliuintel @EikanWang


dvrogozh:
@dbollapr, linter fails. Please, run:
```
pip install lintrunner
lintrunner -a
```
This will modify sources. Create a commit and push to the PR.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -11,13 +11,823 @@
...
+
+    for(dir_val = 0; dir_val < 2; ++dir_val) {
+      if(desc.which_dir != 2 && dir_val != desc.which_dir) continue;
+      auto kernel_name_dir = kernel_name + ((dir_val == 0) ? "_fwd" : "_bwd");
+      char scale_buf[64];
+      std::snprintf(scale_buf, sizeof(scale_buf), "%a", dir_val == 0 ? desc.fwd_scale :
+                                                                       dim == 0 ? desc.bwd_scale : 1.0);
+      std::vector<std::string> fft_build_opts = { "-DKERNEL_NAME=" + kernel_name_dir,
+                                  "-DREGSIZE=" + std::to_string(num_regs),
+                                  "-DFFT_LEN=" + std::to_string(desc.fft_len[dim]),
```
dvrogozh: This one does not seem actually used:
```
$ fgrep -rsn FFT_LEN src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
525:                                  "-DFFT_LEN=" + std::to_string(desc.fft_len[dim]),
```

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -11,13 +11,823 @@
...
+                                  };
+      if constexpr (std::is_same_v<T, float>) {
+        fft_build_opts.push_back("-DDFT_SINGLE_PRECISION");
+      } else if constexpr (std::is_same_v<T, double>) {
+        fft_build_opts.push_back("-DDFT_DOUBLE_PRECISION");
+      } else {
+        throw std::runtime_error("Unsupported data type");
+      };
+
+      auto exe_bundle = syclexp::build(src_bundle, syclexp::properties{syclexp::build_options{fft_build_opts}});
```
dvrogozh: @EikanWang , what do you think regarding this approach to build FFT kernel at runtime?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -11,13 +11,823 @@
...
+                                  "-DOUTER_BATCH_BWD_DIST=" + std::to_string(outer_batch_bwd_dist),
+                                  "-DINNER_BATCH_BWD_DIST=" + std::to_string(inner_batch_bwd_dist)
+                                  };
+      if constexpr (std::is_same_v<T, float>) {
+        fft_build_opts.push_back("-DDFT_SINGLE_PRECISION");
+      } else if constexpr (std::is_same_v<T, double>) {
+        fft_build_opts.push_back("-DDFT_DOUBLE_PRECISION");
+      } else {
+        throw std::runtime_error("Unsupported data type");
+      };
```
dvrogozh, line 617: DFT_SINGLE_PRECISION
```suggestion
      if constexpr (std::is_same_v<T, float>) {
        // no define needed - that's default
      } else if constexpr (std::is_same_v<T, double>) {
        fft_build_opts.push_back("-DDFT_DOUBLE_PRECISION");
      } else {
        throw std::runtime_error("Unsupported data type");
      };
```

EikanWang:
Why do we need to implement it rather than oneMKL?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
EikanWang: @dvrogozh , I cannot capture the motivation. What does this pr serve for?

riverliuintel:
@EikanWang  This is for clean open source build solution that doesn't depend on oneMKL package.

src/ATen/native/xpu/SpectralOps.cpp
```diff
@@ -26,6 +27,9 @@ Tensor _fft_c2c_xpu(
     bool forward) {
   TORCH_CHECK(self.is_complex());
 
+  if (native::xpu::_is_fft_size_supported_sycl(self, dim)) {
+    return native::xpu::_fft_c2c_sycl(self, dim, normalization, forward);
+  }
```
EikanWang: It means that the sycl-based priority is higher than the oneMKL. So, I suggest that the code should be guarded by `#if defined(USE_ONEMKL_XPU)`

src/ATen/native/xpu/SpectralOps.cpp
EikanWang:
```suggestion
#if defined(USE_ONEMKL_XPU)
return native::xpu::_fft_c2c_mkl(self, dim, normalization, forward);
#else
if (native::xpu::_is_fft_size_supported_sycl(self, dim)) {
  return native::xpu::_fft_c2c_sycl(self, dim, normalization, forward);
}
Tensor out_cpu = native::_fft_c2c_mkl(
self.to(Device(at::kCPU)), dim, normalization, forward);
return out_cpu.to(Device(at::kXPU));
#endif // USE_ONEMKL_XPU
```

src/ATen/native/xpu/SpectralOps.cpp
```diff
@@ -43,6 +47,10 @@ Tensor& _fft_c2c_xpu_out(
     Tensor& out) {
   TORCH_CHECK(self.is_complex());
 
+  if (native::xpu::_is_fft_size_supported_sycl(self, dim)) {
+    return native::xpu::_fft_c2c_sycl_out(
+        self, dim, normalization, forward, out);
+  }
```
EikanWang: ditto

src/ATen/native/xpu/sycl/FFTKernelFunctor.h
```diff
@@ -33,6 +33,40 @@ void _fft_conjugate_copy_kernel(
...
+    const Tensor& self,
+    int64_t normalization,
+    IntArrayRef norm_sizes,
+    IntArrayRef dims);
+
+Tensor promote_fft_input(const Tensor& input);
+
+} // namespace impl
+
+TORCH_XPU_API bool _is_fft_size_supported_sycl(
```
EikanWang: `TORCH_XPU_API` should not be necessary.

src/ATen/native/xpu/sycl/FFTKernelFunctor.h
```diff
@@ -33,6 +33,40 @@ void _fft_conjugate_copy_kernel(
...
+    const Tensor& orig_self,
+    IntArrayRef dim);
+
+TORCH_XPU_API Tensor _fft_c2c_sycl(
```
EikanWang: ditto

src/ATen/native/xpu/sycl/FFTKernelFunctor.h
```diff
@@ -33,6 +33,40 @@ void _fft_conjugate_copy_kernel(
...
+    const Tensor& self,
+    IntArrayRef dim,
+    int64_t normalization,
+    bool forward);
+
+TORCH_XPU_API Tensor& _fft_c2c_sycl_out(
```
EikanWang: ditto

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
EikanWang: Oh, I captured the point.

EikanWang: @dbollapr , may I know the motivation to JIT the binary at runtime rather than an AOT build?

EikanWang (dismissed):
## PR Review: #3962

### Summary
Adds a SYCL-based JIT-compiled FFT kernel for 2D complex-to-complex batched transforms with limited size support (512, 768). The kernel intercepts `_fft_c2c_xpu` before MKL/CPU fallback for supported sizes.

### Correctness

**Memory leak in `_fft_with_size_sycl`** (`sycl/FFTKernelFunctor.cpp`): `commit<T>` allocates `exe_bundle` objects via `new` but `free_descriptor` is never called. Every FFT call leaks up to 6 `sycl::kernel_bundle<executable>` allocations. Either call `free_descriptor(queue, desc)` after `compute`, or make `fft_descriptor` RAII (destructor calls cleanup). An exception between `commit` and cleanup would also leak, so RAII is preferred.

**JIT compilation on every call**: `commit` calls `create_kernel_bundle_from_source` and `syclexp::build` on every invocation of `_fft_with_size_sycl`. Online SYCL compilation is expensive (potentially seconds). There is no caching of the compiled kernel bundle. For a performance-oriented kernel, this needs a cache keyed on `(device, dtype, fft_len, direction)`. Without it, this path will likely be slower than MKL for any workload that calls FFT more than once.

**`_fft_c2c_sycl` does not handle multi-pass chunking**: The MKL path in `_fft_c2c_mkl` iterates in chunks of `mkl_max_ndim=3` dimensions, ping-ponging buffers when `dim.size() > 3`. The SYCL path calls `_exec_fft_sycl` with all sorted dims at once, but the kernel only supports exactly 2D (as gated by `_is_fft_size_supported_sycl` requiring `dim.size() == 2`). If the gating check is relaxed in the future, the single-pass approach will break for higher-dimensional transforms. This is a latent design mismatch that should be documented or handled defensively.

**Scale application is asymmetric**: In `commit`, twiddle factor scaling is applied only for `dim == 0`:
```cpp
if (dim == 0)
    scale = (i == 0) ? static_cast<T>(desc.fwd_scale) : static_cast<T>(desc.bwd_scale);
```
But `fwd_scale` and `bwd_scale` are always `1.0` (set in `_fft_with_size_sycl`), and normalization is applied separately by `_fft_apply_normalization` after the kernel. This works today, but the dead `SCALE` build option and scale-in-twiddle mechanism is confusing and fragile — future callers might set non-unity scales and get incorrect results because the second dimension does not apply them.

### XPU-Specific Risks

**No synchronization after compute**: `compute` returns a `sycl::event` but the caller ignores it. The code calls `queue.throw_asynchronous()` which only checks for *previous* async errors but does NOT wait for the submitted kernels to complete. If downstream code reads the output tensor immediately (which PyTorch does), this is a race condition. Either call `prev_ev.wait()` in `compute` or `.wait()` on the returned event in `_fft_with_size_sycl`.

**SLM size validation uses wrong unit**: The check `desc.slm_size[dim] = fact0 * fact1 * 2 * sizeof(T)` computes bytes, but the kernel allocates `sycl::local_accessor<T, 1>(desc.facts[dim][0] * desc.facts[dim][1] * 2, h)` which is in *elements* (multiplied by `sizeof(T)` implicitly). The SLM check is correct, but the mismatch between the descriptor's byte-based field and the element-based accessor is confusing and could cause bugs if someone extends this code.

### Dispatch & Registration

**SYCL path intercepts before MKL unconditionally**: For supported sizes, the SYCL kernel always wins over MKL. There is no way for users to opt out of the JIT-compiled path. If the SYCL kernel has correctness or performance issues, users are stuck. Consider an environment variable escape hatch or at minimum document this priority in a comment.

### Testing

**Only `norm=None` is tested**: The test hardcodes `norm_modes = (None,)`. The normalization logic in `_fft_apply_normalization` handles `"forward"`, `"backward"`, and `"ortho"` modes. All should be tested, especially since the scale application mechanism is non-trivial.

**No batched tensor test**: The kernel handles batched inputs (tensors with more than 2 dimensions where `dim=(-2, -1)`), but the test only uses 2D tensors `(s0, s1)`. A batched test like `torch.randn(4, 512, 768)` with `dim=(-2, -1)` is needed.

**No non-contiguous input test**: The kernel has explicit contiguity checks (`need_contiguous`). Test with a transposed or strided input to verify correctness.

**No double precision test**: The kernel has a `DFT_DOUBLE_PRECISION` code path but the test only uses `torch.cfloat`.

**No inverse-then-forward roundtrip test**: A basic property `ifft2(fft2(x)) ≈ x` would catch normalization and direction errors.

**Tolerance of 1e-3 is loose for float32 FFT**: For 512-point FFT of random data, float32 FFT implementations typically achieve ~1e-5 relative error. Using 1e-3 may mask correctness issues.

### Performance

**JIT compilation overhead dominates**: Every call to `_fft_c2c_sycl` triggers a full online compilation of the SYCL kernel source. For a 512x512 FFT, the kernel computation itself is microseconds, but JIT compilation could be milliseconds to seconds. This makes the kernel unusable in any scenario where FFT is called repeatedly (training loops, inference). A compiled kernel cache is essential for this to be a net improvement over MKL.

### Recommendation
**Request Changes**

Blocking issues: memory leak of `exe_bundle` allocations, missing synchronization (potential race), and lack of kernel caching making this strictly worse than MKL for repeated calls. Test coverage is also insufficient — missing batched, non-contiguous, double precision, and normalization mode tests.

EikanWang (commented):
Clarification on the JIT compilation point: I understand this PR targets a limited set of shapes (512, 768). The kernel caching concern is less critical if the intent is a narrow, specialized path — but it is still worth noting that every call to `_fft_c2c_sycl` with these shapes pays the full JIT cost (no amortization across calls). If the expected usage is repeated FFTs at these sizes (e.g., in a training loop), a cache would still be needed for this to be faster than MKL in practice.

The remaining blocking issues (memory leak of `exe_bundle`, missing synchronization after `compute`) and the test coverage gaps still apply regardless of the limited shape support.

EikanWang:
@guangyey , please help review the sync logic.

src/ATen/native/xpu/SpectralOps.cpp
dvrogozh: @EikanWang, do we currently build and validate `USE_ONEMKL_XPU=0` when building against oneAPI DLE?

src/ATen/native/xpu/SpectralOps.cpp
dvrogozh: It depends on relative performance and support level of the new code. If we can deliver better or the same performance with the direct support of the new code, then we can consider this as an optimization path. In such a case I would suggest to leave is as is but also have environment variable to skip it and use mkl implementation instead.

dbollapr:
@EikanWang thank you for the review, please find the response below,
### Correctness
**Memory leak in _fft_with_size_sycl (sycl/FFTKernelFunctor.cpp):** Good catch, I missed adding the called to `free_descriptor`. As you suggested I added a destructor to the `fft_descriptor` to handle the cleanup.

**JIT compilation on every call:** This has been my concern as well. On the MKL side we have an internal caching mechanism that helps retrieve a kernel that has been JIT compiled in an earlier call. I am open to suggestions here, should I work on enabling some sort of caching mechanism for this kernel or is there any plan on your side to enable such cache? 

**_fft_c2c_sycl does not handle multi-pass chunking:** Not handling the multi-pass chunking was deliberate as pointed out in the `is_fft_size_supported_sycl` check. I added a comment as a reminder to handle this case if the support is extended.

**Scale application is asymmetric:** Scale is usually provided to oneMKL using the `desc.set_value(FWD/BWD_SCALE, scale_value)` API, and this scale is only applied once in any of the 'N' dimensional FFT computation. An example, for a 2D size of N1xN2, lets say the backward scale is `1/(N1xN2)`, this scale is applied once while computing either FFT-N1 or FFT-N2, applying it to both dimensions will lead to incorrect results. This SYCL kernel also employs the same `desc.fwd_scale/bwd_scale` params which can be set accordingly and are applied during the `0`'th dim computation. I see that this scaling option/feature is not used in the MKL FFT calls and it is entirely handled by `_fft_apply_normalization`. A suggestion would be that if the actual FFT dimensions are <= 3, use the scaling feature of MKL and not use `_fft_apply_normalization` which is an extra pass over the memory.

### XPU-Specific Risks
**No synchronization after compute:** I tried to keep the structure similar to the MKL dispatch and I found no `wait()` on the event in there, so I was guessing some sort of `queue.wait()` might be happening elsewhere. If I am mistaken and sync is needed here I can add it.

**SLM size validation uses wrong unit:** Updated the SLM size to just use number of elements everywhere instead of bytes at one place and number of elements in the other.

### Dispatch & Registration
**SYCL path intercepts before MKL unconditionally:** The SYCL kernel should be functional on all supported Intel GPUs with decent performance, so it could be the first priority as it is contained within here with no dependency and if the user sees a functional or perf issue we can have an env variable as you suggested to fall back to using MKL GPU(if available). 

### Testing:
* Extended the test to run double precision.
* Added batch, `torch.randn(4, s0, s1)` as you suggested. 
* Added a transposed input tensor to trigger `need_contiguous` path.
* Kept the norm=None, as you mentioned the normalization is not handled by the kernel but by `_fft_apply_normalization` so I dont see any value in adding other norm configs especially given that adding them increasing the testing time by 4x.
* Updated the `rtol` to `1e-5`.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: @EikanWang AOT build for FFT requires templating the kernel for all supported sizes leading to multiple kernels being created which increases the size of the library. If more workloads need support this will only increase the size further. Using online compilation helps keep the library size constant.

riverliuintel:
Can we prioritize and speed up this PR land in torch-xpu-ops? @guangyey please help to review this PR.

guangyey:
Sure, I will take some time to review this.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+// Bundles are owned here via shared_ptr for the lifetime of the process; the
+// fft_descriptor holds a shared reference rather than owning them. The set of
+// distinct keys is tiny (supported sizes x dims x directions x precisions per
+// device), so no eviction is needed for now.
+class FftKernelBundleCache {
+ public:
+  using bundle_t = sycl::kernel_bundle<sycl::bundle_state::executable>;
+
+  static FftKernelBundleCache& instance() {
+    // Intentionally leaked (never deleted). The cache owns
```
dvrogozh, line 46: The `bundle_t` sycl objects you save here hold references to sycl context and devices as all these objects under the hood are reference countable. What we work with on our level are tiny wrappers subject to very specific rules. See https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reference-semantics. This means that the context is unlikely deleted before destructor for the cache would work as cache items hold the references.

Do you observe any issues if you will do a proper tear down? In my mind cache itself should be kept as `std:unique_ptr` and cache items should be subject for appropriate destruction procedures which I believe are automatic. So just having cache as `std:unique_ptr` should be enough.

dvrogozh (dismissed):

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
mkrainiuk: Unfortunately we discovered compiler issue specific to creating kernels from source by using "online" compiler API. We've reported the problem to Intel Compiler team. Based on our experiments it looks like the small leak is the best workaround that we have for now.

dvrogozh: Ah, that explains the implementation. @dbollapr , can you, please, update the comment in the sources stating TODO and explaining that this is a workaround due to the issue in the sycl runtime implementation. We also need an issue filed at https://github.com/intel/llvm for that referenced in the comment. This way this code piece will be also maintainable by those who did not directly participate in the discussion.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+
+  void put(const std::string& key, std::shared_ptr<bundle_t> bundle) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    cache_.emplace(key, std::move(bundle));
+  }
+
+  // NOTE: the critical section is guarded by std::lock_guard (RAII).
+  // It acquires the mutex on construction and releases it in its destructor
+  // when `lock` goes out of scope at the end of get()/put(). Adding a manual
+  // unlock() would cause a double-unlock bug.
```
dvrogozh: I don't see this comment necessary. It basically explain the basics about mutex and guard behavior. Those who will be updating this code are supposed to know that already.

```suggestion
```

src/ATen/native/xpu/SpectralOps.cpp
dvrogozh: My opinion that going forward we should have this SYCL implementation of FFT a default having higher priority over MKL and anything else as it's open source GPU capable implementation maintainable by anyone in the community. Before making it a default:
1. We need to verify it's performance vs. MKL
2. The https://github.com/intel/torch-xpu-ops/pull/3962/changes#r3661139579 likely needs to be resolved

In a meanwhile I suggest to introduce environment variable, for example `USE_SYCL_FFT=1|0`, to allow explicitly enable or disable this feature. I suggest that for now this should be OFF by default (`=0`). This way we will also be able to have test coverage for both MKL and SYCL scenarios.

@EikanWang , @guangyey , are you ok with such approach?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+       }).wait();
+    }
+  }
+}
+
+template <typename T>
+void commit(sycl::queue& q, fft_descriptor& desc) {
+  if (!q.get_device().is_gpu()) {
+    throw std::runtime_error("Device is not a GPU");
+  }
```
guangyey:
```suggestion
```
It must be on gpu.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+        desc.twidl_table[dim][i] = static_cast<const void*>(twidl_buf);
+      } else {
+        twidl_buf = (T*)desc.twidl_table[dim][i];
+      }
+
+      auto ker =
+          TwiddleTableKernel2FactsFunctor<T>(fact0, fact1, twidl_buf, scale);
+      q.submit([&](sycl::handler& h) {
+         h.parallel_for(sycl::range<2>(fact0, fact1), ker);
+       }).wait();
```
guangyey: why needs `wait` here.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+  const void* twidl_table[3][2] = {{nullptr}}; // [dimension][direction]
+  std::array<std::array<size_t, 3>, 3> local_work_size{
+      {{1, 1, 1}, {1, 1, 1}, {1, 1, 1}}};
+  std::array<std::array<size_t, 3>, 3> global_work_size{
+      {{1, 1, 1}, {1, 1, 1}, {1, 1, 1}}};
+  // Shared references to cached executable bundles; ownership stays with
+  // FftKernelBundleCache. [dimension][direction]
+  std::shared_ptr<sycl::kernel_bundle<sycl::bundle_state::executable>>
+      exe_bundle[3][2];
+  sycl::queue* queue = nullptr; // non-owning, for resource cleanup
```
guangyey: I think fft_descriptor doen't neeed to know the information of queue. Could we remove it?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+        continue;
+
+      T scale = static_cast<T>(1.0);
+      if (dim == 0)
+        scale = (i == 0) ? static_cast<T>(desc.fwd_scale)
+                         : static_cast<T>(desc.bwd_scale);
+
+      T* twidl_buf = nullptr;
+      if (!desc.external_workspace) {
+        twidl_buf = (T*)malloc_device(2 * fact0 * fact1 * sizeof(T), q);
```
guangyey: Don't use `malloc_device` directly here. Please use `c10::xpu::XPUCachingAllocator::raw_alloc(size)` and paird with `c10::xpu::XPUCachingAllocator::raw_delete(ptr)`, or use `at::getDeviceAllocator(at::kXPU)->allocate` for DataPtr which has an auto-lifetime cycle.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
guangyey: Could you paste the JIRA or issue link here to help us understand what compiler bug it is that leads us to use memory leak here.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+    cache_.emplace(key, std::move(bundle));
+  }
+
+  // NOTE: the critical section is guarded by std::lock_guard (RAII).
+  // It acquires the mutex on construction and releases it in its destructor
+  // when `lock` goes out of scope at the end of get()/put(). Adding a manual
+  // unlock() would cause a double-unlock bug.
+ private:
+  std::mutex mutex_;
+  std::unordered_map<std::string, std::shared_ptr<bundle_t>> cache_;
```
guangyey:
```suggestion
  std::unordered_map<std::string, bundle_t cache_;
```
I don't think we need `shared_ptr` here. `bundle_t` should be cheap to assign.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+        desc.global_work_size[dim][2]};
+    sycl::range<3> local_range{
+        desc.local_work_size[dim][0],
+        desc.local_work_size[dim][1],
+        desc.local_work_size[dim][2]};
+    sycl::nd_range<3> ndrange(global_range, local_range);
+
+    const T* input = (dim == 0) ? in : out;
+    prev_ev = q.submit([&](sycl::handler& h) {
+      h.depends_on(prev_ev);
```
guangyey: why need this dependence. In PyTorch, we only use the in-order queue.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+        write_output(out_batch, inReg, group_id, local_id, FACT1, FACT0, BATCH, BWD_DIST, i, threads_in_group);
+    }
+}
+
+)""";
+
+constexpr int supported_sizes[][3] = {{512, 32, 16}, {768, 32, 24}};
+
+template <typename T>
+struct TwiddleTableKernel2FactsFunctor {
```
guangyey, line 431: Is there any room to cache it to improve performance?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+    throw std::runtime_error("Workspace pointer is null");
+  }
+
+  // Delete any previously allocated twiddle factor buffers if this overrides
+  // workspace from internal to external.
+  if (!desc.external_workspace && desc.queue) {
+    for (int dim = 0; dim < 3; ++dim) {
+      for (int dir = 0; dir < 2; ++dir) {
+        if (desc.twidl_table[dim][dir]) {
+          sycl::free(
```
guangyey: use `raw_delete` instead.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+      } else if constexpr (std::is_same_v<T, double>) {
+        fft_build_opts.push_back("-DDFT_DOUBLE_PRECISION");
+      } else {
+        throw std::runtime_error("Unsupported data type");
+      };
+
+      // Build options fully determine the compiled binary, so use them (plus
+      // the device index) as the cache key. A '\x1f' separator keeps distinct
+      // option lists from colliding.
+      std::string cache_key = std::to_string(at::xpu::current_device());
```
guangyey: We don't need the device index in `cache_key` because in PyTorch all sycl devices share the same sycl context.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,942 @@
...
+      }
+
+      auto& bundle_cache = FftKernelBundleCache::instance();
+      auto cached = bundle_cache.get(cache_key);
+      if (!cached) {
+        if (!src_bundle) {
+          src_bundle = syclexp::create_kernel_bundle_from_source(
+              q.get_context(), syclexp::source_language::sycl, kernel_src);
+        }
+        auto exe_bundle = syclexp::build(
```
guangyey: Here, it is possible for the same source code to be built multiple times concurrently. Currently, it may be OK, we could improve it in future.

guangyey (commented):
Hi @dbollapr 

Thanks for your patience. I think @dvrogozh said reasonable. We should intorduce a `USE_SYCL_FFT` build option to enable SYCL FFT support and keep disabled by default.

The scope of this PR is quite large for me to review effectively. Would it be possible to split it into a few smaller PRs? For example:
1. introduce `USE_SYCL_FFT` and update 'mkl/SpectralOps.cpp` and `SpectralOps.cpp`. I think we could land this part first.
2. in 2nd PR, introduce the design of `FftKernelBundleCache`. 
3. introduce the design of `fft_descriptor`
4. intorduce the twiddle-factor computation
5. introduce online-verion fft kernel

Do you think this breakdown would be reasonable and convenient for you as well?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
mkrainiuk: While I was about to open github issue, I already see the PR that fixes it: https://github.com/intel/llvm/pull/22785

mkrainiuk: For tracking purposes here is a compiler bug: https://github.com/intel/llvm/issues/22786

guangyey: Thanks, I see.

src/ATen/native/xpu/SpectralOps.cpp
dbollapr: Added an env variable `USE_SYCL_SPECTRAL` and enabled it in the test too, 083e09eece9af08d1934506495bc856b2ab6e2c3

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Added the `TODO` info in the comment, 083e09eece9af08d1934506495bc856b2ab6e2c3

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Removed the comment, 083e09eece9af08d1934506495bc856b2ab6e2c3

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Done 083e09eece9af08d1934506495bc856b2ab6e2c3

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Removed `wait`, 083e09eece9af08d1934506495bc856b2ab6e2c3

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: So are you suggesting to directly pass the queue instead of storing it in the `fft_descriptor`?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Updated 083e09eece9af08d1934506495bc856b2ab6e2c3

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Removed 083e09eece9af08d1934506495bc856b2ab6e2c3

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: The twiddle table kernel is very small and should not be a bottleneck. It is consistent with MKL and can be cached if it turns out to be a bottleneck.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Done 083e09eece9af08d1934506495bc856b2ab6e2c3

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: If I understood right, the context stays the same across multiple calls so we can safely remove this line?

dbollapr:
@guangyey thank you for the review, I tried addressing the issues in 083e09eece9af08d1934506495bc856b2ab6e2c3.
As for the scope of the PR, I honestly think it can only be split in to two parts,

1. SpectralOps changes, `fft_descriptor`, twiddle calculation, online kernel etc. 
2. And the `FFTKernelBundleCache`.

This way the kernel is functional and can be verified through testing independent of the caching. Further breaking (1) into small PRs leads to non-functional/untested pieces being added to the repo which I believe is undesirable.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
guangyey: The current implementation already passed the queue.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,939 @@
...
+
+ private:
+  std::int64_t fact0_;
+  std::int64_t fact1_;
+  T* twidl_buf_;
+  T scale_;
+};
+
+template <typename T>
+void calculate_twiddle_factors(sycl::queue& q, fft_descriptor& desc) {
```
guangyey, line 457: I see we always pass the queue so it should be OK to remove it from the descriptor.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
guangyey: In PyTorch, all sycl device shares the same. So the kernel is built for device 0 can be used in device 1 directly. So we don't need `device_index` information in the key.
Is there any difference in the code source between different devices that I don't capture?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
guangyey: Thanks!

guangyey (commented):
@dbollapr Could you please help resolve the conflicts as well.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,939 @@
...
+  using bundle_t = sycl::kernel_bundle<sycl::bundle_state::executable>;
+
+  static FftKernelBundleCache& instance() {
+    // Intentionally leaked (never deleted). The cache owns
+    // sycl::kernel_bundle objects whose destructor's reference the
+    // sycl::context. During process teardown the SYCL runtime's
+    // context may already be destroyed, this is an issue with
+    // sycl runtime and tracked by intel/llvm#22786.
+    // TODO: Allocate the cache on stack once intel/llvm#22786 is fixed
+    // and available in oneapi compiler.
```
dvrogozh: Don't make reader guessing the link to the issue.
```suggestion
    // sycl runtime and tracked by https://github.com/intel/llvm/issues/22786.
    // TODO: Allocate the cache on stack once sycl runtime issue is fixed
    // and fix will be available in oneapi compiler.
```

src/ATen/native/xpu/SpectralOps.cpp
```diff
@@ -26,6 +30,12 @@ Tensor _fft_c2c_xpu(
     bool forward) {
   TORCH_CHECK(self.is_complex());
 
+  // The SYCL FFT implementation is only enabled if `USE_SYCL_SPECTRAL` is set to 1.
+  static const char* enable_sycl_fft = getenv("USE_SYCL_SPECTRAL");
```
dvrogozh: As this code block already repeats twice, can you create a function which handles reading option and returns bool? I also suggest to consider a helper to map user inputs to handle option better:
```
inline bool to_bool(std::string str) {
    static const std::unordered_map<std::string, bool> bool_map = {
        {"1", true},  {"0", false},
        {"on", true}, {"off", false},
        {"true", true}, {"false", false}
    };

    auto it = bool_map.find(str);
    if (it != bool_map.end()) {
        return it->second;
    }
    return false;
}

inline bool use_sycl_spectral() {
  static const char* USE_SYCL_SPECTRAL = std::getenv("USE_SYCL_SPECTRAL");
  return USE_SYCL_SPECTRAL && to_bool(USE_SYCL_SPECTRAL);
}
```

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dvrogozh: Cache we implement here has mutex locks for `get` and `put` operations, so adding same key should be handled without any undefined behavior or race conditions. Building same code with sycl extension is subject to correctness handling on sycl runtime side and I guess this should be already handled correctly. So this seems to be indeed a point for future optimization rather than having any negative impact.

Thus, @dbollapr, please, add in-source TODO comment for that.

test/xpu/test_spectral_ops_xpu.py
```diff
@@ -132,12 +138,46 @@ def _test_fft_half_and_chalf_not_power_of_two(self, device, dtype, op):
     self._compare_xpu_cpu(xpu_sized, cpu_sized, t)
 
 
+@ops(
```
dvrogozh: Does this test bypass torch ops registration and makes sure we test sycl implementation? or it needs to be adjusted with `USE_SYCL_SPECTRAL`?

Also, I guess there should be other already existing tests for the torch op or ops using this kernel. It would be nice if those tests actually verify all the implementation paths, i.e. having a matrix for `USE_SYCL_SPECTRAL` false and true.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Removed queue from `fft_descriptor`, f05a2d6469eac48c309b1b78b2d9b86f0bbc5085

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Updated, had to use `std::optional` as `bundle_t` is not default constructible, 5eb55547393f530c360da4c6bfed5db45b0e3ea7

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: No difference in code from different devices, I was just making sure. Thank you for the clarification, removed in f05a2d6469eac48c309b1b78b2d9b86f0bbc5085

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Added TODO comment, f05a2d6469eac48c309b1b78b2d9b86f0bbc5085

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Updated comment, f05a2d6469eac48c309b1b78b2d9b86f0bbc5085

src/ATen/native/xpu/SpectralOps.cpp
dbollapr: Added a function to handle reading the env var, f05a2d6469eac48c309b1b78b2d9b86f0bbc5085

test/xpu/test_spectral_ops_xpu.py
dbollapr: This test requires `USE_SYCL_SPECTRAL` to be set.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,943 @@
...
+        exe_bundle[dim][dir].reset();
+      }
+    }
+  }
+};
+
+// Sort transform dimensions by input layout, for best performance
+// exclude_last is for onesided transforms where the last dimension cannot be
+// reordered
+static DimVector _sort_dims(
```
guangyey: @dbollapr It seems this is duplicated with `mkl/SpectralOps.cpp`.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,943 @@
...
+    compute(queue, desc, in_data, out_data, desc.which_dir);
+  };
+
+  if (input.scalar_type() == ScalarType::Float ||
+      input.scalar_type() == ScalarType::ComplexFloat) {
+    run(float{});
+  } else {
+    run(double{});
+  }
+  queue.throw_asynchronous();
```
guangyey, line 819: Just to confirm the motivation here: is the purpose simply to report any errors immediately, so we can verify that `compute` worked correctly? However, here is an async progress that `compute` has not completed. Is it OK to remove it.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,943 @@
...
+        for(int i=0;i<rows;++i) {
+            const DFT_FTYPE *base = in + i * stride_to_next_row * DIST_TO_NEXT_THREAD * 2;
+            inReg[i][0] = base[offset * 2];
+            inReg[i][1] = base[offset * 2 + 1];
+        }
+    }
+}
+
+inline void write_output(DFT_FTYPE *out, sycl::marray<DFT_FTYPE, 2> *inReg, int group_id, int local_id,
+                         int rows, int stride_to_next_row, int batch, int odist, int fact1s_idx, int threads_in_group) {
```
guangyey: threads is a CUDA-term, is it ok we rename it to workitems? this is a nice to have, not a blocker.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,943 @@
...
+
+inline void fft_32(sycl::marray<DFT_FTYPE, 2> *in, sycl::marray<DFT_FTYPE, 2> *out, int is = 1, int id = 0, int os = 1, int od = 0)
+{
+    sycl::marray<DFT_FTYPE, 2> y[32];
+    FFT_FN(32, 4, 8);
+}
+
+inline void read_input(const DFT_FTYPE *in, sycl::marray<DFT_FTYPE, 2> *inReg, int group_id,
+                       int local_id, int rows, int stride_to_next_row, int batch, int idist) {
+    size_t offset = group_id * idist + local_id * DIST_TO_NEXT_THREAD;
```
guangyey: thread is a CUDA-term, is it ok we rename it to workitem? this is a nice-to-have, not a blocker.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,943 @@
...
+            const DFT_FTYPE *base = in + i * stride_to_next_row * DIST_TO_NEXT_THREAD * 2;
+            inReg[i][0] = base[offset * 2];
+            inReg[i][1] = base[offset * 2 + 1];
+        }
+    }
+}
+
+inline void write_output(DFT_FTYPE *out, sycl::marray<DFT_FTYPE, 2> *inReg, int group_id, int local_id,
+                         int rows, int stride_to_next_row, int batch, int odist, int fact1s_idx, int threads_in_group) {
+    size_t offset = group_id * odist + fact1s_idx * threads_in_group * DIST_TO_NEXT_THREAD + local_id * DIST_TO_NEXT_THREAD;
```
guangyey: ditto

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,943 @@
...
+
+    #pragma unroll
+    for(int i=0;i<rows;++i) {
+        DFT_FTYPE *base = out + i * stride_to_next_row * DIST_TO_NEXT_THREAD * 2;
```
guangyey: ditto

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,943 @@
...
+          "-DFACT1=" + std::to_string(fact1),
+          "-DBATCH=" + std::to_string(batch),
+          "-DFWD_DIST=" + std::to_string(fwd_dist),
+          "-DBWD_DIST=" + std::to_string(bwd_dist),
+          "-DNUM_FACT1S=" + std::to_string(num_fact1s),
+          "-DDIR_VAL=" + std::to_string(dir_val),
+          "-DSCALE=" + std::string(scale_buf),
+          "-DFFT_FACT0=" + fact0_fn,
+          "-DFFT_FACT1=" + fact1_fn,
+          "-DDIST_TO_NEXT_THREAD=" + std::to_string(dist_to_next_thread),
```
guangyey: ditto, not a blocker

guangyey (commented):
Overall design LGTM. Thanks for your update!

I just need some time to go through the kernels introduced in this PR, but the review should be quick.

test/xpu/test_spectral_ops_xpu.py
guangyey: I think @dvrogozh expects this test to be run twice, one time for `USE_SYCL_SPECTRAL=0` to test mkl path, and another time for `USE_SYCL_SPECTRAL=1` to test sycl path. Otherwise, we don't have tests to cover the mkl path when this PR lands. One thing we need to note is current design, we only parse `USE_SYCL_SPECTRAL` the first time in the same process. If we need to cover both paths, the test would be run in a different process.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Changed to `work_item`, 52a1155a56a128b66a6bf2a9dd93510db59351a8

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Changed to `work_item`, 52a1155a56a128b66a6bf2a9dd93510db59351a8

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Consolidated `_sort_dims` like other common functions(`_dft_scale` etc), 52a1155a56a128b66a6bf2a9dd93510db59351a8

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: I tried to follow the same structure as `_fft_c2c_mkl` function from `xpu/mkl/SpectralOps.cpp` which also performs a `queue.throw_asynchronous();`, so I am not sure of the original intent why it was placed in `_fft_c2c_mkl`

guangyey: Haha, I think the original `_fft_c2c_mkl` code is unnecessary.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+                         : static_cast<T>(desc.bwd_scale);
+
+      T* twidl_buf = nullptr;
+      if (!desc.external_workspace) {
+        twidl_buf = static_cast<T*>(c10::xpu::XPUCachingAllocator::raw_alloc(
+            2 * fact0 * fact1 * sizeof(T)));
+        if (twidl_buf == nullptr) {
+          throw std::runtime_error(
+              "Failed to allocate device memory for twiddle factors");
+        }
```
guangyey:
```suggestion
        TORCH_CHECK(!twidl_buf, "Failed to allocate device memory for twiddle factors")
```

guangyey: In PyTorch, we don't use `throw std::runtime_error` directly; instead, use `TORCH_CHECK` or `TORCH_INTERNAL_ASSERT`.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+    sycl::queue& q,
+    fft_descriptor& desc,
+    const T* in,
+    T* out,
+    int dir) {
+  const char* dir_suffix = (dir == 0) ? "_fwd" : "_bwd";
+  sycl::event prev_ev;
+  for (auto dim = 0; dim < desc.fft_len.size(); ++dim) {
+    if (desc.external_workspace && desc.twidl_table[dim][dir] == nullptr) {
+      throw std::runtime_error(
```
guangyey: torch_check(!desc.external_workspace && desc.twidl_table[dim][dir], "")

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+      q.submit([&](sycl::handler& h) {
+        h.parallel_for(sycl::range<2>(fact0, fact1), ker);
+      });
+    }
+  }
+}
+
+template <typename T>
+void commit(sycl::queue& q, fft_descriptor& desc) {
+  if (desc.fft_len.size() < 1 || desc.fft_len.size() > 3) {
```
guangyey: torch_internal_assert(desc.fft_len.size() >=1 && desc.fft_len.size() <= 3, "")

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+    if (slm_size * static_cast<std::int64_t>(sizeof(T)) >
+        q.get_device().get_info<sycl::info::device::local_mem_size>()) {
+      throw std::runtime_error("Required SLM size exceeds device limits");
+    }
+    desc.slm_size[dim] = slm_size;
+    desc.twidl_table_size[dim] = fact0 * fact1 * 2;
+    desc.external_workspace_size += desc.twidl_table_size[dim] * sizeof(T) *
+        ((desc.which_dir == 2) ? 2 : 1);
+
+    if (fact0 == 0 || fact1 == 0) {
```
guangyey: ditto

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+        break;
+      }
+    }
+
+    auto fact0 = desc.facts[dim][0];
+    auto fact1 = desc.facts[dim][1];
+    auto slm_size = fact0 * fact1 * 2;
+    if (slm_size * static_cast<std::int64_t>(sizeof(T)) >
+        q.get_device().get_info<sycl::info::device::local_mem_size>()) {
+      throw std::runtime_error("Required SLM size exceeds device limits");
```
guangyey: ditto

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+          "-DOUTER_BATCH_FWD_DIST=" + std::to_string(outer_batch_fwd_dist),
+          "-DINNER_BATCH_FWD_DIST=" + std::to_string(inner_batch_fwd_dist),
+          "-DOUTER_BATCH_BWD_DIST=" + std::to_string(outer_batch_bwd_dist),
+          "-DINNER_BATCH_BWD_DIST=" + std::to_string(inner_batch_bwd_dist)};
+      if constexpr (std::is_same_v<T, float>) {
+        // no define needed - that's default
+      } else if constexpr (std::is_same_v<T, double>) {
+        fft_build_opts.push_back("-DDFT_DOUBLE_PRECISION");
+      } else {
+        throw std::runtime_error("Unsupported data type");
```
guangyey: torch_internal_assert(false, "")

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+  }
+  if (!desc.external_workspace) {
+    calculate_twiddle_factors<T>(q, desc);
+  }
+}
+
+template <typename T>
+void set_workspace(T* workspace, sycl::queue& q, fft_descriptor& desc) {
+  if (workspace == nullptr) {
+    throw std::runtime_error("Workspace pointer is null");
```
guangyey: ditto

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+}
+
+inline void fft_32(sycl::marray<DFT_FTYPE, 2> *in, sycl::marray<DFT_FTYPE, 2> *out, int is = 1, int id = 0, int os = 1, int od = 0)
+{
+    sycl::marray<DFT_FTYPE, 2> y[32];
+    FFT_FN(32, 4, 8);
+}
+
+inline void read_input(const DFT_FTYPE *in, sycl::marray<DFT_FTYPE, 2> *inReg, int group_id,
+                       int local_id, int rows, int stride_to_next_row, int batch, int idist) {
```
guangyey: batch is unused?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+        for(int i=0;i<rows;++i) {
+            const DFT_FTYPE *base = in + i * stride_to_next_row * DIST_TO_NEXT_WORK_ITEM * 2;
+            inReg[i][0] = base[offset * 2];
+            inReg[i][1] = base[offset * 2 + 1];
+        }
+    }
+}
+
+inline void write_output(DFT_FTYPE *out, sycl::marray<DFT_FTYPE, 2> *inReg, int group_id, int local_id,
+                         int rows, int stride_to_next_row, int batch, int odist, int fact1s_idx, int work_items_in_group) {
```
guangyey: batch seems unsed.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+          "-DKERNEL_NAME=" + kernel_name_dir,
+          "-DREGSIZE=" + std::to_string(num_regs),
+          "-DFACT0=" + std::to_string(fact0),
+          "-DFACT1=" + std::to_string(fact1),
+          "-DBATCH=" + std::to_string(batch),
+          "-DFWD_DIST=" + std::to_string(fwd_dist),
+          "-DBWD_DIST=" + std::to_string(bwd_dist),
+          "-DNUM_FACT1S=" + std::to_string(num_fact1s),
+          "-DDIR_VAL=" + std::to_string(dir_val),
+          "-DSCALE=" + std::string(scale_buf),
```
guangyey: is `SCALE` unused?

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+          "%a",
+          dir_val == 0   ? desc.fwd_scale
+              : dim == 0 ? desc.bwd_scale
+                         : 1.0);
+      std::vector<std::string> fft_build_opts = {
+          "-DKERNEL_NAME=" + kernel_name_dir,
+          "-DREGSIZE=" + std::to_string(num_regs),
+          "-DFACT0=" + std::to_string(fact0),
+          "-DFACT1=" + std::to_string(fact1),
+          "-DBATCH=" + std::to_string(batch),
```
guangyey: ditto.

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
```diff
@@ -9,15 +9,941 @@
...
+        inner_batch_bwd_dist = desc.bwd_strides[1] * desc.fft_len[1];
+        dist_to_next_work_item = desc.fwd_strides[1];
+        desc.global_work_size[dim][0] =
+            desc.local_work_size[dim][0] * desc.fft_len[0];
+        desc.global_work_size[dim][1] = desc.batch;
+        if (desc.fft_len.size() == 3) {
+          outer_batch_fwd_dist = desc.fwd_dist;
+          outer_batch_bwd_dist = desc.bwd_dist;
+          desc.global_work_size[dim][1] = desc.fft_len[2];
+          desc.global_work_size[dim][2] = desc.fft_len[2] * desc.batch;
```
guangyey: The agent told me that this should be `desc.batch`, I think it is reasonable but I am not sure. Could you please help me double confirm.

src/ATen/native/xpu/SpectralOps.cpp
```diff
@@ -16,8 +16,46 @@
...
+
+  auto it = bool_map.find(str);
+  if (it != bool_map.end()) {
+    return it->second;
+  }
+  return false;
+}
+
+bool use_sycl_spectral() {
+  static const bool enabled = [] {
```
guangyey: For a better ut coverage test, let us change it to `const bool` and dispatch at runtime each time. I think it should not be a bottleneck.

guangyey:
```suggestion
  const bool enabled = [] {
```

test/xpu/test_spectral_ops_xpu.py
```diff
@@ -132,12 +138,46 @@ def _test_fft_half_and_chalf_not_power_of_two(self, device, dtype, op):
...
+                )
+
+
 def _compare_xpu_cpu(self, xpu_result, cpu_result, t):
     self.assertEqual(xpu_result.device, t.device)
     self.assertEqual(xpu_result.is_complex(), cpu_result.is_complex())
     self.assertEqual(xpu_result, cpu_result, exact_dtype=False)
 
 
+TestFFT.test_fft_c2c_sycl_kernel = _test_fft_c2c_sycl_kernel
```
guangyey: This isn't a good place to put the test here. `test/xpu/test_spectral_ops_xpu.py` is a file copy from the upstream `test/test_spectral_ops.py` We should add the new test case under `test/regressions/`. We can use `@mock.patch` to disable/enable `USE_SYCL_SPECTRAL` either to cover SYCL and MKL path.
`

guangyey (commented):
PR will be accepted. I have no more suggestions.

test/xpu/test_spectral_ops_xpu.py
dbollapr: Moved the test to `test/regressions`, 12dfaa84a7d03242ece7aed764ed5e474614fad1

src/ATen/native/xpu/SpectralOps.cpp
dbollapr: Done, 5b67bc7440a71c838c1281472ef322a648706b80

dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Replaced `throw` instances with `TORCH_CHECK` and `TORCH_INTERNAL_ASSERT`, 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Removed unused batch, 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Removed, 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: 5b67bc7440a71c838c1281472ef322a648706b80

src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp
dbollapr: Updated, 5b67bc7440a71c838c1281472ef322a648706b80

guangyey (approved):
Thanks for your update!

guangyey:
@dbollapr Thanks for your patience. Let's wait for a moment to see if there are any more comments from @dvrogozh and @EikanWang

guangyey:
I will land this PR. @dvrogozh If you have any other comments. please levae it and we could handle it in a separate PR.

dvrogozh:
> I will land this PR. @dvrogozh If you have any other comments. please levae it and we could handle it in a separate PR.

I don't have any more comments for this PR. Last week was a bit crazy for me - I missed to re-review it. Thank you for taking care of it, @guangyey!

Issue #3358, opened 2026-04-16
[v.2.12.0] Release Tracker


chuanqi129:
We cut a [release branch](https://github.com/intel/torch-xpu-ops/tree/release/2.12) for the 2.12.0 release.

Our plan from this point from this point is roughly:

- Phase 1 (until 27/4/26): work on finalizing the release branch

- Phase 2 (after 27/4/26): perform extended integration/stability/performance testing based on Release Candidate builds.

This issue is for tracking cherry-picks to the release branch.

Refer https://github.com/pytorch/pytorch/issues/180506

chuanqi129:
Link to landed trunk PR (if applicable):
* NA

Link to release branch PR:
* https://github.com/intel/torch-xpu-ops/pull/3360

Criteria Category:
* Release only changes

chuanqi129:
Link to landed trunk PR (if applicable):
* https://github.com/intel/torch-xpu-ops/pull/3406

Link to release branch PR:
* https://github.com/intel/torch-xpu-ops/pull/3408

Criteria Category:
* Critical

chuanqi129:
closing this release is delivered

PR #3965, opened 2026-06-11
[Skill] Add release-branching skill for automating release branch setup

chuanqi129:
## Summary

- Adds a Claude skill (`release-branching`) that automates the torch-xpu-ops release branch setup process when PyTorch cuts a new release branch.
- Covers the full workflow: branch creation from pytorch-pinned commit, release tracker issue, workflow-defaults PR, and tracker comment.
- Templates and commands derived from the 2.11 and 2.12 release patterns.

### DIFF
--- a/.claude/skills/release-branching/SKILL.md
+++ b/.claude/skills/release-branching/SKILL.md
@@ -4,8 +4,6 @@ description: >
   Use when setting up a new torch-xpu-ops release branch corresponding to a
   PyTorch release. Covers branch creation, release tracker issue, workflow
   default changes PR, and tracker comment.
-placeholders:
-  - release_branch  # e.g. release/2.12
 ---
 
 # XPU Release Branching
@@ -82,7 +80,7 @@ Type: Task
 ```markdown
 We cut a [release branch](https://github.com/intel/torch-xpu-ops/tree/BRANCH) for the MAJOR_MINOR.0 release.
 
-Our plan from this point from this point is roughly:
+Our plan from this point is roughly:
 
 - Phase 1 (until <PHASE1_DEADLINE>): work on finalizing the release branch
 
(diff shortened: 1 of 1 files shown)


chuanqi129:
@copilot+claude-opus-4.6 review this PR with skill-writer

EikanWang (changes requested):
## Review

### Correctness

**"from this point from this point" duplication (line 89 of SKILL.md)**

The issue body template contains `Our plan from this point from this point is roughly:` — a duplicated phrase. While this matches the existing tracker issues (#3358, #2898), the skill should fix the typo for future releases rather than perpetuating it. An agent following this template will create new issues with the same error.

**`PREV_MINOR` defined but never used (line 30)**

The Terminology section defines `PREV_MINOR = 2.11 (previous minor version)` but it is never referenced anywhere in the subsequent steps. Either remove it or document where it's needed.

**`Type: Task` is not actionable (line 79)**

Step 3 lists `Type: Task` as a property to set on the issue, but `gh issue create` has no `--type` flag, and the actual tracker issues don't have a type label set. The agent will either fail or be confused. Remove this line or replace with the correct mechanism if one exists.

### Suggestions

**Missing milestone creation command** — Step 3 says "Create a milestone if it does not exist yet" but provides no command. The agent would need something like:
```bash
gh api repos/intel/torch-xpu-ops/milestones -f title="PTMAJOR_MINOR" -f state=open
```

**No error handling for pre-existing resources** — If the release branch already exists (e.g., re-running after a partial failure), Step 2's `gh api` call will fail with a 422. A brief note on idempotency would help the agent handle retries.

**Placeholder substitution ambiguity** — Bash commands use bare text like `release/MAJOR_MINOR` rather than a consistent template notation. Consider using `<MAJOR_MINOR>` (matching the date placeholders in the issue body) or `${MAJOR_MINOR}` for clarity.

EikanWang (approved):

PR #3970, opened 2026-06-11
Migrate ATen XPU kernel registerations to stock PyTorch

guangyey:
# Motivation
This PR will help fix the build when https://github.com/pytorch/pytorch/pull/181233 landed.
Will remove codegen from torch-xpu-ops in follow up PRs.


src/ATen/native/xpu/Distributions.cpp
```diff
@@ -83,7 +83,7 @@ Tensor _s_gamma_xpu(const Tensor& alpha, std::optional<Generator> gen_) {
   return ret;
 }
 
-Tensor _sample_dirichlet_xpu(
```
EikanWang, line 86: Is the change related to this PR?

src/ATen/native/xpu/RangeFactories.cpp
```diff
@@ -32,7 +32,7 @@
 namespace at {
 
 namespace native {
-Tensor& arange_out_xpu(
```
EikanWang, line 35: ditto

src/ATen/native/xpu/XPUFallback.cpp
EikanWang, line 1: @etaf , please review this file.

src/CMakeLists.txt
```diff
@@ -7,8 +7,6 @@
 # http://www.apache.org/licenses/LICENSE-2.0
 
 # Define archive static library target - torch_xpu_ops
-include(${TORCH_XPU_OPS_ROOT}/cmake/Codegen.cmake)
```
EikanWang, line 10: I suppose we should also remove `Codegen.cmake`, right?

EikanWang (approved):

src/CMakeLists.txt
guangyey: We have many code need to be removed, will follow up in a separate PR.

src/ATen/native/xpu/Distributions.cpp
guangyey: Have to change the name to align other backend's style. See https://github.com/pytorch/pytorch/pull/181233/changes#diff-2f3dbd85efb9b5172f2264eedd3be47dd765e6ab7cc8bf3ade5e62c28ae35991L6542

src/ATen/native/xpu/RangeFactories.cpp
guangyey: Have to change the name to align with other backend's style. See https://github.com/pytorch/pytorch/pull/181233/changes#diff-2f3dbd85efb9b5172f2264eedd3be47dd765e6ab7cc8bf3ade5e62c28ae35991R794-R797

src/ATen/native/xpu/XPUFallback.cpp
etaf: This change will break the `debug` functionality that we want to replace a xpu kernel with cpu kernel. But the missing xpu impl kernel can still be fallback to cpu.

The original design is append the code of `XPUFallback.template` at the end of the generated `RegisterXPU.cpp` so that the later registered cpu kernel can override the xpu kernel we want to debug. See https://github.com/intel/torch-xpu-ops/blob/407856869e0be494ae9c06b47481e91a085642e5/cmake/Codegen.cmake#L93
But since we're removing the `Codegen.cmake`, we may need to consider if we still need the debug functionality.
If not currently change is ok.

guangyey: Yes, we only preserve the fallback functionality and drop the debug override functionality. After migrating the registrations to upstream PyTorch, `RegisterXPU.cpp` is compiled directly into `libtorch_xpu.so`, so we can no longer guarantee that our fallback code is appended before `RegisterXPU.cpp` is compiled.

That said, there is still a workaround. We can disable the XPU kernel registration in `native_functions.yaml` and register it through `XPUFallback.cpp` instead. This would provide behavior similar to the previous debug mechanism, although it is less convenient than the original approach.

EikanWang: @etaf , @guangyey , at the current stage, do you think we still need the CPU-fallback feature for debugging?

etaf: We can do this in python level like
```
with ForceXPUOpFallbackToCPU({"aten::div.Tensor"}):
    z = model(x, y)
```
to force fallback a XPU op.

guangyey:
The failures caused by https://github.com/pytorch/pytorch/pull/186928 not related this PR.

src/ATen/native/xpu/XPUFallback.cpp
guangyey: I'm fine with enabling it if necessary. Currently, for this PR, we need to remove it because there is no longer a guaranteed registration order, and the debug functionality depends on that ordering.

guangyey:
Some failures were detected in this PR. I have already filed an issue (https://github.com/intel/torch-xpu-ops/issues/3994) and added a temporary dynamic skip as a workaround.

To be honest, I don't currently have the bandwidth to carefully investigate whether these failures are related to this PR, although I suspect they are not.

At this point, I think the highest priority is to land this PR and restore the torch-xpu-ops CI. Another challenge is that I'm not sure of the best way to handle or skip the distributed test failures `../../../../test/distributed/fsdp/test_fsdp_core.py::TestParityWithDDPXPU::test_mixture_of_experts_offload_true_shard_grad_op_xpu`. cc @chuanqi129

guangyey:
Land this PR to restore the CI.

PR #3979, opened 2026-06-12
Replace deprecated AT_ERROR and AT_ASSERT macros

cyyever:
Replaces all remaining uses of the deprecated `AT_ERROR` and `AT_ASSERT` macros (deprecated upstream in `c10/util/Exception.h`, see pytorch/pytorch#20287):

- `AT_ERROR(...)` -> `TORCH_CHECK(false, ...)` (4 sites)
- `AT_ASSERT(...)` -> `TORCH_INTERNAL_ASSERT(...)`, or `TORCH_CHECK` in `Copy.cpp` to match upstream `Copy.cu` (9 sites)

Each site was checked against its upstream CUDA counterpart; where upstream has already modernized (`Copy.cu`, `MaxUnpooling.cu`, `SummaryOps.cu`, mkl `SpectralOps.cpp`) the replacement matches it exactly. Where upstream still uses `AT_ASSERT` (`Reduce.cuh`, `EmbeddingBag.cu`, `ReduceOps.cpp`, `AdaptiveAveragePooling.cu`, `SparseCUDATensorMath.cu`), the documented replacement `TORCH_INTERNAL_ASSERT` is used instead of copying the deprecated macro; semantics are identical.

Intentional divergences from upstream, beyond the macro rename:
- Dropped `TORCH_INTERNAL_ASSERT(grad_input_values.is_xpu())` in `SparseTensorMathKernels.cpp`: trivially true, the tensor is created two lines above with XPU options.
- Dropped `TORCH_INTERNAL_ASSERT(input_width > 0 && output_width > 0)` in `UpSampleLinear1dKernels.cpp`: already enforced by the structured op meta check (`upsample_1d_common_check`).
- Folded `if (cond) { TORCH_CHECK(false, ...) }` into single `TORCH_CHECK(!cond, ...)` calls in `SummaryOpsKernels.cpp` and `MaxUnpoolingKernels.cpp`; messages unchanged.

Error messages and failure behavior are unchanged at every site.

Test: none (mechanical macro replacement; no behavior change)


guangyey (approved):
Thanks!

PR #3980, opened 2026-06-12
[CI] Sync benchmark suite configs

mengfei25:
## Summary

Sync benchmark suite configuration with the latest PyTorch benchmark expectations.

## To fix
```
[2026-06-11 14:26:30] xpu  eval  deit_tiny_patch16_224.fb_in1k
[2026-06-11 14:26:30] Traceback (most recent call last):
[2026-06-11 14:26:30]   File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/benchmarks/dynamo/common.py", line 4772, in run
[2026-06-11 14:26:30]     ) = runner.load_model(
[2026-06-11 14:26:30]   File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/benchmarks/dynamo/timm_models.py", line 310, in load_model
[2026-06-11 14:26:30]     model_name in self._emulate_precision_casts
[2026-06-11 14:26:30]   File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/benchmarks/dynamo/timm_models.py", line 178, in _emulate_precision_casts
[2026-06-11 14:26:30]     return self._config["emulate_precision_casts"]
[2026-06-11 14:26:30] KeyError: 'emulate_precision_casts'
```

## What changed

- Remove `CamemBert` from the HuggingFace skip list now that the related PyTorch issue is resolved
- Refresh TIMM model batch-size divisor overrides
- Add `vit_base_patch14_dinov2.lvd142m` to `emulate_precision_casts` for bf16 eager/inductor parity
- Add Detectron2 Mask R-CNN variants to the TorchBench tolerance list
- Keep model-list alignment scoped to same-name upstream lists:
  - `huggingface_models_list.txt`
  - `timm_models_list.txt`
  - `torchbench_models_list.txt`

## Motivation

Keeping these benchmark config overrides and same-name model lists aligned with upstream behavior avoids carrying stale skips, prevents unnecessary batch-size workarounds, and captures model-specific precision/tolerance settings needed for stable E2E benchmark validation.


.ci/benchmarks/timm_models.yaml
```diff
@@ -17,21 +17,10 @@ skip:
...
     gluon_xception65: 2
     mobilevit_s: 2
-    pnasnet5large: 2
-    poolformer_m36: 2
-    resnest101e: 2
     swin_base_patch4_window7_224: 2
-    swsl_resnext101_32x16d: 2
-    vit_base_patch16_224: 2
-    volo_d1_224: 2
-    jx_nest_base: 4
```
chuanqi129, line 34: Do we need those changes? Does it indicate that we will remove those models or those models batchsize will be larger than before after this change?

mengfei25: This is pytorch change, I think we just need keep the diffs what must need for us, such as high tolerance for removed models. I think we should not keep batch size diff, it only affect perf which we can use same when comparison.

mengfei25: And now, we only keep 2 diffs with pytorch.
1. keep pytorch removed models, but still run in our side.
2. restore high tolerance for the pytorch removed models

chuanqi129 (approved):

.ci/benchmarks/timm_models.yaml
chuanqi129: Fine to me. CC @bjarzemb to aware those models bs change after this PR land

PR #3982, opened 2026-06-12
Fix minor CMake issues

cyyever:
Cleanups from a CMake audit: the include-dir dedup in run_sycl.cmake referenced a misspelled variable and never ran; GNU ld's --start-group/--end-group were passed to MSVC link on Windows (ignored with warnings); FindONEMKL accepted *-NOTFOUND libraries and still reported success, deferring a partial oneMKL install to an obscure link error instead of failing at configure time. Also removes an undefined variable reference, a stray get_filename_component() argument, and a duplicated source listing, none of which affect the build output.

Authored with Claude Code.

### DIFF
--- a/cmake/Modules/FindONEMKL.cmake
+++ b/cmake/Modules/FindONEMKL.cmake
@@ -74,6 +74,10 @@ foreach(LIB_NAME IN LISTS MKL_LIB_NAMES)
     NAMES ${LIB_NAME}
     HINTS ${ONEMKL_LIB_DIR}
     NO_CMAKE_PATH NO_CMAKE_ENVIRONMENT_PATH)
+  if(NOT ${LIB_NAME}_library)
+    message(WARNING "${LIB_NAME} not found in ${ONEMKL_LIB_DIR}!!")
+    return()
+  endif()
   list(APPEND ONEMKL_LIBRARIES ${${LIB_NAME}_library})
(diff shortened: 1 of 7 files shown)


cmake/Modules/FindONEMKL.cmake
```diff
@@ -74,6 +74,10 @@ foreach(LIB_NAME IN LISTS MKL_LIB_NAMES)
     NAMES ${LIB_NAME}
     HINTS ${ONEMKL_LIB_DIR}
     NO_CMAKE_PATH NO_CMAKE_ENVIRONMENT_PATH)
+  if(NOT ${LIB_NAME}_library)
+    message(WARNING "${LIB_NAME} not found in ${ONEMKL_LIB_DIR}!!")
+    return()
```
guangyey, line 79: Good catch!
Could you help update https://github.com/intel/torch-xpu-ops/blob/8ca427ab514e0e0861a793e927865fcb5dfdd5f1/cmake/ONEMKL.cmake#L22-L24 to something like 
"oneMKL not found or installation is incomplete; see warnings above for details."

guangyey (approved):

guangyey (approved):
Thanks for your update!

guangyey:
Cmake change only, low risk to land it.

PR #3983, opened 2026-06-12
[CI] Enhance the auto merge workflow on ci check and permission scope

chuanqi129:
## Summary

Three targeted fixes to the auto merge workflow:

### 1. CI check deduplication — use latest trigger only

When a workflow is re-triggered on the same commit (e.g., manual re-run), `checks.listForRef` returns check runs from **all** runs, including stale ones. Previously, an old failed run would block the merge even after a successful re-run.

Fix: deduplicate check runs by name, keeping only the one with the latest `started_at` timestamp.

### 2. Minimize MERGE_TOKEN scope — use default token for comments

Previously the "Merge PR" step used `MERGE_TOKEN` for both the merge API call and posting result comments. Now:
- **Merge step**: uses `secrets.MERGE_TOKEN` (only for `pulls.merge`)
- **Report step**: uses default `GITHUB_TOKEN` (for posting the result comment)

This follows least-privilege: the PAT's elevated scope is never used for comment posting. `contents: write` was also removed from the workflow `permissions` block since it was only needed when the default token served as merge fallback.

### 3. Fix script injection in Report Merge Result step

The Report step previously interpolated `steps.merge.outputs.message` directly into a JavaScript template literal via `${{ }}`. If the API error message contained backticks or `${}` patterns, the resulting JavaScript would break or execute unintended expressions.

Fix: pass outputs through environment variables (`env` + `process.env`), which is the standard GitHub Actions mitigation for untrusted interpolation.

## Test Plan

Validated YAML syntax. Functional testing requires:
- A PR where CI was re-triggered (to verify dedup picks the latest run)
- A `/merge` comment (to verify the split merge/comment steps work)

### DIFF
--- a/.github/workflows/auto_merge_by_comment.yml
+++ b/.github/workflows/auto_merge_by_comment.yml
@@ -6,7 +6,6 @@ on:
 
 permissions:
   checks: read
-  contents: write
   issues: write
   pull-requests: write
   statuses: read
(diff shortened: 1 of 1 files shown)


chuanqi129:

CuiYifeng (approved):

mengfei25 (approved):

chuanqi129:

PR #3984, opened 2026-06-12
Remove suppression of -Werror

orrangetabby17:
This PR fixes an issue where the -Werror flag could be unintentionally suppressed when building with WERROR=1.
For example, a simplified `CMAKE_HOST_FLAGS` might look like:`-Werror;-Wno-error;-Werror`
After applying `REMOVE_DUPLICATES` in [run_sycl.cmake](https://github.com/intel/torch-xpu-ops/blob/24fa8b1ea6025b2f268790e09eda5b79b85119aa/cmake/Modules/FindSYCL/run_sycl.cmake#L73), the flags become:`-Werror;-Wno-error`
As a result, the final `-Werror` is supressed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,6 @@ cmake_minimum_required(VERSION 3.27 FATAL_ERROR)
 
 # Avoid SYCL compiler error
 if(NOT WIN32)
-  string(APPEND CMAKE_CXX_FLAGS " -Wno-error")
   if("$ENV{IS_XPU_CI}" STREQUAL "1")
     string(APPEND CMAKE_CXX_FLAGS " -Werror=unused-variable")
   endif()


orrangetabby17:
@copilot /pr-review

guangyey (approved):

guangyey:
I understand that CI would capture the warning as an error after this PR landed no matter whether `WERROR=1` was added, right?

orrangetabby17:
> I understand that CI would capture the warning as an error after this PR landed no matter whether `WERROR=1` was added, right?

When `WERROR=1` was not added, 
`CMAKE_HOST_FLAGS` might look like: `-Wno-error;-Werror`
`REMOVE_DUPLICATES` wont influence `-Werror`, and the `-Werror` wont be suppressed.

In FLAGS like `-Werror;-Wno-error;-Werror`, the first flag `-Werror` depends on var`WERROR=1`
The second `-Werror` is introduced in 
https://github.com/intel/torch-xpu-ops/blob/8ca427ab514e0e0861a793e927865fcb5dfdd5f1/cmake/BuildFlags.cmake#L56

This PR ensures that warnings continue to be treated as errors, which is the expected behavior in torch-xpu-ops, regardless of whether WERROR=1 is enabled.

chunhuanMeng (approved):

Issue #2200, opened 2025-10-21
support flash attention op on XPU device


Zjq9409:
### 🚀 The feature, motivation and pitch

Running the same Attention model on B60 and 4090D, it was found that the flash attention operator on XPU was not implemented and the SDPA operator was used. B60 is much slower than 4090D.
4090D:
```
                                                   Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg     Self CUDA   Self CUDA %    CUDA total  CUDA time avg       CPU Mem  Self CPU Mem    # of Calls  
void pytorch_flash::flash_fwd_kernel<Flash_fwd_kerne...         0.00%       0.000us         0.00%       0.000us       0.000us      36.261ms        62.36%      36.261ms     292.430us           0 B           0 B           124  
```
B60:
```
                                                   Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg      Self XPU    Self XPU %     XPU total  XPU time avg    # of Calls  
aten::_scaled_dot_product_fused_attention_overrideab...         0.82%       6.596ms        11.46%      92.334ms     699.498us     175.266ms        76.53%     175.384ms       1.329ms           132  
```
The compile mode used has no CPU overhead on 4090D, but there is still CPU overhead on B60.

### Additional context

Minimize reproducing code
```
# Lets define a helpful benchmarking function:
import torch
import torch.nn as nn
import torch.nn.functional as F
device = "cuda"   # or "xpu"
import torch.utils.benchmark as benchmark

def benchmark_torch_function_in_microseconds(f, *args, **kwargs):
    t0 = benchmark.Timer(
        stmt="f(*args, **kwargs)", globals={"args": args, "kwargs": kwargs, "f": f}
    )
    return t0.blocked_autorange().mean * 1e6

# Lets define the hyper-parameters of our input
batch_size = 32
max_sequence_len = 1024
num_heads = 32
embed_dimension = 32

dtype = torch.float16
query = torch.rand(batch_size, num_heads, max_sequence_len, embed_dimension, device=device, dtype=dtype)
key = torch.rand(batch_size, num_heads, max_sequence_len, embed_dimension, device=device, dtype=dtype)
value = torch.rand(batch_size, num_heads, max_sequence_len, embed_dimension, device=device, dtype=dtype)

print(f"The default implementation runs in {benchmark_torch_function_in_microseconds(F.scaled_dot_product_attention, query, key, value):.3f} microseconds")

[...]

PR #3986, opened 2026-06-12
Skip meta dispatch flash_attention_forward OpInfo on XPU

### LINKED ISSUE #3848: _flash_attention_forward not implemented for XPU — meta test failures
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_meta_xpu.TestMetaXPU,test_dispatch_symbolic_meta_outplace_torch_ops_aten__flash_attention_forward_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_meta_xpu.TestMetaXPU,test_dispatch_meta_outplace_torch_ops_aten__flash_attention_forward_xpu_float16

## Root Cause

The `aten::_flash_attention_forward` operator is not registered for the XPU device. When the meta dispatch tests try to execute this op on an XPU tensor, it fails with a "not implemented" error, which gets wrapped as `Exception: Original OpInfo is broken`.

The underlying error is:
```
The operator 'aten::_flash_attention_forward' is not currently implemented for the XPU device.
```

This is separate from the flex_attention export tests (Group 1), which fail due to graph structure mismatch. Here, the op itself cannot run on XPU.

**Fix:** Either:
1. Implement `_flash_attention_forward` for XPU (substantial effort).
2. Add these test patterns to the skip list in `skip_list_common.py` (the `_flash_attention_` pattern already skips flash attention in `test_ops_xpu.py` but not in `test_meta_xpu.py`).

## Reproducer

```bash
source ~/.bashrc && cd third_party/torch-xpu-ops/test/xpu && python -m pytest -sxv test_meta_xpu.py -k "test_dispatch_symbolic_meta_outplace_torch_ops_aten__flash_attention_forward_xpu_float16"
```

```bash
source ~/.bashrc && cd third_party/torch-xpu-ops/test/xpu && python -m pytest -sxv test_meta_xpu.py -k "test_dispatch_meta_outplace_torch_ops_aten__flash_attention_forward_xpu_float16"
```

## Representative Error

```
Exception: Original OpInfo is broken

Caused by sample input at index 0: SampleInput(input=Tensor[size=(4, 11, 4, 8),
device="xpu:0", dtype=torch.float16, contiguous=False], ...)

The operator 'aten::_flash_attention_forward' is not currently implemented for
the XPU device. Please open a feature on https://github.com/intel/torch-xpu-ops/issues.
You can set the environment variable `PYTORCH_ENABLE_XPU_FALLBACK=1` to use the
CPU implementation as a fallback for XPU unimplemented operators.
```

pponikox:
This patch fixes #3848 by adding a targeted XPU test-side skip for flash attention meta dispatch OpInfo cases that currently fail because aten::_flash_attention_forward is not implemented on XPU.


wpietka (approved):

PatrykWilczewski (approved):
Is this temporary fix? Will be _flash_attention_forward implemented in the future?

pponikox:
> Is this temporary fix? Will be _flash_attention_forward implemented in the future?

Yes, this is only temporary fix because `_flash_attention_forward` is under development process.
For more info see: https://github.com/pytorch/pytorch/pull/181559, https://github.com/intel/torch-xpu-ops/issues/2200, or https://github.com/intel/torch-xpu-ops/issues/2853 😉

BBBela (commented):
Is my understanding correct that this PR (https://github.com/pytorch/pytorch/pull/181559) would fix the failing test cases? Or it is more complex and the https://github.com/pytorch/pytorch/pull/181559 is just one of more changes to enable `flash_attention_forward`?

Because if https://github.com/pytorch/pytorch/pull/181559 fixes the test cases, then I think we should simply wait for it to be merged instead of adding this temporary test fix that will need to be reverted once https://github.com/pytorch/pytorch/pull/181559 is merged.

pponikox:
> Is my understanding correct that this PR ([pytorch/pytorch#181559](https://github.com/pytorch/pytorch/pull/181559)) would fix the failing test cases? Or it is more complex and the [pytorch/pytorch#181559](https://github.com/pytorch/pytorch/pull/181559) is just one of more changes to enable `flash_attention_forward`?

I'm not sure in 100% that it will fix all failing test cases because I didn't analyzed it too much. I'll check it out.

> Because if https://github.com/pytorch/pytorch/pull/181559 fixes the test cases, then I think we should simply wait for it to be merged instead of adding this temporary test fix that will need to be reverted once https://github.com/pytorch/pytorch/pull/181559 is merged.

Agree. For context: I added this patch because the mentioned https://github.com/pytorch/pytorch/pull/181559 was added long time ago and wasn't sure how long will it take to merge it.
However now I can see that recently it has some movement so probably you're right 😉

pponikox:
I checked the pytorch/pytorch#181559 and it seems that it should address the root cause for the exact skip reason in this PR (meta dispatch on XPU for `_flash_attention_forward`, `float16`). pytorch/pytorch#181559 adds XPU support for `_flash_attention_forward` and `_flash_attention_backward`, and also updates the meta-side handling for XPU in the flash attention path.
That ultimately means that this patch is a temporary fix until pytorch/pytorch#181559 is merged.

Having that said, should I close this PR without merging + add a comment in parent issue?
Or leave both this PR and parent issue as is and return to it when pytorch/pytorch#181559 is merged?
@BBBela what do you think?

BBBela:
> Having that said, should I close this PR without merging + add a comment in parent issue?
> Or leave both this PR and parent issue as is and return to it when [pytorch/pytorch#181559](https://github.com/pytorch/pytorch/pull/181559) is merged?

I think for now we can leave both this PR and parent issue as is and wait for [pytorch/pytorch#181559](https://github.com/pytorch/pytorch/pull/181559). Once the upstream PR is merged and failing test cases are passing we can close both this PR and the issue.

pponikox:
While waiting for the https://github.com/pytorch/pytorch/pull/181559 (still nor merged), a new changes was introduced: https://github.com/intel/torch-xpu-ops/pull/4278, that completely removed the tests and replaced them with a wrapper importing upstream using `XPUImportCtx`, which ultimately makes my changes no longer relevant.

Given the above, and the fact that this patch would most likely be unnecessary after the  https://github.com/pytorch/pytorch/pull/181559 is merged anyway, I decided that if the root cause still occurs, it would be better to address it from scratch rather than deal with merge conflicts and try to adjust the new approach.

Therefore, I'm closing this task.

Issue #3030, opened 2026-03-11
[Bug Skip] test/test_modules.py::TestModuleXPU::test_cpu_gpu_parity_nn_ConvTranspose2d_xpu_complex32 failed with


shangerxin:
### Description

The UT failed on torch public CI.

FAILED CONSISTENTLY: test/test_modules.py::TestModuleXPU::test_cpu_gpu_parity_nn_ConvTranspose2d_xpu_complex32
Stopping at first consistent failure
The following tests failed consistently: ['test/test_modules.py::TestModuleXPU::test_cpu_gpu_parity_nn_ConvTranspose2d_xpu_complex32']

### Root Cause
Certain uses of torch.narrow() are not supported by the backend compiler for XPU in PyTorch's Inductor.

### Error Message

```
==================================== RERUNS ====================================
______ TestModuleXPU.test_cpu_gpu_parity_nn_ConvTranspose2d_xpu_complex32 ______
Traceback (most recent call last):
  File "/var/lib/jenkins/pytorch/test/test_modules.py", line 628, in test_cpu_gpu_parity
    check_backward(cpu_outputs, gpu_outputs)
  File "/var/lib/jenkins/pytorch/test/test_modules.py", line 617, in check_backward
    self.assertEqual(cpu_grad_input, gpu_grad_input)
  File "/opt/conda/envs/py_3.10/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 4363, in assertEqual
    raise error_metas.pop()[0].to_error(  # type: ignore[index]
AssertionError: Tensor-likes are not close!

Mismatched elements: 1 / 96 (1.0%)
Greatest absolute difference: 0.0001220703125 at index (1, 2, 1, 1) (up to 1e-05 allowed)
Greatest relative difference: 0.005130767822265625 at index (1, 2, 1, 1) (up to 0.001 allowed)

The failure occurred for item [0]

To execute this test, run the following from the base repo dir:
    python test/test_modules.py TestModuleXPU.test_cpu_gpu_parity_nn_ConvTranspose2d_xpu_complex32

This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0
----------------------------- Captured stderr call -----------------------------
[...]

PR #3991, opened 2026-06-12
Add complex32 ConvTranspose2d reference regression for XPU

gplutop7:
Fixes:
https://github.com/intel/torch-xpu-ops/issues/3030

Adds a focused regression test for ConvTranspose2d complex32 operator on XPU.

The upstream parity test (test_cpu_gpu_parity_nn_ConvTranspose2d_xpu_complex32) 
is marked expectedFailure due to fundamental backend path differences:
- CPU uses slow_conv_transpose2d
- XPU uses oneDNN deconv operator

These distinct implementations yield divergent complex32 results beyond strict 
parity tolerances.

This regression validates XPU quality using an alternative metric: comparing 
XPU(complex32) output against CPU(complex64) reference.


tszulist-hbn (approved):

test/regressions/test_conv_transpose_complex32.py
BBBela, line 1: I would suggest to make the file name shorter and more general.

BBBela (commented):
Generally it looks good to me. Just one naming comment below 😉 

Would it be possible to extend those tests for `1d` and `3d` versions also? It looks like there is the same underlying problem. The `2d` and `3d` versions are already skipped upstream and for the `1d` skip there is a PR waiting. So maybe we could generalize this regression test for all the dimensions? What do you think?

test/regressions/test_conv_transpose_complex32.py
gplutop7: Done

gplutop7:
> Generally it looks good to me. Just one naming comment below 😉
> 
> Would it be possible to extend those tests for `1d` and `3d` versions also? It looks like there is the same underlying problem. The `2d` and `3d` versions are already skipped upstream and for the `1d` skip there is a PR waiting. So maybe we could generalize this regression test for all the dimensions? What do you think?

I’d prefer to keep this PR scoped to the current 2d regression, since it addresses issue 3030 specifically and there is no mention of 1d/3d there. I’ll open a separate issue for extending the coverage to other ConvTranspose dimensions.

pbielak (approved):

gplutop7:
Added label: disable_all - unrelated unit tests are failing.`

PR #3995, opened 2026-06-13
[TEST ONLY] distributed UT test on bmg

zxd1997066:


zxd1997066:
using https://github.com/intel/torch-xpu-ops/pull/5147

PR #4033, opened 2026-06-17
ci: use MERGE_TOKEN for all write operations in auto merge workflow

chuanqi129:
## Summary

Follow-up to #4031. The org restricts GITHUB_TOKEN from **all** write operations, not just reactions. Every step that calls `issues.createComment` also fails with `403: Resource not accessible by integration`.

## Root Cause

Same org-level restriction as #4031, but affecting comment posting (not just reactions). The "Check Reviews" step failed when trying to post a "not enough approvals" comment ([log](https://github.com/intel/torch-xpu-ops/actions/runs/27671120314/job/81835475454#step:4:45)).

## Fix

- Use `secrets.MERGE_TOKEN` for all steps that perform write operations (reactions + comments)
- Downgrade GITHUB_TOKEN permissions to read-only since it is no longer used for writes

## Test Plan

Trigger `/merge` on a PR and verify all steps (reaction, review check comments, CI status comments, merge result comments) succeed.

### DIFF
--- a/.github/workflows/auto_merge_by_comment.yml
+++ b/.github/workflows/auto_merge_by_comment.yml
@@ -12,8 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       checks: read
-      issues: write         # reactions + status comments via GITHUB_TOKEN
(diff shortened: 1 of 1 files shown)


chuanqi129:
Superseded — rebasing fix onto latest main.

PR #4058, opened 2026-06-18
Replace local min/max pool helpers with sycl::min/sycl::max

cyyever:
AveragePool2dKernels, AveragePool3dKernels, and ReplicationPaddingKernels each defined file-local `int` `min`/`max` (`imin`/`imax`) helpers. Replace them with `sycl::min`/`sycl::max`, the SYCL device convention, matching the cleanup direction of recent commits that removed local arithmetic-helper clones (#3943, #3944).

`sycl::min`/`sycl::max` don't coerce arguments, so operands are explicitly cast to a common type — the destination variable's type (`int` / `index_t` / `int64_t`). This matches the old `int` helpers' results for all real inputs.

### DIFF
--- a/src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
+++ b/src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
@@ -27,14 +27,6 @@
 namespace at::native {
 namespace xpu {
 
-inline int min(int a, int b) {
-  return a <= b ? a : b;
-}
-
-inline int max(int a, int b) {
-  return a >= b ? a : b;
-}
-
 template <typename scalar_t, typename accscalar_t, typename index_t>
 struct AvgPool2dKernelFunctor {
   void operator()(sycl::nd_item<1> item) const {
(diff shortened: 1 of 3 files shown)


pbielak:
@cyyever Why did you choose to use `sycl::min` / `sycl::max` instead of `std::min` / `std::max`? A quick grep over the repo shows that the `std::` variant is used in other places. 

If we decide to go with the `sycl::` variants, we should replace them in the entire repo to keep it consistent.

EikanWang (approved):

EikanWang:
LGTM

EikanWang:
@pbielak , we always prefer to use sycl native functions rather than std.

pbielak:
@EikanWang then this PR should also replace all other `std::min, std::max` calls with `sycl::` variants to keep things consistent across the repository

cyyever:
@pbielak I agree to change more std::{min,max} callers, but that will be a bigger PR. Meanwhile this PR focus on removing the helpers.

guangyey (commented):
Hi @cyyever why use `std::<one>` instead of `syc::<one>`, we recommend to use `sycl::<one>` in sycl kernel.

cyyever:
@guangyey Fixed

src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
```diff
@@ -353,13 +349,15 @@ struct AvgPool2dChannelsLastBackwardKernelFunctor {
           // figure out the pooling size
           int hstart = ph * stride_h_ - pad_h_;
           int wstart = pw * stride_w_ - pad_w_;
-          int hend = min(hstart + kernel_h_, height_ + pad_h_);
-          int wend = min(wstart + kernel_w_, width_ + pad_w_);
+          int hend =
+              sycl::min(hstart + kernel_h_, static_cast<int>(height_ + pad_h_));
```
guangyey, line 353: nit, narrowing cast.

src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
```diff
@@ -458,13 +456,15 @@ struct AvgPool2dBackwarKernelFunctor {
...
```
guangyey, line 460: narrowing cast

src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
```diff
@@ -458,13 +456,15 @@ struct AvgPool2dBackwarKernelFunctor {
...
+          int wend =
+              sycl::min(wstart + kernel_w_, static_cast<int>(width_ + pad_w_));
```
guangyey, line 462: ditto

guangyey (approved):

PR #4075, opened 2026-06-18
More CMake cleanups

cyyever:
Configure-time only cleanups:

- **FindSYCLToolkit**: the run-failure branch tested `test_result`, which was never set (dead code); now checks the actual run result and prints a diagnostic like the compile path. Drops the unused `test_result` export.
- **FindSYCL**: remove the no-op `CMAKE_CURRENT_LIST_DIR` recompute (a pre-2.8.3 FindCUDA leftover; CMake keeps it in sync with `CMAKE_CURRENT_LIST_FILE`).
- **FindXCCL**: check `XCCL_INCLUDE_ONEAPI_DIR` explicitly and append it only after the not-found gate, instead of relying on a `-NOTFOUND`-terminated list being false.
- **BuildFlags**: MSVC was getting GNU-spelled per-config flags (`-g -O0`); each compiler now gets native flags (`/Od /Z7`, `/O2 /Z7` for MSVC).
- **src/CMakeLists.txt**: dedup the `USE_ONEMKL_XPU` block that was copy-pasted in BuildOnLinux/BuildOnWindows.


guangyey (approved):

cmake/BuildFlags.cmake
```diff
@@ -41,14 +41,22 @@ macro(set_build_flags)
...
       list(APPEND SYCL_HOST_FLAGS /std:${CPP_STD})
       list(APPEND SYCL_HOST_FLAGS /MD)
       list(APPEND SYCL_HOST_FLAGS /EHsc) # exception handling
       # SYCL headers warnings
       list(APPEND SYCL_HOST_FLAGS /wd4996) # allow usage of deprecated functions
       list(APPEND SYCL_HOST_FLAGS /wd4018) # allow signed and unsigned comparison
+
+      # -fsycl-host-compiler-options never sees CMAKE_CXX_FLAGS_<CONFIG>, so
+      # spell per-config flags here. /Z7 (not /Zi) avoids parallel PDB contention.
+      if(CMAKE_BUILD_TYPE MATCHES Debug)
```
guangyey, line 55: @chunhuanMeng could you please help review this PR.

chunhuanMeng (approved):

PR #4087, opened 2026-06-19
Use test_higher_order_ops.py directly from upstream.

BBBela:
The upstream `dynamo/test_higher_orded_ops.py` has been adjusted to be used with XPU.
There is no more need to keep the local copy of this test file as we can use the upstream one directly.

## Changes

- This PR changes torch-xpu-ops CI test launch to use `dynamo/test_higher_orded_ops.py` directly from upstream instead of the local copy.
- The local copy of `dynamo/test_higher_orded_ops.py` is no longer needed so is removed entirely.

## Comparison

I double checked if this PR doesn't loose any test cases or bring any regressions. Here is some short summary:

| state | previously | current |
| - | - | - |
| TOTAL | 234 | 238 | 
| PASSED | 218 | 222 | 
| FAILED | 9 | 0 | 
| SKIPPED | 6 | 15 | 
| XFAIL | 1 | 1 | 

All the previously failing tests are passing with the upstream version - they were caused by changes made to upstream test functions that were not ported to local tests. So those were false negatives.

Some tests were added in the upstream, while they were not enabled in torch-xpu-ops. That is why the total number of tests increased by 4 test cases - all passing on XPU.

Test cases that were previously skipped locally are also skipped in the upstream.

Some test cases were passing in the torch-xpu-ops while they are skipped temporarily on Linux due to some PyTorch problem described in the provided issue.

## Issues with skipped label

As after changes from this PR, there are no more failing test cases from `test_higher_order_ops.py`, there are no issues that needs adjustment.


test/xpu/skip_list_common.py
```diff
@@ -323,7 +325,7 @@
     "dynamo/test_compiler_bisector_xpu.py": None,
     "dynamo/test_deviceguard_xpu.py": None,
     "dynamo/test_functions_xpu.py": None,
-    "dynamo/test_higher_order_ops_xpu.py": None,
+    f"{PYTORCH_TEST_DIR}/dynamo/test_higher_order_ops.py": None,
```
Silv3S, line 333: Great idea 👍There will be less false-positives related to test misalignment.

One thing I'd consider is dropping entire `test_higher_order_ops.py`. Maybe we could leave this file as wrapper for upstream tests, so if there will be any new test that requires XPU-specific version, we will use it for test override

Silv3S (approved):

test/xpu/skip_list_common.py
BBBela: > One thing I'd consider is dropping entire test_higher_order_ops.py.

That was my initial idea - I thought about leaving the original test file and just making it a wrapper to call the upstream test file. Although @daisyden recommended to use the upstream file directly, so I followed this approach.

I totally understand your point that it is easier to make potential changes locally, but on the other hand, this way we have one source of truth without any duplication. And those local changes will be ported to upstream in the end anyway. I suppose that this is part of the whole process of merging torch-xpu-ops with PyTorch.

Silv3S: Sure

test/xpu/skip_list_common.py
```diff
@@ -6,6 +6,8 @@
 #
 # http://www.apache.org/licenses/LICENSE-2.0
 
+PYTORCH_TEST_DIR = "../../../../test"
```
tsocha, line 9: This does not look like a good solution.
You can clone `torch-xpu-ops` just to run tests with pytorch from pypi.
In such case the path will be incorrect.

You can experiment with `__file__` attribute but it's kinda hacky.

tsocha (changes requested):

test/xpu/skip_list_common.py
BBBela: The whole test module from `torch-xpu-ops` and the whole `torch-xpu-ops` CI is already based on this approach. It already heavily depends on the relative path to PyTorch, so this PR doesn't introduce anything new.
This is a design assumption that `torch-xpu-ops` is always used as a submodule of PyTorch, not standalone.

I don't say it is great, but that is commonly used and adopted assumption in this project.

tsocha (approved):

BBBela:
Hello @daisyden,
I would like to know your opinion and your feedback regarding this PR. Please review when you have time. Thanks! 😉

daisyden (approved):
LGTM

test/xpu/skip_list_common.py
```diff
@@ -328,7 +330,7 @@
     "dynamo/test_compiler_bisector_xpu.py": None,
     "dynamo/test_deviceguard_xpu.py": None,
     "dynamo/test_functions_xpu.py": None,
-    "dynamo/test_higher_order_ops_xpu.py": None,
+    f"{PYTORCH_TEST_DIR}/dynamo/test_higher_order_ops.py": None,
```
pbielak, line 333: Please check also other skip list files, if they need to be modified too

BBBela: That is a good point, thanks for this remainder!
I checked and this is the only skip list file that uses the `test_higher_order_ops_xpu.py`, so nothing else needs to be adjusted.

pbielak (approved):

Issue #4020, opened 2026-06-16
[release/2.14] Distribution sampling/moment statistical assertion failures on XPU


AKloniecki:
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_cauchy_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_normal_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_logisticnormal_sample_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_multivariate_normal_moments_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_uniform_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_generalized_pareto_sample_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_studentT_sample_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_wishart_sample_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_exponential_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_halfcauchy_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_chi2_sample_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_distributions_xpu.TestDistributionsXPU,test_wishart_moments_xpu

## Root Cause

Distribution sampling and moment-verification tests fail with `AssertionError` on XPU. The failures fall into two patterns:

1. **"Tensor-likes are not close"** — gradient or icdf/quantile comparisons show large absolute/relative differences (e.g., `scale.grad` mismatch of 64.5 in `test_cauchy_xpu`, moment mismatches in `test_wishart_moments_xpu`). All 25/25 elements mismatched in several tests, indicating a systematic numerical divergence rather than occasional precision loss.

[...]

PR #4089, opened 2026-06-19
RNG state patching in distribution tests

gplutop7:
Fixed: 
https://github.com/intel/torch-xpu-ops/issues/4020

Patch `torch.get_rng_state()` and `set_rng_state()` to redirect to XPU equivalents (`torch.xpu.get/set_rng_state()`) in the test harness. This ensures replay-based tests work correctly on XPU without modifying upstream test code.

Tests Fixed (RNG Replay)
- `test_cauchy_xpu`
- `test_exponential_xpu`
- `test_halfcauchy_xpu`
- `test_normal_xpu`
- `test_uniform_xpu`

### DIFF
--- a/test/xpu/test_distributions_xpu.py
+++ b/test/xpu/test_distributions_xpu.py
@@ -13,6 +13,8 @@
 # Owner(s): ["module: intel"]
 
 
+from unittest import mock
+
 from torch.testing._internal.common_device_type import instantiate_device_type_tests
 from torch.testing._internal.common_utils import run_tests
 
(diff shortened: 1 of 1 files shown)


test/xpu/test_distributions_xpu.py
```diff
@@ -79,6 +81,22 @@
...
 from torch.testing._internal.common_utils import set_rng_seed
 
+_orig_test_distributions_set_up = TestDistributions.setUp
+
+
+def _xpu_test_distributions_set_up(self):
+    _orig_test_distributions_set_up(self)
+    patcher = mock.patch.multiple(
+        "torch",
+        get_rng_state=lambda: torch.xpu.get_rng_state(),
```
guangyey, line 91: Where cuda calls `get_rng_state` and `set_rng_state`?

gplutop7: CUDA uses torch.cuda.get_rng_state() / torch.cuda.set_rng_state(). The patch redirects the CPU-only APIs to XPU equivalents so tests work correctly.

BBBela (dismissed):
Currently in the upstream those tests are not instantiated for cuda and xpu, so they are not even launched:
[CODE FRAGMENT](https://github.com/pytorch/pytorch/blob/c004e2dc8082e016df1893e3f42a831ea3f12c75/test/distributions/test_distributions.py#L7332-L7340)

That is not only xpu-specific problem, but both cuda and xpu. There is a need for some generic solution that effectively calls `torch.get_rng_state()` for CPU device and `torch.<device>.get_rng_state()` for accelerators.

MPS simply skips them as expected failure. I suppose that fix for CUDA will also fix this issue for XPU.

PR #4096, opened 2026-06-22
[CI] Remove CamemBert from huggingface benchmark model list

wdziurdz:
## Summary

`CamemBert` fails at model load (`eager_fail_to_run`) in **every** E2E accuracy config. The benchmark harness resolves a model-list entry to a class via `getattr(transformers, "<name>")`, and the currently pinned `transformers==5.10.1` exports no bare `CamemBert` symbol (only `CamembertModel` / `CamembertForMaskedLM` / etc.), so the lookup raises:

```
AttributeError: module transformers has no attribute CamemBert
```

This happens before any compilation runs, so it is purely a stale model-list entry — not a model or backend issue.

## Why remove (not rename)

Upstream PyTorch already **removed** `CamemBert` from `benchmarks/dynamo/huggingface_models_list.txt` in [pytorch/pytorch#164815](https://github.com/pytorch/pytorch/pull/164815) ("[Benchmark] cleanup huggingface models") and did **not** move it into `EXTRA_MODELS`, so the model is no longer benchmarked upstream. The entry survives only in this torch-xpu-ops overlay list, where it now produces a guaranteed eager failure on every run. Removing it matches upstream and clears the spurious `eager_fail_to_run`.

## Change

Delete the single dead `CamemBert,32` line from `.ci/benchmarks/huggingface_models_list.txt`.


Silv3S (approved):

pbielak (approved):

wdziurdz:
@Silv3S @pbielak These failures don't seem related to the changes. Could we re-run the tests, or is there another way to merge this?

Silv3S:
failures are clearly not introduced by this patch, so we could just force merge it. Let's wait for review from QA side before the merge

pbielak:
@chuanqi129 @daisyden @mengfei25 please review

chuanqi129 (approved):

Issue #3595, opened 2026-05-10
[release/2.13] test_get_device_properties_tensor_device missing XPU attribute


daisyden:
## Bug Description

`test_get_device_properties_tensor_device` fails because `torch._C._XpuDeviceProperties` does not have `max_threads_per_multi_processor` attribute. The test was ported from CUDA where `CudaDeviceProperties` has this attribute.

## Affected Tests

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.dynamo.test_functions_xpu.FunctionTests,test_get_device_properties_tensor_device

## Error Message

```
AttributeError: 'torch._C._XpuDeviceProperties' object has no attribute 'max_threads_per_multi_processor'
```

## Context

Surfaced by PR #3383 (CI run 25534217826) which ported `test_functions_xpu.py`.

bjarzemb:
I have this issue also on Windows with Intel XPU (Arc B580) in release 2.13 tests.

bjarzemb:
I have this issue also on LNL (windows) in release 2.13 tests.

DamJanusz:
Prepared test fix in upstream https://github.com/pytorch/pytorch/pull/190323

PR #4122, opened 2026-06-23
Use `test_functions.py` directly from upstream.

BBBela:
This is the next step after initial one from https://github.com/intel/torch-xpu-ops/pull/4087

The upstream `dynamo/test_functions.py` has been adjusted to be used with XPU. There is no more need to keep the local copy of this test file as we can use the upstream one directly.

## Changes

- This PR changes torch-xpu-ops CI test launch to use `dynamo/test_functions.py` directly from upstream instead of the local copy. - The local copy of `dynamo/test_functions.py` is no longer needed so is removed entirely.

## Comparison

I double checked if this PR doesn't loose any test cases or bring any
regressions. Here is some short summary:

| state | previously | current |
| - | - | - |
| TOTAL | 499 | 523 | 
| PASSED | 488 | 512 | 
| FAILED | 2 | 0 | 
| SKIPPED | 7 | 9 | 
| XFAIL | 2 | 2 | 

All test cases that were so far passing are passing also with upstream version.

One previously skipped test case is now passing with upstream version.

From two previously failing test cases, one is passing and second one is skipped with the upstream version.

The additional test cases (the difference between local number of tests and upstream number of tests) are new test cases that were not launched in torch-xpu-ops CI and are skipped or passing with the upstream version.

## Affected already existing issues

There are two issues that would be affected by this PR:
1. https://github.com/intel/torch-xpu-ops/issues/3842 - would be fixed by this PR (test misalignment)
2. https://github.com/intel/torch-xpu-ops/issues/3595 - there is already a PR prepared that would have to be ported to PyTorch directly. I considered this case, but we will have to make those changes directly in the upstream now or later. Supposed this issue would be fixed in torch-xpu-ops, it still needs to be ported to PyTorch directly eventually. So both ways there are upstream changes needed, the case is only the timing.


Silv3S (approved):

pbielak (approved):

BBBela:
Hello @daisyden,
please look at this PR - it is the same approach as for the `test_higher_order_ops_xpu.py` in the previous PR (https://github.com/intel/torch-xpu-ops/pull/4087). Thanks! 😄

daisyden:
Hi, BBBela, could you help to compare the test cases number of the file before and after this PR?

BBBela:
> Hi, BBBela, could you help to compare the test cases number of the file before and after this PR?

Hi @daisyden,
I already put the comparison and numbers in the PR description.

PR #4132, opened 2026-06-24
Fuse GroupNorm forward into a single kernel

jianyizh:
## Summary

Replace the 3-kernel GroupNorm forward path (RowwiseMoments + ComputeFusedParams + TensorIterator) with a single fused kernel. Avoids reading X twice from global memory.

- Small DS path: sub-group-per-group Welford, no SLM/barriers
- Large DS path: workgroup batched Welford with vec4 tree reduction, optional SLM coefficients

Bandwidth on Arc B580: 386-515 GB/s for large DS (vs 24-193 baseline), 51-297 GB/s for small DS (vs 52-146 baseline).

### DIFF
--- a/src/ATen/native/xpu/sycl/GroupNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/GroupNormKernels.cpp
@@ -639,7 +639,8 @@ struct GNFusedForwardMediumFunctor {
 #pragma unroll
         for (int v = 0; v < VEC_SIZE; v++) {
           const int pos = (sg_lid + li * SIMD) * VEC_SIZE + v;
-          const index_t c = pos / S_;
+          const index_t c = static_cast<index_t>(
+              s_divider_.div(static_cast<unsigned_index_t>(pos)));
           T_ACC gv = gamma_
               ? static_cast<T_ACC>(gamma_[g_offset + c])
               : T_ACC(1);
@@ -685,9 +686,13 @@ struct GNFusedForwardMediumFunctor {
         mean_(mean),
         rstd_(rstd),
         mean_acc_(mean_acc),
-        rstd_acc_(rstd_acc) {}
+        rstd_acc_(rstd_acc),
+        s_divider_(static_cast<unsigned_index_t>(S)) {}
 
  private:
+  // See GNFusedForwardFunctor's s_divider_ for rationale (magic-number
+  // multiply+shift for 32-bit index_t, plain division fallback for 64-bit).
+  using unsigned_index_t = std::make_unsigned_t<index_t>;
   index_t D_;
   index_t S_;
   index_t DS_;
@@ -702,6 +707,7 @@ struct GNFusedForwardMediumFunctor {
   T* rstd_;
   T_ACC* mean_acc_;
   T_ACC* rstd_acc_;
+  at::detail::IntDivider<unsigned_index_t> s_divider_;
 };
 
 // Fused GroupNorm forward kernel: combines reduction + normalization
--- a/src/ATen/native/xpu/sycl/GroupNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/GroupNormKernels.cpp
@@ -1088,7 +1088,7 @@ void group_norm_kernel_impl(
   // Small-DS path: single vec4 per lane (covers DS where lanes <= SIMD).
   // WG = 1 SG, flat mapping over (N, G) with grid-stride loop.
   auto try_single_vec = [&](auto index_tag) -> bool {
-    if (simd != 16)
+    if (simd != 32)
       return false;
     using index_t = decltype(index_tag);
     if (DS % FUSED_VEC_SIZE != 0)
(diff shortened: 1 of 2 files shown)


jianyizh:
@torchxpubot review

jianyizh:
@torchxpubot lint

jianyizh:
@torchxpubot review

jianyizh:
@torchxpubot review

jianyizh:
@torchxpubot review

src/ATen/native/xpu/sycl/GroupNormKernels.cpp
```diff
@@ -365,6 +366,674 @@ bool can_use_vectorization(T* p, int vec_size) {
...
+    }
+    sycl::group_barrier(item.get_group());
+    const T_ACC g_mean = broadcast_[0];
+    const T_ACC g_rstd = broadcast_[1];
+
+    // === Precompute a[c], b[c] in SLM (if budget allows) ===
+    const index_t g = ng % G_;
+    const index_t g_offset = g * D_;
+    if (use_slm_coeff_) {
+      if (gamma_ != nullptr && beta_ != nullptr) {
```
CuiYifeng: How about adding two template parameters like `has_gamma` and `has_beta` to avoid runtime check (at the cost of larger binary)? I noticed that there are multiple similar runtime check for `gamma_/beta_`.

CuiYifeng (commented):
Can we refactor these three functors with a shared params struct like `GNForwardParams`?

jianyizh:
> Can we refactor these three functors with a shared params struct like `GNForwardParams`?

I tried use one struct for the small and medium cases, but it becomes hard to read

src/ATen/native/xpu/sycl/GroupNormKernels.cpp
jianyizh: Now the fused kernel always requires gamma and beta defined. I think this can cover most cases.

jianyizh:
@torchxpubot review

jianyizh:
@torchxpubot review

src/ATen/native/xpu/sycl/GroupNormKernels.cpp
```diff
@@ -365,6 +368,658 @@ bool can_use_vectorization(T* p, int vec_size) {
...
+    // - 1).
+    const index_t g = flat_base & (G_ - 1);
+    const index_t total_groups = N_ * G_;
+    const index_t stride =
+        static_cast<index_t>(item.get_group_range(0)) * GROUPS_PER_SG;
+
+    // Pre-load gamma/beta (g is invariant across iterations since
+    // stride % G == 0 for power-of-2 G and GROUPS_PER_SG).
+    T_ACC my_gamma[VEC_SIZE];
+    T_ACC my_beta[VEC_SIZE];
```
CuiYifeng: Can we avoid using `my_*`? The prefix `my_` does not convey semantics. `gamma_v/beta_v` may be a better choice.

src/ATen/native/xpu/sycl/GroupNormKernels.cpp
```diff
@@ -388,24 +1043,253 @@ void group_norm_kernel_impl(
...
+  }
+
+  int64_t thread_slots = syclGpuEuCount() * syclGpuHWThreadsPerEU();
+
+  // Small-DS path: single vec4 per lane (covers DS where lanes <= SIMD).
+  // WG = 1 SG, flat mapping over (N, G) with grid-stride loop.
+  auto try_small_path = [&](auto index_tag) -> bool {
+    if (!gamma_beta_defined)
+      return false;
+    if (simd != 32)
```
CuiYifeng: How about using SIMD32 here?

```suggestion
    if (simd != SIMD32)
```

src/ATen/native/xpu/sycl/GroupNormKernels.cpp
```diff
@@ -388,24 +1043,253 @@ void group_norm_kernel_impl(
...
+          sycl::range<1>(n_wgs * wg_sz),
+          sycl::range<1>(wg_sz),
+          queue,
+          syclex::properties{syclex::sub_group_size<SIMD32>},
+          kfn);
+    };
+    switch (lanes) {
+      case 1:
+        launch(std::integral_constant<int, 1>{});
+        break;
```
CuiYifeng: In my understanding, such cases (`D * H * W == vec_size`) are usually host-bound (except N is very large). Do we have any real cases that have benefited? We'd better making a trade off between the coverage of the shape and the binary size.

CuiYifeng (changes requested):
Please check the comments. How about separating the detection and execution of each type of functor?

src/ATen/native/xpu/sycl/GroupNormKernels.cpp
jianyizh: I removed D * H * W == vec_size, I see D * H * W == 16 in real workload

src/ATen/native/xpu/sycl/GroupNormKernels.cpp
jianyizh: fixed

src/ATen/native/xpu/sycl/GroupNormKernels.cpp
jianyizh: fixed

CuiYifeng (approved):

LuFinch (approved):

jianyizh:
@torchxpubot merge -f

jianyizh:
@torchxpubot merge -f

PR #4145, opened 2026-06-25
[xccl] Drop oneCCL v1 (C++) backend, keep only v2 C API

Chao1Han:
Previously the xccl backend supported two parallel paths gated by the USE_CCL_V2 env var: the original oneCCL v1 C++ API (`ccl::*` namespace, backed by libccl.so.1) and the newer v2 C API (`oneccl*` functions, backed by libccl.so.2). The v1 path was the default, the v2 path was opt-in.

This change removes the v1 path entirely and makes v2 the only path, mirroring the NCCL backend layout (`ProcessGroupNCCL` calling the C API directly).


cmake/XCCL.cmake
```diff
@@ -22,7 +22,13 @@ if(NOT __XCCL_INCLUDED)
   set_property(
     TARGET torch::xccl PROPERTY INTERFACE_INCLUDE_DIRECTORIES
     ${XCCL_INCLUDE_DIR})
+  # oneCCL declares all C-API symbols (oneccl*) with __attribute__((weak)) in
+  # <oneapi/ccl.h>. Under the default --as-needed link policy this would cause
+  # the libccl shared objects to be dropped from libtorch_xpu.so dependencies
+  # (no strong undefined reference is generated), and the weak symbols would
+  # resolve to NULL at runtime -> segfault on the first oneccl* call.
+  # Wrap them with --no-as-needed,...,--as-needed so they remain in NEEDED.
```
blazej-smorawski: @Chao1Han we could change this attribute I think. I remember I added it initially so it would be possible to have the single package with both APIs, but as the old one is not needed anymore we could remove this weak linkage.

Chao1Han: Agree, dropping weak is the right move now that v1 is gone. The currently shipping oneAPI 2026.0 still has weak CCL_C_API in <oneapi/ccl.h>. If oneCCL drops the weak attribute in a future release we can remove `--no-as-needed` here

src/xccl/xccl.h
```diff
@@ -74,44 +68,17 @@ inline std::string reduceOpToString(c10d::ReduceOp op) {
...
-            return std::tolower(c);
-          });
-      return value == "on" || value == "yes" || value == "1";
-    }
-    return false;
-  }();
-  return cachedValue;
-}
-
 inline const std::string& getVersionString() {
```
zhangxiaoli73, line 71: No backward compability?

src/xccl/ProcessGroupXCCL.cpp
```diff
@@ -615,20 +615,31 @@ std::shared_ptr<xcclComm_t> ProcessGroupXCCL::initXCCLComm(
...
-      rank_,
-      device,
-      q,
-      deviceKey,
-      store_.get(),
+  onecclUniqueId xcclID;
+  if (rank_ == 0 || (singleP2POp && p2pRank == 0)) {
+    onecclGetUniqueId(&xcclID);
+  }
+  broadcastUniqueXCCLID(
```
guangyey, line 623: `xcclCommCounter_` was incremented in `broadcastUniqueXCCLID`, no data race protection, is this fine?

cmake/Modules/FindXCCL.cmake
```diff
@@ -47,22 +47,15 @@ find_file(
...
-)
-
+# Find XCCL v2 C-API library (libccl.so.2.0).
 find_file(
-  XCCL_LIBRARY_2_0
+  XCCL_LIBRARY
   NAMES libccl.so.2.0
   HINTS ${XCCL_LIBRARY_DIR}
   NO_DEFAULT_PATH
 )
```
guangyey, line 58:
```cmake
set(CCL_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES})
set(CMAKE_FIND_LIBRARY_SUFFIXES ".so.2.0")
find_library(XCCL_LIBRARY NAMES ccl
  HINTS ${XCCL_LIBRARY_DIR}
  NO_DEFAULT_PATH
)
set(CMAKE_FIND_LIBRARY_SUFFIXES ${CCL_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES})
```
another choice

src/xccl/ProcessGroupXCCL.cpp
Chao1Han: ProcessGroup is single-threaded in the eager execution model, and the counter only needs to produce a store key that is consistent across ranks, So the increment being outside the lock is intentional and matches NCCL implementation.

cmake/Modules/FindXCCL.cmake
Chao1Han: updated

cmake/Modules/FindXCCL.cmake
```diff
@@ -10,7 +10,7 @@
 # XCCL_FOUND               : True if the system has the XCCL library.
 # XCCL_INCLUDE_DIR         : Include directories needed to use XCCL.
 # XCCL_LIBRARY_DIR         ：The path to the XCCL library.
-# XCCL_LIBRARY             : XCCL library fullname.
+# XCCL_LIBRARY             : Full path to libccl.so.2.0 (oneCCL v2 C API).
```
guangyey: forgot update?

Chao1Han: I keep it because torch print it https://github.com/pytorch/pytorch/blob/14c2be6563899240a522d14383e76b49c8c9d028/cmake/Summary.cmake#L175

guangyey:
```cmake
# XCCL_LIBRARY             : XCCL library fullname. (oneCCL v2 C API).
```

Chao1Han: updated

src/xccl/xccl.h
```diff
@@ -120,16 +87,12 @@ namespace c10d {
 
 struct XCCLStream {
```
guangyey: Is it possible to remove `XCCLStream` entirely?

Chao1Han: I think so, still a few redundant structure for compatibility with the two APIs. Let me clean those up as well.

guangyey: It's fine in a follow up PR.

guangyey (approved):

Chao1Han:
@torchxpubot merge

zhangxiaoli73 (approved):

Chao1Han:
@torchxpubot merge

Chao1Han:
@torchxpubot merge -f "distributed tests passed"

PR #4146, opened 2026-06-25
Fix off-by-one in deform_conv2d 32/64-bit indexing threshold

cyyever:
Aligns the deform_conv2d 32/64-bit indexing threshold with upstream torchvision: gate on `> std::numeric_limits<int32_t>::max()` instead of `> ((int64_t)1 << 31)`.

CUDA uses `> std::numeric_limits<int32_t>::max()`; the XPU port diverged by one (used `2**31`). With the XPU check, a count of exactly `2**31` stays on the 32-bit path, narrows to `INT_MIN` in the kernel's `int` launch parameter, and the strided loop `index < n` never runs — silent wrong gradient/output. Also drops the bare `(1 << 31)` sign-bit UB at `compute_grad_input`.


guangyey (approved):
Thanks!

guangyey:
"lint" + "build" passed

PR #3990, opened 2026-06-12
[test] Fix XPU von Mises sample test deterministic failure under fixed pytest seed

tszulist-hbn:
Fixes https://github.com/intel/torch-xpu-ops/issues/3977
Closes https://github.com/intel/torch-xpu-ops/issues/4004

## Root Cause

`calc_execution_policy` computed `counter_offset` using the `UNROLL` template
parameter as the stride multiplier:

```cpp
uint64_t counter_offset =
    ((total_elements - 1) / (group_size * num_groups * UNROLL) + 1) * UNROLL;
```

However, every call to `dist_func_` inside the kernel always advances the Philox counter by `rand4_engine_calls` (4) regardless of the unroll factor. When `unroll_factor != rand4_engine_calls`, the offset was wrong: kernel launches could receive overlapping or incorrectly spaced Philox sub-streams, producing statistically biased output at certain generator positions.

Additionally, `distribution_nullary_kernel` was calling `calc_execution_policy(numel)` (defaulting to `UNROLL=4`) rather than `calc_execution_policy<unroll_factor>(numel)`, so the number-of-groups calculation was decoupled from the actual kernel unroll factor.

## Fix

- Make `calc_execution_policy` a template function parameterized on `UNROLL` (defaulting to `rand4_engine_calls`).
- Replace `* UNROLL` with `* rand4_engine_calls` in the `counter_offset` formula so the stride always matches actual Philox consumption.
- Pass `unroll_factor` explicitly when calling `calc_execution_policy` from `distribution_nullary_kernel`.

## Validation

```bash
PYTORCH_TEST_WITH_SLOW=1 python -m pytest -xvs \
  third_party/torch-xpu-ops/test/xpu/test_distributions_xpu.py::TestDistributionsXPU::test_vonmises_sample_xpu
# 1 passed
```

PR #4156, opened 2026-06-25
Use correct unroll_factor depending on dtype in `RreluWithNoiseKernels.cpp`

BBBela:
Use correct `unroll_factor` depending on used dtype.
The `_rrelu_with_noise_xpu_train` function calls the following kernel:
```cpp
template <typename scalar_t, int unroll_factor, typename transform_t>
struct RreluWithNoiseKernelFunctor
```
with `unroll_factor == 2` when `std::is_same_v<scalar_t, double>` is true and with `unroll_factor == 4` otherwise. But the kernel parameters are always defined using `calc_execution_policy` with default template parameter which is `UNROLL == 4`.
This PR makes the behavior aligned. We always use the same `unroll_factor` for the same data type.

The same approach is used on CUDA: [CODE FRAGMENT](https://github.com/pytorch/pytorch/blob/e683636150f35c3972bde18ac610291f3e9e577f/aten/src/ATen/native/cuda/RreluWithNoise.cu#L81)


BBBela:
This PR waits for https://github.com/intel/torch-xpu-ops/pull/3990 to be merged first.

BBBela:
I thought about adding some regression test, but I am not sure if it is required and necessary in this case. Especially that this behavior depends on the hardware specifics regarding number of Execution Units and Work Items. But I can add one if needed that verifies the offset is different for different dtypes.
```python
    # A philox4x32 block yields 2 doubles but 4 floats, so at equal numel the
    # double kernel needs twice as many blocks and must reserve twice as much
    # generator offset. Reserving with the default unroll factor of 4 rather
    # than the 2 the double kernel actually uses makes both dtypes reserve the
    # same amount, under-advancing the generator for double.
    def test_double_reserves_more_offset_than_float(self):
        # Must exceed twice the GPU's work-item count for the two candidate
        # unroll factors to disagree: at one loop iteration per work item both
        # round up to the same offset and hide the bug.
        numel = 1 << 24

        def reserved_offset(dtype):
            gen = torch.Generator(device="xpu")
            gen.manual_seed(12345)
            x = torch.full((numel,), -1.0, device="xpu", dtype=dtype)
            noise = torch.zeros_like(x)
            torch._C._nn.rrelu_with_noise(x, noise, 0.0, 1.0, True, gen)
            torch.xpu.synchronize()
            return gen.get_offset()

        self.assertGreater(
            reserved_offset(torch.float64), reserved_offset(torch.float32)
        )
```

PatrykWilczewski (approved):

pbielak (approved):

BBBela:
@torchxpubot merge

PR #4165, opened 2026-06-26
Replace enable_if SFINAE with C++20 requires-clauses

cyyever:
The project compiles as C++20, so std::enable_if return-type SFINAE under src/ can be rewritten as requires-clauses, which read more clearly and keep the return type unobscured. Constraints are preserved exactly, so overload selection is unchanged. In Reduce.h the dummy `enable_if<...>::type* = nullptr` parameter is dropped (all call sites relied on its default).

Authored with assistance from an AI coding agent.


guangyey (approved):
Thanks!

guangyey:
lint and build pass. land it immediately

PR #4166, opened 2026-06-26
Use upstream version for test_native_functions and test_modules

Silv3S:
This PR ports 2 test files, to use upstream ones

- **test_native_functions_xpu.py**
These tests are device-agnostic. No changes in pass rate after migration
- **test_modules_xpu.py**
Remove test overrides to use original upstream functions without workarounds


test/xpu/test_modules_xpu.py
```diff
@@ -13,172 +13,8 @@
 # Owner(s): ["module: intel"]
```
pbielak, line 13: Such setup will only launch XPU-specific test, whereas when running the upstream file directly, we would also add CPU-only tests. Check the time difference - if small, please use the upstream file directly.

pbielak (changes requested):

BBBela (approved):
Thanks!
I believe that we can leave the thin wrapper for now and it can always be removed in the future as next steps.

test/xpu/test_modules_xpu.py
Silv3S: CPU suite takes 4 minutes. It's not much, but if we continue with this approach it will significantly slow down CI

pbielak (approved):

Silv3S:
All tests from `test_native_functions_xpu` and `test_modules_xpu` passed in CI

PR #4174, opened 2026-06-29
Add diagonal_copy, expand_copy, slice_copy.Tensor to XPU yaml

xuhancn:
These three _copy variants for core view ops (diagonal, expand, slice.Tensor) were missing from the XPU native functions yaml, while all other core view _copy variants (alias_copy, as_strided_copy, permute_copy, select_copy.int, split_with_sizes_copy, squeeze_copy, unsqueeze_copy, view_copy) were present.

This inconsistency caused test_views_op_having_view_copy to fail on XPU because torch.ops.aten.diagonal_copy, torch.ops.aten.expand_copy, and torch.ops.aten.slice_copy.Tensor were not properly available.

All three use CompositeExplicitAutogradNonFunctional dispatch (no XPU-specific kernel needed), matching the pattern of existing _copy variants in the XPU yaml.

### DIFF
--- a/yaml/native/native_functions.yaml
+++ b/yaml/native/native_functions.yaml
@@ -9051,6 +9051,27 @@
   tags: view_copy
   autogen: select_copy.int_out
 
+- func: diagonal_copy(Tensor self, int offset=0, int dim1=0, int dim2=1) -> Tensor
+  variants: function
+  dispatch:
+    CompositeExplicitAutogradNonFunctional: diagonal_copy
+  tags: view_copy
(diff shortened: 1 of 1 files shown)


xuhancn:
Closing: the yaml-based fix is no longer needed. Upstream torch-xpu-ops restructured (removed yaml codegen entirely). _copy variants (CompositeExplicitAutogradNonFunctional) are now handled by pytorch's main codegen — no torch-xpu-ops changes required. The pytorch-side PR pins the latest intel/main commit.

PR #4189, opened 2026-06-30
Use `test_embedding.py` directly from upstream

pbielak:
The upstream `nn/test_embedding.py` already has `allow_xpu=True` in its `instantiate_device_type_tests` call. We can run it directly without the local `_xpu.py` wrapper.

Before: 113 passed, 29 skipped (37.19s) 
After: 277 passed, 42 skipped (34.99s) - CPU tests added, but unnecessary runtime patching removed


Silv3S (approved):
LGTM

tszulist-hbn (approved):

PR #4190, opened 2026-06-30
sdpa: tests: adjust  head-dim threshold to backend limit

BartoszKokoszko:
The XPU SDPA flash-attention backend currently accepts head dimensions up to 256 (kXPUFlashAttentionMaxHeadDim = 256), but the XPU test still encoded 192 as the maximum supported value. This made the negative-path assertion stale: the test expected a RuntimeError at 193, while the backend correctly allowed it.

Update the test to use 256 as the supported boundary so it validates the real contract: 256 succeeds and 257 fails with “No available kernel”.


DamJanusz (approved):

jmamzax (approved):

BartoszKokoszko:
https://github.com/pytorch/pytorch/blob/1a0a3c99f32b4c5f7b1aba0d6336427f323f20cc/test/test_transformers.py#L5311

PR #4191, opened 2026-06-30
Add test class to `test_init.py`

pbielak:
The upstream `nn/test_init.py` file also contains the `TestNNInitDeviceType`
class which is not tested against XPU. This commit updates the thin
wrapper to also execute tests on XPU.

Before: 31 passed, 28 subtests passed (6.66s)
After:  53 passed, 56 subtests passed (9.26s)

### DIFF
--- a/test/xpu/nn/test_init_xpu.py
+++ b/test/xpu/nn/test_init_xpu.py
@@ -8,6 +8,7 @@
 
 # Owner(s): ["module: intel"]
 
+from torch.testing._internal.common_device_type import instantiate_device_type_tests
 from torch.testing._internal.common_utils import run_tests
 
 try:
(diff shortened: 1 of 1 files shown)


test/xpu/skip_list_common.py
```diff
@@ -14,7 +14,7 @@
     "nn/test_convolution_xpu.py": None,
     "nn/test_dropout_xpu.py": None,
     "nn/test_embedding_xpu.py": None,
-    "nn/test_init_xpu.py": None,
+    f"{PYTORCH_TEST_DIR}/nn/test_init.py": None,
```
Silv3S: I think we should leave thin wrapper to instantiate and run `TestNNInitDeviceTypeXPU`. These tests were not ported so they are not invoked regardless of this patch being merged.

pbielak: Done

Silv3S (approved):

tszulist-hbn (approved):

PR #4193, opened 2026-06-30
Use `nn/test_module_hooks.py` directly from upstream

pbielak:
The upstream `nn/test_module_hooks.py` can be directly called without the local `_xpu.py` wrapper.

Before: 14 passed (5.87s)
After: 53 passed (1.68s) -- additional test class is launched


Silv3S (approved):
It's strange that 14 tests are 3.5 times slower than 53 if they test same functionality, but I reproduced same results. Is this overhead caused by XPUPatchForImport?

pbielak:
@Silv3S yes, it's the overhead from `XPUPatchForImport` - the op_db rewrite (i.e. iterating through the `common_method_invocations.py` file) takes a lot of time

tszulist-hbn (approved):

PR #4209, opened 2026-07-02
Migrate `test_custom_ops_xpu.py` to use `XPUImportCtx`

Silv3S:
Replace the full copy of upstream `test/test_custom_ops.py` with a thin
wrapper that imports the upstream module via `XPUImportCtx` and only
overrides XPU-specific differences:

- `TestCustomOpAPI.test_split_device`: rewrite to use accelerator-
  agnostic API (`torch.accelerator.current_accelerator().type`)
- `TestCustomOpAPI.test_multi_types`: rewrite to use accelerator-
  agnostic multi-device registration
- `TestCustomOpTesting.test_opcheck_opinfo`: override to run opcheck
  against `custom_op_db` on XPU
- `TestGenerateOpcheckTests`: apply `run_on_accelerator` skip decorator
  to 3 pinned-memory tests (no body change, just re-wrap with
  `unittest.skipUnless(TEST_ACCELERATOR, ...)`)

Test results:

  Before:  271 passed, 14 skipped, 2 xfailed in 23.45s
  After:   351 passed,  8 skipped, 2 xfailed in 40.45s
  \--------------------------------------------------------
  Delta:  +80 passed, -6 skipped

Newly passing (previously skipped, 6):
  - TestCustomOp::test_duplicate_impl
  - TestCustomOpAPI::test_library_register_autocast
  - TestCustomOpAPI::test_library_register_autocast_list_input
  - TestCustomOpAPI::test_library_register_autocast_low_level
  - TestCustomOpAPI::test_library_register_autocast_multiple_times
  - TestCustomOpAPI::test_library_register_autocast_multiple_times_different_devices

Newly passing (inherited from upstream, not previously run, 74):
  - 39 test_custom_op_inplace_tag_* / test_custom_op_out_tag_* tests
  - 17 test_pyobject_dispatch_* / test_register_autograd_* tests
  - test_custom_op_register_fake_ordinary_op
  - test_custom_op_tags_{dedup_pt2_compliant,list,single,tuple}
  - test_library_impl_does_not_enable_pyobject_dispatch_by_default
  - test_library_register_autograd_out_tag_low_level
  - test_mutated_optional_arg_default_none
  - test_mutated_version_bump_does_not_fill_all_defaults
  - test_opcheck_cpu_stride_mismatch_xpu
  - test_opcheck_preserves_pinned_memory_by_default
  - test_opcheck_preserves_pinned_memory_for_schema_check
  - test_register_fake_does_not_leak_on_failure
  - test_safe_aot_autograd_check_checks_gradients_for_non_leaf_inputs
  - test_safe_schema_check_copy_inputs_preserves_pinned_memory_and_copies
  - test_aot_autograd_copy_does_not_check_pinned_memory_for_wrapper_subclass


pbielak (approved):

pbielak:
@Silv3S Please provide the list of new passed and skipped tests in the PR description

Silv3S:
Until tests are aligned in upstream I changed approach to use `XPUPatchForImport` and overwrite CUDA-specific cases

test/xpu/test_custom_ops_xpu.py
```diff
@@ -11,5093 +11,177 @@
...
-        with self.assertRaisesRegex(RuntimeError, "Autograd has not been implemented"):
-            op([y, x])
-        custom_ops._destroy(f"{TestCustomOp.test_ns}::foo")
-        del foo
 
-        @custom_ops.custom_op(f"{TestCustomOp.test_ns}::foo")
-        def foo(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
-            raise NotImplementedError
+@skipIfTorchDynamo("Expected to fail due to no FakeTensor support; not a bug")
+def _test_multi_types(self):
```
pbielak: Update this function to take `device` as an argument (might require moving the function to another class that is instantiated via `instantiate_device_type_tests`). We could also keep `device_types=("cpu", "cuda", "xpu")` for the `torch.library.custom_op` call - it will be easier to merge it back to upstream later

test/xpu/test_custom_ops_xpu.py
```diff
@@ -11,5093 +11,177 @@
...
+    x = torch.randn(3)
+    y = f(x)
+    self.assertEqual(y, x.sin())
+    x = x.xpu()
+    y = f(x)
+    self.assertEqual(y, x.sin())
+
+
+@skipIfTorchDynamo("Expected to fail due to no FakeTensor support; not a bug")
+def _test_split_device(self):
```
pbielak, line 63: ditto - add `device` argument and use it instead of hardcoding `f.register_kernel("xpu")` and `x.xpu()`

test/xpu/test_custom_ops_xpu.py
```diff
@@ -11,5093 +11,177 @@
...
+    self.assertEqual(xpu_call_count, 0)
+
+    x = x.xpu()
+    y = f(x)
+    self.assertEqual(y, x.sin())
+    self.assertEqual(cpu_call_count, 1)
+    self.assertEqual(xpu_call_count, 1)
+
+
+def _test_opcheck_preserves_pinned_memory_for_schema_check(self):
```
pbielak: why is this test even guarded by CUDA specific skipIfs? the function body doesn't seem to be device specific

pbielak (changes requested):

BBBela (commented):
As the `XPUImportCtx` is already introduced into main branch - can we use it instead of `XPUPatchForImport` in this PR or the old version is required in this case?

test/xpu/test_custom_ops_xpu.py
Silv3S: Class for these 2 test cases added

test/xpu/test_custom_ops_xpu.py
Silv3S: Maybe it was guarding from when running CPU-only. Test extended to run also on XPU

test/xpu/test_custom_ops_xpu.py
```diff
@@ -3895,1209 +67,84 @@ def f(x: Tensor) -> Tensor:
...
-
-optests.generate_opcheck_tests(
-    MiniOpTestOther,
-    ["aten", "mini_op_test"],
-    _MINIOPTEST_FAILURES_DICT_PATH,
-    test_utils=optests.generate_tests.DEPRECATED_DEFAULT_TEST_UTILS,
+TestGenerateOpcheckTests.test_opcheck_preserves_pinned_memory_for_schema_check = run_on_cuda_or_xpu(
+    inspect.unwrap(
+        TestGenerateOpcheckTests.test_opcheck_preserves_pinned_memory_for_schema_check
+    )
```
BBBela, line 163: If my understanding is correct, the `inspect.unwrap()` strips all the decorators - right?
So not only the `@unittest.skipIf(not TEST_CUDA, "pinned CPU memory requires CUDA")` is dropped, but also all the other decorators. I don't think we should do that. That is risky, because it can drop decorators that change the test behavior. My opinion is that we should either use some mechanism that allows to drop just the one particular decorator or simply, like previously, leave the whole test function body copied and the one decorator adjusted, while still the final and desired solution is to change the decorator in the upstream.

test/xpu/test_custom_ops_xpu.py
```diff
@@ -11,3881 +11,53 @@
...
+        TestLibrarySourceLocation,
+        TestOpProfiles,
+        TestTypeConversion,
+    )
+
 
+# Temporary class to test custom op API with XPU device type.
+# The original TestCustomOpAPI class is not parametrized by device.
+# When testcases in upstream are generialized, this class can be removed.
+class TestDeviceCustomOpAPI(TestCase):
```
BBBela: I think we should use what is already present in the upstream file `test_custom_ops.py` - the `TEST_ACCELERATOR` approach.
So instead of creating new class and instantiating it for XPU, we could do what is already done in the upstream like that:
```python
    @unittest.skipIf(not TEST_ACCELERATOR, "requires accelerator")
    def test_multi_types(self):
        @torch.library.custom_op(
            "_torch_testing::f", mutates_args=(), device_types=("cpu", torch.accelerator.current_accelerator().type)
        )
```

test/xpu/skip_list_common.py
```diff
@@ -280,16 +280,7 @@
...
-        # https://github.com/intel/torch-xpu-ops/issues/3644
-        # CPU-only tests, flaky under pytest-xdist worksteal scheduling
-        "test_backward_dict_grad_for_nontensor",
-        "test_backward_dict_invalid_keys",
-        "test_backward_dict_requires_keys_for_input_optional_tensors",
-        "test_backward_dict_requires_keys_for_input_tensors",
-        "test_backward_grads_are_tensor_or_none",
-        "test_backward_impl_on_existing_op",
-    ),
+    "test_custom_ops_xpu.py": None,
```
BBBela: Why removing those entries from `skip_list_common.py`?
Please correct me if I am wrong, but from what I understand, this PR version still launches those tests with `pytest-xdist`, so the problem is still present, just randomly passed the tests during migration. The fix would be to launch those tests with upstream CI mechanism that does not use the `pytest-xdist`, right?

BBBela (commented):
One general thought I have is that there are several tests (for example `test_library_register_torch_dispatch_low_level`) where the previous local copy changed the `"cuda"` entries to `"xpu"` entries. This example test in current version uses the upstream version directly that has the `"cuda"` hardcoded back again. In those tests it shouldn't functionally matter, because those tests test CPU execution, while `"cuda"` or `"xpu"`  keys are just for registration, but still, if we want to make the upstream tests generic, I think it shouldn't be abandoned, but somehow generalized or at least marked for future generalization.

test/xpu/skip_list_common.py
Silv3S: It passes locally both with(out) `pytest-xdist`. Also [logs](https://productionresultssa14.blob.core.windows.net/actions-results/9c80d28f-95fc-4f2a-9104-9e12097f4d01/workflow-job-run-adfc8544-934e-51ce-b227-3a2eea52ec0e/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-07-13T12%3A27%3A16Z&sig=GSS3y7k9wlbX0fY4B6mRLbjx1v4LTvPqmvBcNqMyTqA%3D&ske=2026-07-13T14%3A32%3A24Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-07-13T10%3A32%3A24Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-07-13T12%3A17%3A11Z&sv=2025-11-05) from this PR CI:

- 2026-07-13T12:05:33.3208723Z [gw7] [ 91%] PASSED test_custom_ops_xpu.py::TestCustomOp::test_backward_dict_grad_for_nontensor
- 2026-07-13T12:05:33.3640488Z [gw7] [ 92%] PASSED test_custom_ops_xpu.py::TestCustomOp::test_backward_dict_invalid_keys 
- 2026-07-13T12:05:33.3906250Z [gw7] [ 92%] PASSED test_custom_ops_xpu.py::TestCustomOp::test_backward_dict_requires_keys_for_input_optional_tensors 
- 2026-07-13T12:05:29.3839933Z [gw2] [  1%] PASSED test_custom_ops_xpu.py::TestCustomOp::test_backward_dict_requires_keys_for_input_tensors
- 2026-07-13T12:05:29.3932733Z [gw2] [  2%] PASSED test_custom_ops_xpu.py::TestCustomOp::test_backward_grads_are_tensor_or_none 
- 2026-07-13T12:05:29.4399567Z [gw2] [  4%] PASSED test_custom_ops_xpu.py::TestCustomOp::test_backward_impl_on_existing_op

test/xpu/test_custom_ops_xpu.py
Silv3S: Yes `inspect.unwrap` strips all decorators, order doesn't matter. As you said - we might diverge from upstream, but I don't think that is big risk. I'd rather expect discrepancy in test body due to outdated copy in torch-xpu-ops than solely due to any decorator change. I'd keep current approach with `inspect.unwrap` as intermediate step, until tests are fully upstreamed

test/xpu/skip_list_common.py
BBBela: That is weird to me - those tests were skipped for some reason, this reason hasn't been removed and now they are passing. Unless the reason has disappeared? Unless we find why it started passing, I think we should keep it skipped as it may be still flaky under some specific, unpredictable (or rather unknown) condition - to avoid it starting failing again and repeating investigation of the problem again.

test/xpu/test_custom_ops_xpu.py
BBBela: In my opinion it is much bigger risk to drop all the decorators than to have those several tests diverging from upstream. Unwrapping may drop some important decorators like skips for specific OS, or skips due to some important condition, or specifying list of dtypes, etc. Some tests have several decorators. I am not sure if this unwrapping wouldn't also drop the parametrization decorators if used in wrong way.

I am sure that you have checked the test functions that you unwrapped and it is not a problem in this case. I am rather talking about principle. Generally, unwrapping all decorators is potentially destructive, so I would avoid using it even when we checked the particular function we unwrap. At least to prevent someone looking at it and using it without deeper knowledge.

BBBela (approved):
Looks good to me, thanks!

test/xpu/test_custom_ops_xpu.py
BBBela: Resolving after offline discussion with @pbielak

tszulist-hbn (approved):

test/xpu/test_custom_ops_xpu.py
```diff
@@ -11,5090 +11,171 @@
...
-        if typ is float:
-            return [3.14]
-        if typ is bool:
-            return [True]
-        if typ is str:
-            return ["foo"]
-        if torch.distributed.is_available():
-            from torch.distributed.distributed_c10d import GroupName
+run_on_accelerator = unittest.skipUnless(
+    TEST_ACCELERATOR, "pinned CPU memory requires CUDA or XPU"
```
astachowiczhabana, line 149: The upstream methods have a second decorator on top of the CUDA skip (PYTORCH_CUDA_MEMCHECK, "is_pinned uses failure to detect pointer property").
  inspect.unwrap() strips both, and run_on_accelerator only puts back one of them, so the PYTORCH_CUDA_MEMCHECK guard quietly disappears. Other XPU tests in this repo
  that touch is_pinned (e.g. test_torch_xpu.py:5734) keep this guard explicitly. Was dropping it here intentional?

astachowiczhabana (changes requested):

astachowiczhabana (approved):

test/xpu/test_custom_ops_xpu.py
pbielak: The `PYTORCH_CUDA_MEMCHECK` variable is used to selectively skip test when running on CUDA. On our side, we don't need this flag. Let's skip it here.

pbielak:

pbielak (approved):

pbielak:

PR #4214, opened 2026-07-03
Simplify xpu::pair

cyyever:
Reverts the std::pair replacement and instead simplifies `at::xpu::pair`.

`std::pair` is not trivially copyable, so it broke `sycl::shift_group_left` (which requires `is_trivially_copyable_v<T> || is_vec<T>`) — see the linux-build failure. `xpu::pair` is the SYCL analog of `thrust::pair` (used on CUDA for the same `__shfl_sync` requirement), so it must stay.

This PR trims the unused API surface from `xpu::pair`:
- converting constructors from `pair<U1,U2>` and `std::pair<U1,U2>` (no callers)
- `make_pair` (no callers)
- `get<N>` (no callers)
- `operator!=` (auto-derived from `operator==` in C++20)

The type remains trivially copyable (defaulted copy ctor, no user copy/move/dtor) and API-compatible with existing call sites.


guangyey (commented):
`std::pair` is not trivially copyable, this is the reason why we need `xpu::pair`. See https://github.com/intel/torch-xpu-ops/pull/635

cyyever:
@guangyey You're right, but we can change the PR to simplify the current implementation.

guangyey (approved):
Thanks!

PR #4215, opened 2026-07-03
[xccl] add oneccl bench scripts

Chao1Han:
Add latency benchmarks scripts for oneCCL collectives


test/xpu/distributed/bench_script/bench_ccl_allgather_latency.cpp
```diff
@@ -0,0 +1,326 @@
+/*
```
zhangxiaoli73, line 1: I think you don't need to mention c_api here. No C++ API later.

zhangxiaoli73 (approved):

test/xpu/distributed/bench_script/bench_ccl_allgather_latency.cpp
Chao1Han: renamed

chuanqi129 (approved):

PR #4225, opened 2026-07-03
Migrate `test_utils_xpu.py` from file copy to `XPUImportCtx`-based import

pbielak:
Overridden tests:
- `test_format_traceback_short`: widen filename regex to match `test_utils_xpu.py` (not needed once file moves upstream).
- `test_infer_device_state_recursive_multi_device`: upstream asserts `f"Device types: ['{dev_type}', 'meta']"` which relies on alphabetical order (`'cuda' < 'meta'`); fails on XPU (`'meta' < 'xpu'`). Override checks both types are present without assuming order. The old file-copy never hit this because `TEST_MULTIGPU` (CUDA-only) was always False.
- add `allow_xpu=True` to `TestDeviceUtils` instantiation

Before: 52 passed, 0 failed, 6,075 skipped, 10 subtests passed 
After:    55 passed, 0 failed, 6,072 skipped, 10 subtests passed 
(+3 passed: multi-device checkpoint tests now run via @deviceCountAtLeast(2) instead of CUDA-only TEST_MULTIGPU, and test_get_default_device_more_xpu enabled via allow_xpu=True. 0 regressions)

### DIFF
--- a/test/xpu/test_utils_xpu.py
+++ b/test/xpu/test_utils_xpu.py
@@ -11,1087 +11,132 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 # Owner(s): ["module: intel"]
-
-import multiprocessing
-import os
-import random
-import shutil
-import subprocess
-import sys
-import tempfile
-import textwrap
-import traceback
-import unittest
-import warnings
-from typing import Any
+# ruff: noqa: F401
 
 import torch
-import torch.cuda
-import torch.nn as nn
-import torch.utils.cpp_extension
-import torch.utils.data
-from torch._utils import try_import
-from torch._utils_internal import deprecated
-from torch.testing._internal.common_cuda import TEST_MULTIGPU
-from torch.testing._internal.common_device_type import (
-    instantiate_device_type_tests,
-    onlyCPU,
-    ops,
-)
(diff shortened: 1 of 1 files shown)


test/xpu/test_utils_xpu.py
```diff
@@ -11,1087 +11,132 @@
...
-        self.assertExpectedInline(
-            torch._utils.render_call(torch.sum, [torch.randn(100)], {"dim": 0}),
-            """torch.sum(tensor([...], size=(100,)), dim=0)""",
+def _test_format_traceback_short(self):
+    try:
+        raise RuntimeError
+    except RuntimeError as e:
+        self.assertRegex(
+            format_traceback_short(e.__traceback__),
+            r".*test_utils(?:_xpu\d?)?\.py:\d+ in _test_format_traceback_short",
```
BBBela: Why do we need the additional `\d?` fragment in this regex versus what was previously in this file?

pbielak: Thanks for the catch - fixed

BBBela (approved):
Thanks!

Silv3S (approved):

pbielak:

PR #4226, opened 2026-07-03
Extend complex32 ConvTranspose regression to 1D and 3D for XPU

gplutop7:
Extends the complex32 ConvTranspose regression tests to cover 1D and 3D dimensions, 
complementing the existing 2D coverage from PR #3991.

Problem
ConvTranspose1d, ConvTranspose2d, and ConvTranspose3d operators exhibit complex32 parity failures between CPU and XPU backends due to fundamental implementation differences:
- CPU uses `slow_conv_transpose{1,2,3}d` kernels
- XPU uses oneDNN deconv operators

These distinct implementations produce divergent complex32 results beyond strict parity tolerances (rtol=1e-3, atol=1e-5). The upstream parity tests are marked expectedFailure or skipped for complex32.

Solution
This PR adds focused regression tests for ConvTranspose1d and ConvTranspose3d (extending existing ConvTranspose2d coverage) that validate XPU quality using an alternative metric: comparing XPU(complex32) output against CPU(complex64) reference.


BBBela (commented):
For me it looks like a good candidate to use the parameterized test. Something like that:
```python
class TestConvTransposeComplex32Reference(TestCase):
    @parametrize(
        "conv_op, module_cls, x_shape",
        [
            subtest((F.conv_transpose1d, nn.ConvTranspose1d, (2, 4, 3)), name="1d"),
            subtest((F.conv_transpose2d, nn.ConvTranspose2d, (2, 4, 3, 4)), name="2d"),
            subtest((F.conv_transpose3d, nn.ConvTranspose3d, (2, 4, 3, 4, 5)), name="3d"),
        ],
    )
    def test_complex32_xpu_close_to_ref64(self, conv_op, module_cls, x_shape):
```

What do you think about such approach? It looks like even less code duplication, doesn't it?

gplutop7:
> For me it looks like a good candidate to use the parameterized test. Something like that:
> 
> ```python
> class TestConvTransposeComplex32Reference(TestCase):
>     @parametrize(
>         "conv_op, module_cls, x_shape",
>         [
>             subtest((F.conv_transpose1d, nn.ConvTranspose1d, (2, 4, 3)), name="1d"),
>             subtest((F.conv_transpose2d, nn.ConvTranspose2d, (2, 4, 3, 4)), name="2d"),
>             subtest((F.conv_transpose3d, nn.ConvTranspose3d, (2, 4, 3, 4, 5)), name="3d"),
>         ],
>     )
>     def test_complex32_xpu_close_to_ref64(self, conv_op, module_cls, x_shape):
> ```
> 
> What do you think about such approach? It looks like even less code duplication, doesn't it?

@BBBela Good idea - applied the suggestion.

BBBela (commented):
Thanks!
Do we still need those helper methods when we switched to the parameterized approach?
It looks like it would be simpler to just make one test function. Currently the extracted helpers are only used once.

gplutop7:
> Thanks! Do we still need those helper methods when we switched to the parameterized approach? It looks like it would be simpler to just make one test function. Currently the extracted helpers are only used once.

Thanks for the suggestion. I agree that after the parametrization refactor these helpers are currently single-use. I’d still prefer to keep them, because they separate input construction from validation logic, which makes the test flow easier to read and maintain. This structure is also more extension-friendly if we add more ConvTranspose variants, extra cases, or dimension-specific checks later.

If you prefer, I can inline them, but I thought keeping this separation gives us better clarity and a cleaner path for future test growth.

BBBela:
> If you prefer, I can inline them, but I thought keeping this separation gives us better clarity and a cleaner path for future test growth.

Those are regression tests, not the main suite, so I don't expect much development and new cases here to be honest. I would prefer to inline them and extract only when new cases would come that require the same logic.

test/regressions/test_conv_transpose_complex32.py
```diff
@@ -8,35 +8,47 @@
...
+
 import torch
 import torch.nn.functional as F
-from torch.testing._internal.common_utils import run_tests, TestCase
-
-
-class TestConvTranspose2dComplex32Reference(TestCase):
-    def test_conv_transpose2d_complex32_xpu_close_to_ref64(self):
+from torch import nn
+from torch.testing._creation import make_tensor
```
BBBela:
```suggestion
from torch.testing import make_tensor
```
From what I know the `torch.testing` is the official API for `make_tensor`.

test/regressions/test_conv_transpose_complex32.py
```diff
@@ -8,35 +8,47 @@
 
 # Owner(s): ["module: intel"]
 # Regression test for https://github.com/intel/torch-xpu-ops/issues/3030
```
BBBela: I am wondering if this comment wouldn't fit better just above the test function or test class. Currently it refers to the whole file, while there may be added more regression test cases for different regressions in the future.

test/regressions/test_conv_transpose_complex32.py
```diff
@@ -8,35 +8,47 @@
...
-
-        xpu_out = F.conv_transpose2d(
-            x_cpu.to("xpu"), w_cpu.to("xpu"), b_cpu.to("xpu")
-        ).cpu()
-        ref64 = F.conv_transpose2d(
+        in_channels = x_shape[1]
+        x_cpu = make_tensor(x_shape, dtype=torch.complex32, device="cpu")
+        module_cpu = module_cls(in_channels, 5, 3).to(dtype=torch.complex32)
+        w_cpu = module_cpu.weight.detach()
+        b_cpu = module_cpu.bias.detach()
```
BBBela: Do we need to use this `module_cpu`? From what I know, in this case the `module_cls()` call will produce the real values by default and then casting them to complex numbers makes the imaginary part equal to 0 for every element - are there reasons to change the previous approach to this one?
Then we could drop the `module_cls` parameter and instead of that parameterize with `conv_op`, `x_shape` and `weights_shape` for example?

test/regressions/test_conv_transpose_complex32.py
```diff
@@ -8,35 +8,47 @@
...
             b_cpu.to(torch.complex64),
         )
-
-        self.assertEqual(
-            xpu_out.to(torch.complex64),
-            ref64,
-            atol=atol,
-            rtol=rtol,
-        )
+        self.assertEqual(xpu_out.to(torch.complex64), ref64, atol=3e-3, rtol=1e-3)
```
BBBela: That is interesting that previous version required `atol = 2e-2` and now we can use the `3e-3`? Why is that? That is almost the order of magnitude change in tolerance.

test/regressions/test_conv_transpose_complex32.py
gplutop7: Fixed

test/regressions/test_conv_transpose_complex32.py
gplutop7: Fixed

test/regressions/test_conv_transpose_complex32.py
gplutop7: Fixed.

test/regressions/test_conv_transpose_complex32.py
gplutop7: Good point. There was no real benefit in using module_cls() here, so I removed it and simplified the parametrization to the values actually needed by the test.

BBBela (commented):
Thanks! Looks good to me.

The one remaining thing is that I cannot see those added test cases in CI - could you please check why they are not launched in our CI?

My guess is that maybe there is some difference between using `@instantiate_parametrized_tests` decorator and simply calling `instantiate_parametrized_tests(TestConvTransposeComplex32Reference)` - but I am not sure. I just checked that in another regression test file it is done that way and it works and shows in our CI.

BBBela:
The added test cases were still not present in the CI logs, so I removed the `disable_build` label and retriggered CI.
It looks like with the `disable_build` label, some different source code was used for tests instead of the version from this PR.

BBBela (approved):
Thanks!

kdrozd-dev (approved):

gplutop7:
Added disable_all label because this PR focuses on adding two new test cases in TestConvTransposeComplex32Reference. The failing tests are not related to this PR.

PR #4240, opened 2026-07-06
Extract `XPUImportCtx` base class from `XPUPatchForImport`

pbielak:
Minimal import-only context manager that handles sys.path, suppresses `instantiate_device_type_tests`, and optionally replaces TestCase — without op-db rewrites or CUDA->XPU decorator translation.


BBBela (approved):
From what I see the `XPUPatchForImport` behavior is effectively unchanged, yet we extract the minimal version as base class `XPUImportCtx`.
Looks good to me, thanks for this change! 😉

Silv3S (approved):

PR #4244, opened 2026-07-06
Migrate `test_decomp_xpu.py` to use `XPUImportCtx`

### LINKED ISSUE #4218: [split from #4211] test_decomp_xpu: linalg_vector_norm XPU backward produces near-zero Jacobian
### 🐛 Describe the bug

**Affected tests (1):**
- `TestDecompXPU.test_quick_core_backward_linalg_vector_norm_xpu_float64`

**Root cause:** PyTorch PR #185735 (commit `db9d1d7e8df`) added `aten.linalg_vector_norm` to the core ATen decomposition table and added a new gradcheck test `test_quick_core_backward`. The test passes on CPU and CUDA upstream but fails on XPU with a dramatic Jacobian mismatch:
- numerical: `-1.0000`
- analytical: `-3.2369e-10` (effectively zero)

The decomposition breaks `linalg_vector_norm` into `pow`, `sum`, etc. The backward of one of these sub-ops on XPU produces near-zero instead of the correct gradient. A likely contributing factor is PR `ba09082f036` which enabled SYCL Native Fast Math on XPU — fast-math can reduce float64 backward precision. The specific broken sub-op (most likely `pow` backward or `norm_backward`) needs isolated investigation.

**Fix path:**
1. Bisect to identify which sub-op backward returns wrong gradients on XPU for `float64`
2. Either fix the sub-op or disable fast-math for grad-check-sensitive paths
3. If the issue is fast-math, consider scoping it away from `float64` backward ops

**Reproducer:**
```bash
cd third_party/torch-xpu-ops/test/xpu && \
  python -m pytest -sxv test_decomp_xpu.py -k "test_quick_core_backward_linalg_vector_norm_xpu_float64"
```

**Representative error:**
```
GradcheckError: Jacobian mismatch for output 0 with respect to input 0,
numerical:tensor([[-1.0000]], device='xpu:0', dtype=torch.float64)
analytical:tensor([[-3.2369e-10]], device='xpu:0', dtype=torch.float64)
```

### Versions

- PyTorch version: `2.14.0a0+gita0f2bdc`
- Good commit: `de829759c7e622b806788914ed98a6160b2c5f1c`
- Bad commit: `a0f2bdc50c8590762d0c9ff73bd1801898707001`
- Upstream fix commit: `e356f51370f` (PR #188591, 2026-07-02)
- Hardware: Intel Data Center GPU Max 1100

pbielak:
Replace the full copy of upstream `test/test_decomp.py` with a wrapper
that imports the upstream module via `XPUImportCtx` and only overrides
XPU-specific differences:

- Tolerance table patches (3 entries in `op_assert_ref_tol_table`)
- `CROSS_REF_EXCLUDE_SET`: add `max_pool2d_with_indices_backward` for XPU
- `CROSS_REF_BACKWARD_EXCLUDE_SET`: add `max_pool1d/max_pool2d` for XPU
- `core_backward_failures`: un-skip `logaddexp` (passes on XPU) by patching
  `_op_overrides` directly, since `@skipOps` parses entries at class
  definition time
- `TestDecomp.do_cross_ref`: migrate `(device_type, None, op.name)` test_key
  for device-specific exclusion matching
- `DecompOneOffTests`: override 6 methods to remove `@onlyCUDA` and use
  the `device` argument from `instantiate_device_type_tests`
- remove redundant local `HasDecompTest` expect symlinks:
   - `test/xpu/expect/HasDecompTest.test_aten_core_operators.expect`
   - `test/xpu/expect/HasDecompTest.test_has_decomposition.expect`
- rename RMSNorm one-off test to upstream-style naming:
  `test_rms_norm_decomp_accelerator` -> `test_rms_norm_decomp_gpu`
- drop two stricter-than-upstream tolerance overrides after validation:
   - `(torch.bfloat16, torch.ops.aten._batch_norm_with_update.default): 2e-7`
   - `(torch.float16, torch.ops.aten.mv.default): 1e-5`

Test results:

  Before:  15 failed, 8384 passed,  751 skipped, 51 xfailed
  After:   13 failed, 8397 passed,  755 skipped, 48 xfailed
  \---------------------------------------------------------
  Delta:   -2 failed,  +13 passed,   +4 skipped, -3 xfailed
    
Newly passing tests (8):
  - TestDecompXPU::test_binary_cross_entropy_with_logits_decomp_xpu
  - TestDecompXPU::test_hann_window_decomp_xpu
  - TestDecompXPU::test_quick_core_backward_addcdiv_xpu_float64 (previous XFAIL)
  - TestDecompXPU::test_quick_core_backward_lerp_xpu_float64 (previous XFAIL)
  - TestDecompXPU::test_quick_core_backward_linalg_vector_norm_xpu_float64 (previous FAIL)
  - TestDecompXPU::test_quick_core_backward_logaddexp_xpu_float64 (previous ERROR)
  - TestDecompXPU::test_quick_core_backward_nn_functional_binary_cross_entropy_with_logits_xpu_floa
t64 (previous XFAIL)
  - DecompOneOffTestsXPU::test_addmm_out_dtype_decomp_xpu

Newly passing (inherited from upstream, not from migration):
  - HasDecompTest::test_conv1d_decomposition
  - HasDecompTest::test_mm_decompose_mm_dde
  - DecompOneOffTestsXPU::test_fused_dropout_compile_extreme_p_xpu
  - DecompOneOffTestsXPU::test_fused_dropout_decomposition_extreme_p_xpu
  - TestDecompXPU::test_batch_norm_eval_emits_rsqrt_xpu

Newly skipped (4, inherited from upstream, all are onlyCPU tests):
  - DecompOneOffTestsXPU::test_addmv_decomp_rejects_mixed_dtype_xpu
  - DecompOneOffTestsXPU::test_linalg_vector_norm_decomp_correctness_xpu
  - TestDecompXPU::test_decomp_crossref_mode_decomposes_mutable_functional_ops_xpu
  - TestDecompXPU::test_decomp_crossref_mode_skips_overloaded_inplace_xpu

Note: RMSNorm rename (test_rms_norm_decomp_accelerator_xpu -> test_rms_norm_decomp_gpu_xpu)
is treated as rename-only and excluded from added/removed accounting.

### DIFF
--- a/test/xpu/test_decomp_xpu.py
+++ b/test/xpu/test_decomp_xpu.py
@@ -5,1378 +5,354 @@
 # You may obtain a copy of the License at
 #
 # http://www.apache.org/licenses/LICENSE-2.0
+#
+# Portions of this file are derived from PyTorch
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: BSD-3-Clause
 
-# Owner(s): ["module: decompositions"]
+# Owner(s): ["module: intel"]
+# ruff: noqa: F401
 
-
-import functools
-import itertools
-import re
 import unittest
-from collections import defaultdict
-from functools import partial
 
-import torch._inductor.decomposition
-import torch.autograd
-from torch import Tensor
-from torch._decomp import core_aten_decompositions, decomposition_table
+import torch
 from torch._dispatch.python import enable_python_dispatcher
-from torch._export.utils import _is_cia_op
 from torch._ops import DispatchKey
-from torch.testing import make_tensor
-from torch.testing._internal.common_cuda import SM70OrLater, tf32_off
+from torch.testing._internal.common_cuda import SM70OrLater
 from torch.testing._internal.common_device_type import (
     instantiate_device_type_tests,
-    onlyCPU,
-    onlyNativeDeviceTypes,
-    ops,
-    skip,
-    skipOps,
-    xfail,
+    onlyOn,
 )
-from torch.testing._internal.common_methods_invocations import op_db
-from torch.testing._internal.common_modules import module_db, modules
 from torch.testing._internal.common_utils import (
-    is_iterable_of_tensors,
     run_tests,
     skipIfCrossRef,
     skipIfTorchDynamo,
-    suppress_warnings,
     TEST_WITH_ASAN,
-    TEST_WITH_SLOW,
-    TestCase,
-    unMarkDynamoStrictTest,
 )
-from torch.utils import _pytree as pytree
-from torch.utils._python_dispatch import TorchDispatchMode
-from torch.utils._pytree import tree_flatten, tree_map, tree_unflatten
+from torch.utils._pytree import tree_map
 
-device_type = (
-    acc.type if (acc := torch.accelerator.current_accelerator(True)) else "cpu"
-)
+try:
+    from xpu_test_utils import retarget_outermost_onlycuda_to_onlyon, XPUImportCtx
+except Exception:
+    from .xpu_test_utils import retarget_outermost_onlycuda_to_onlyon, XPUImportCtx
 
-aten = torch.ops.aten
-
-
-# TODO: this isn't going to work with non-aten namespaces
-def overload_to_aten_name(op):
-    return op._schema.name.split("::")[1]
-
-
-# All operators that can have decomp tests
-decomposition_names = {
-    overload_to_aten_name(k)
-    for k in decomposition_table
-    if isinstance(k, torch._ops.OpOverload)
-}
-core_decomposition_names = {
-    overload_to_aten_name(k)
-    for k in core_aten_decompositions()
-    if isinstance(k, torch._ops.OpOverload) and not _is_cia_op(k)
-}
-_decomp_test_ops = [
-    op
-    for op in op_db
-    if op.aten_name in decomposition_names
-    or op.aten_backward_name in decomposition_names
-]
-_decomp_test_ops_core_autograd = [
-    op
-    for op in op_db
-    if op.aten_name in core_decomposition_names and op.supports_autograd
-]
-_sdpa_op_info = [op for op in op_db if "scaled_dot_product_attention" in op.aten_name]
-
-
-def diff_arg(arg, requires_grad=True):
-    def is_differentiable_arg(arg):
-        if requires_grad:
-            return arg.requires_grad
-        else:
-            return arg.is_floating_point() or arg.is_complex()
-
-    if is_iterable_of_tensors(arg):
-        if all(is_differentiable_arg(a) for a in arg):
-            return True
-        if all(not is_differentiable_arg(a) for a in arg):
-            return False
-        raise RuntimeError("NYI: The test runner can't handle this")
-    return isinstance(arg, Tensor) and is_differentiable_arg(arg)
-
-
-# Version of autograd.grad with some differences:
-#   - pytree inputs is allowed (but leaves of the pytree have to all
-#     be tensors)
-#   - if an input is not used as part of derivatives, we will return a
-#     zero-filled tensor for the result
-def _autograd_grad(
-    outputs, inputs, grad_outputs=None, retain_graph=False, create_graph=True
-):
-    inputs, inputs_spec = tree_flatten(inputs)
-    diff_inputs = tuple(inp for inp in inputs if inp.requires_grad)
-    if grad_outputs is None:
-        diff_outputs = tuple(out for out in outputs if out.requires_grad)
-    else:
-        diff_grad_outputs = [
-            (out, go) for out, go in zip(outputs, grad_outputs) if out.requires_grad
-        ]
-        if len(diff_grad_outputs) == 0:
-            diff_outputs, grad_outputs = (), ()
-        else:
-            diff_outputs, grad_outputs = zip(*diff_grad_outputs)
-    grad_inputs = torch.autograd.grad(
-        diff_outputs,
-        diff_inputs,
-        grad_outputs,
-        retain_graph=retain_graph,
-        create_graph=create_graph,
-        allow_unused=True,
+with XPUImportCtx(False):
+    from test_decomp import (
+        CROSS_REF_BACKWARD_EXCLUDE_SET,
+        CROSS_REF_EXCLUDE_SET,
+        DecompOneOffTests,
+        decomposition_names,
+        HasDecompTest,
+        normalize_op_input_output,
+        op_assert_ref_tol_table,
+        ref_vjp_no_create,
+        TestDecomp,
     )
-    result = []
-    grad_inputs_iter = iter(grad_inputs)
-    for inp in inputs:
-        if inp.requires_grad:
-            grad_input = next(grad_inputs_iter)
-            if grad_input is None:
-                result.append(torch.zeros_like(inp))
-            else:
-                result.append(grad_input)
-        else:
-            result.append(torch.zeros_like(inp))
-    return tree_unflatten(result, inputs_spec)
 
-
-def _as_tuple(val):
-    if isinstance(val, tuple):
-        return val
-    return (val,)
-
-
-def ref_vjp_no_create(f, *primals):
-    result = f(*primals)
-
-    def wrapped(cotangents):
-        return _autograd_grad(
-            _as_tuple(result),
-            primals,
-            _as_tuple(cotangents),
-            create_graph=False,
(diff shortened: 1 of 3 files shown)


BBBela (commented):
The description says that there are new passing test cases from `HasDecompTest` test class:
- HasDecompTest::test_conv1d_decomposition
- HasDecompTest::test_mm_decompose_mm_dde

but this PR doesn't import the `HasDecompTest` into the `test_decomp_xpu.py` file. How is it possible that they are launched when the test class is not imported?

test/xpu/test_decomp_xpu.py
```diff
@@ -5,1378 +5,432 @@
...
+        (
+            torch.bfloat16,
+            torch.ops.aten.reflection_pad2d_backward.default,
+        ): 5e-2,  # upstream 5e-3 -> 5e-2
         (
             torch.bfloat16,
             torch.ops.aten._batch_norm_with_update.default,
-        ): 2e-7,  # adjust tolerance for xpu
+        ): 2e-7,  # upstream 5e-7 -> 2e-7
+        (torch.float16, torch.ops.aten.mv.default): 1e-5,  # upstream 2e-5 -> 1e-5
```
BBBela: Why do we need stricter tolerances than in upstream? Can't we just use the upstream looser tolerances directly instead of additional overrides for those two entries?

test/xpu/test_decomp_xpu.py
```diff
@@ -5,1378 +5,432 @@
...
-            return torch.sinh(output)
 
-        normalized_shape_arg = (3, 3, 3)
-        input_tensor = torch.randn(3, 3, 3, device=device, requires_grad=True)
-        weight_tensor = torch.randn(3, 3, 3, device=device, requires_grad=True)
 
-        def forward_pass_fn():
-            return rms_norm_sinh(input_tensor, normalized_shape_arg, weight_tensor)
+# Renamed from test_rms_norm_decomp_cuda to test_rms_norm_decomp_accelerator
+DecompOneOffTests.test_rms_norm_decomp_accelerator = _test_rms_norm_decomp_accelerator
```
BBBela:
```suggestion
DecompOneOffTests.test_rms_norm_decomp_gpu = _test_rms_norm_decomp_gpu
```
Just a suggestion - I mostly saw that in upstream the `cuda` hardcoded test names were changed to `gpu` rather than `accelerator` when making them generic, but I haven't checked all the occurrences, so leaving this comment just for consideration 😉

test/xpu/test_decomp_xpu.py
```diff
@@ -5,1378 +5,432 @@
...
+# Remove the upstream CUDA-only version
+if hasattr(DecompOneOffTests, "test_rms_norm_decomp_cuda"):
+    delattr(DecompOneOffTests, "test_rms_norm_decomp_cuda")
```
BBBela, line 323: Do we need those lines? Without them this test case would be simply skipped on XPU, right? Just a suggestion of simplification, can stay as is 😉

BBBela (commented):
When we use the `HasDecompTest` class directly from upstream, then I suppose we don't need anymore those two symbolic links in torch-xpu-ops repo:
- [file1](https://github.com/intel/torch-xpu-ops/blob/main/test/xpu/expect/HasDecompTest.test_aten_core_operators.expect)
- [file2](https://github.com/intel/torch-xpu-ops/blob/main/test/xpu/expect/HasDecompTest.test_has_decomposition.expect)

Could you please check that and remove those symlinks if they are not needed anymore?

pbielak:
> The description says that there are new passing test cases from `HasDecompTest` test class:
> 
> * HasDecompTest::test_conv1d_decomposition
> * HasDecompTest::test_mm_decompose_mm_dde
> 
> but this PR doesn't import the `HasDecompTest` into the `test_decomp_xpu.py` file. How is it possible that they are launched when the test class is not imported?

@BBBela lintrunner must have removed it as an unused import at some point - fixed

pbielak:
> When we use the `HasDecompTest` class directly from upstream, then I suppose we don't need anymore those two symbolic links in torch-xpu-ops repo:
> 
> * [file1](https://github.com/intel/torch-xpu-ops/blob/main/test/xpu/expect/HasDecompTest.test_aten_core_operators.expect)
> * [file2](https://github.com/intel/torch-xpu-ops/blob/main/test/xpu/expect/HasDecompTest.test_has_decomposition.expect)
> 
> Could you please check that and remove those symlinks if they are not needed anymore?

The symlinked files are not needed anymore - I removed them

test/xpu/test_decomp_xpu.py
pbielak: Sure - I updated the code to use the upstream tolerances in these two cases

test/xpu/test_decomp_xpu.py
pbielak: Updated

test/xpu/test_decomp_xpu.py
pbielak: When renaming the test function, this would show up as another skipped tests, which would make the test statistics more complicated - I kept this attribute removal

BBBela (approved):
Thanks! Looks good to me now 😄

test/xpu/test_decomp_xpu.py
```diff
@@ -5,1378 +5,428 @@
...
-instantiate_device_type_tests(TestDecomp, globals(), only_for="xpu", allow_xpu=True)
 
+# Removed @onlyCUDA; converted @unittest.skipIf decorator into an in-body
+# skip using self.skipTest so the device type is determined at runtime from
+# the `device` argument rather than at decoration time from a module-level
+# variable. Replaced hardcoded kernel name check with regex to work across
+# backends (XPU/CUDA).
+def _test_rms_norm_decomp_gpu(self, device):
+    if torch.device(device).type == "cuda" and not SM70OrLater:
+        self.skipTest("triton requires CUDA with SM>=7.0")
```
Silv3S, line 291:
```suggestion
```
You want to keep this CUDA specific check so it's easier to upstream?

test/xpu/test_decomp_xpu.py
```diff
@@ -5,1378 +5,428 @@
...
 
+# ======================================================================
+# 1. Tolerance table patches (XPU-specific entries)
+# ======================================================================
 
-def _as_tuple(val):
-    if isinstance(val, tuple):
-        return val
-    return (val,)
+op_assert_ref_tol_table.update(
```
Silv3S, line 57: I think we should have dedicated variable for XPU-specific tolerance bumps, because updating this `op_assert_ref_tol_table`  in upstream will affect other backends. For example `op_assert_ref_tol_table_xpu` that is joined if test runs on XPU

test/xpu/test_decomp_xpu.py
```diff
@@ -5,1378 +5,428 @@
...
-    torch.complex32: (0.001, 1e-5),
-    torch.complex64: (1.3e-6, 1e-5),
-    torch.complex128: (1e-7, 1e-7),
-}
-# Returns the "default" rtol and atol for comparing scalars or
-# tensors of the given dtypes.
+from torch._dispatch.python import enable_python_dispatcher
+from torch._ops import DispatchKey
+from torch.testing._internal.common_utils import skipIfTorchDynamo
+from torch.utils._pytree import tree_map
```
Silv3S: Let's move all imports up

test/xpu/test_decomp_xpu.py
```diff
@@ -5,1378 +5,428 @@
...
+
+    with torch._dynamo.utils.preserve_rng_state():
+        exp_ref = inp.exponential_()
+    exp = torch._refs.exponential(inp)
+
+    self.assertEqual(exp, exp_ref)
+    self.assertFalse(exp.isinf().any())
+
+
+DecompOneOffTests.test_exponential_non_inf = _test_exponential_non_inf
```
Silv3S:
```suggestion
import inspect
DecompOneOffTests.test_exponential_non_inf = inspect.unwrap(DecompOneOffTests.test_exponential_non_inf)
```
As test body is already generic we can just drop only_cuda decorator and use upstream test directly

tszulist-hbn (approved):

test/xpu/test_decomp_xpu.py
pbielak: Exactly

test/xpu/test_decomp_xpu.py
pbielak: I will leave a comment, so when upstreaming we need to refactor this table to be device-specific

pbielak:

PR #4270, opened 2026-07-08
Use if constexpr for compile-time-constant branches 

cyyever:
Converts runtime `if` checks on compile-time-constant template parameters to `if constexpr`, so the dead branch is discarded at compile time instead of emitted.

- `FusedSgdKernels.cpp`: `depth > 2`
- `LogAddExpKernels.cpp`: `min`
- `LogcumsumexpKernel.cpp`: `min`
- `MaxUnpoolingKernels.cpp`: `is_channels_last_`
- `SortingRadixSelect.h`: `Order`


guangyey (approved):
Thanks!

PR #4276, opened 2026-07-08
Replace NumericLimits.h with upstream generic version

cyyever:
The XPU `NumericLimits.h` was a stale hand-written copy (11 explicit specializations, macros, `from_bits`). Replace it with the upstream generic version.


guangyey (approved):
Good cleanup!

PR #4280, opened 2026-07-08
Simplify map_options via c10::CppTypeToScalarType

cyyever:
c10::CppTypeToScalarType can replace the if-else chain in map_options.


guangyey (approved):
Nice!

PR #4285, opened 2026-07-09
Use aminmax and align index_put with upstream CUDA

cyyever:
Aligns XPU summary/indexing kernels with upstream CUDA where XPU lagged:

- SummaryOps (histc/bincount): `.min()+.max()` -> `aminmax()`; `int64_t`
  instead of `ScalarTypeToCPPType<kLong>::type`; bincount dim/negativity
  check as `TORCH_CHECK` + `if constexpr`. (matches `cuda/SummaryOps.cu`)
- IndexingUtils `wrapIndexOnce`: `at::aminmax` + `at::_assert_async` (no
  host sync) instead of `.item()`+`TORCH_CHECK_INDEX`; `std::move`
  `makeLinearIndex`'s tuple args. (matches `cuda/Indexing.cu`)

Behavior change: `wrapIndexOnce` out-of-bounds indices now report via an
async, generic device assert instead of an immediate, detailed host-side
error. Indices are still wrapped via `remainder(dim_size)`, so there is no
out-of-bounds memory access — only the diagnostic UX changes. This matches
upstream CUDA's trade-off.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ATen/native/xpu/sycl/IndexingUtils.h
+++ b/src/ATen/native/xpu/sycl/IndexingUtils.h
@@ -25,8 +25,9 @@ static Tensor wrapIndexOnce(
   // we don't need to check range in backward - if there were out of bounds
   // indices forward should already have errored out
   if (index.numel() != 0 && check_range) {
-    auto max_idx = index.max().item<int64_t>();
-    auto min_idx = index.min().item<int64_t>();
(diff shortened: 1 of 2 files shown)


guangyey (approved):
Thanks!

PR #4286, opened 2026-07-09
Register and implement _spsolve for XPU backend

### LINKED ISSUE #4148: Register _spsolve for XPU backend

Register and implement **_spsolve** on XPU.

### Native function fragment (from native_functions.yaml)
```yaml
- func: _spsolve(Tensor A, Tensor B, *, bool left=True) -> Tensor
  python_module: sparse
  dispatch:
    SparseCsrCUDA: _sparse_csr_linear_solve
```

### Goals
- Register the above native signature for the XPU backend and add proper dispatch mapping
- Implement the XPU kernel for `_spsolve` (or provide an efficient device-side fallback if native kernel is not feasible)
- Ensure unit tests covering:
  - Sparse CSR input matrices with various sparsity patterns
  - Both `left=True` and `left=False` cases
  - Edge cases (singular matrices, empty inputs, identity matrices)
  - Correctness against CPU/CUDA reference results

### Alternatives

_No response_

### Additional context

_No response_

pponikox:
This patch fixes #4148 by implementing `torch.sparse.spsolve` (`aten::_spsolve`) for the SparseCsrXPU backend. Input validation and supported cases mirror the CUDA (cuDSS) backend: square A, 1-D b, and left=true only (consistent with the documented torch.sparse.spsolve API). left=False, non-1-D b, and size mismatches raise the same errors.

#### Summary

`_spsolve` solves a square linear system `A @ x = b` where `A` is a sparse CSR matrix and `b` is a dense vector.
The CUDA backend uses NVIDIA's cuDSS direct sparse solver, which has no equivalent on XPU. Following the existing pattern for sparse-CSR XPU ops in the same file (triangular_solve, addmv, baddbmm, which all materialize A.to_dense() and delegate to a dense op), the newly added `_sparse_csr_linear_solve_xpu` performs an on-device dense solve via `at::linalg_solve`.

**This PR requires the companion PyTorch change that registers the XPU dispatch keys in `native_functions.yaml` which is done in: https://github.com/pytorch/pytorch/pull/189372.**

#### Testing
Due to the fact that any new tests for `_spsolve` would fail on XPU build until registration PR (https://github.com/pytorch/pytorch/pull/189372) is merged, a decision was made to move ALL new tests to a separate PR (https://github.com/intel/torch-xpu-ops/pull/5076) which will be merged at the end (after this and registration PR). They were previously checked and confirmed passing for XPU and skipping for CPU.
Thanks to that there should be no blockers nor build failures related to adding `_spsolve` implementation.


src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
```diff
@@ -514,4 +515,29 @@ std::tuple<Tensor&, Tensor&> triangular_solve_out_sparse_csr_xpu(
...
+  TORCH_CHECK(b.layout() == kStrided, "b must be a strided tensor");
+  // dim check
+  TORCH_CHECK(b.dim() == 1, "b must be a 1D tensor");
+  TORCH_CHECK(b.size(0) == A.size(0), "linear system size mismatch.");
+  TORCH_CHECK(A.size(0) == A.size(1), "A must be a square matrix");
+  TORCH_CHECK(A.dtype() == b.dtype(), "A and b must have the same dtype");
+  TORCH_CHECK(
+      left == true, "only left == true is supported by the Sparse CSR backend");
+
+  Tensor b_copy = b.contiguous();
```
BBBela: Is this enforcing contiguity on `b` necessary here? I mean - doesn't `at::linalg_solve()` alredy handle that internally? Not a problem, just asking to avoid unnecessarily calling the same here and in  `at::linalg_solve()`. Nothing blocking - just wanted to ask 😉

test/xpu/test_sparse_csr_xpu.py
BBBela, line 1: I have one general question regarding those added tests - aren't they duplicated with some already existing tests in upstream? I mean, the `_spsolve` operator is already registered for CUDA, so I suppose that tests are also added in the upstream PyTorch.

Could you please check if the  `_spsolve` is already tested in upstream or not? If it is, do those newly added tests perform more comprehensive checks?

BBBela (commented):
In general the solution looks good to me - I just have some considerations regarding added tests in general. Please look below 😉

src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
```diff
@@ -514,4 +515,29 @@ std::tuple<Tensor&, Tensor&> triangular_solve_out_sparse_csr_xpu(
...
+  TORCH_CHECK(A.is_sparse_csr(), "A must be a CSR matrix");
+  TORCH_CHECK(b.layout() == kStrided, "b must be a strided tensor");
+  // dim check
+  TORCH_CHECK(b.dim() == 1, "b must be a 1D tensor");
+  TORCH_CHECK(b.size(0) == A.size(0), "linear system size mismatch.");
+  TORCH_CHECK(A.size(0) == A.size(1), "A must be a square matrix");
+  TORCH_CHECK(A.dtype() == b.dtype(), "A and b must have the same dtype");
+  TORCH_CHECK(
+      left == true, "only left == true is supported by the Sparse CSR backend");
+
```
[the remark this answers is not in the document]
pponikox: Updated to mirror CUDA semantics more closely.

src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
pponikox: You're right.
I checked that CUDA calls `b.contiguous()` because it hands a raw pointer to `cuDSS`, which requires a packed column-major buffer. Our XPU fallback routes through `at::linalg_solve()`, which normalizes/handles non-contiguous inputs internally, so the extra copy was unnecessary. Thus I've updated this code and pass `b` directly.

test/xpu/test_sparse_csr_xpu.py
pponikox: From what I've found out upstream has exactly one `_spsolve` test: `test_linalg_solve_sparse_csr_cusolver` in `test_sparse_csr.py`:
* It is gated by @unittest.skipIf(not TEST_CUDA_CUDSS, "The test requires cudss") and further self-skips = it only ever runs on CUDA with cuDSS — it will never test the XPU backend.
* Its coverage is narrower: it iterates sample_inputs_linalg_solve, compares against dense torch.linalg.solve, and checks two error paths ("b must be a 1D tensor" and the empty-other case).

The new XPU tests IMO are more comprehensive:
* `test_spsolve_xpu`: multiple structured matrices (identity, 1×1, SPD, larger SPD, diagonal, tridiagonal), a residual check (A @ x ≈ b), a CPU cross-reference, and an empty (0×0) system.
* `test_spsolve_xpu_errors`: left=False, non-1D b, size mismatch, and singular-matrix (LinAlgError) — none of which the upstream test covers.

So even though there's conceptual overlap (both validate against a dense solve), but no real duplication: the upstream test can't cover XPU, and the new tests check materially more behavior. 
IMO we could leave the newly added tests as-is :wink:

BBBela: Thank you for this detailed analysis!
I see your point and it makes sense to add those additional tests, although I have several considerations:
- Currently if we add those test cases, they will fail consistently because they need the upstream registration PR to be merged.
- I am not sure if we should add those xpu-only tests to the main test file or somewhere else.

@CuiYifeng, @guangyey - could you please help with advice regarding those two topics?
1. If we add test cases in this PR, they will fail consistently until the upstream registration is added [HERE](https://github.com/pytorch/pytorch/pull/189372). Should we first merge this PR without test cases, then merge the upstream registration and then add additional tests or what is the preferable way?
2. Where should we put the additional xpu-only tests? To the main `test_sparse_csr_xpu.py` test file or to `regressions/` directory for now? And should we add them locally or directly in upstream along with the registration PR [HERE](https://github.com/pytorch/pytorch/pull/189372)

BBBela (commented):
Added comment in the previous thread https://github.com/intel/torch-xpu-ops/pull/4286#discussion_r3795376325

test/xpu/test_sparse_csr_xpu.py
pponikox: Oh, I understand your concern now. You're right to point out that the upstream PR needs to be merged first in order to tests to pass. On the other hand we need this PR first so the upstream has something to register so we're in a weird cyclic dependency 😉 IMO it would be best to skip those tests for now and add them as a separate PR when op implementation and registration are merged.
However let's wait for opinion from @CuiYifeng, @guangyey

guangyey: We can follow this flow:

1. Land the implementation first and temporarily skip the UT. Attach a test snapshot as evidence that the UT passes.
2. Register the op in `native_functions.yaml`.
3. Submit a separate PR to re-enable the skipped UT.

CuiYifeng: Agree with Guangye's proposal.

pponikox: Ok, so I'll follow that plan. I double checked locally if the tests are not failing with the newest main branch versions for both PyTorch (with spsolve registration) and XPU (with current PR diff):
image
Since everything is still looking good then I also decided to run the On-demand CI build - [LINK](https://github.com/intel/torch-xpu-ops/actions/runs/32460258583) - where the same TestSparseCSR tests also passed for XPU and were skipped for CPU: 
```
2026-08-21T14:14:19.5554827Z test_sparse_csr_xpu.py::TestSparseCSRCPU::test_spsolve_xpu_cpu_float32 
2026-08-21T14:14:19.5632695Z [gw1] [ 18%] SKIPPED test_sparse_csr_xpu.py::TestSparseCSRCPU::test_spsolve_xpu_cpu_float32 
...
2026-08-21T14:14:19.5974497Z test_sparse_csr_xpu.py::TestSparseCSRCPU::test_spsolve_xpu_cpu_float64 
2026-08-21T14:14:19.6074252Z [gw1] [ 18%] SKIPPED test_sparse_csr_xpu.py::TestSparseCSRCPU::test_spsolve_xpu_cpu_float64 
...
2026-08-21T14:14:19.6389666Z test_sparse_csr_xpu.py::TestSparseCSRCPU::test_spsolve_xpu_errors_cpu_float32 
2026-08-21T14:14:19.6654515Z [gw1] [ 18%] SKIPPED test_sparse_csr_xpu.py::TestSparseCSRCPU::test_spsolve_xpu_errors_cpu_float32 
...
2026-08-21T14:14:19.6796148Z test_sparse_csr_xpu.py::TestSparseCSRCPU::test_spsolve_xpu_errors_cpu_float64 
2026-08-21T14:14:19.7020418Z [gw1] [ 18%] SKIPPED test_sparse_csr_xpu.py::TestSparseCSRCPU::test_spsolve_xpu_errors_cpu_float64 
...
2026-08-21T14:16:07.7248108Z test_sparse_csr_xpu.py::TestSparseCSRXPU::test_spsolve_xpu_errors_xpu_float32 
2026-08-21T14:16:07.7326775Z [gw2] [ 86%] PASSED test_sparse_csr_xpu.py::TestSparseCSRXPU::test_spsolve_xpu_errors_xpu_float32 
...
2026-08-21T14:16:07.7453266Z test_sparse_csr_xpu.py::TestSparseCSRXPU::test_spsolve_xpu_errors_xpu_float64
2026-08-21T14:16:07.7677665Z [gw2] [ 86%] PASSED test_sparse_csr_xpu.py::TestSparseCSRXPU::test_spsolve_xpu_errors_xpu_float64 
...
2026-08-21T14:16:07.7754840Z test_sparse_csr_xpu.py::TestSparseCSRXPU::test_spsolve_xpu_xpu_float32 
2026-08-21T14:16:07.8187476Z [gw2] [ 86%] PASSED test_sparse_csr_xpu.py::TestSparseCSRXPU::test_spsolve_xpu_xpu_float32 
...
2026-08-21T14:16:07.8249517Z test_sparse_csr_xpu.py::TestSparseCSRXPU::test_spsolve_xpu_xpu_float64 
2026-08-21T14:16:07.8897805Z [gw2] [ 86%] PASSED test_sparse_csr_xpu.py::TestSparseCSRXPU::test_spsolve_xpu_xpu_float64 
```
That being said I'm updating this PR to contain only production code changes and move tests addition to a separate PR.

pponikox: Done:
* This PR contains only logic for implementing _spsolve for XPU backend (description updated).
* All unit tests were moved to a separate PR: https://github.com/intel/torch-xpu-ops/pull/5076
* The unit tests PR is in draft. It will be marked as ready for review when this patch and the op registration in PyTorch are both merged.

I'm resolving this convo.
Please re-review this PR and approve if current solution is acceptable 😉

guangyey (approved):
Defer to @CuiYifeng

src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
```diff
@@ -514,4 +515,28 @@ std::tuple<Tensor&, Tensor&> triangular_solve_out_sparse_csr_xpu(
...
+// Solves the linear system A @ x = b, where A is a sparse CSR matrix and b is a
+// dense right-hand side. This mirrors the semantics of the CUDA (cuDSS) backend
+// in aten/src/ATen/native/sparse/cuda/SparseCsrTensorMath.cu. XPU has no direct
+// sparse solver available, so we fall back to a dense solve that still runs
+// entirely on the device.
+Tensor _sparse_csr_linear_solve_xpu(
+    const Tensor& A,
+    const Tensor& b,
+    const bool left) {
+  // layout check
```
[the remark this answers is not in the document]
pponikox: This is expected. Unit tests were moved to a separate PR: https://github.com/intel/torch-xpu-ops/pull/5076 in order to avoid build errors caused by the fact that this op wouldn't be registered in PyTorch (also done in a separate PR: https://github.com/pytorch/pytorch/pull/189372).

BBBela (approved):
Looks good to me now, thanks!

src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp
```diff
@@ -514,4 +515,28 @@ std::tuple<Tensor&, Tensor&> triangular_solve_out_sparse_csr_xpu(
...
+  TORCH_CHECK(A.is_sparse_csr(), "A must be a CSR matrix");
+  TORCH_CHECK(b.layout() == kStrided, "b must be a strided tensor");
+  // dim check
+  TORCH_CHECK(b.dim() == 1, "b must be a 1D tensor");
+  TORCH_CHECK(b.size(0) == A.size(0), "linear system size mismatch.");
+  TORCH_CHECK(b.size(0) == A.size(1), "linear system size mismatch.");
+  TORCH_CHECK(A.dtype() == b.dtype(), "A, x, and b must have the same dtype");
+  TORCH_CHECK(
+      left == true, "only left == true is supported by the Sparse CSR backend");
```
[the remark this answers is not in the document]
pponikox: Done.

CuiYifeng (approved):
LGTM.

pponikox:
Unit test failures in CI are NOT related to my changes in this PR:
```
op_ut,third_party.torch-xpu-ops.test.xpu.profiler.test_profiler_xpu.TestProfiler,test_cpu_annotation_overlap
op_ut,,third_party.torch-xpu-ops.test.xpu.test_dataloader_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_serialization_xpu.TestSerialization,test_serialization_backwards_compat
```
Thus I'm merging this PR.

pponikox:
@torchxpubot merge -f

Issue #3546, opened 2026-04-30
[split from #3093] NestedTensor SDPA on XPU has no viable backend (jagged path)


tszulist-hbn:
### 🐛 Describe the bug

test_cases:
  - op_ut,third_party.torch-xpu-ops.test.xpu.test_nestedtensor_xpu.TestNestedTensorSubclassXPU,test_sdpa_with_packed_in_proj_xpu_bfloat16
---

## Parent Issue

Split from #3093 — XPU does not support NestedTensor for SDPA operations.

## Affected Test Cases

- `test_nestedtensor_xpu.py::TestNestedTensorSubclassXPU::test_sdpa_with_packed_in_proj_xpu_bfloat16`

(The closely-related `test_dummy_mha_with_nt_use_legacy_api_{True,False}_xpu` cases mentioned in the parent issue are already skipped on XPU since torch-xpu-ops commit `7625ae9b` via `@skipXPUIf(True, "XPU does not support NestedTensor for SDPA operations.")`. They no longer reproduce.)

## Root Cause

`F.scaled_dot_product_attention` on jagged-NestedTensor inputs is dispatched to `jagged_scaled_dot_product_attention` in `torch/nested/_internal/sdpa.py`, which calls `_select_sdp_backend` (line 293). On XPU all four candidate backends are rejected:

| Backend | Why it fails on XPU |
|---|---|
| `CUDNN_ATTENTION` | `torch._C._can_use_cudnn_attention` is guarded `#ifdef USE_CUDA` only in `torch/csrc/Module.cpp` (always returns `False` on XPU). |
| `FLASH_ATTENTION` | `aten/src/ATen/native/transformers/xpu/sdp_utils.cpp::can_use_flash_attention` includes `check_nested_tensor` (in `sdp_utils_cpp.h`), which unconditionally rejects nested inputs. CUDA's version splits constraints into a separate nested branch and accepts them. |
| `EFFICIENT_ATTENTION` | `torch._C._can_use_mem_efficient_attention` is guarded `#ifdef USE_CUDA` only. |
| `MATH` | This specific test produces non-contiguous q/k/v (`# these are non-contiguous to trigger _is_safe_to_get_storage_as_tensor()`); `_can_use_math_sdpa_jagged` rejects them via the `transpose(1,2).is_contiguous()` check. |

[...]

Issue #3140, opened 2026-03-24
[upstream_ut]  RuntimeError: FlashAttentionForwardXPU does not only support dropout > 0.0 yet 



daisyden:
Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_transformers_xpu.TestSDPACudaOnlyXPU,test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_1_seq_len_q_256_seq_len_k_256_head_dim_64_is_causal_True_dropout_p_0_22_float16_scale0_fused_kernel0_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_transformers_xpu.TestSDPACudaOnlyXPU,test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_8_seq_len_q_256_seq_len_k_256_head_dim_32_is_causal_False_dropout_p_0_22_float16_scale0_fused_kernel0_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_transformers_xpu.TestSDPACudaOnlyXPU,test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_1_seq_len_q_1024_seq_len_k_256_head_dim_32_is_causal_False_dropout_p_0_22_float16_scale_l1_fused_kernel0_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_transformers_xpu.TestSDPACudaOnlyXPU,test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_8_seq_len_q_256_seq_len_k_256_head_dim_32_is_causal_False_dropout_p_0_22_float16_scale_l1_fused_kernel0_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_transformers_xpu.TestSDPACudaOnlyXPU,test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_8_seq_len_q_256_seq_len_k_256_head_dim_64_is_causal_True_dropout_p_0_22_float16_scale0_fused_kernel0_xpu_float16
op_ut,third_party.torch-xpu-ops.test.xpu.test_transformers_xpu.TestSDPACudaOnlyXPU,test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_1_seq_len_q_1024_seq_len_k_1024_head_dim_64_is_causal_True_dropout_p_0_22_float16_scale0_fused_kernel0_xpu_float16
[...]

PR #4289, opened 2026-07-09
Migrate test_nestedtensor_xpu.py to use `XPUImportCtx`

tszulist-hbn:
Replace the full copy of upstream test/test_nestedtensor.py with a wrapper that imports upstream test classes via  `XPUImportCtx`. The previous version was a 9,157-line copy requiring manual syncing with upstream. This wrapper imports five upstream test classes and instantiates them for XPU using `XPUImportCtx` for minimal import patching.

Test results:
Before: 3 failed, 1404 passed, 439 skipped (1846 tests)
After:  3 failed, 1415 passed, 435 skipped (1853 tests)
\-----------------------------------------------------------
Delta:  +11 passed, -4 skipped (+7 tests)

Remaining failures (3, unchanged from before - same tests, not fixed by this PR):
```text
TestNestedTensorSubclassXPU::test_sdpa_compile_xpu_bfloat16
TestNestedTensorSubclassXPU::test_profiler_sequence_nr_xpu
TestNestedTensorSubclassXPU::test_sdpa_xpu_bfloat16
```

New passing tests (11):

New tests exercised for the first time (6):
```text
TestNestedTensorDeviceTypeXPU::test_jagged_max_extreme_values_xpu_int64
TestNestedTensorDeviceTypeXPU::test_jagged_min_extreme_values_xpu_int64
TestNestedTensorSubclassXPU::test_compile_jagged_mean_omits_outer_size_stride_guards_xpu_float32
TestNestedTensorSubclassXPU::test_torch_function_metadata_fast_path_exact_callables_xpu_float32
TestNestedTensorSubclassXPU::test_where_scalar_broadcast_on_in_graph_constructed_njt_scalar_input_other_xpu_float3
TestNestedTensorSubclassXPU::test_where_scalar_broadcast_on_in_graph_constructed_njt_scalar_input_self_xpu_float32
```

Previously skipped, now passing (5):
```text
TestNestedTensorSubclassXPU::test_compile_padded_dense_conversion_preserves_metadata_cache_xpu_float32
TestNestedTensorSubclassXPU::test_compile_preserves_metadata_cache_xpu_float32
TestNestedTensorSubclassXPU::test_compile_with_dynamic_max_seq_len_xpu_float32
TestNestedTensorSubclassXPU::test_compile_with_dynamic_min_seq_len_xpu_float32
TestNestedTensorSubclassXPU::test_compile_with_propagated_dynamic_max_seq_len_xpu_float32
```

Renamed skipped tests
- `test_sdpa_with_packed_in_proj`:

This test is unconditionally skipped on XPU (`skipIfXpu`), and this PR drops the previous `dtypesIfXPU(torch.bfloat16)` override so the test is overridden in a single place. Consequently the skipped variants change name, because XPU now inherits the upstream dtype list. Upstream gates it on `SM80OrLater`:

```python
@dtypes(*([torch.float16, torch.bfloat16, torch.float32] if SM80OrLater
          else [torch.float16, torch.float32]))
```

On an XPU host (no CUDA device) `SM80OrLater` is `False`, so the list resolves to `float16, float32`. No coverage changes - the test was skipped before and is skipped now.

Removed (1):
```text
TestNestedTensorSubclassXPU::test_sdpa_with_packed_in_proj_xpu_bfloat16
```

Added (2):
```text
TestNestedTensorSubclassXPU::test_sdpa_with_packed_in_proj_xpu_float16
TestNestedTensorSubclassXPU::test_sdpa_with_packed_in_proj_xpu_float32
```


test/xpu/test_nestedtensor_xpu.py
```diff
@@ -5,9139 +5,68 @@
...
-            (2, 3, 6, 7), device=device, dtype=torch.float32
-        )
-        scalar = 0.0
+# Re-export CUDA-only tests to run on both CUDA and XPU
+# These tests have @onlyCUDA in upstream but work on XPU
+def _unwrap(fn):
+    """Unwrap decorated function to remove @onlyCUDA."""
+    while hasattr(fn, "__wrapped__"):
+        fn = fn.__wrapped__
+    return fn
```
BBBela: I am not sure about this `_unwrap` function. It removes all the `@wraps` layers, effectively removing not only the `@onlyCUDA` decorator, but potentially much more. I think we should simply keep the local copies of those test functions that have `@onlyCUDA` decorator in the upstream but they pass on XPU and should be launched on XPU. And the desired final solution is to change those `@onlyCUDA` decorator to `@onlyOn(["cuda", "xpu"])` in the upstream directly.

BBBela (commented):
I have several questions/topics below:

1. Why was `TestNestedInt` test class dropped entirely? Are the statistics in the description up to date with current version of PR? Because it looks like the 2 test cases from `TestNestedInt` were lost during migration and I don't see any comment saying that it is known and desired solution.
2. I am worried about so many test cases that were previously passing and now they are skipped. I think we should keep local copies of those not to loose coverage, make necessary changes upstream and only then remove the local copies.
3. Was it investigated why `test_sdpa_xpu_float16 ` started failing? It doesn't seem right that one test started failing and we just leave it in that state? I think it should be investigated why it started failing first, then find the cause. If that is valid that it started failing, then at least some issue should be added to track this.

test/xpu/test_nestedtensor_xpu.py
```diff
@@ -5,9138 +5,283 @@
...
-        answer = torch.tensor(100.0, device=device, dtype=dtype).expand((2, 5))
-        self.assertEqual(nt[0], answer)
-        nt[1, 1, :].fill_(200.0)
-        answer = torch.tensor(200.0, device=device, dtype=dtype).expand(4)
-        self.assertEqual(nt[1, 1, :], answer)
+TestNestedTensorDeviceType.test_nested_tensor_dense_elementwise = (
+    _test_nested_tensor_dense_elementwise
+)
+TestNestedTensorDeviceType.test_bmm_cuda = _test_bmm_cuda
+TestNestedTensorDeviceType.test_layer_norm.dtypes = {
```
pbielak: For these `dtypes` overrides, please use the `dtypesIfXPU` decorator here (also use the same names as in the previous file version, i.e. `torch.float, torch.half`)

test/xpu/test_nestedtensor_xpu.py
```diff
@@ -5,9138 +5,283 @@
...
-        self.assertEqual(nt.select(2, 0)[1], x1.select(1, 0))
-        self.assertRaises(IndexError, lambda: nt.select(2, 5))
-        # make sure indexing returns a view
-        nt[0].fill_(100.0)
-        answer = torch.tensor(100.0, device=device, dtype=dtype).expand((2, 5))
-        self.assertEqual(nt[0], answer)
-        nt[1, 1, :].fill_(200.0)
-        answer = torch.tensor(200.0, device=device, dtype=dtype).expand(4)
-        self.assertEqual(nt[1, 1, :], answer)
+TestNestedTensorDeviceType.test_nested_tensor_dense_elementwise = (
```
pbielak: Set these overrides directly after the function definition

test/xpu/test_nestedtensor_xpu.py
tszulist-hbn: Done

test/xpu/test_nestedtensor_xpu.py
tszulist-hbn: Done

test/xpu/test_nestedtensor_xpu.py
tszulist-hbn: Done

BBBela:
Hi, thanks for the updates!

Before I do another review pass, could you please address the questions I raised in my previous comment ([COMMENT](https://github.com/intel/torch-xpu-ops/pull/4289#pullrequestreview-4670308571))? I want to make sure we're aligned on those points first.

Thanks! 😄

tszulist-hbn:
**1. `TestNestedInt` dropped** — Good catch, my miss. It's a plain `TestCase` (not device-typed), so it slipped through the wrapper conversion. Restored it: added it to the import block + `instantiate_parametrized_tests(TestNestedInt)` (a bare import gets stripped as unused by the linter). Both tests pass again.

**2. Passing → skipped tests** — I checked each `skipIfXpu` against the old copy, and none of them were actually passing on XPU before:

| Test | Old copy on XPU |
|---|---|
| `test_record_stream` | `@onlyCUDA` — never ran |
| `test_sdpa_with_packed_in_proj` | `@skipXPUIf(True)` — already skipped |
| `test_sdpa_flop_counter` | `@skipXPUIf(True)` — already skipped |
| `test_dummy_mha_with_nt` | `@skipXPUIf(True)` — already skipped |
| `test_sdpa_compile` | ran but was failing → now skipped |

So the `skipIfXpu` calls just carry over skips that already existed inline in the deleted file — no new coverage lost. Totally agree with the principle in general though. The only real change is `test_sdpa_compile` (failed → skipped) — happy to file an issue to keep it visible if you'd rather.

**3. `test_sdpa_xpu_float16` failing** — That variant doesn't exist; `test_sdpa` only runs bfloat16 on XPU (both before and after). The real remaining failure is `test_sdpa_xpu_bfloat16`, and it's `FAILED` in both the before and after runs — pre-existing, not introduced here (same for `test_profiler_sequence_nr_xpu`). Net effect is -1 failed. I'll open tracking issues for those two so they don't get forgotten.

test/xpu/test_nestedtensor_xpu.py
```diff
@@ -5,9140 +5,293 @@
...
-        def grad_test_func(values_1, values_2, offsets):
-            nt_1 = torch.nested.nested_tensor_from_jagged(values_1, offsets)
-            nt_2 = torch.nested.nested_tensor_from_jagged(values_2, offsets)
-            nt_3 = torch.cat([nt_1, nt_2], dim=-1)
-            return nt_3.values()
-
-        assert gradcheck(
-            grad_test_func,
-            inputs=(values_1, values_2, offsets),
-            check_batched_grad=Fal
```
BBBela: Shouldn't we use the `dtypesIfXPU` decorator here also? As @pbielak suggested? I see that it was changed already in one place from `.dtypes` to `dtypesIfXPU` - right?

BBBela (commented):
Thank you a lot for this detailed explanation. Now everything is clear to me!

It basically looks good to me now, but please look below at one minor comment - just wanted to make sure if that is intentional or some oversight?

test/xpu/test_nestedtensor_xpu.py
BBBela: I don't know why GitHub shows different code fragment in the `Conversation` tab and different in the `Files changed` tab - at least for me 😮 I meant the lines `+275` to `+284`, not the `-275` to `-284`

tszulist-hbn: Good point, agreed. I'll switch these to `dtypesIfXPU` for consistency with `test_layer_norm` above and to avoid poking `.dtypes` directly. Note `test_sdpa_compile` and `test_sdpa_with_packed_in_proj` are `skipIfXpu`, so their dtype lines don't actually do anything on XPU. I'll convert `test_sdpa` to `dtypesIfXPU(torch.bfloat16)` and drop the redundant lines on the two skipped ones.

tszulist-hbn: Done

BBBela: I am still wondering why was `test_sdpa_compile` skipped in this PR?
From what I understand from the PR description it was failing and now is skipped. Why? I mean if there is a problem or we don't know the problem, then we should leave it failing, right?

BBBela (commented):
I have one additional general thought (not introduced by this PR though, but related to this test file) - regarding the tests decorated with `@dtypesIfXPU(torch.bfloat16)` - is that correct? I mean it seems a little weird that all the backends test it against `float16` and `float32` and on XPU we test it only against `bfloat16`? CUDA does the opposite - all the previous generations are tested with `float16` and `float32` and the newer are additionally tested with `bfloat16`. Is it true that XPU doesn't support `float16` and `float32` for those tests? Because I suspect that the initial idea was to extend the dtypes tested on XPU versus default, not to limit dtypes to only `bfloat16`. Have you maybe investigated that or do you have some more knowledge on that topic than me?

pbielak:
> **1. `TestNestedInt` dropped** — Good catch, my miss. It's a plain `TestCase` (not device-typed), so it slipped through the wrapper conversion. Restored it: added it to the import block + `instantiate_parametrized_tests(TestNestedInt)` (a bare import gets stripped as unused by the linter). Both tests pass again.
> 
> **2. Passing → skipped tests** — I checked each `skipIfXpu` against the old copy, and none of them were actually passing on XPU before:
> 
> Test	Old copy on XPU
> `test_record_stream`	`@onlyCUDA` — never ran
> `test_sdpa_with_packed_in_proj`	`@skipXPUIf(True)` — already skipped
> `test_sdpa_flop_counter`	`@skipXPUIf(True)` — already skipped
> `test_dummy_mha_with_nt`	`@skipXPUIf(True)` — already skipped
> `test_sdpa_compile`	ran but was failing → now skipped
> So the `skipIfXpu` calls just carry over skips that already existed inline in the deleted file — no new coverage lost. Totally agree with the principle in general though. The only real change is `test_sdpa_compile` (failed → skipped) — happy to file an issue to keep it visible if you'd rather.
> 
> **3. `test_sdpa_xpu_float16` failing** — That variant doesn't exist; `test_sdpa` only runs bfloat16 on XPU (both before and after). The real remaining failure is `test_sdpa_xpu_bfloat16`, and it's `FAILED` in both the before and after runs — pre-existing, not introduced here (same for `test_profiler_sequence_nr_xpu`). Net effect is -1 failed. I'll open tracking issues for those two so they don't get forgotten.

@tszulist-hbn If a class was not wrapped by an `instatiate_*(...)` before, we shouldn't introduce it now. If the import gets removed by the lintrunner, we can add `# ruff: noqa: F401` at the beginning of the file (right after the copyright header) - this should solve "unused imports"

test/xpu/test_nestedtensor_xpu.py
```diff
@@ -5,9139 +5,246 @@
...
-        for chunk in chunked:
-            self.assertFalse(chunk.is_contiguous())
+@dtypes(torch.bfloat16)
+def _test_sdpa_compile_xpu(self, device, dtype):
+    bypass_device_restrictions = self.bypass_device_restrictions
+    self.bypass_device_restrictions = True
+    try:
+        return _test_sdpa_compile(self, device, dtype)
+    finally:
+        self.bypass_device_restrictions = bypass_device_restrictions
```
BBBela: I think that is overcomplicating things. In my opinion for such case we should simply copy-paste the whole test function and change the necessary decorators instead of using such workarounds.

test/xpu/test_nestedtensor_xpu.py
```diff
@@ -5,9139 +5,246 @@
...
+# @skipIfTorchDynamo are both dropped; skipIfTorchDynamo is re-applied below.
+# The xfailIfWindows guard is intentionally not restored: it is Windows-only and
+# fires only for float32, while XPU runs bfloat16 (via dtypesIfXPU) exclusively.
+TestNestedTensorSubclass.test_sdpa_backwards = skipIfTorchDynamo(
+    "SDPA test compiles internally"
+)(
+    skipCUDAIf(not SM70OrLater, "GPU capability is < SM70")(
+        onlyOn(["cuda", "xpu"])(
+            dtypesIfXPU(torch.bfloat16)(
+                inspect.unwrap(TestNestedTensorSubclass.test_sdpa_backwards)
```
BBBela: I think we should not overcomplicate this. My opinion is that we should simply make the copy-paste of this test function and change only the necessary decorators.

test/xpu/test_nestedtensor_xpu.py
```diff
@@ -5,9139 +5,246 @@
...
-
-        # TODO: Use this approach when unbind is functional
-        # unbinded_nt = nt.unbind()
-        # for i in range(starts.shape[0]):
-        #     self.assertEqual(torch.arange(starts[i], starts[i] + lengths[i], device=device, dtype=torch.int64), unbinded_nt[i])
-        for i in range(starts.shape[0]):
-            self.assertEqual(
-                torch.arange(
-                    starts[i], starts[i] + lengths[i], device=device, dtype=torch.int64
-
```
BBBela: Why do we need to set the dtypes for XPU if we eventually skip this test for all the dtypes?

test/xpu/test_nestedtensor_xpu.py
```diff
@@ -5,9139 +5,246 @@
...
-# Helper function for test_dummy_mha_with_nt
-@torch.fx.wrap
-def convert_dense_to_nested_tensor_legacy(values):
-    offsets = torch.arange(
-        0, values.shape[0] * values.shape[1] + 1, values.shape[1], device=values.device
+with XPUImportCtx(False):
+    from test_nestedtensor import (
+        convert_jagged_to_nested_tensor,
+        convert_nt_to_jagged,
+        random_nt_noncontiguous_pair,
```
BBBela:
```suggestion
```
It looks like those imports are currently unused.

BBBela (commented):
I re-raise my previous comment from here: [COMMENT](https://github.com/intel/torch-xpu-ops/pull/4289#pullrequestreview-4685659472)

Regarding the PR description - there are 5 tests listed and described as "No longer failing tests", yet the table says that there were 3 failing test cases previously and there are still 3 failing test cases now. That doesn't sum up - could you please verify if the description is correct and up to date with the real state?

test/xpu/test_nestedtensor_xpu.py
tszulist-hbn: Done

test/xpu/test_nestedtensor_xpu.py
tszulist-hbn: Done

test/xpu/test_nestedtensor_xpu.py
tszulist-hbn: Done

test/xpu/test_nestedtensor_xpu.py
tszulist-hbn: Done

AKloniecki (approved):

tszulist-hbn:
> I have one additional general thought (not introduced by this PR though, but related to this test file) - regarding the tests decorated with `@dtypesIfXPU(torch.bfloat16)` - is that correct? I mean it seems a little weird that all the backends test it against `float16` and `float32` and on XPU we test it only against `bfloat16`? CUDA does the opposite - all the previous generations are tested with `float16` and `float32` and the newer are additionally tested with `bfloat16`. Is it true that XPU doesn't support `float16` and `float32` for those tests? Because I suspect that the initial idea was to extend the dtypes tested on XPU versus default, not to limit dtypes to only `bfloat16`. Have you maybe investigated that or do you have some more knowledge on that topic than me?

@BBBela. For `test_sdpa_with_packed_in_proj` specifically: this test is wrapped in `skipIfXpu(...)`, so it never actually executes on XPU regardless of dtype. `dtypesIfXPU(torch.bfloat16)` here is purely a test-collection trick to keep a single skipped test id (`..._xpu_bfloat16`) instead of three (float16/bfloat16/float32), matching what was previously skipped. It has no effect on real coverage since the body never runs.

For the other `dtypesIfXPU(torch.bfloat16)` usages in this file (e.g. `test_sdpa`, and a few tests in TestNestedTensorDeviceType), your concern is valid: those tests do run on XPU, and `dtypesIfXPU` replaces rather than extends the upstream dtype list, so XPU only gets bfloat16 coverage while CUDA gets float16/float32 (+bfloat16 on newer GPUs). That's pre-existing behavior unrelated to this PR's `XPUImportCtx` refactor, so I'd rather not change it here to keep this change surgical. I'll follow up separately to check whether XPU actually lacks float16/float32 support for these SDPA paths or if this was just an oversight when the file was ported within the scope of https://github.com/intel/torch-xpu-ops/issues/3546

pbielak (changes requested):
Please move all `retarget_...` calls to one place, same for all `dtypesIfXPU` and same for all `skipXPUIf` decorations.

tszulist-hbn:
> Please move all `retarget_...` calls to one place, same for all `dtypesIfXPU` and same for all `skipXPUIf` decorations.

Done. Consolidated into three blocks:
- Retarget block (`test_nestedtensor_xpu.py:51-66`) - added `test_pin_memory`, which was previously stranded mid-file.
- `dtypesIfXPU` block (:68-84) - now holds `test_layer_norm`, `test_layer_norm_breaking`, `test_sdpa`, and `test_sdpa_with_packed_in_proj`.
- `skipIfXpu` block (:86-101) - `test_record_stream`, `test_sdpa_with_packed_in_proj`, `test_sdpa_flop_counter`, `test_dummy_mha_with_nt`.

One semantic note: `test_sdpa_with_packed_in_proj` previously had both decorators applied in a single nested call (`skipIfXpu(dtypesIfXPU(...)(...))`). Splitting it across the two blocks preserves the same order - `dtypesIfXPU` applied first (inner), then `skipIfXpu` reassigning over the already-decorated attribute. The method overrides that define replacement functions stay in place below, since each needs its own body.

pbielak (approved):

BBBela (commented):
Thanks for all the answers and changes! 😉 
It mostly looks good to me now, I have just the last concern below.

Regarding the `test_sdpa_with_packed_in_proj` - if XPU doesn't support this feature then the test should be simply skipped for XPU. I see no point in setting the `dtypesIfXPU` and then skipping the test anyway. I suggest to just skip this test case without changing the dtypes for XPU. That would also resolve my second concern regarding splitting applying dtypes and skipping this test case in different places of refactored file. One test case should be overridden in one place in my opinion for clarity.

tszulist-hbn:
> Thanks for all the answers and changes! 😉 It mostly looks good to me now, I have just the last concern below.
> 
> Regarding the `test_sdpa_with_packed_in_proj` - if XPU doesn't support this feature then the test should be simply skipped for XPU. I see no point in setting the `dtypesIfXPU` and then skipping the test anyway. I suggest to just skip this test case without changing the dtypes for XPU. That would also resolve my second concern regarding splitting applying dtypes and skipping this test case in different places of refactored file. One test case should be overridden in one place in my opinion for clarity.

I looked into this carefully. There's a subtlety: dropping `dtypesIfXPU` here doesn't leave the skip behavior unchanged.

The upstream test is decorated with:
```
@dtypes(*([torch.float16, torch.bfloat16, torch.float32] if SM80OrLater
          else [torch.float16, torch.float32]))
```
The dtype list is gated on `SM80OrLater`, which is a CUDA capability check. On a pure XPU machine (no CUDA device present) it evaluates to False, so the "all" default collapses to `[float16, float32]` - bfloat16 is not included.

So without `dtypesIfXPU(torch.bfloat16)`, the skipped variant reported for XPU changes from `test_sdpa_with_packed_in_proj_xpu_bfloat16` to `test_sdpa_with_packed_in_proj_xpu_float16` + `..._xpu_float32`. All variants are still skipped by `skipIfXpu` either way - nothing runs but the skip fingerprint in the logs changes, which is why I kept the `dtypesIfXPU` line: it pins the XPU variant to a single, backend-appropriate dtype instead of inheriting a CUDA-capability-dependent list.

I agree the two-place override (dtype pinning + skip) is less clean than one place. If you'd prefer I drop `dtypesIfXPU` entirely for this case, I'm happy to it just means the expected skipped entries become `_xpu_float16`/`_xpu_float32`, and we'd want any skip-list/reference to match. Let me know which you'd prefer.

BBBela (commented):
> If you'd prefer I drop dtypesIfXPU entirely for this case, I'm happy to it just means the expected skipped entries become _xpu_float16/_xpu_float32, and we'd want any skip-list/reference to match. Let me know which you'd prefer.

I vote for dropping the `dtypesIfXPU` entirely. My approach is that if xpu does not support some feature at all, then it doesn't matter if it doesn't support it for `float16` or it doesn't support it for `float32`. It just simply does not support it. Introducing specific dtypes for anyway skipped test for me looks confusing. And I don't see any problem in fact that the names of skipped tests will change as long as it is mentioned in the PR description.

But if you prefer, we can leave the `dtypesIfXPU` and re-think this when porting to upstream, but then please at least merge the skip and dtypes into one override - that would be sufficient for me.

tszulist-hbn:
Alright, I'll leave it there for now. `test_sdpa_with_packed_in_proj_xpu_bfloat16` will be handled in https://github.com/intel/torch-xpu-ops/issues/3546, and that it is currently blocked by https://github.com/intel/torch-xpu-ops/issues/3140. I'll address it once I'm working on it.

BBBela (approved):
Thank you! Looks good to me now.

pbielak:

PR #4299, opened 2026-07-10
Reuse upstream Math.h for XPU special functions

cyyever:
MathExtensions.h duplicated upstream Math.h with std:: rewritten as sycl::; upstream already compiles on the XPU device, so use it directly and drop the copies. Keep calc_digamma, which has an XPU-only pi_t precision knob.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


guangyey:
I think this is a good point worth discussing. From a code reuse perspective, we should definitely recommend using the upstream implementation. However, from the perspective of SYCL device code, we should prefer `sycl::math` functions over `std:: math` functions. In this case, which guideline should take precedence? @EikanWang, What do you think about?

cyyever:
can SYCL compilers automatically covert ``std`` math functions to its version? If so, it may be desirable to use the upstream version except for some numerical cases.

guangyey:
According to the SYCL extension spec for [C/C++ Standard Library support](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/C-CXX-StandardLibrary.rst), the compiler guarantees that most `std:: functions` are available in device code. Based on that, I think we should follow these guidelines:

1. If `std::one` is supported in device code, prefer `std::one`.
2. If fast-math behavior is desired, use `sycl::native::one` (similar to CUDA's `__one` intrinsics).
3. Use `sycl::one` only when the corresponding `std::one` is not supported in device code.

cc @EikanWang @jianyizh Any idea?

jianyizh:
implementation of std::{math_func} may change by compiler / compile flags. sycl::{math_func} provides the accurate one. So we prefer using sycl::{math_func}. However, I noticed that we used std::sin and std::cos in this file, they should be sycl::sin

EikanWang:
@cyyever , yes, the sycl compiler is able to convert `std::{math_func}` to a sycl-based implementation. However, the conversion is not clearly documented. We observed that some std functions are lowered to `sycl::{math_func}`, while some are lowered to `sycl::native::{math_func}`. The accuracy of `sycl::{math_func}` is better than `sycl::native`, while its performance is not. So, we have a general rule that uses `sycl::{math_func}` by default to guarantee accuracy and follows the community practice of using `sycl::native::{math_func}` limitedly to trade off accuracy and performance.

Back to `std`, the motivation for deprecating is that we intend to control the accuracy and performance behavior fully to avoid discrepancies. So far, the `std` behaviors in SYCL come from our observation but not from the spec. So, we are happy to discuss the practice, as the programming paradigm and developer user experience are so important. 

If we believe `std`-based programming could deliver a better developer user experience, we should suggest the compiler team improve it. @cyyever , any thoughts?

cyyever:
@EikanWang I haven't checked the conversion rules for the SYCL compiler (assuming that it is open-sourced), but all that matter are numerical difference and performance. Also note the ``constexpr`` support, that means C++20 std::math function calls can be computed at build time for some cases (constexpr inputs), which is an obvious optimization.

guangyey:
To be honest, I think this would be difficult to enforce `sycl::math` and hard to avoid `std::math` in practice. We often reuse internal helper functions from PyTorch, and those helpers may already use `std:: math` functions. To replace them with `sycl:: math` functions, we would need to copy the upstream implementation and modify it. This has two drawbacks:

1. It's not always easy to identify which internal helpers use `std:: math` functions (oversight).
2. It introduces code duplication and creates a maintenance burden, as well as a potential divergence from the upstream implementation over time.

I also looked into the compiler implementation. For example, `sycl::sin` dispatches to `__spirv_ocl_native_sin` when fast math is enabled, and to `__spirv_ocl_sin` otherwise (see [here](https://github.com/intel/llvm/blob/c27d1ac8c54217be8424c82d0b1d4cc0e13a2c37/sycl/include/sycl/detail/builtins/math_functions.hpp#L100)). In contrast, `std::sin` always dispatches to `__spirv_ocl_sin` (see [here](https://github.com/intel/llvm/blob/c27d1ac8c54217be8424c82d0b1d4cc0e13a2c37/sycl/include/sycl/stl_wrappers/__sycl_cmath_wrapper_impl.hpp#L89)).
In this case, `std::sin` is indeed what we need. However, I remember we already disabled fast math on our side; then they are equivalent (for scalar types).

From a long-term perspective, I think the compiler should guarantee that `std:: math` functions are equivalent to `sycl:: math` functions for device code, with the behavior under fast-math clearly documented. Otherwise, we cannot ensure that all shared code in PyTorch uses `sycl:: math` functions. We'd end up copying upstream code just to replace `std::` with `sycl::`, which significantly increases the maintenance burden and the risk of diverging from upstream over time.

EikanWang:
>I also looked into the compiler implementation. For example, sycl::sin dispatches to __spirv_ocl_native_sin when fast math is enabled, and to __spirv_ocl_sin otherwise (see [here](https://github.com/intel/llvm/blob/c27d1ac8c54217be8424c82d0b1d4cc0e13a2c37/sycl/include/sycl/detail/builtins/math_functions.hpp#L100)). In contrast, std::sin always dispatches to __spirv_ocl_sin (see [here](https://github.com/intel/llvm/blob/c27d1ac8c54217be8424c82d0b1d4cc0e13a2c37/sycl/include/sycl/stl_wrappers/__sycl_cmath_wrapper_impl.hpp#L89)).
In this case, std::sin is indeed what we need. However, I remember we already disabled fast math on our side; then they are equivalent (for scalar types).

That's the point. Under the PyTorch context, we disable fast-math-related flags by default, just like other accelerators, and invoke fast math functions through `sycl::native` on purpose.

EikanWang:
> From a long-term perspective, I think the compiler should guarantee that std:: math functions are equivalent to sycl:: math functions for device code, with the behavior under fast-math clearly documented. Otherwise, we cannot ensure that all shared code in PyTorch uses sycl:: math functions. We'd end up copying upstream code just to replace std:: with sycl::, which significantly increases the maintenance burden and the risk of diverging from upstream over time.

@guangyey , let's submit an issue to the Intel LLVM community to make it happen.

EikanWang:
> Also note the constexpr support, that means C++20 std::math function calls can be computed at build time for some cases (constexpr inputs), which is an obvious optimization.

For this optimization, I think it has been in the SYCL compiler regardless of `std::{math_func}` and `sycl::{math_func}`. 

Besides that, I think we agree that `std::{math_func}` and `sycl::{math_func}` should exhibit equivalent behavior for the same inputs, unless SYCL explicitly specifies different semantics, right? @cyyever @guangyey @jianyizh

EikanWang:
Specific to this PR, I think it is the right change to reuse code as much as possible from the design perspective. But before landing it, we need to get confirmation from the compiler team.

guangyey:
> > Also note the constexpr support, that means C++20 std::math function calls can be computed at build time for some cases (constexpr inputs), which is an obvious optimization.
> 
> For this optimization, I think it has been in the SYCL compiler regardless of `std::{math_func}` and `sycl::{math_func}`.
> 
> Besides that, I think we agree that `std::{math_func}` and `sycl::{math_func}` should exhibit equivalent behavior for the same inputs, unless SYCL explicitly specifies different semantics, right? @cyyever @guangyey @jianyizh

I think so. Otherwise, this will become a significant maintenance burden for us.

guangyey:
> @guangyey , let's submit an issue to the Intel LLVM community to make it happen.

Sure.

cyyever:
Will you check fastmath flag?

guangyey:
> Will you check fastmath flag?

Yes, I'll check with the SYCL compiler team, get their feedback, and then move this PR forward.

guangyey:
@EikanWang @jianyizh Our proposal has been accepted by SYCL compiler. I think we could stop porting the std::xxx math functions to sycl::xxx ones. Any other concerns about this PR?

jianyizh (approved):

guangyey:
build fail

cyyever:
@guangyey Help me trigger CI?

guangyey:
Done.

src/ATen/native/xpu/sycl/MathExtensions.h
```diff
@@ -1223,6 +892,11 @@ static inline C10_HOST_DEVICE scalar_t bessel_j1_forward(scalar_t x) {
       +5.32278620332680085395e+18,
   };
 
+  // J1 is odd. Fold the sign out so the branches below only ever see |x|;
+  // upstream reflects by recursing, which SYCL device code forbids.
+  const scalar_t sign = (x < scalar_t(0.0)) ? scalar_t(-1.0) : scalar_t(1.0);
```
guangyey, line 897: Good catch!

guangyey (approved):

guangyey:

PR #4301, opened 2026-07-10
Use if constexpr and constexpr in XPU kernels

cyyever:
Convert runtime if/ternary on compile-time-constant template params and type traits to if constexpr, and static const/static inline to constexpr where the initializer is constant.


guangyey (approved):
Thanks!

PR #4303, opened 2026-07-10
Fix stale found-flag in radix select findPattern

cyyever:
findPattern zeroed slot RADIX_SIZE instead of slots 0/1 (found-flag, value), so
the loop read a stale flag left by countRadixUsingMask and could return early
with the wrong value. Zero slots 0/1 (matches upstream CUDA
SortingRadixSelect.cuh).


guangyey (approved):
Good catch!

guangyey:
cc @yucai-intel

Issue #4110, opened 2026-06-23
[RFC]Performance Optimization Skills Suite on Intel Platforms


liangan1:
### 🚀 The feature, motivation and pitch

# Performance Optimization Skills Suite on Intel Platforms  

## Task List
- Kernel performance analysis: build the evidence chain based on the kernel implementation and Unitrace counter.   
#4306 
- ASM code extract and source code mapping: 
 #4176 
 #4177 

- Hardware workflow skill PRs to cover the EU analysis path:
 #4307 adds `eu-utilization-triage`, the ComputeBasic three-state entry workflow for routing stall / ILP / TLP follow-up analysis.
#4308 adds `eu-stall-attribution`, the workflow for high XVE_STALL attribution using stall counters, per-IP data, ASM extraction, and ASM/source mapping.
#4309 adds `eu-ilp-coissue`, the workflow for low co-issue / pipe imbalance analysis with ComputeBasic pipe counters and ASM evidence.
#4310 adds `eu-tlp-occupancy`, the workflow for deciding when more thread-level parallelism helps versus worsens Stage-2 Pipe/Send contention.
---

## 1. Motivation

Given a PyTorch model or an operator, the goal of perfomrance optimization is to improve the hardware resource ultization to approach the roofline goal. This process need a lot of expertise. For example, when you want optmize the attention layer in the LLM modesl on the BMG-580, you need to familar with the flash attention algorithm, roofline model, perforamnce analyze tools and hardware architechture. Besides, the optmization process is iterative and time consuming, you need to repeat the workflow multiple times to approache the roofline: collect the performance counters with unitrace tools, find the rootcasue of the hostspot from the insights and then apply the optmization changes. 

The agent AI provide a good practice to reduce the mannual effort and make the optmization process more scalable. In this RFC, we provide a **top-down** methodology to do the performance optmization and will build a skills suite to make this process to be reusable. 

Two things make this an RFC rather than a tuning handbook:

1. **Knowledge vs Skill are separated.** *Knowledge* is static fact (peak compute, bandwidth, GRF size, DPAS shape) referenced by skills; *skills* are executable actions (measure or rewrite). Any SKU-specific number lives in `knowledge/`, never hard-coded in a skill  
2. **The Skills must be data driven.** *Data Driven* means that the rootcasue should based on the insights from the profier trace and unitrace counter. Performance gain only with hyper parameter tuning w/o insights is not a good practice and should not be part of skills.   

---

[...]

PR #4306, opened 2026-07-10
Add kernel perf analysis skill

liangan1:
Per #4110 
# Summary
 Adds `.claude/skills/kernel-perf-analysis/SKILL.md` for source-code based roofline analysis of one hotspot operator/target kernel; includes device-agnostic workflow and XPU/unitrace ComputeBasic mapping; 

# Workflow 
This skill is part of the performance skills suite. It firstly build the roofline model based on the kernel implementation and finds the gap between the roofline result and Unitrace counters.  And then call other skills, e.g., stall reason analysis for further analysis.

### DIFF
--- a/.claude/skills/kernel-perf-analysis/SKILL.md
+++ b/.claude/skills/kernel-perf-analysis/SKILL.md
@@ -145,11 +145,11 @@ Use this compact report structure:
 
 | Layer | Required content |
 |---|---|
-| L0 Identity / launch | Target kernel, launch grid, call count, time share |
-| L1 Compute state | Stall / active / co-issue / occupancy |
-| L2 Pipe / execution mix | Scalar/vector/tensor pipe utilization, load-store pressure, issue profile |
-| L3 Memory hierarchy | Intra-core on-chip, inter-core on-chip, global memory, queue, TLB/page behavior |
-| L4 Verdict | Primary bottleneck, secondary bottleneck, next measurable action |
+| Level 0 Identity / launch | Target kernel, launch grid, call count, time share |
+| Level 1 Compute state | Stall / active / co-issue / occupancy |
+| Level 2 Pipe / execution mix | Scalar/vector/tensor pipe utilization, load-store pressure, issue profile |
+| Level 3 Memory hierarchy | Intra-core on-chip, inter-core on-chip, global memory, queue, TLB/page behavior |
+| Level 4 Verdict | Primary bottleneck, secondary bottleneck, next measurable action |
 
 Assign evidence level:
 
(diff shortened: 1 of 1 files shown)


.claude/skills/kernel-perf-analysis/SKILL.md
```diff
@@ -0,0 +1,395 @@
...
+| Target kernel | Runtime kernel that implements the hotspot operator |
+| Reproducer | Minimal command or script that launches the target kernel |
+| Backend evidence | Profiler/counter artifacts for the target kernel only |
+
+If the target kernel is ambiguous, first collect or inspect profiler output to map the hotspot operator to the runtime kernel name.
+
+## Device-Agnostic Workflow
+
+### Step 0: Pick One Hotspot Operator
+
```
jianyizh, line 39: Step 0 run the reproducer / microbenchmark. Then start from one selected hotspot op.

.claude/skills/kernel-perf-analysis/SKILL.md
```diff
@@ -0,0 +1,395 @@
...
+| Synchronization | Barriers, subgroup reductions, memory fences, atomic operations |
+
+Derive these quantities when applicable:
+
+| Quantity | Meaning |
+|---|---|
+| `input_bytes` | Required input tensor bytes |
+| `output_bytes` | Required output tensor bytes |
+| `parameter_bytes` | Weights, bias, scale, metadata, descriptors |
+| `source_logical_read_bytes` | Element bytes read by the implementation before cacheline or transaction effects |
```
jianyizh: We also want to know the minimal read/write  byte from op semantics.  Maybe agent can raise algorithm level optimization

.claude/skills/kernel-perf-analysis/SKILL.md
```diff
@@ -0,0 +1,395 @@
...
+Aggregate only rows or events matching the target kernel. Weight percentage and rate counters by kernel time from the same measurement window.
+
+Collect these groups when available:
+
+| Measurement group | Meaning |
+|---|---|
+| Runtime and launch | Kernel time, call count, launch grid, frequency, device busy |
+| Compute state | Stall, active, issue utilization, occupancy, pipe mix, vector/tensor-core utilization |
+| Intra-core on-chip memory | Local/shared bytes or events, local-cache transactions, bank conflicts |
+| Inter-core on-chip cache | Shared cache hits/misses/transactions/stalls |
```
jianyizh: Here you mean load store cache and shared local memory for intra core and last level cache for inter core?

liangan1: Yes.

.claude/skills/kernel-perf-analysis/SKILL.md
```diff
@@ -0,0 +1,395 @@
...
+### XPU Target Kernel and Launch Grid
+
+| Abstract field | XPU / ComputeBasic source |
+|---|---|
+| Runtime kernel name | unitrace CSV `Kernel` |
+| Execution width | Kernel signature such as `SIMD32`, or ASM |
+| Global grid | Kernel signature or source launch code |
+| Local workgroup | Kernel signature or source launch code |
+| Workgroup count | `GPGPU_THREADGROUP_COUNT` |
+| Kernel time | `GpuTime[ns]` |
```
lchen2331: `GpuTime[ns]` is available only in the Unitrace `-q` output. However, the timing reported by `-q` has significant overhead, especially for small kernels, while the device time reported by `-d` is more accurate.

gemm_kernel | -q timing inflation over -d
-- | --
M=1, K=1024, N=256 | 2.54× (+154%)
M=1190, K=2048, N=2048 | 1.08× (+8%)

.claude/skills/kernel-perf-analysis/SKILL.md
liangan1: Implemented in da61fde0: Step 0 now explicitly runs the reproducer or microbenchmark first, then uses its profiler output to select the hotspot operator and target kernel.

.claude/skills/kernel-perf-analysis/SKILL.md
liangan1: Implemented in da61fde0: the roofline model now separates semantic minimal read/write bytes from implementation logical traffic and records the comparable implementation-extra terms. The workflow treats a large extra term as evidence for algorithm-level opportunities such as removing intermediates, repeated passes, or fusion, while requiring aliasing and in-place updates to be explained first.

.claude/skills/kernel-perf-analysis/SKILL.md
liangan1: Implemented in da61fde0: the device-agnostic definitions now state that intra-core covers explicit local/shared storage plus the per-core load/store-cache path, while inter-core is the shared last-level/device cache. The XPU mapping remains explicit: SLM plus LOAD_STORE_CACHE for intra-Xe-core behavior, and L3 for inter-Xe-core behavior.

.claude/skills/kernel-perf-analysis/SKILL.md
liangan1: Implemented in da61fde0: repeated unitrace `-d` device duration is now the primary source for absolute latency, throughput, and roofline-gap claims. `-q` `GpuTime[ns]` is explicitly limited to weighting/alignment within the same counter collection and is not used for primary short-kernel performance conclusions. The output template was updated to enforce the same distinction.

PR #4309, opened 2026-07-10
Add EU ILP co-issue skill

liangan1:
## Summary
Adds `.claude/skills/eu-ilp-coissue/SKILL.md`, a focused workflow for Intel GPU/XPU EU instruction-level parallelism and co-issue analysis.

This skill is used when stall is not the dominant deficit, but active EU cycles are mostly single-pipe and `XVE_MULTIPLE_PIPE_ACTIVE` is low. It identifies whether independent source-code work can be moved so execution windows overlap on more than one pipe.

## Workflow
```mermaid
flowchart TD
    A["Input: repro-cmd + optional asm-dir/ip-source-table"] --> B["Step 1: Read ComputeBasic counters"]
    B --> C["Measure active, co-issue, ALU0/ALU1/ALU2 utilization"]
    C --> D{"Step 2: Is ILP headroom likely?"}
    D -->|"S0/stall high"| STALL["route -> eu-stall-attribution"]
    D -->|"S1 high + S2 low + one pipe underused"| ASM["Step 3: inspect ASM/source regions"]
    D -->|"co-issue already healthy or true serial dependency"| STOP["stop: no ILP-first fix"]
    ASM --> E{"Step 4: match source-code pattern"}
    E -->|"FP burst then matrix/tensor work"| P1["Rescale-GEMM interleave"]
    E -->|"matrix chain with deferred independent prep"| P2["move epilogue/address/metadata prep between tiles"]
    E -->|"convert/store chain"| P3["epilogue split store"]
    E -->|"address math just before SEND"| P4["address pre-computation"]
    E -->|"repeated math waits"| P5["loop unroll for cross-iteration ILP"]
    E -->|"math phase then matrix phase"| P6["math-matrix interleave"]
    P1 --> CARD["Step 5: emit ILP card"]
    P2 --> CARD
    P3 --> CARD
    P4 --> CARD
    P5 --> CARD
    P6 --> CARD
    CARD --> V["Step 6: validate: time, ComputeBasic, ASM, spill/pressure checks"]
```

## Co-Issue Model

Co-issue is the measurable result; ILP is the source-code technique. The EU issues one instruction at a time, but multi-cycle execution windows from adjacent instructions can overlap on different execution pipes. The goal is to move independent source-level work so the hot loop has useful instructions for more than one pipe in flight.

## Source-Code Patterns

| Evidence | Source-code pattern | Before | After | Expected counter movement |
|---|---|---|---|---|
| Long FP `mul/add/math.*` burst, then matrix/tensor work | Rescale-GEMM interleave | Rescale a whole tile, then run the matrix tile | Split rescale into smaller blocks and run each block next to the corresponding matrix tile | Co-issue share rises because FP work overlaps matrix-pipe work |
| Long matrix/tensor chain with no independent FP or INT work nearby | Deferred epilogue/address preparation | Finish the whole matrix chain before preparing epilogue, addresses, or next-tile metadata | Move independent epilogue prep, address prep, or next-tile metadata between matrix tiles without breaking true dependencies | Idle FP/INT pipe utilization improves while matrix utilization stays stable |
| Convert plus store chain at loop end | Epilogue split store | Convert all outputs, then store, or store rows serially with no overlapped prep | Split convert/store into chunks and overlap next address/prep work with current convert/store sequence | Epilogue single-pipe region shrinks; check SEND pressure and write-side counters |
| Address math appears immediately before each load/store SEND | Address pre-computation | Compute `ptr + offset` just before SEND, then wait for data | Precompute next addresses during current independent math or matrix work; issue SEND earlier when legal | ALU1 address work and SEND latency overlap with useful work; Stage-1/SBID waits may fall |
| Math function loop has repeated same-operation waits | Loop unroll for cross-iteration ILP | Issue one expensive math op, then wait for its result before the next independent element | Unroll independent elements so several math ops are in flight on different registers | Distance stalls fall; math pipe utilization and co-issue opportunity improve |
| Softmax/math phase and following matrix phase are strictly sequential | Math-matrix interleave | Finish all `exp/mul/reduce` work before starting the next matrix accumulation | Start matrix work for completed rows/blocks while later rows/blocks continue math work | MATH/FP work overlaps matrix-pipe work; total phase time can shrink |

## Evidence Rules

- Do not use this skill when S0/stall is dominant; run `eu-stall-attribution` first.
- Reuse `ip-source-table` from `eu-stall-attribution` when available; do not repeat ASM/source mapping unnecessarily.
- Do not break a hardware-optimized matrix chain blindly. Place independent work around it only when dependencies allow.
- Every recommendation must identify the source loop/epilogue to change, the idle pipe it should feed, and the validation counters.
- Added ILP can increase register pressure; validate time, `XVE_MULTIPLE_PIPE_ACTIVE`, idle-pipe utilization, SEND pressure, and spill/GRF behavior.

## Outputs

The workflow emits an ILP card containing:
- co-issue share and single-pipe active share
- busy/idle pipe evidence
- ASM/source region evidence
- matched source-code pattern
- recommended source change
- expected counter movement
- validation command/counters

## Validation

- Markdown diagnostics passed.
- PR contents are limited to `.claude/skills/eu-ilp-coissue/SKILL.md`.

### DIFF
--- a/.claude/skills/eu-ilp-coissue/SKILL.md
+++ b/.claude/skills/eu-ilp-coissue/SKILL.md
@@ -1,0 +1,145 @@
+---
+name: eu-ilp-coissue
+description: Analyze Xe2 EU instruction-level parallelism and co-issue for one XPU target kernel. Use when stall is not dominant but XVE_MULTIPLE_PIPE_ACTIVE is low, ALU pipe utilization is imbalanced, or ASM evidence is needed to find interleave opportunities.
+---
+
+# EU ILP and Co-Issue Analysis
+
+Analyze low co-issue in one XPU target kernel and identify whether instruction-level parallelism can raise execution from mostly single-pipe active cycles to multi-pipe active cycles.
+
+This skill is for the S1-to-S2 problem: the EU is active, but too much active time uses only one physical pipe.
+
+## When to Use
+
+Use this skill when:
+- `XVE_STALL` is not the dominant deficit.
+- `XVE_MULTIPLE_PIPE_ACTIVE` or co-issue share is low.
+- Per-pipe ALU utilization shows an idle pipe with plausible work to overlap.
+- `eu-utilization-triage` routes to ILP/co-issue.
+- `eu-stall-attribution` identifies Stage-1 dependency distance and routes here with ASM/source evidence.
+
+Do not use this skill when S0/stall dominates; run `eu-stall-attribution` first. Do not use it when the idle pipe has no independent work available.
+
+## Inputs
+
+| Input | Meaning |
+|---|---|
+| `repro-cmd` | Command that launches the target kernel |
+| `kernel-filter` | Target kernel filter |
+| `ComputeBasic` counters | Co-issue, active, and per-pipe utilization counters |
+| `asm-dir` | Optional existing ASM artifacts |
+| `ip-source-table` | Optional source mapping from `eu-stall-attribution` |
+
+## Workflow
+
+### Step 1: Measure Co-Issue and Pipe Utilization
+
+Collect or read `ComputeBasic` for the target kernel.
+
+Key metrics:
+- `XVE_ACTIVE`
+- `XVE_MULTIPLE_PIPE_ACTIVE`
+- `XVE_INST_EXECUTED_ALU0_ALL_UTILIZATION`
+- `XVE_INST_EXECUTED_ALU1_ALL_UTILIZATION`
+- `XVE_INST_EXECUTED_ALU2_ALL_UTILIZATION`
+- SEND and issue counters when available
+
+Derived values:
+
+```text
+single_pipe_active = XVE_ACTIVE - XVE_MULTIPLE_PIPE_ACTIVE
+coissue_share = XVE_MULTIPLE_PIPE_ACTIVE / XVE_ACTIVE
+```
+
+Identify the idle or underused pipe.
+
+### Step 2: Decide Whether ILP Headroom Exists
+
+| Observation | Meaning |
+|---|---|
+| S0 high | Not an ILP-first case; run stall attribution |
+| S1 high and S2 low | ILP/co-issue headroom likely exists |
+| One pipe busy, another pipe low | Candidate for interleaving independent work |
+| All useful pipes already balanced | ILP headroom may be small |
+
+Stop if co-issue is already healthy or if the candidate work is truly serial.
+
+### Step 3: Inspect ASM for Interleave Opportunities
+
+If an `ip-source-table` from `eu-stall-attribution` exists, reuse it. Otherwise extract ASM with `extract-xpu-kernel-asm` and map hot regions with `asm-source-mapping` when source-level evidence is needed.
+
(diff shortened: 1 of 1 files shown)


.claude/skills/eu-ilp-coissue/SKILL.md
```diff
@@ -0,0 +1,151 @@
...
+Collect or read `ComputeBasic` for the target kernel.
+
+Key metrics:
+- `XVE_ACTIVE`
+- `XVE_MULTIPLE_PIPE_ACTIVE`
+- `XVE_INST_EXECUTED_ALU0_ALL_UTILIZATION`
+- `XVE_INST_EXECUTED_ALU1_ALL_UTILIZATION`
+- `XVE_INST_EXECUTED_ALU2_ALL_UTILIZATION`
+- SEND and issue counters when available
+
```
jianyizh, line 62: The utilization metric does not include math instructions. Maybe metric like XVE_INST_EXECUTED_ALU0_ALL， XVE_INST_EXECUTED_ALU1_ALL， XVE_INST_EXECUTED_ALU2_ALL， XVE_INST_EXECUTED_FP32， XVE_INST_EXECUTED_INT32， XVE_INST_EXECUTED_MATH， XVE_INST_EXECUTED_SEND_ALL... can provide more info,

.claude/skills/eu-ilp-coissue/SKILL.md
```diff
@@ -0,0 +1,151 @@
...
+| Observation | Meaning |
+|---|---|
+| S0 high | Not an ILP-first case; run stall attribution |
+| S1 high and S2 low | ILP/co-issue headroom likely exists |
+| One pipe busy, another pipe low | Candidate for interleaving independent work |
+| All useful pipes already balanced | ILP headroom may be small |
+
+Stop if co-issue is already healthy or if the candidate work is truly serial.
+
```
jianyizh, line 83: Not sure if we can have interleave opportunities, this part is handled by igc. What we normally do is check source code the see if there is redundant computation

liangan1: Yes. You are right; I may need to add "reduce redundant computation" into this part. For the interleave instructions, the IGC should do it as much as possible, but it may also miss some patterns. The rescale overlap with the PV gemm in the flash attention is a good example. https://github.com/intel/sycl-tla/commit/7998afd98785d2b6f4c6a6b31c212d82890fb89b

.claude/skills/eu-ilp-coissue/SKILL.md
liangan1: Implemented in 51913b41. The workflow now separates ComputeBasic co-issue/pipe-time metrics from raw ALU0/1/2, SEND, and issued events, and adds an optional VectorEngineProfile collection for FP32, INT32, MATH, SEND, and relevant XMX counts. It also states that ALU1 excludes extended math, so MATH must be checked before declaring ALU1 idle, and that raw events must not be compared directly with utilization percentages.

LuFinch (approved):

PR #4311, opened 2026-07-10
Migrate `test_testing_xpu.py` to use `XPUImportCtx`

pbielak:
Replace the full copy of upstream `test/test_testing.py` with a wrapper that imports the upstream module via `XPUImportCtx` and only overrides XPU-specific differences:

- Restore `instantiate_parametrized_tests` and `instantiate_device_type_tests` bindings in the imported module: XPUImportCtx patches these to DO_NOTHING during import, but `test_testing.py` captures module-local references that remain stale. Several test methods call these functions on locally-defined classes and would silently produce wrong results without this fixup.
- Override `test_get_supported_dtypes` to add the `xpu` branch so `dtypesIfXPU` is checked when running on XPU device (upstream only handles cpu/cuda).
- Instantiate `TestAssertCloseMultiDevice`, `TestMakeTensor` and  `TestOpInfoSampleFunctions` for both cuda and xpu (upstream only instantiates for cuda).

Test results:

  Before: 2060 passed, 17 skipped, 12 xfailed
  After:  3953 passed, 36 skipped, 12 xfailed
  \-----------------------------------------------
  Delta: +1893 passed, +19 skipped,  0 xfailed

Newly passing test (1893):
  - TestTestingCPU::test_callable_msg_cpu (new upstream test)
  - TestMakeTensor::* (727 tests; added `allow_xpu=True`)
  - TestOpInfoSampleFunctions::* (1160 tests; added `allow_xpu=True`)

Newly skipped tests (19):
  - TestOpInfoSampleFunctions::* (19 tests; previously not launched)


test/xpu/test_testing_xpu.py
```diff
@@ -1,2970 +1,114 @@
...
-
-        actual = torch.testing.make_tensor(
-            shape, dtype=dtype, device=device, low=low, high=high
-        )
-        expected = torch.full(shape, low, dtype=dtype, device=device)
-
-        torch.testing.assert_close(actual, expected)
-
-
 instantiate_device_type_tests(TestMakeTensor, globals())
```
BBBela: I know that this is copied directly from the previous version, but as far as I know, without the `allow_xpu=True` this won't instantiate tests for XPU - is that correct? So I am wondering if maybe we should enable those tests?

It looks like they are not currently tested in our CI and on the first sight it looks like they should have - what do you think?

test/xpu/test_testing_xpu.py
```diff
@@ -1,2970 +1,114 @@
...
-        dtypes=OpDTypes.none,
+try:
+    from xpu_test_utils import XPUImportCtx
+except Exception:
+    from .xpu_test_utils import XPUImportCtx
+
+with XPUImportCtx(False):
+    import test_testing as _test_testing_mod
+    from test_testing import (
+        expectedFailureMeta,
```
BBBela, line 30: There is upstream `TestEnvironmentDefFlag` test class that was not present previously and is not ported currently. It looks like it is device-agnostic, but maybe we could add short info about intentionally dropping this test class to the PR description for clarity and being explicit?

test/xpu/test_testing_xpu.py
pbielak: I added `allow_xpu=True` to the classes that pass. There are two classes which currently fail - adding info about that to the RFC issue.

test/xpu/test_testing_xpu.py
pbielak: Class added - all test pass

BBBela (approved):
Thanks for the changes and updates!

astachowiczhabana (approved):
LGTM

pbielak:

PR #4312, opened 2026-07-10
Fix read/write races in topk single-workgroup kernel

cyyever:
TensorTopKSingleWgKernel.cpp reads `smem` then the next radix pass re-zeros it
(or callers reuse it) without a `group_barrier` — a read/write data race that can
give torn reads and wrong topk results. Add WAR barriers (matches upstream CUDA
`TensorTopK.cu`, which has `__syncthreads` at these points).

Test: none (not yet built/run on hardware; matched to upstream CUDA)


guangyey (approved):
I think this definetly is a bug.

guangyey:
@jianyizh Could you please help review this PR.

jianyizh (approved):

guangyey:

PR #4315, opened 2026-07-12
Wire pull.yml workflow_dispatch inputs for bot rerun -b

chuanqi129:
## Summary

The `@torchxpubot rerun -b <ref>` command dispatches `pull.yml` via `workflow_dispatch` with `pytorch_ref` and `pr_number` inputs, but `pull.yml` never consumed them:

- Checkout defaulted to the dispatch ref (`main`), not the PR branch
- `github.base_ref` is empty for `workflow_dispatch`, so the `pytorch` parameter passed to all downstream reusable workflows was an empty string
- `github.event.pull_request.draft == false` evaluated to false (null context), blocking build/test jobs
- `github.event.pull_request.number` was empty, breaking the live label read

This PR fixes all four issues:

1. Checkout `refs/pull/{pr_number}/head` when triggered via `workflow_dispatch`
2. Chain `github.event.inputs.pytorch_ref` into all 7 `pytorch:` parameter expressions, with `'main'` as final fallback
3. Add `|| github.event_name == 'workflow_dispatch'` to all 3 draft-check gates
4. Use `github.event.pull_request.number || github.event.inputs.pr_number` for live label reads

No changes to `bot.yml` -- the dispatch logic there was already correct.

Test: none (workflow plumbing only; testable by commenting `@torchxpubot rerun -b main` on any PR)

### DIFF
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -638,7 +638,7 @@ jobs:
                   owner: context.repo.owner,
                   repo: context.repo.repo,
                   workflow_id: 'pull.yml',
-                  ref: pr.data.base.ref,
+                  ref: pr.data.head.ref,
                   inputs: {
                     pytorch_ref: pytorchRef,
                     pr_number: String(prNumber)
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -189,7 +189,8 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          LABELS=$(gh pr view ${{ github.event.pull_request.number }} \
+          PR_NUM="${{ github.event.pull_request.number || github.event.inputs.pr_number }}"
+          LABELS=$(gh pr view "$PR_NUM" \
             --repo ${{ github.repository }} \
             --json labels --jq '[.labels[].name] | join(",")' 2>/dev/null || echo "")
           echo "Live PR labels: $LABELS"
@@ -202,7 +203,7 @@ jobs:
           echo "force_win_ut=$(has windows_ci)" >> "$GITHUB_OUTPUT"
 
   conditions-filter-win:
-    if: ${{ github.repository_owner == 'intel' && github.event.pull_request.draft == false }}
+    if: ${{ github.repository_owner == 'intel' && (github.event.pull_request.draft == false || github.event_name == 'workflow_dispatch') }}
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     outputs:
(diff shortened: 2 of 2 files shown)


chuanqi129:
@torchxpubot review

chuanqi129:
@torchxpubot review

BBBela (commented):
I have two following considerations:

---

#### Would calling `@torchxpubot rerun -b <branch>` use the `torch-xpu-ops` PR branch?

If my understanding is correct, then the `@torchxpubot rerun -b <branch>` effectively triggers the `workflow_dispatch`. So the `github.event_name` is `workflow_dispatch`. Next, the  build calls `prepare_pytorch.py` script that sets the `torch-xpu-ops` repo based on the following condition ([CODE LINK](https://github.com/intel/torch-xpu-ops/blob/53bfa8a844b9dea399506b592609d979d83174ee/.github/scripts/prepare_pytorch.py#L104-L110)):
```python
    if github_event == "pull_request":
        # Copy the checked-out PR source
        src = workspace / "torch-xpu-ops"
        shutil.copytree(src, third_party_dir, symlinks=True)
    else:
        git("clone", torch_xpu_ops_repo, str(third_party_dir))
        git("-C", str(third_party_dir), "checkout", torch_xpu_ops_commit)
```
So if the `github.event_name` is `workflow_dispatch`, then the else branch is taken and `torch-xpu-ops` repo is being cloned into the `torch_xpu_ops_commit` which defaults to `main` instead of PR branch - right?
So effectively the `@torchxpubot rerun -b <branch>` command takes the specified PyTorch branch, but it uses the `torch-xpu-ops` main branch instead of the PR branch.

---

#### Would calling `@torchxpubot rerun -b <branch>` build for all the default architectures?

Taking the above into consideration, the `github.event_name` is `workflow_dispatch` when calling `@torchxpubot rerun -b <branch>`. In the `_linux_build.yml` file, the `TORCH_XPU_ARCH_LIST` is being set to `pvc` when `github.event_name` is `pull_request` ([CODE FRAGMENT](https://github.com/intel/torch-xpu-ops/blob/6786179d22b483b79b781ca41d7919833c4bdd56/.github/workflows/_linux_build.yml#L140-L143)).
```bash
          # only build pvc for CI
          if [ "${{ github.event_name }}" == "pull_request" ];then
            export TORCH_XPU_ARCH_LIST='pvc'
          fi
```
So if the `github.event_name` is `workflow_dispatch`, the `TORCH_XPU_ARCH_LIST` isn't set at all, so the build falls back to the default list of architectures. That increases the build time.

What do you think? Please let me know if I am missing something important.

PR #4316, opened 2026-07-13
Use mkl_intel_thread instead of mkl_gnu_thread on Linux

CuiYifeng:
This PR is to replace `mkl_gnu_thread` with `mkl_intel_thread + iomp5` for oneMKL XPU part (built with SYCL compiler) to match PyTorch's FindMKL.cmake behavior for non-GCC compilers.


cmake/ONEMKL.cmake
```diff
@@ -27,6 +27,21 @@ set(TORCH_XPU_OPS_ONEMKL_INCLUDE_DIR ${ONEMKL_INCLUDE_DIR})
 
 set(TORCH_XPU_OPS_ONEMKL_LIBRARIES ${ONEMKL_LIBRARIES})
 
+# mkl_intel_thread requires libiomp5 (Intel OpenMP runtime), which is shipped
+# with the compiler, not MKL.  Find it from the SYCL compiler library directory.
+if(NOT WIN32)
+  find_library(
+    IOMP5_LIBRARY
+    NAMES libiomp5.so
```
guangyey:
```suggestion
    NAMES iomp5
```

guangyey (commented):
Currently, PyTorch XPU still uses GCC as the host compiler, and then `libtorch_cpu` is built with `mkl_gnu_thread`. Shouldn't we align with the CPU party and also use the GNU threading runtime here?

CuiYifeng:
> Currently, PyTorch XPU still uses GCC as the host compiler, and then `libtorch_cpu` is built with `mkl_gnu_thread`. Shouldn't we align with the CPU party and also use the GNU threading runtime here?

@guangyey The threading library `mkl_gnu_thread`/`mkl_intel_thread` of `libtorch_xpu` is used by oneMKL XPU part (MKL SYCL libraries). In my understanding, the threading usage of MKL will be:

- XPU part -> mkl_intel_thread
- CPU part (including fallback from XPU) -> PyTorch's selection

I have updated the PR description.

cmake/ONEMKL.cmake
CuiYifeng: Good catch. Updated.

guangyey:
> @guangyey The threading library `mkl_gnu_thread`/`mkl_intel_thread` of `libtorch_xpu` is used by oneMKL XPU part (MKL SYCL libraries). In my understanding, the threading usage of MKL will be:
> 
> * XPU part -> mkl_intel_thread
> * CPU part (including fallback from XPU) -> PyTorch's selection
> 
> I have updated the PR description.

I understand; I mean that we should use the one that PyTorch selects for CPU.

CuiYifeng:
> > @guangyey The threading library `mkl_gnu_thread`/`mkl_intel_thread` of `libtorch_xpu` is used by oneMKL XPU part (MKL SYCL libraries). In my understanding, the threading usage of MKL will be:
> > 
> > * XPU part -> mkl_intel_thread
> > * CPU part (including fallback from XPU) -> PyTorch's selection
> > 
> > I have updated the PR description.
> 
> I understand; I mean that we should use the one that PyTorch selects for CPU.

Let's hold this PR util a conclusion is reached.

PR #4317, opened 2026-07-13
Remove dead std_var_out/std_var_mean_out from ReduceOps

cyyever:
These four statics (warn_invalid_degrees_of_freedom, std_var_out, std_var_mean_out, options_to_value_type) are unreachable。 they are a stale fork of upstream's implementation. std/var/std_mean/var_mean already dispatch to upstream (native_functions.yaml lists "CPU, CUDA, XPU: std_out" etc.), and upstream's std_var_out whitelists XPU explicitly, reaching the XPU kernel through std_var_stub. 

Also drops the five ATen/ops headers used only by the deleted code. No behavior change: every TORCH_CHECK/TORCH_WARN in the deleted block, including the degrees-of-freedom warning, exists on upstream's live path.

Test: test/xpu/test_ops_xpu.py

---
This PR was authored with the assistance of an AI coding agent (Claude Code).


guangyey (approved):
Thanks!

PR #4330, opened 2026-07-13
Simplify SYCL build-flag and Windows CMake setup

cyyever:
Behavior-preserving CMake cleanup.

`BuildFlags.cmake` grows the SYCL flag lists with `list(APPEND)` (and the space-joined
`SYCL_OFFLINE_COMPILER_CG_OPTIONS` with `string(APPEND)`) instead of the older
`set(VAR ${VAR} item)` idiom, matching the adjacent `SYCL_HOST_FLAGS` block.

`BuildOnWindows.cmake` hoists the `torch_xpu_ops` target creation that was duplicated
across both `BUILD_SEPARATE_OPS` branches, drops the single-use `setup_common_libraries`
macro, partitions the SYCL sources with `list(FILTER)` instead of a hand-rolled
`foreach`/`if`, and expresses whole-archive linking with the portable
`$<LINK_LIBRARY:WHOLE_ARCHIVE,...>` generator expression (CMake 3.24+; this project
already requires 3.27) instead of manually appending `-WHOLEARCHIVE:` link options.
On Windows the host CXX compiler is MSVC, for which CMake maps the feature back to
`/WHOLEARCHIVE:`.

The two-library split working around the Windows 4GB static-library limit is left intact.

Test: no functional change, so no new test. Equivalence was checked locally on Windows
(MSVC 19.51, Ninja, CMake 4.4).

Flag accumulation -- byte-for-byte compare of the old and new idioms, including the
semicolon-quoted `-Xspirv-translator` element and `string(APPEND)` on an unset variable:

```
cmake -P eq.cmake   # set(VAR ${VAR} item) vs list(APPEND) -> IDENTICAL
```

Whole-archive linking -- confirming an unreferenced symbol in a static library still
lands in the consuming DLL under both spellings:

```
cmake -G Ninja -DCMAKE_CXX_COMPILER=cl -S . -B build
cmake --build build          # link with /MAP, grep the map for the symbol
# old  -WHOLEARCHIVE:$<TARGET_FILE:>   -> symbol present
# new  $<LINK_LIBRARY:WHOLE_ARCHIVE>   -> symbol present
```

The full PyTorch XPU build was not run (this host has no Intel GPU and no PyTorch source
tree); CI covers that.

Authored with AI assistance.


guangyey (approved):
Thanks!

guangyey:
Hi @cyyever  could you please help resolve the conflicts.

cyyever:
@guangyey fixed

PR #4332, opened 2026-07-14
Fix SIMD macro hygiene and cover index_add's high-contention path

cyyever:
Three independent cleanups around the `SIMD` constant in the indexing kernels, plus benchmark coverage for the path they guard.

**`SIMD` macro to `inline constexpr`** (`IndexKernelUtils.h`). The macro had no `#undef`, so it leaked into every including TU and could rewrite `GroupReduceUtils.h`'s `template <typename T, int SIMD, int DIM>` parameter to `int 32` and break the build. The include chains don't cross today, so it's latent, not live. A constexpr can't clash: a template parameter shadows a constant, but nothing shadows a macro. All use sites are in `namespace at::native::xpu` and there is no `#ifdef SIMD`, so name lookup and the preprocessor both still resolve.

**Remove dead `#define NUM_THREADS 1024;`** (`DepthwiseConv2dKernels.cpp`). Zero uses -- the trailing semicolon means any use would expand to `foo(1024;)` and fail to compile. `KernelUtils.h` already carries the working `SYCL_NUM_THREADS`.

**Cover `index_add`'s high-contention path** (`indexing.index_add.py`, `microbench_summary.py`). #2294 added a shared-local-memory accumulation path for high-contention embedding backward, but the benchmark only ran distinct `linspace` indices -- zero contention -- so that path had no regression coverage. It now also folds all `step` indices onto 8 positions (64 accumulations each). Indices use deterministic `arange`, not `randint`, so contention cannot drift between runs and manufacture false regressions. `distinct_indices` is registered in `microbench_summary.py`'s column list, or the two contention levels collapse into duplicate rows.

Note: deliberately leaves the `SYCL_REQD_SUB_GROUP_SIZE(32)` attributes on `VectorizedGatherKernel`/`IndexFuncLargeIndexFunctor` untouched -- removing them is an unmeasured performance change and belongs in its own PR.

Test: test/microbench/indexing.index_add.py (requires an XPU device; the C++ changes are covered by existing indexing tests, e.g. `pytest test/xpu/test_ops_xpu.py -v -k "index_add or index_select or gather"`)

---

### DIFF
--- a/.github/scripts/microbench_summary.py
+++ b/.github/scripts/microbench_summary.py
@@ -44,7 +44,7 @@ def parse_logs(log_dir: str, get_backward: bool = False) -> pd.DataFrame:
         "case_name", "datatype", "op_name", "shape", "channels_last", "dim",
         "output_size", "P", "reduce", "kernel_size", "stride", "replacement",
         "num_samples", "scale_factor", "mode", "padding_mode", "align_corners",
-        "shifts", "affine", "backward", "time(us)"
+        "shifts", "affine", "distinct_indices", "backward", "time(us)"
     ]
 
     for log_file in glob.glob(os.path.join(log_dir, "*.log")):
--- a/src/ATen/native/xpu/sycl/DepthwiseConv2dKernels.cpp
+++ b/src/ATen/native/xpu/sycl/DepthwiseConv2dKernels.cpp
@@ -31,7 +31,6 @@
 #endif
 
 namespace at::native::xpu {
-#define NUM_THREADS 1024;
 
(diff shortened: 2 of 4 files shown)


guangyey (approved):

guangyey:
@RUIJIEZHONG66166  could you please help review this PR?

RUIJIEZHONG66166 (approved):

guangyey:
@torchxpubot merge -f

guangyey:
@torchxpubot merge -f

PR #4334, opened 2026-07-14
index_add/index_reduce: raise catchable error on out-of-range index

### LINKED ISSUE #4284: [xpu-alignment] index_add with negative out-of-bounds index causes XPU kernel SIGABRT instead of Python RuntimeError
### 🐛 Describe the bug

`torch.index_add` with a negative index aborts the Python process on XPU instead of raising a catchable Python out-of-bounds error.

This was found during XPU alignment for upstream PyTorch issue https://github.com/pytorch/pytorch/issues/185885. The upstream issue is an Inductor eager-parity bug: compiled `torch.index_add` accepts an invalid negative index while eager mode raises. The XPU result is a related but different failure: even eager XPU execution lets the invalid index reach the XPU kernel, where a `SYCL_KERNEL_ASSERT` aborts the process.

### Reproducer

```python
import torch

print("torch", torch.__version__)
print("xpu available", torch.xpu.is_available())

device = "xpu"
inp = torch.zeros(4, device=device)
source = torch.ones(1, device=device)
neg_index = torch.tensor([-1], device=device, dtype=torch.long)

inp.index_add(0, neg_index, source)
```

### Actual behavior

The process aborts with an XPU kernel assertion instead of raising a Python exception:

```text
AssertHandler::printMessage
/__w/pytorch/pytorch/third_party/torch-xpu-ops/src/ATen/native/xpu/sycl/Indexing.cpp:914:
operator(): Assertion `dstIndex < dstAddDimSize_` failed.
```

The local alignment run classified this as `related-failure`.

### Expected behavior

XPU should report a catchable Python error for the invalid index, matching eager CPU semantics instead of terminating the process.

Note that the exact error depends on the rank of `self` (verified on CPU):

- 1-D `self`, as in the reproducer above:

```text
IndexError: index out of range in self
```

- `self` with more than one dimension, e.g. `inp = torch.zeros(4, 3)` with `source = torch.ones(1, 3)`:

```text
RuntimeError: index -1 is out of bounds for dimension 0 with size 4
```

The reproducer in this issue uses a 1-D `self`, so the expected result is the `IndexError` form. A previous revision of this section quoted the multi-dimensional `RuntimeError` message, which does not match the reproducer. Tests should assert the message corresponding to the rank under test rather than accepting either error type.

### Alignment metadata

- Upstream source: https://github.com/pytorch/pytorch/issues/185885
- Source type: upstream issue
- Scan window: 2026-06-01 to 2026-06-07
- Local XPU result: related-failure
- Routed area: `intel/torch-xpu-ops`, XPU indexing / kernel assertion behavior

### Notes

This issue should track the XPU hard-abort behavior. It should not be treated as fully resolved by a host-side synchronous bounds check unless the proposed fix also preserves expected async behavior and aligns with CUDA-style error handling. A broader fix at the XPU assertion/runtime layer may be preferable if the same `SYCL_KERNEL_ASSERT` abort pattern affects other kernels.

### Versions

```text
PyTorch version: 2.14.0.dev20260623+xpu
Python version: 3.10.12
Is XPU available: True
Relevant path: third_party/torch-xpu-ops/src/ATen/native/xpu/sycl/Indexing.cpp
```

sgurwinderr:
## Summary
On XPU, `index_add`/`index_reduce` with an out-of-range index aborts the process with an uncatchable `SIGABRT` (in assert-enabled builds) or silently produces a wrong result (in release/NDEBUG builds), instead of raising a Python-catchable error as CPU/CUDA do. This adds a host-side bounds check so out-of-range indices are reported catchably on all build types.

## Root cause
An out-of-bounds (or negative) index reaches the kernel, where `dstIndex` is `unsigned` in the 32-bit indexing path. A bad index wraps to a huge value and trips the device-side `SYCL_KERNEL_ASSERT`, which calls `abort()` → `SIGABRT`. Because it fires on-device it cannot be caught from Python, and in release builds where the assert is compiled out the bad index instead silently corrupts/no-ops.

## Fix
Add a host-side bounds check (`at::aminmax` + `TORCH_CHECK_INDEX`) in both `index_add_kernel` and `index_reduce_func_xpu_template`, placed after the `self_` view is built and before kernel launch. Out-of-range indices now raise a catchable `IndexError`/`RuntimeError` matching eager CPU/CUDA. Added `#include <ATen/ops/aminmax.h>`. Empty-index calls are gated by `index.numel() > 0`.

Valid indices are `[0, dim_size - 1]`. **Negative indices are rejected, not wrapped**, matching eager CPU: `aten/src/ATen/native/TensorAdvancedIndexing.cpp` checks `(self_i >= 0) && (self_i < self_dim_size)` on both the sliced and the 1-D fast path for `index_add` and `index_reduce`. The XPU kernels cannot wrap either — `IndexFuncSmallIndexFunctor` and the large-index functor compute an unsigned `dstIndex` and only assert `dstIndex < dstAddDimSize_`. The error message reuses CPU's `index out of range in self` wording so tests matching on that substring behave identically on XPU.

Perf note: the check uses `.item<int64_t>()`, introducing a host-device sync per call — accepted as the cost of correctness parity. Happy to move it device-side behind a flag if maintainers object to the sync.

## Test plan
`test_index_add_reduce_out_of_range_index` in `test/xpu/test_indexing_xpu.py`: indices `4`, `-1` and `-5` on dim size 4 raise a catchable error for both `index_add` and `index_reduce`, the same indices are cross-checked as rejected on CPU, and a valid index (`3`) still produces the correct result.

Verified on Arc Pro B60 (from-source build):
- Out-of-range → `IndexError` (catchable), process does not abort.
- Stock build: same out-of-range index silently returns a wrong result (release build), confirming the pre-fix hazard.

CPU reference behavior confirmed on stock PyTorch: `torch.zeros(4).index_add(0, torch.tensor([-1]), torch.ones(1))` → `IndexError: index out of range in self`.

## Risk
Low blast radius. Adds a validation-only guard on the XPU `index_add`/`index_reduce` path; correct in-range usage is unchanged aside from an added sync. No effect on CPU/CUDA.

Note: touches the same `index_add`/`index_reduce` code as open PR #3061 (small-index refactor); will need a rebase/coordination if that lands first.

Fixes #4284.

### DIFF
--- a/src/ATen/native/xpu/sycl/Indexing.cpp
+++ b/src/ATen/native/xpu/sycl/Indexing.cpp
@@ -28,6 +28,7 @@ DISABLE_RETURN_TYPE_WARNING_BEGIN
 #include <ATen/native/xpu/sycl/SortingKernels.h>
 #include <ATen/native/xpu/sycl/pstl/PSTLFunctions.h>
 #include <ATen/ops/_sparse_coo_tensor_with_dims_and_tensors.h>
+#include <ATen/ops/aminmax.h>
 #include <ATen/ops/arange.h>
 #include <ATen/ops/empty.h>
 #include <ATen/ops/gather.h>
@@ -1227,6 +1228,19 @@ void index_reduce_add_xpu_template(
   const Tensor self_ = (result.dim() == 0) ? result.view(1) : result;
   const Tensor source_ = (source.dim() == 0) ? source.view(1) : source;
 
+  // Host-side bounds check so an out-of-range index raises a catchable
+  // RuntimeError/IndexError instead of tripping the device-side
+  // SYCL_KERNEL_ASSERT (which aborts the process). Matches eager CPU/CUDA.
+  if (index.numel() > 0) {
+    auto [idx_min, idx_max] = at::aminmax(index);
+    int64_t dim_size = self_.size(dim);
+    TORCH_CHECK_INDEX(
+        idx_max.item<int64_t>() < dim_size &&
+            idx_min.item<int64_t>() >= -dim_size,
+        "index_add(): index out of range: expected indices in [",
+        -dim_size, ", ", dim_size - 1, "]");
+  }
+
   TORCH_CHECK(
       result.dim() <= XPU_MAX_TENSORINFO_DIMS,
       "tensor has too many (>",
@@ -1498,6 +1512,19 @@ void index_reduce_func_xpu_template(
   // Scalars are treated as 1-d tensor
   Tensor self_ = (result.dim() == 0) ? result.view(1) : result;
   Tensor source_ = (source.dim() == 0) ? source.view(1) : source;
+
+  // Host-side bounds check so an out-of-range index raises a catchable
+  // RuntimeError/IndexError instead of tripping the device-side
+  // SYCL_KERNEL_ASSERT (which aborts the process). Matches eager CPU/CUDA.
+  if (index.numel() > 0) {
+    auto [idx_min, idx_max] = at::aminmax(index);
+    int64_t dim_size = self_.size(dim);
+    TORCH_CHECK_INDEX(
+        idx_max.item<int64_t>() < dim_size &&
+            idx_min.item<int64_t>() >= -dim_size,
+        "index_reduce(): index out of range: expected indices in [",
+        -dim_size, ", ", dim_size - 1, "]");
+  }
 
   TORCH_CHECK(
       result.dim() <= XPU_MAX_TENSORINFO_DIMS,
--- a/src/ATen/native/xpu/sycl/Indexing.cpp
+++ b/src/ATen/native/xpu/sycl/Indexing.cpp
@@ -1230,15 +1230,16 @@ void index_reduce_add_xpu_template(
 
   // Host-side bounds check so an out-of-range index raises a catchable
   // RuntimeError/IndexError instead of tripping the device-side
-  // SYCL_KERNEL_ASSERT (which aborts the process). Matches eager CPU/CUDA.
+  // SYCL_KERNEL_ASSERT (which aborts the process). Negative indices are
+  // rejected rather than wrapped, matching eager CPU/CUDA.
   if (index.numel() > 0) {
     auto [idx_min, idx_max] = at::aminmax(index);
     int64_t dim_size = self_.size(dim);
     TORCH_CHECK_INDEX(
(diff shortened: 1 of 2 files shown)


laifenxiawucha:
Thanks! Minor detail: eager CPU rejects negative indices (including -1). Could you update the check to >= 0 and add -1 to the test cases? This ensures XPU matches CPU and covers the reproducer.

sgurwinderr:
Good catch — confirmed and fixed in ab09d83.

You're right that CPU rejects every negative index. `aten/src/ATen/native/TensorAdvancedIndexing.cpp` checks `(self_i >= 0) && (self_i < self_dim_size)` on both the sliced and the 1-D fast path, for `index_add` and `index_reduce` alike, and stock PyTorch confirms it:

```
>>> torch.zeros(4).index_add(0, torch.tensor([-1]), torch.ones(1))
IndexError: index out of range in self
```

Worth noting the XPU kernels can't wrap either: `IndexFuncSmallIndexFunctor` and the large-index functor compute an unsigned `dstIndex` and only assert `dstIndex < dstAddDimSize_`, so `-1` becomes a huge unsigned value rather than `dim_size - 1`. My earlier "valid `-1` still works" claim came from a release build where the assert was compiled out — it was an unchecked OOB access, not correct wrapping. Thanks for catching it.

Changes:
- `idx_min >= -dim_size` → `idx_min >= 0` in both `index_reduce_add_xpu_template` and `index_reduce_func_xpu_template`.
- Message now reuses CPU's `index out of range in self` wording, so upstream tests matching on that substring behave the same on XPU.
- Test loop extended to `(4, -1, -5)`, with a CPU cross-check asserting the same rejection; the positive case now uses an in-range index.
- PR description corrected — it previously claimed negative indices were unaffected.

sgurwinderr:
@laifenxiawucha Can you please review this?

laifenxiawucha:
@sgurwinderr  Thanks for the fix! I've corrected the Expected behavior in #4284 — the quoted message was the multi-dimensional one and didn't match its 1-D reproducer. Sorry about that.
On CPU the error depends on the rank of self: 1-D raises `IndexError: index out of range in self`, higher ranks raise `RuntimeError: index -1 is out of bounds for dimension 0 with size 4`. This PR always raises the 1-D form. Could you cover both and assert the exact message for each? The current `assertRaises((IndexError, RuntimeError))` accepts either and wouldn't catch this. And this would also benefit from other reviewers.

PR #4339, opened 2026-07-14
Add helper function for replacing `onlyCUDA` with `onlyOn`

pbielak:
- Add `retarget_outermost_onlycuda_to_onlyon()` to `xpu_test_utils.py`: replaces an outermost `onlyCUDA` wrapper with `onlyOn(["cuda", "xpu"])` by introspecting the closure, avoiding the need to duplicate test bodies.

- Migrate `test_cublas_baddbmm_large_input` in `test_matmul_cuda_xpu.py`: drop the full local copy and use `retarget_outermost_onlycuda_to_onlyon` on the upstream method. Remove unused imports (`unittest`, `IS_JETSON`, `MI200_ARCH`, `parametrize`, `runOnRocmArch`).

### DIFF
--- a/test/xpu/test_matmul_cuda_xpu.py
+++ b/test/xpu/test_matmul_cuda_xpu.py
@@ -12,7 +12,6 @@
 
 # Owner(s): ["module: intel"]
 
-import unittest
 
 import torch
 from torch.testing._internal.common_device_type import (
@@ -23,19 +22,12 @@ from torch.testing._internal.common_device_type import (
     tol as xtol,
     toleranceOverride,
 )
-from torch.testing._internal.common_utils import (
-    IS_JETSON,
-    MI200_ARCH,
-    parametrize,
-    run_tests,
-    runOnRocmArch,
-    TestCase,
-)
+from torch.testing._internal.common_utils import run_tests, TestCase
 
 try:
-    from xpu_test_utils import XPUImportCtx
+    from xpu_test_utils import retarget_outermost_onlycuda_to_onlyon, XPUImportCtx
 except Exception:
-    from .xpu_test_utils import XPUImportCtx
+    from .xpu_test_utils import retarget_outermost_onlycuda_to_onlyon, XPUImportCtx
 
 with XPUImportCtx(False):
     from test_matmul_cuda import TestMatmulCuda, TestMixedDtypesLinearCuda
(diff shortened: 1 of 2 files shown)


test/xpu/nn/test_pooling_xpu.py
```diff
@@ -44,6 +44,11 @@
         TestPoolingNN,
         TestPoolingNNDeviceType,
     )
+    from torch.testing._internal.common_device_type import dtypesIfXPU
+
+    TestPoolingNNDeviceType.test_max_pool_nan_inf = dtypesIfXPU(
+        torch.half, torch.float, torch.double
+    )(TestPoolingNNDeviceType.test_max_pool_nan_inf)
```
BBBela: Do we need this to be imported and overrode inside the `XPUPatchForImport` context manager?

pbielak: Yes, we do - otherwise the dtypesIfXPU will raise a "dtypes redefinition" error (caused by the auto-replacement of dtypesIfCUDA->dtypesIfXPU).

I will try to migrate this file to use XPUImportCtx to avoid confusion

BBBela: So is my understanding correct that we need to import it within `XPUPatchForImport` ctx manager, but we wouldn't have to with the `XPUImportCtx`, right?

pbielak: Exactly

BBBela: Thanks for clarification! 😉

BBBela (approved):
Thanks!

Silv3S (approved):

test/xpu/nn/test_pooling_xpu.py
```diff
@@ -46,6 +49,92 @@
...
+# 2. Retarget @onlyCUDA to @onlyOn(["cuda", "xpu"])
+# ======================================================================
+
+TestPoolingNNDeviceType.test_adaptive_avg_pool2d_backward_large_index_offsets = largeTensorTest(
+    "10GB", device="xpu"
+)(
+    retarget_outermost_onlycuda_to_onlyon(
+        TestPoolingNNDeviceType.test_adaptive_avg_pool2d_backward_large_index_offsets
+    )
+)
```
BBBela: I would not overcomplicate the porting.
I mean, we settled that we change the outermost `@onlyCUDA` to `@onlyOn(["cuda", "xpu"])` and I think we should not overcomplicate things and simply copy the whole test function for other cases.
This double decorator applying in my opinion becomes more complex than simple test function copy.

BBBela (commented):
My general opinion is that we shouldn't mix introducing new helper function and porting next file. The `test_matmul_cuda_xpu.py` seems reasonable as the only change is starting using the introduced helper function, but for me porting the `test_pooling_xpu.py` should be split to another PR.

test/xpu/nn/test_pooling_xpu.py
pbielak: Done

BBBela (approved):
Thanks!

pbielak:

PR #4340, opened 2026-07-14
Migrate `test_tensor_creation_ops_xpu.py` to use `XPUImportCtx`

pbielak:
Replace the full copy of upstream `test/test_tensor_creation_ops.py` with a wrapper that imports upstream via `XPUImportCtx` and keeps only XPU-specific deltas.

Overridden test methods:
- `test_float_to_int_conversion_finite`: add XPU divergent-conversion guards alongside existing HIP/CUDA branches.
- `test_empty_full`: add "xpu" device-type branch for strided layout.
- `test_as_tensor`: remove `@onlyCPU`, add XPU device branches for tensor aliasing and numpy-to-device copy tests.
- `test_tensor_ctor_device_inference`: add `xpu` device-context branch for sparse constructor inference.
- `test_randperm_device_compatibility`: replace `@onlyCUDA` with `@onlyOn(["cuda", "xpu"])` and generalize device references.
- `test_tensor_device`: add explicit XPU branch and CUDA-parity checks

Decorator-only adjustments:
- `dtypesIfXPU`: `test_signal_window_functions`, `test_logspace_device_vs_cpu`, `test_logspace_base2`, `test_logspace_special_steps`, `test_logspace`, `test_uniform_from_to`.
- `largeTensorTest("40GB", "xpu")`: `test_randperm_large`.

Retarget outermost `@onlyCUDA` to `@onlyOn(["cuda", "xpu"])`:
- `test_cat_channels_last_large_inputs`, `test_cat_out_memory_format`, `test_cat_stack_cross_devices`, `test_cat`, `test_new_tensor_device`, `test_range_factories_64bit_indexing`, `test_device_rounding`, `test_copy_from_tensor_mult_devices`, `test_copy_from_dlpack_mult_devices`, `test_unsupported_alias_mult_devices`.
- `test_device_rounding` additionally gets `dtypesIfXPU(half, float, double)`.

Keep XPU-only added tests:
- `test_zeros_large`, `test_ones_large` (2^31-1 element allocation).

Test results:

  Before: 596 passed, 89 skipped, 28 subtests passed
  After:  631 passed, 85 skipped, 28 subtests passed
  \-----------------------------------------------
  Delta:   +35 passed,  -4 skipped,  0 regressions

Newly passing tests (35):
  - XPU tests enabled by migration overrides (SKIPPED -> PASSED):
    - TestTensorCreationXPU::test_as_tensor_xpu
    - TestTensorCreationXPU::test_new_methods_requires_grad_xpu
    - TestTensorCreationXPU::test_offset_scalar_cast_xpu
    - TestTensorCreationXPU::test_simple_scalar_cast_xpu
    - TestTensorCreationXPU::test_stack_out_xpu
    - TestTensorCreationXPU::test_stack_xpu
    - TestTensorCreationXPU::test_tensor_from_sequence_xpu
  - CPU TestFromBlob coverage now instantiated from upstream:
    - TestFromBlobCPU::test_2d_cpu_{complex128,complex64,float16,float32,float64,int16,int32,int64,int8,uint8}
    - TestFromBlobCPU::test_basic_cpu_{complex128,complex64,float16,float32,float64,int16,int32,int64,int8,uint8}
    - TestFromBlobCPU::test_default_dtype_cpu
    - TestFromBlobCPU::test_explicit_device_cpu
    - TestFromBlobCPU::test_shared_memory_cpu_{float32,int64}
    - TestFromBlobCPU::test_strides_cpu_{float32,float64,int32,int64}

Newly skipped tests (3):
  - TestFromBlobCPU::test_invalid_device_cpu
  - TestTensorCreationXPU::test_tensor_factory_gpu_type_inference_xpu
  - TestTensorCreationXPU::test_tensor_factory_gpu_type_xpu


pbielak:
Waiting for https://github.com/intel/torch-xpu-ops/pull/4339 to be merged

test/xpu/test_tensor_creation_ops_xpu.py
```diff
@@ -10,5281 +10,438 @@
...
-        double_min = torch.finfo(torch.double).min
-        double_max = torch.finfo(torch.double).max
-
-        if dtype == torch.bfloat16:
-            min_val = -3.389531389251535e38
-            max_val = 3.389531389251535e38
+# ======================================================================
+# 1. Override upstream test methods for XPU migration adjustments
+# ======================================================================
+@onlyNativeDeviceTypes
```
BBBela, line 208: Could we add empty lines after those section comments? I believe that is how it was done for previous migration PRs - to make the ported files consistent in terms of formatting. Nothing blocking - up to you 😉

test/xpu/test_tensor_creation_ops_xpu.py
```diff
@@ -10,5281 +10,438 @@
...
+            "test_unsupported_alias_mult_devices",
+        ],
+    ),
+):
+    for _method_name in _method_names:
+        setattr(
+            _cls,
+            _method_name,
+            retarget_outermost_onlycuda_to_onlyon(getattr(_cls, _method_name)),
+        )
```
BBBela: I know that it is a good way not to repeat the same code several times and overriding every test function separately, but on the other hand, for this temporary test migration my opinion is that we should be explicit and make the overrides as simple as possible - separately for every test function. Not blocking - up to you 😉

test/xpu/test_tensor_creation_ops_xpu.py
```diff
@@ -10,5281 +10,438 @@
...
+        self.assertIsNot(y, torch.as_tensor(y, device="cuda"))
+        y_cuda = y.to("cuda")
+        self.assertIs(y_cuda, torch.as_tensor(y_cuda))
+        self.assertIs(y_cuda, torch.as_tensor(y_cuda, device="cuda"))
+
+    if torch.xpu.is_available():
+        self.assertIsNot(y, torch.as_tensor(y, device="xpu"))
+        y_xpu = y.to("xpu")
+        self.assertIs(y_xpu, torch.as_tensor(y_xpu))
+        self.assertIs(y_xpu, torch.as_tensor(y_xpu, device="xpu"))
```
BBBela: It looks like it could be made generic instead of branching on device type, right?

test/xpu/test_tensor_creation_ops_xpu.py
```diff
@@ -10,5281 +10,438 @@
...
+        self.assertEqual(torch.tensor(n, device="cuda"), n_astensor)
+        n_astensor[0][2] = 250.9
+        self.assertNotEqual(torch.tensor(n, device="cuda"), n_astensor)
+
+    if torch.xpu.is_available():
+        n = np.random.randn(5, 6)
+        n_astensor = torch.as_tensor(n, device="xpu")
+        self.assertEqual(torch.tensor(n, device="xpu"), n_astensor)
+        n_astensor[0][2] = 250.9
+        self.assertNotEqual(torch.tensor(n, device="xpu"), n_astensor)
```
BBBela: It looks like it could be made generic instead of branching on device type, right?

test/xpu/test_tensor_creation_ops_xpu.py
```diff
@@ -10,5281 +10,438 @@
...
-            ).t()
-            self.assertFalse(non_contiguous_tensor.is_contiguous())
-            with torch.random.fork_rng(devices=rng_device, device_type="xpu"):
-                res = torch.randperm(n, dtype=torch.long, device=device)
-            torch.randperm(n, out=non_contiguous_tensor)
-            self.assertEqual(non_contiguous_tensor, res)
-            self.assertEqual(res.sort().values.long(), torch.arange(n, device=device))
+            if self.device_type == "xpu":
+                with torch.xpu.device(device):
+                    self.assertEqual(op(values.cpu()).device, torch.device("cpu"))
```
BBBela: It looks like it could be made generic instead of branching on device type, right?

test/xpu/test_tensor_creation_ops_xpu.py
```diff
@@ -10,5281 +10,438 @@
...
-        original = torch.arange(5, device="xpu")
+# ======================================================================
+# 4. Add XPU-only large tensor creation tests
+# ======================================================================
+@dtypes(*all_types_and_complex_and(torch.half, torch.bool, torch.bfloat16))
+@largeTensorTest(
+    lambda self, device, dtype: (2**31) * torch.tensor([], dtype=dtype).element_size()
+)
+def _test_zeros_large(self, device, dtype):
+    output = torch.zeros(2**31 - 1, device=device, dtype=dtype)
```
BBBela:
```suggestion
    _ = torch.zeros(2**31 - 1, device=device, dtype=dtype)
```
This way we avoid introducing the unused variable that may cause future linter warnings.

test/xpu/test_tensor_creation_ops_xpu.py
```diff
@@ -10,5281 +10,438 @@
...
-        self.assertNotEqual(original.data_ptr(), tensor.data_ptr())
+TestLikeTensorCreation.test_zeros_large = _test_zeros_large
 
 
+@dtypes(*all_types_and_complex_and(torch.half, torch.bool, torch.bfloat16))
+@largeTensorTest(
+    lambda self, device, dtype: (2**31) * torch.tensor([], dtype=dtype).element_size()
+)
+def _test_ones_large(self, device, dtype):
+    output = torch.ones(2**31 - 1, device=device, dtype=dtype)
```
BBBela:
```suggestion
    _ = torch.ones(2**31 - 1, device=device, dtype=dtype)
```
This way we avoid introducing the unused variable that may cause future linter warnings.

BBBela (commented):
Shouldn't we add some xpu-specific branch to the `def test_tensor_device(self, devices):` test function? Currently it is a no-op for xpu device type. It is not a problem with porting but could be resolved along this PR or with separate PR to make future porting straightforward - what do you think?

test/xpu/test_tensor_creation_ops_xpu.py
pbielak: Done

test/xpu/test_tensor_creation_ops_xpu.py
pbielak: Done

test/xpu/test_tensor_creation_ops_xpu.py
pbielak: Done

test/xpu/test_tensor_creation_ops_xpu.py
pbielak: Done

test/xpu/test_tensor_creation_ops_xpu.py
pbielak: Done

test/xpu/test_tensor_creation_ops_xpu.py
pbielak: Done

test/xpu/test_tensor_creation_ops_xpu.py
pbielak: Done

BBBela (approved):
Thanks for the changes!

I am only still wondering if we shouldn't change the `if self.device_type in ("cuda", "xpu"):` approach to `if self.device_type != "cpu":` as it is more generic to exclude the cpu than to provide specific list of device types. Then any new device can launch this test without any changes to test function body - what do you think?

pbielak:
> Thanks for the changes!
> 
> I am only still wondering if we shouldn't change the `if self.device_type in ("cuda", "xpu"):` approach to `if self.device_type != "cpu":` as it is more generic to exclude the cpu than to provide specific list of device types. Then any new device can launch this test without any changes to test function body - what do you think?

@BBBela I explicitly used `if self.device_type in ("cuda", "xpu"):` because in the previous version these tests were only executed on CUDA. The test functions are either guarded by `@onlyNativeDeviceTypes` (which means: `NATIVE_DEVICES = ('cpu', 'cuda', 'xpu', 'meta', 'mps', 'mtia', torch._C._get_privateuse1_backend_name())`) or there is no decorator at all (so all devices will be run). We don't know if these code blocks will run on `mps` or other device types, so **not to break** anything later in upstream, I decided to narrow it to CUDA and XPU.

tszulist-hbn (approved):

pbielak:

PR #4346, opened 2026-07-15
Probe the SYCL compiler for the fast-math-disabling flag

cyyever:
The SYCL compiler enables fast-math by default, which strips the inf/NaN checks our kernels rely on (flash-attention masking, CTC `zero_infinity`, inf-norm dispatch). The build disables it, but the flag lived in the per-compiler branches (`-fno-fast-math` / `/fp:strict`) coupled to the `{GNU, MSVC}` whitelist — widening that whitelist without a matching branch silently re-enables fast-math.

This hoists it into one block that picks the platform spelling, verifies the SYCL compiler accepts it via `CHECK_SYCL_FLAG`, and `FATAL`s otherwise. Applied flags are unchanged for the supported GNU/MSVC builds (kept before `/Qfma` / `-ffp-contract=fast` to preserve the strict-FP-plus-FMA ordering).

---


guangyey (approved):
LGTM.

PR #4351, opened 2026-07-15
Use mutable_data_ptr for written buffers in unique kernels

cyyever:
The non-const `data_ptr<T>()` calls in the unique kernels all target buffers that the pstl sort/scan/unique routines write in place, so spell them `mutable_data_ptr<T>()` (the read-only inputs already use `const_data_ptr`). `data_ptr<T>()` is the legacy non-explicit accessor `TensorBase` is migrating away from. No behavior change.

---


guangyey (approved):
Nice!

Issue #4326, opened 2026-07-13
test_tensor_creation_ops_xpu.py::TestTensorCreationXPU::test_block_diag_scipy_xpu AssertionError


bjarzemb:
### 🐛 Describe the bug

`test_tensor_creation_ops_xpu.py::TestTensorCreationXPU::test_block_diag_scipy_xpu` fails on Windows with Intel XPU (Arc B580) with `AssertionError: Object comparison failed: torch.int64 != torch.int32`

torch-xpu commit hash | torch-xpu commit date | status 
-- | -- | -- 
a597fd33 | Jun 30 | passed
d87e686f | Jul 8 | failed

### Affected Test Cases

```
test_tensor_creation_ops_xpu.py::TestTensorCreationXPU::test_block_diag_scipy_xpu
```

### Error Message

<details>

<summary>Click to expand traceback</summary>

```
_______________ TestTensorCreationXPU.test_block_diag_scipy_xpu _______________
[gw0] win32 -- Python 3.12.13 C:\Users\gta\miniforge3\envs\202607091418_140e4d58_32.0.101.8826_2026.0.0.644\python.exe
Traceback (most recent call last):
  File "C:\Users\gta\miniforge3\envs\202607091418_140e4d58_32.0.101.8826_2026.0.0.644\Lib\unittest\case.py", line 58, in testPartExecutor
    yield
  File "C:\Users\gta\miniforge3\envs\202607091418_140e4d58_32.0.101.8826_2026.0.0.644\Lib\unittest\case.py", line 634, in run
    self._callTestMethod(testMethod)
  File "C:\Users\gta\miniforge3\envs\202607091418_140e4d58_32.0.101.8826_2026.0.0.644\Lib\unittest\case.py", line 589, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "C:\Users\gta\miniforge3\envs\202607091418_140e4d58_32.0.101.8826_2026.0.0.644\Lib\site-packages\torch\testing\_internal\common_utils.py", line 3590, in wrapper
    method(*args, **kwargs)
  File "C:\Users\gta\miniforge3\envs\202607091418_140e4d58_32.0.101.8826_2026.0.0.644\Lib\site-packages\torch\testing\_internal\common_device_type.py", line 627, in instantiated_test
    result = test(self, **param_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4352, opened 2026-07-15
Prevent numpy error for block diag on Windows when numpy>=2.0

jmamzax:
From version 2.0 NumPy on Windows uses int64 as default integer: https://numpy.org/devdocs/numpy_2_0_migration_guide.html#windows-default-integer. Checking NumPy version will prevent from the error `AssertionError: Object comparison failed: torch.int64 != torch.int32` when NumPy will be updated on CI. Issue: https://github.com/intel/torch-xpu-ops/issues/4326

### DIFF
--- a/test/xpu/test_tensor_creation_ops_xpu.py
+++ b/test/xpu/test_tensor_creation_ops_xpu.py
@@ -485,10 +485,16 @@ class TestTensorCreation(TestCase):
             torch.float32,
         ]
 
+        # numpy < 2.0 on Windows used int32 as the default integer type
+        # numpy >= 2.0 uses int64 everywhere
+        scipy_int_type = (
+            torch.int32
+            if IS_WINDOWS and np.lib.NumpyVersion(np.__version__) < "2.0.0"
+            else torch.int64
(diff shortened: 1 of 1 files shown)


BBBela (approved):
Looks good to me, thanks!

Functionally it is okay, but I thought about separating this conditional check - something like that:
```python
        # numpy < 2.0 on Windows used int32 as the default integer type
        # numpy >= 2.0 uses int64 everywhere
        scipy_int_type = (
            torch.int32
            if IS_WINDOWS and np.lib.NumpyVersion(np.__version__) < "2.0.0"
            else torch.int64
        )

        expected_scipy_types = [
            torch.float64,
            scipy_int_type,
            torch.complex128,
            torch.float64,
        ]
```

But that is just a suggestion - approving anyway 😉

PatrykWilczewski (approved):

jmamzax:
After changes in xpu-ops and pytorch: https://github.com/pytorch/pytorch/pull/189478/changes there is no longer need for this fix as problem was already fixed in pytorch.

PR #4354, opened 2026-07-15
Migrate `test_aot_joint_with_descriptors_xpu.py` to use XPUImportCtx.

BBBela:
Replace the full copy of upstream test/functorch/test_aot_joint_with_descriptors_xpu.py with a wrapper that imports upstream via XPUImportCtx and keeps only XPU-specific deltas.

## Changes

- change one occurrence of the hardcoded "cuda" device type to "xpu"

## Additional changes

- Added `__init__.py` file to the `functorch` directory to allow relative import of `xpu_test_utils`.
- Added `functorch` path to the PATH in `XPUImportCtx`.

## Test results
| | FAILED | PASSED | SKIPPED | XFAIL |
| - | - | - | - | - |
| Before | 0 | 18 | 0 | 0 |
| After | 0 | 21 | 0 | 0 |
| Delta| 0 | +3 | 0 | 0 |

## Test results delta analysis

Newly passing tests (3):
- TestAOTJointWithDescriptors::test_annotate_fn_anchors_nested_functional_call (new)
- TestAOTJointWithDescriptors::test_annotate_fn_anchors_pure_tensor_ops (new)
- TestAOTJointWithDescriptors::test_annotate_fn_nested_with_module_forward (new)

### DIFF
--- a/test/xpu/functorch/test_aot_joint_with_descriptors_xpu.py
+++ b/test/xpu/functorch/test_aot_joint_with_descriptors_xpu.py
@@ -21,12 +21,12 @@ from torch.nn.attention.flex_attention import create_block_mask, flex_attention
 from torch.testing._internal.common_utils import run_tests, TEST_CUDA, TEST_XPU
 
 try:
-    from xpu_test_utils import XPUImportCtx
-except Exception:
     from .xpu_test_utils import XPUImportCtx
+except Exception:
+    from ..xpu_test_utils import XPUImportCtx
 
 with XPUImportCtx(False):
-    from functorch.test_aot_joint_with_descriptors import (
+    from test_aot_joint_with_descriptors import (
         graph_capture,
         TestAOTJointWithDescriptors,
     )
--- a/test/xpu/functorch/test_aot_joint_with_descriptors_xpu.py
+++ b/test/xpu/functorch/test_aot_joint_with_descriptors_xpu.py
@@ -5,1256 +5,144 @@
 # You may obtain a copy of the License at
 #
 # http://www.apache.org/licenses/LICENSE-2.0
-
-# flake8: noqa
+#
+# Portions of this file are derived from PyTorch
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: BSD-3-Clause
 
 # Owner(s): ["module: intel"]
-
-from contextlib import ExitStack
+# ruff: noqa: F401
 
 import torch
 import torch.fx.traceback as fx_traceback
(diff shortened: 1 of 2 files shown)


test/xpu/functorch/test_aot_joint_with_descriptors_xpu.py
```diff
@@ -5,1256 +5,144 @@
...
+# ======================================================================
 
-        model = MultiOutputModule()
-        inputs = (torch.randn(4, 3),)
 
-        with ExitStack() as stack:
-            joint_with_descriptors = aot_export_joint_with_descriptors(
-                stack, model, inputs, decompositions=decomposition_table
-            )
+def _test_preserve_annotate_flex_attention(self):
```
pbielak, line 47: When porting this function to upstream, we will need to:
- add a `device` parameter
- add the `onlyOn(["cuda", "xpu"])` decorator
- add an `instatiate_device_type_tests` call to this class

To make the upstreaming easier in future, please add the mentioned changes in this PR. You could also consider, to extract this function to a new class e.g.,`TestAOTJointWithDescriptorsDeviceType` and run `instantiate_...` on this class only.

pbielak (changes requested):
With the mentioned changes (i.e. if a class will be extracted) the test names will change - please add an information about these in the PR description and check skip lists for this function. Thanks!

test/xpu/functorch/test_aot_joint_with_descriptors_xpu.py
BBBela: Good point, thanks!
Although I checked that using:
```python
device_type = (
    acc.type if (acc := torch.accelerator.current_accelerator(True)) else "cpu"
)
```
is a more common way in upstream for tests that aren't currently instantiated with `instatiate_device_type_tests` unless you have any objection on that?

pbielak: Sure, we can use that solution

BBBela: Applied changes, please look at them.

BBBela:
I added the following changes to fix issues discovered by the CI results:
- Added `__init__.py` file to the `functorch` directory to allow relative import of `xpu_test_utils`.
- Added `functorch` path to the PATH in `XPUImportCtx`.

pbielak (approved):

kdrozd-dev (approved):

BBBela:
@torchxpubot merge

PR #4359, opened 2026-07-16
add ut classification and test refactory skills under valiation folde…

daisyden:
…r, added extract-basic-info skills and scripts


daisyden:
@torchxpubot review

Issue #4256, opened 2026-07-07
[release/2.14] test_transformers_xpu.py::TestSDPACudaOnlyXPU::test_fused_attention_vs_math_ref_grads_cudagraph_batch_size* RuntimeError: flash_attention_forward: Torch XPU was not compiled with SYCLTLA support.


bjarzemb:
### 🐛 Describe the bug

48 UTs `test_transformers_xpu.py::TestSDPACudaOnlyXPU::test_fused_attention_vs_math_ref_grads_cudagraph_batch_size*` fail on Windows with Intel XPU (Arc B580) with `RuntimeError: flash_attention_forward: Torch XPU was not compiled with SYCLTLA support.`.

torch-xpu commit hash | torch-xpu commit date | status | notes
-- | -- | -- | --
d67a8778 | Apr 23 | skipped | 
bc294243 | Jun 9 | skipped | release branch
390cceb8 | Jun 16 | failed | 
a597fd33 | Jun 30 | failed | 

### Affected Test Cases

```
test_transformers_xpu.py::TestSDPACudaOnlyXPU::test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_1_seq_len_q_1024_seq_len_k_1024_head_dim_32_is_causal_False_dropout_p_0_0_float16_scale0_fused_kernel0_xpu_float16
test_transformers_xpu.py::TestSDPACudaOnlyXPU::test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_1_seq_len_q_1024_seq_len_k_1024_head_dim_32_is_causal_False_dropout_p_0_0_float16_scale_l1_fused_kernel0_xpu_float16
test_transformers_xpu.py::TestSDPACudaOnlyXPU::test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_1_seq_len_q_1024_seq_len_k_1024_head_dim_32_is_causal_True_dropout_p_0_0_float16_scale0_fused_kernel0_xpu_float16
test_transformers_xpu.py::TestSDPACudaOnlyXPU::test_fused_attention_vs_math_ref_grads_cudagraph_batch_size_1_seq_len_q_1024_seq_len_k_1024_head_dim_32_is_causal_True_dropout_p_0_0_float16_scale_l1_fused_kernel0_xpu_float16
[...]

PR #4368, opened 2026-07-16
Add PLATFORM_SUPPORTS_FLASH_ATTENTION_XPU for flash attention tests

jmamzax:
https://github.com/intel/torch-xpu-ops/issues/4256 Test with flash attention are wrongly executed on Windows which don't support flash attention.


PatrykWilczewski (approved):

BBBela (commented):
The `PLATFORM_SUPPORTS_FLASH_ATTENTION` already has the `if TEST_XPU:` branch. I think it should be adjusted directly there instead of using separate `PLATFORM_SUPPORTS_XPU_FLASH_ATTENTION`.

[MENTIONED CODE FRAGMENT](https://github.com/pytorch/pytorch/blob/c46c8102908fb51e24b3a1ce4c5e9eb64236f565/torch/testing/_internal/common_cuda.py#L114-L131)

One additional comment is that, from what I know, there is an assumption in the codebase that there are no both CUDA and XPU devices on the same platform. So we have always only one of them currently accessible.

jmamzax:
@BBBela Thanks for pointing this out. I change it to PLATFORM_SUPPORTS_FLASH_ATTENTION_XPU which is implemented in common_xpu.py and is already used in some tests.

test/xpu/higher_order_ops/test_invoke_subgraph_xpu.py
```diff
@@ -55,9 +55,7 @@
 
 nested_compile_region = torch.compiler.nested_compile_region
 
-PLATFORM_SUPPORTS_XPU_FLASH_ATTENTION = (
-    torch.xpu.is_available() and torch._C._is_flash_attention_available()
-)
+from torch.testing._internal.common_xpu import PLATFORM_SUPPORTS_FLASH_ATTENTION_XPU
```
BBBela: Let's move this import just below the `from torch.testing._internal.common_utils` to preserve the alphabetical order of imports.

BBBela (approved):
Thanks! 😉

I still think that the long term solution would be to adjust the `PLATFORM_SUPPORTS_FLASH_ATTENTION` directly in upstream to handle the XPU case, but I think that for now we can make it that way and adjust upstream during porting - especially as you noticed that this solution is already used in the torch-xpu-ops codebase, so anyway it needs to be ported.

Just please look at the one comment below.

jmamzax:
@torchxpubot merge -f

PR #4370, opened 2026-07-17
Support multi-config generators in the SYCL CMake build

cyyever:
Multi-config generators (Ninja Multi-Config, Visual Studio) did not build correctly: per-config flags keyed off `CMAKE_BUILD_TYPE`, which is empty under multi-config, and FindSYCL object paths used the deprecated `CMAKE_CFG_INTDIR`, which stays `.` under Ninja Multi-Config so every configuration collided on the same outputs.

Three commits, in review order:

1. **flags** - per-config host/kernel flags become `$<CONFIG:>` genexes. `TORCH_XPU_OPS_FLAGS` stays config-agnostic and defers to PyTorch's per-config flags (`CMAKE_CXX_FLAGS_DEBUG`, `MSVC_Z7_OVERRIDE`) instead of duplicating/fighting them.
2. **paths** - per-config object and dependency subpaths under multi-config generators only (single-config layout unchanged); same-named sources from different directories are disambiguated by a source-dir hash rather than the old FindCUDA mirrored-subdirectory scheme.
3. **guard** - settle `USE_SYCLTLA` before its `include()` so debug builds skip it.

Authored with Claude Code assistance.


guangyey (commented):
@chunhuanMeng could you please help review this PR.

chunhuanMeng (approved):

guangyey (approved):
No bandwidth to go through this PR.

PR #4371, opened 2026-07-17
Turn Philox and Reduce #define constants into inline constexpr and reuse uniform helpers

cyyever:
Replace the object-like macros in `Philox4x32.h` and `Reduce.h` (`RAND_2POW32_INV*`, `RAND_2POW53_INV_DOUBLE`, `DEFAULT_OUTPUT_VEC_SIZE`) with `inline constexpr`. Pi constants come from `std::numbers`, powers of two use hex-float literals, and the duplicated uniform-mapping expressions now reuse `_rand_uniform`/`_rand_uniform_double_hq`. All values are bit-exact with the old literals and -O2 codegen is unchanged; the lgamma table's log(2!) entry keeps its literal with a NOLINT.

Test: none (build-only refactor, covered by existing random/reduce tests)

### DIFF
--- a/src/ATen/native/xpu/sycl/Philox4x32.h
+++ b/src/ATen/native/xpu/sycl/Philox4x32.h
@@ -47,6 +47,7 @@
 
 #include <ATen/core/DistributionsHelper.h>
 #include <ATen/native/xpu/sycl/MemoryAccess.h>
+#include <numbers>
 
 namespace at {
 namespace native {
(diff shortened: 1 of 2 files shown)


guangyey (approved):
Nice!

guangyey:
@torchxpubot review "Please check if the code change is identical"

guangyey:
@torchxpubot ut-check

PR #4374, opened 2026-07-17
Fix local atomicMax doing fetch_add

cyyever:
The `sycl_local_ptr<int32_t/int64_t>` `atomicMax` overloads called `fetch_add`, so racing threads accumulated instead of taking the max. Reached via MultinomialKernel's `foundPos` update (upstream CUDA uses `atomicMax(&foundPos, cat)` at the same spot).

Test: none (2-line fix; covered by existing multinomial tests, failure needs a same-bucket thread race)


guangyey (approved):
Good catch!

guangyey (commented):
cc @chunhuanMeng

chunhuanMeng (approved):

guangyey:
@torchxpubot check-ut

PR #4378, opened 2026-07-17
LayerNorm : Handling if batch size is zero

xiaowangintel:
# Summary
Align XPU `native_layer_norm` with the CUDA fix from https://github.com/pytorch/pytorch/pull/28614.
Skip the forward kernel launch when `M == 0` (empty batch dimension) to avoid submitting
an invalid `nd_range(0, wg_size)` to the SYCL runtime.


CuiYifeng (commented):
In CUDA's fix, both LayerNorm fwd/bwd kernels are handled for batch size 0. Why not add a check for LayerNorm bwd kernel?
Please also add a simple test case in test/regression, thanks.

xiaowangintel:
> In CUDA's fix, both LayerNorm fwd/bwd kernels are handled for batch size 0. Why not add a check for LayerNorm bwd kernel? Please also add a simple test case in test/regression, thanks.

Based on layer_norm_backward_kernel and layer_norm_backward_xpu, the boundary checks for M > 0 in the backward path are already complete. This scenario is also covered by the upstream PyTorch test suite.

The test passed on previous hardware platforms, but it now fails on the new hardware.

CuiYifeng (approved):
Thanks for the explanation. Please update the PR title to indicate which op is the fix for.

xiaowangintel:
> Thanks for the explanation. Please update the PR title to indicate which op is the fix for.

Done

chunhuanMeng (approved):

CuiYifeng:
@torchxpubot ut-check

CuiYifeng:
@torchxpubot merge -f

PR #4380, opened 2026-07-17
Fix output column index in upsample bilinear2d NotAlign backward

cyyever:
Both branches of `UpsampleBilinear2dBackwardNotAlignKernelFunctor` computed the output column as `(point_w - input_height_) / (2 * input_width_)`; the subtrahend was copy-pasted from the row computation one line above. For portrait inputs (height > width) with `align_corners=False` and integer scales, every input pixel accumulated gradient from the wrong output column; square and landscape shapes coincidentally floored to the correct index, which is why existing tests passed. This fast path is XPU-specific with no CUDA counterpart.

Test: test/regressions/test_upsample_bilinear_bwd.py (new portrait case; existing square case was blind to the bug)


pbielak:
Already done in PR: https://github.com/intel/torch-xpu-ops/pull/4333

guangyey (approved):

jianyizh (approved):

PR #4385, opened 2026-07-17
Pass macro definitions to sycl kernel code

guangyey:
# Motivation
https://github.com/intel/torch-xpu-ops/pull/3629 introduces `SYCL_COMPILER_VERSION` for use in device code, but @NeoZhangJianyu found that it doesn't work as expected.

The reason is that this macro is currently only added to the host compilation. For SYCL device code, compile definitions must be passed through `SYCL_compile_definitions`, so `SYCL_COMPILER_VERSION` is not visible during device compilation.

We fix it in this PR and guard it in `Sleep.cpp` via `#error` syntax.

# Additional Context
Why the CI can't catch the error in https://github.com/pytorch/pytorch/pull/189527?
I found that the test was skipped due to CI lacks of pyzes:
```bash
2026-07-16T05:57:24.5881011Z test_xpu_expandable_segments.py::TestXpu::test_sleep <- test/test_xpu.py SKIPPED [0.0002s] (pyzes is required for this test) [ 43%]
```
See https://github.com/pytorch/pytorch/actions/runs/29469731264/job/87584586906?pr=189527#step:13:3160

### DIFF
--- a/cmake/BuildFlags.cmake
+++ b/cmake/BuildFlags.cmake
@@ -37,12 +37,22 @@ macro(set_build_flags)
     return()
   endif()
   set(SYCL_HOST_FLAGS)
+  set(SYCL_DEVICE_COMPILE_DEFINITIONS)
   set(SYCL_KERNEL_OPTIONS)
   set(SYCL_COMPILE_FLAGS ${SYCL_FLAGS})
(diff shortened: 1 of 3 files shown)


EikanWang (approved):

guangyey:
Build pass. However, clock functionality still has an issue on BMG. I already submitted a JIRA `CMPLRLLVM-76989` to track the issue.

PR #4402, opened 2026-07-21
[SYCL free func] Add enqueue functions for SYCL free function and convert the first kernel repeat_interleave to SYCL free function style

NeoZhangJianyu:
This is the first PR for SYCL free function style code:
1. enhance the sycl_kernel_submit() to support SYCL free function.
2. support oneAPI 26.1 and older (26.0).
3. support windows with 26.1 and older (26.0).

The UT cases of repeat_interleave are passed on local test.


chuanqi129:
@torchxpubot check-ut

tye1:
Pls. note that the perf regression in 26.0 nd_launch is known and has been resolved in 26.1 nd_launch

src/comm/DeviceProperties.h
```diff
@@ -50,6 +54,28 @@ static int64_t syclMaxWorkGroupSize(
...
+  auto q = c10::xpu::getCurrentXPUStream(dev_id).queue();
+  auto ctxt = q.get_context();
+  auto dev = q.get_device();
+  auto exe_bndl =
+      ::syclexp::get_kernel_bundle<kptr, ::sycl::bundle_state::executable>(
+          ctxt);
+  ::sycl::kernel k = exe_bndl.template ext_oneapi_get_kernel<kptr>();
+  return k.get_info<::sycl::info::kernel_device_specific::work_group_size>(dev);
+#else
+  auto& ctx = c10::xpu::get_device_context();
```
fengyuan14, line 73: Do we need a ctx for `get_kernel_info`? If there is a variant API of `get_kernel_info` which doesn't require ctx? Because the device object should imply the information of ctx?

src/comm/SYCLHelpers.h
```diff
@@ -15,10 +15,14 @@ DISABLE_SYCL_DEPRECATED_WARNING_BEGIN
 // host-only compilation (without -fsycl).
 #define SYCL_DISABLE_FSYCL_SYCLHPP_WARNING
 #include <comm/Scalar.h>
+#include <sycl/ext/oneapi/work_group_static.hpp> // Remove it once the header is exposed by sycl.hpp
```
fengyuan14: Could you check if the header has be included in sycl.hpp?

src/comm/SYCLHelpers.h
```diff
@@ -234,6 +238,104 @@ static inline void sycl_kernel_submit(
...
+        ::sycl::nd_range<dim>(global_range, local_range),
+        syclexp::properties{syclexp::work_group_scratch_size(slm_sz)}};
+    syclexp::nd_launch(q, cfg, syclexp::kernel_function<kptr>, args...);
+  } else {
+    syclexp::nd_launch(
+        q,
+        ::sycl::nd_range<dim>(global_range, local_range),
+        syclexp::kernel_function<kptr>,
+        args...);
+  }
```
fengyuan14, line 337: Please check if we can unify and remove the if-else with set `syclexp::properties{syclexp::work_group_scratch_size(0)` for the current else path.

src/comm/DeviceProperties.h
NeoZhangJianyu: There are three functions:
```
get_kernel_info(const context &Ctx)

get_kernel_info(const context &Ctx, const device &Dev)

get_kernel_info(const queue &Q)
```

get_kernel_info(const context &Ctx, const device &Dev) has special comments:
```
// Device-specific query - uses fast kernel cache (O(1) lookup) via the
// ABI-exported get_kernel_info_impl, which performs spec-mandated validation
// and dispatches to get_kernel_device_specific_info inside the library.
```
The original code has been verified with good performance.

But we don't know the performance impact by other API.

src/comm/SYCLHelpers.h
NeoZhangJianyu: yes, it's included in sycl.hpp in both 26.0 and 26.1. remove it

src/comm/DeviceProperties.h
NeoZhangJianyu: Current code is verified with good functionality and performance.
If change the API, need performance evaluation again.
We will optimize the code later.

src/comm/SYCLHelpers.h
NeoZhangJianyu: Yes, it's possible to optimize this code for compact.

Current code is verified with good functionality and performance.
If change it, we need performance evaluation again.

We will optimize the code later.

src/comm/DeviceProperties.h
fengyuan14: Thanks. Please add the comment in source code to explain why we choose the variant.

src/comm/SYCLHelpers.h
fengyuan14: Please add the comment for the TODO.

src/comm/DeviceProperties.h
NeoZhangJianyu: Yes, update with the comment

src/comm/SYCLHelpers.h
NeoZhangJianyu: yes, update with TODO

fengyuan14 (approved):

tye1:
@gujinghui  @EikanWang  pls. help review and comment, thanks.

src/ATen/native/xpu/sycl/RepeatKernel.cpp
```diff
@@ -14,42 +14,30 @@
...
 
 namespace at::native::xpu {
+
 template <typename index_t>
-struct RepeatInterleaveKernelFunctor {
-  void operator()(sycl::nd_item<1> item) const {
-    auto rep_ptr = rep_data_;
-    auto cum_ptr = cum_data_;
-    auto res_ptr = res_data_;
+SYCL_EXT_ONEAPI_FUNCTION_PROPERTY((syclexp::nd_range_kernel<1>))
```
gujinghui, line 19: Where is the `syclexp` namespace defined? If it's not a std namespace, we should re-define it here?

src/comm/SYCLHelpers.h
```diff
@@ -234,6 +237,107 @@ static inline void sycl_kernel_submit(
...
+// TODO: unify and remove the if-else for slm_sz
+template <auto* kptr, typename... Kargs>
+static inline void sycl_kernel_submit(
+    int64_t global_range,
+    int64_t local_range,
+    ::sycl::queue q,
+    int slm_sz,
+    Kargs... args) {
+#if defined(SYCL_COMPILER_VERSION) && SYCL_COMPILER_VERSION < 20260100
+  sycl::context ctxt = q.get_context();
```
gujinghui, line 268: confused me. Why do we sometimes need to use ::sycl, while here the leading :: is not needed?

gujinghui (approved):

mengfei25:
@torchxpubot check-ut

src/comm/SYCLHelpers.h
NeoZhangJianyu: The leading :: in ::sycl::xxx is the C++ global scope resolution operator. It forces the compiler to look for the sycl namespace starting from the absolute global root, ignoring any local namespaces or classes that might accidentally shadow or hide sycl in the current scope. Resolves naming conflicts if your code defines a local class, variable, or nested namespace named sycl. Guarantees that you are using the official standard library/API namespace rather than a custom implementation.

chuanqi129:
@torchxpubot check-ut

tye1:
@torchxpubot check-ut

tye1:
Per @NeoZhangJianyu, first case passed in local test. The second case xfailed in local, should be unrelated to the code change in this PR.

tye1:
@torchxpubot merge -f

tye1:
@torchxpubot merge -f

PR #4408, opened 2026-07-21
Add tolerance override for Muon cross-device state_dict test

### LINKED ISSUE #4263: [release/2.14] Muon optimizer state_dict cross-device numeric tolerance failure
### 🐛 Describe the bug

Cases:
~~op_ut,third_party.torch-xpu-ops.test.xpu.test_optim_xpu.TestOptimRenewedXPU,test_state_dict_cross_device_Muon_xpu_float32~~

## Affected Test Cases

- `test_state_dict_cross_device_Muon_xpu_float32` (`TestOptimRenewedXPU` in `test_optim_xpu.py`)

## Root Cause

The Muon optimizer's `test_state_dict_cross_device` test saves optimizer state on one device and loads it on another, then compares the resulting parameter tensors. The comparison fails with a small numeric tolerance violation: greatest absolute difference of `1.94e-05` (allowed: `1e-05`) and greatest relative difference of `1.88e-05` (allowed: `1.3e-06`). This is a precision issue specific to the Muon optimizer's cross-device state restoration on XPU. The Muon optimizer (which applies Newton-Schulz orthogonalization) accumulates floating-point differences when state tensors are transferred between CPU and XPU, likely due to differences in intermediate computation precision during the orthogonalization steps. This may require either a tolerance override for the XPU Muon cross-device test or investigation into the precision of the Newton-Schulz iteration on XPU.

## Reproducer

```bash
source ~/.bashrc && cd third_party/torch-xpu-ops/test/xpu && python -m pytest -sxv test_optim_xpu.py -k "test_state_dict_cross_device_Muon_xpu_float32"
```

## Representative Error

```
  File "/home/gta/Projects/pytorch/third_party/torch-xpu-ops/test/xpu/../../../../test/test_optim.py", line 1887, in test_state_dict_cross_device
    self.assertEqual(params, params_device)
  File "/home/gta/Projects/pytorch/torch/testing/_internal/common_utils.py", line 4608, in assertEqual
    raise error_metas.pop()[0].to_error(
AssertionError: Tensor-likes are not close!

Mismatched elements: 1 / 6 (16.7%)
Greatest absolute difference: 1.9431114196777344e-05 at index (1, 0) (up to 1e-05 allowed)
Greatest relative difference: 1.8793216440826654e-05 at index (1, 0) (up to 1.3e-06 allowed)

The failure occurred for item [0]
```

### Versions

- PyTorch version: 2.14.0a0+git4dbabff
- Python: 3.10.20 (CI) / 3.12 (local repro)
- OS: Ubuntu 24.04.4 LTS
- XPU available: Yes
- XPU build version: 20260000

AKloniecki:
The Muon optimizer's Newton-Schulz orthogonalization runs in bfloat16 and uses torch.addmm for the iterative polynomial:

    gram_update = addmm(gram, gram, gram, beta=-4.775, alpha=2.0315)
    ortho_grad  = addmm(ortho_grad, gram_update, ortho_grad, beta=3.4445)

The fused addmm kernel rounds differently on CPU vs XPU when the true result falls between two adjacent bf16 representable values.  Both results are valid (within 1 ULP of bf16 precision), but the 1-ULP difference (~3.9e-3 in bf16) amplifies through 5 polynomial iterations, producing additive errors up to ~4.7e-4 in the final float32 parameter update.

Since the error is additive and independent of parameter magnitude, only atol needs relaxation.  rtol stays at the float32 default (1.3e-6).

Tolerance chosen:  atol=7e-4 (70x default), rtol=1.3e-6 (1x default) Empirically validated over 3000 random seeds on Intel Arc B580. Worst observed diff: 4.66e-4; safety margin: 30%.

Fixes: #4263

### DIFF
--- a/test/xpu/test_optim_xpu.py
+++ b/test/xpu/test_optim_xpu.py
@@ -23,8 +23,13 @@ from copy import deepcopy
 
 import torch
 from torch.nn import Parameter
-from torch.testing._internal.common_device_type import TEST_WITH_ROCM
+from torch.testing._internal.common_device_type import (
+    TEST_WITH_ROCM,
+    tol,
+    toleranceOverride,
+)
 from torch.testing._internal.common_dtype import floating_types_and
+from torch.testing._internal.common_methods_invocations import DecorateInfo
 from torch.testing._internal.common_optimizers import (
     _get_optim_inputs_including_global_cliquey_kwargs,
     optim_db,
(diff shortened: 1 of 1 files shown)


BBBela (commented):
Following the advice from [THIS CONVERSATION](https://github.com/intel/torch-xpu-ops/pull/3577#pullrequestreview-4273355596) I think the tolerance override should be applied in the `common_optimizers.py` file in the upstream instead.

AKloniecki:
This PR has been replaced with one in upstream. Closing.

PR #4416, opened 2026-07-21
Migrate `test_sparse_xpu.py` to use `XPUImportCtx`

pbielak:
Replace the full copy of upstream `test/test_sparse.py` with a wrapper
that imports the upstream module via `XPUImportCtx` and only overrides
XPU-specific differences:

- `assertExpected`: local expect-file fallback for XPU print tests
  (`test_print_coalesced`, `test_print_uncoalesced`)
- Retarget `@onlyCUDA` to `@onlyOn(["cuda", "xpu"])`:
  `test_cuda_empty`, `test_storage_not_null`, `test_bmm_oob`,
  `test_same_gpu`,
  `test_coalesce_accepts_large_tensor` (+ `largeTensorTest("30GB", "xpu")`)
- Decorator additions (no body change):
  `test_sparse_matmul` (dtypesIfXPU),
  `test_sparse_addmm` (toleranceOverride)
- `TestSparse` method overrides — device generalization:
  `test_Sparse_to_Sparse_copy_multi_gpu`, `_test_new_device`,
  `test_new_device_single_gpu`, `test_new_device_multi_gpu`,
  `test_factory`, `test_factory_device_type_inference`,
  `test_legacy_new_device`, `test_dtypes`, `test_bmm_deterministic`,
  `test_sparse_mul` (strip outermost skipIf — test works on XPU),
  `_test_empty_like` (widen backend regex to include XPU),
  `test_empty_full` / `_test_empty_full` (override to run on XPU)
- `TestSparseOneOff` overrides — device generalization:
  `test_cuda_from_cpu`, `test_cuda_sparse_cpu_dense_add`
- `TestSparseAny` overrides:
  `test_constructor_pin_memory`, `test_method_pin_memory`,
  `test_constructor_pinned_memory`,
  `test_constructor_mismatched_pinned_memory` (full local overrides
  using `not TEST_CUDA and not TEST_XPU` guard),
  `test_gradcheck_mm` (widen regex to include XPU)
- New XPU-only tests:
  `test_hspmm_out`, `test_hspmm_out_errors`
- Fix `test/xpu/test_sparse_csr_xpu.py`: replace `from test_sparse_xpu
  import (CUSPARSE_SPMM_COMPLEX128_SUPPORTED, ...)` with inline
  definitions, since the migrated `test_sparse_xpu.py` no longer
  exports those constants.
- Remove redundant local expect symlinks:
  `test/xpu/expect/TestSparseCPU.test_print_*.expect` (2 files),
  `test/xpu/expect/TestSparseMeta.test_print_*.expect` (5 files)

Test results:

  Before: 5489 passed,  663 skipped, 0 failed
  After:    5508 passed,  664 skipped, 0 failed
  \----------------------------------------------------------------
  Delta:   +19 passed,   +1 skipped

Newly passing tests from migration (SKIPPED -> PASSED):
  - TestSparseXPU::test_legacy_new_device_xpu
  - TestSparseXPU::test_mm_xpu_{complex128,float64}
  - TestSparseXPU::test_empty_full_requires_grad_{False,True}_xpu_* (17 tests)

Newly passing tests (inherited from upstream, not from migration):
  - TestSparseAnyCPU::test_view_as_complex_cpu_{float16,float32,float64}
  - TestSparseAnyCPU::test_view_as_real_cpu_{complex128,complex32,complex64}
  - TestSparseAnyXPU::test_view_as_complex_xpu_{float16,float32,float64}
  - TestSparseAnyXPU::test_view_as_real_xpu_{complex128,complex32,complex64}
  - TestSparseCPU::test_no_warn_when_check_invariants_is_explicit_cpu_float32
  - TestSparseCPU::test_sspaddmm_wrong_mat_types_error_messages_cpu_{complex128,float64}
  - TestSparseCPU::test_warn_on_sparse_tensor_invariant_checks_disabled_cpu_float32
  - TestSparseXPU::test_no_warn_when_check_invariants_is_explicit_xpu_float32

Newly skipped (inherited from upstream):
  - TestSparseCPU::test_bmm_coo_row_index_alignment_cpu_float64
  - TestSparseXPU::test_bmm_coo_row_index_alignment_xpu_float64
  - TestSparseXPU::test_sspaddmm_wrong_mat_types_error_messages_xpu_{complex128,float64}
  - TestSparseXPU::test_warn_on_sparse_tensor_invariant_checks_disabled_xpu_float32
  - TestSparseCPU::test_empty_full_requires_grad_*_cpu_* (17 tests - skipped because `test_empty_full` is now `@onlyOn(["cuda", "xpu"])`)
  - TestSparseCPU::test_legacy_new_device_cpu (skipped - now `@onlyOn(["cuda", "xpu"])`)

Removed tests (no longer needed - upstream removed sparse bmm Windows skips, so the error-path test is obsolete):
  - test_bmm_windows_error_{cpu,xpu}_float64


test/xpu/test_sparse_xpu.py
```diff
@@ -10,7206 +10,920 @@
...
+            pass
+        else:
+            raise AssertionError(f"unreachable: layout={layout}")
+        self.assertTrue(t.is_pinned())
 
-            def create_invalid_tensor(check_invariants=None):
-                shape = (2, 2)
-                invalid_indices = torch.tensor(
-                    [[0], [3]]
-                )  # colu
```
tszulist-hbn: `test_bmm_windows_error`: `test_bmm` is now skipped on Windows with the reason "bmm sparse-dense XPU is not yet supported", but `test_bmm_windows_error` runs only on Windows XPU and asserts that sparse-dense bmm succeeds (with a comment saying "XPU supports sparse-dense bmm"). These seem to contradict each other - is sparse-dense bmm supported on Windows XPU or not?

tszulist-hbn, line 872: `test_gradcheck_mm`: upstream calls `self.skipTest('NOT IMPL')` after each `assertRaisesRegex` block, but the override drops both, so those parametrizations now report PASSED instead of SKIPPED. Was dropping the `skipTest` intentional? Keeping it would preserve upstream reporting and flip loudly if the op ever gets implemented.

test/xpu/test_sparse_xpu.py
```diff
@@ -10,7206 +10,920 @@
...
+TestSparse.test_bmm_oob = _test_bmm_oob
+
+
+# Strip outermost skipIf decorator (test passes on XPU).
+@coalescedonoff
+@dtypes(torch.double)
+@dtypesIfMPS(torch.float32)
+@unittest.skipIf(TEST_WITH_CROSSREF, "generator unsupported triggers assertion error")
+@gradcheck_semantics()
+def _test_sparse_mul(self, device, dtype, coalesced, gradcheck):
```
tszulist-hbn, line 463: `test_sparse_mul`: The override strips the upstream `@unittest.skipIf(IS_LINUX or IS_MACOS or TEST_WITH_ROCM or IS_WINDOWS, ".../174389")` entirely. Since it's assigned before `instantiate_device_type_tests`, this re-enables the test on the CPU (and MPS) variants generated in this file too, not just XPU and issue 174389 looks platform-agnostic. Did `test_sparse_mul_cpu` pass in your run, or should the un-skip be narrowed to XPU only?

test/xpu/test_sparse_xpu.py
pbielak: Thanks for the catch - XPU supports sparse bmm on Windows. Also the skips for CUDA were already removed in commit: https://github.com/pytorch/pytorch/commit/e96f75b1bf1f1992704b01845907c2171fd9f535 

I pushed another commit to align these tests

pbielak: I'm not sure why upstream keeps these `skipTest` calls there. Let's analyze what the test does: it prepares some inputs for mm / sparse mm and calls the `gradcheck` function. Next, in some configurations the gradcheck calls should raise errors and they actually raise them, so the test should pass. But then the test still adds the unconditional skip, so in the end these tests will finish as `SKIPPED`, while they should be reporting `PASSED`.

I dropped these skips to restore the previous test behaviour. When upstreaming, we should dig into why these skips are present anyway.

Quick git blame shows that this test was added by: https://github.com/pytorch/pytorch/commit/b89fda51cd3899eae1dc9e5a2f2700c4113a4400 - maybe the functionality was implemented in the meantime, so the skips are not necessary anymore.

Let's check that when upstreaming.

test/xpu/test_sparse_xpu.py
pbielak: Thanks for catching this - I narrowed the unskip down only to XPU

BBBela (commented):
I will finish review tomorrow, but below are my initial considerations: 
- It looks like the `test_same_gpu` test wasn't correctly ported. Previously it has `@onlyOn(["cuda", "xpu"])` and after migration it is left with `@onlyCUDA`. It was skipped due to the second condition `@deviceCountAtLeast(2)`, but on platform with 2 XPUs the previous version would run and this PR version would not.
- The `test_empty_full` test locally was gated with `TEST_CUDA or TEST_XPU`, while the upstream version gates on `torch.cuda.is_available()`. So porting it without any override abandons this whole if branch for XPU.
- In `test_sparse_matmul` the previous version has internal if condition `if a.device.type != "cpu":`, while the upstream version has `if a.is_cuda:`, so this porting changes the if conditions evaluation - was that intentional?

pbielak:
> I will finish review tomorrow, but below are my initial considerations:
> 
> * It looks like the `test_same_gpu` test wasn't correctly ported. Previously it has `@onlyOn(["cuda", "xpu"])` and after migration it is left with `@onlyCUDA`. It was skipped due to the second condition `@deviceCountAtLeast(2)`, but on platform with 2 XPUs the previous version would run and this PR version would not.
> * The `test_empty_full` test locally was gated with `TEST_CUDA or TEST_XPU`, while the upstream version gates on `torch.cuda.is_available()`. So porting it without any override abandons this whole if branch for XPU.
> * In `test_sparse_matmul` the previous version has internal if condition `if a.device.type != "cpu":`, while the upstream version has `if a.is_cuda:`, so this porting changes the if conditions evaluation - was that intentional?

@BBBela I addressed your comments:
- `test_same_gpu` fixed
- `test_empty_full` and `_test_empty_full` both fixed
- `test_sparse_matmul` - the `a.is_cuda` just applied a looser tolerance; even with standard setting (i.e. "else branch") it passes on XPU

tszulist-hbn (approved):

BBBela (commented):
- The `test_gradcheck_mm` override looses two `self.skipTest('NOT IMPL')` that are in the upstream tests - this skip doesn't look necessary in this test, but I wanted to make sure if that was intentional? But if removing the skip is not needed to launch this test on XPU, then I think we should leave the exact upstream version where possible to make the future porting easier.
- In `test_coalesce_accepts_large_tensor` the outermost decorator becomes the `@largeTensorTest`. It should not be outermost. It should be below the device gates. I know that for now it doesn't matter, but it is better in my opinion to set the right order in the first place rather than to remember to change the order when migrating to upstream.
- The description says about new, only-xpu test `test_bmm_windows_error`. I cannot see it in the changed test file - why is that?

pbielak:
> * The `test_gradcheck_mm` override looses two `self.skipTest('NOT IMPL')` that are in the upstream tests - this skip doesn't look necessary in this test, but I wanted to make sure if that was intentional? But if removing the skip is not needed to launch this test on XPU, then I think we should leave the exact upstream version where possible to make the future porting easier.
> * In `test_coalesce_accepts_large_tensor` the outermost decorator becomes the `@largeTensorTest`. It should not be outermost. It should be below the device gates. I know that for now it doesn't matter, but it is better in my opinion to set the right order in the first place rather than to remember to change the order when migrating to upstream.
> * The description says about new, only-xpu test `test_bmm_windows_error`. I cannot see it in the changed test file - why is that?

@BBBela Check the "Address review comments 2" commit, where I address your concerns.

I also had to fix the imports in `test/xpu/test_sparse_csr_xpu.py` as this test file was importing constants from the `test_sparse_xpu.py` file, which were removed in this PR.

The PR description is currently out-of-sync. Waiting for CI results to update it.

test/xpu/test_sparse_csr_xpu.py
```diff
@@ -91,11 +92,14 @@
 sys.path.append(os.path.abspath(os.path.dirname(__file__)))
 import operator
 
-from test_sparse_xpu import (
-    CUSPARSE_SPMM_COMPLEX128_SUPPORTED,
-    HIPSPARSE_SPMM_COMPLEX128_SUPPORTED,
+CUSPARSE_SPMM_COMPLEX128_SUPPORTED = (IS_WINDOWS and torch.version.cuda) or (
+    not IS_WINDOWS and not TEST_WITH_ROCM
 )
```
BBBela, line 97: Couldn't we import it directly from the upstream `test_sparse.py` using `XPUImportCtx`?

BBBela (commented):
In general it looks good to me, just one minor suggestion below (not blocking).

Waiting for the CI to complete and for the final description update before approving.

test/xpu/test_sparse_csr_xpu.py
pbielak: We could but that would leave this file as a hybrid between an import-based test file and a hard-copy. I would leave it as is.

BBBela: Okay

BBBela (approved):
Thanks!

pbielak:

PR #4041, opened 2026-06-17
Remove the obsolete op dispatch _safe_softmax_xpu

guangyey:
# Motivation
After https://github.com/pytorch/pytorch/pull/181233 and https://github.com/intel/torch-xpu-ops/pull/3970 landed, we could entirely obsolete op dispatch `_safe_softmax_xpu`. See https://github.com/pytorch/pytorch/blob/7d287fc446e5b99bca5d1d6de467ce5897a223fc/aten/src/ATen/native/native_functions.yaml#L15168-L15171, `_safe_softmax` doesn't need XPU dispatch now.
```yaml
- func: _safe_softmax(Tensor self, int dim, ScalarType? dtype=None) -> Tensor
  dispatch:
    CompositeExplicitAutograd: _safe_softmax
    NestedTensorCPU, NestedTensorHPU, NestedTensorCUDA, NestedTensorXPU: _safe_softmax
```

PR #4417, opened 2026-07-22
Enable fused half_to_float fast-path in _safe_softmax_xpu

cyyever:
`_safe_softmax_kernel` already supports `half_to_float=true`, so route `Half` input requested as `Float` through the fused kernel path instead of first materializing a full float copy of the input. This mirrors the upstream `at::native::softmax` dispatch condition (`is_xpu() && Half && dtype==Float`).


guangyey (approved):
Currently, `_safe_softmax_xpu` is a unreadable code. We plan to remove it in https://github.com/intel/torch-xpu-ops/pull/4041. Anyway, it is fine to add this code.

PR #4726, opened 2026-07-29
[xpu][symm_mem] implement has_allocation() for the XPU allocator

Chao1Han:
SymmetricMemoryAllocator::has_allocation() defaults to false and only the CUDA allocator overrides it, so torch.distributed._symmetric_memory. is_symm_mem_tensor() always returned False on XPU even for buffers allocated through empty_strided_p2p().
test case rely on torch/_inductor/codegen/wrapper.py register


Chao1Han:
duplicate with #4829

PR #4758, opened 2026-07-30
[XPU] Validate replication_pad2d_backward channel dimension

### LINKED ISSUE #4754: [xpu-aglinment] Validate replication_pad2d_backward channel mismatch
### 🐛 Describe the bug

XPU `replication_pad2d_backward` accepts a channel-mismatched `grad_output`
instead of raising the required validation error. The analogous one-dimensional
XPU call raises correctly because its check is shared, but the separate XPU 2D
backward path lacks the corresponding channel guard and can produce undefined
results or invalid memory access.

This was found during XPU alignment for upstream PyTorch PR
https://github.com/pytorch/pytorch/pull/189463, which adds the equivalent 2D
channel check to the CPU and CUDA templates.

### Reproducer

```python
import torch

input_2d = torch.ones(2, 2, 4, 4, device="xpu")
invalid_grad_output = torch.ones(2, 0, 6, 8, device="xpu")

try:
    torch.ops.aten.replication_pad2d_backward(
        invalid_grad_output, input_2d, [2, 2, 1, 1]
    )
except RuntimeError as error:
    assert "gradOutput channel unexpected" in str(error)
else:
    raise AssertionError("replication_pad2d_backward accepted mismatched channels")
```

For comparison, the analogous 1D case already raises the expected error:

```python
torch.ops.aten.replication_pad1d_backward(
    torch.ones(2, 0, 8, device="xpu"),
    torch.ones(2, 2, 4, device="xpu"),
    [2, 2],
)
```

### Actual behavior

The 2D backward path silently accepts the mismatched channel count, while the 1D
path correctly rejects it:

```text
torch version: 2.14.0.dev20260713+xpu
xpu available: True
1d XPU: Correct error raised
2d XPU: No error raised (result shape=torch.Size([2, 2, 4, 4])) - XPU missing channel validation
RESULT: confirmed
```

### Expected behavior

`replication_pad2d_backward` raises `RuntimeError` with `gradOutput channel
unexpected` on a channel mismatch, matching the XPU 1D path and CPU/CUDA eager
semantics, while retaining valid-input behavior.

### Alignment metadata

- Upstream source: https://github.com/pytorch/pytorch/pull/189463
- Source type: upstream PR (CPU/CUDA fixed; XPU 2D path unaligned)
- Local XPU result: confirmed
- Routed area: `intel/torch-xpu-ops`, `replication_pad2d_backward_kernel` in `src/ATen/native/xpu/sycl/ReplicationPaddingKernels.cpp`

### Notes

`replication_pad2d_backward_kernel` validates output width and height but not the
channel/plane dimension; the nearby 3D validation path already checks the
corresponding plane dimension. Compute the channel dimension for the 3D and 4D
input forms and add a `TORCH_CHECK` matching the established 3D and upstream
CPU/CUDA error behavior.

A regression test should be added under `test/regressions/` (mismatch case plus a
valid-input case) and run with `pytest` after an editable XPU build.

### Versions

```text
PyTorch version: 2.14.0.dev20260713+xpu
Is XPU available: True
Relevant path: src/ATen/native/xpu/sycl/ReplicationPaddingKernels.cpp
```

laifenxiawucha:
## Background

XPU `replication_pad2d_backward` validates the spatial dimensions of `grad_output` but not its channel dimension. A mismatch can pass validation and lead to an invalid kernel read instead of a user-facing error.

## Upstream

- [pytorch/pytorch#189463](https://github.com/pytorch/pytorch/pull/189463) adds the analogous channel validation to upstream CPU/CUDA padding code. XPU has a separate 2-D backward implementation, so it needs the guard here.

## Fix

- validate the `grad_output` channel count before launching the XPU kernel
- preserve valid backward behavior
- add mismatch, valid-input, and CPU-parity coverage

## Before and after

| Case | Before | After |
|---|---|---|
| Channel mismatch | Accepted; may read out of bounds | Raises `gradOutput channel unexpected` |
| Matching channels | Works | Still works and matches CPU |

## Testing

Linux XPU `op_regression` in [CI run 34858341206](https://github.com/intel/torch-xpu-ops/actions/runs/34858341206): 3 passed, 0 failed.

Fixes #4754

### DIFF
--- a/src/ATen/native/xpu/sycl/ReplicationPaddingKernels.cpp
+++ b/src/ATen/native/xpu/sycl/ReplicationPaddingKernels.cpp
@@ -547,18 +547,34 @@ void replication_pad2d_backward_kernel(
   const auto padR = padding[1];
   const auto padT = padding[2];
   const auto padB = padding[3];
+  int dimc = 0;
   int dimh = 1;
   int dimw = 2;
 
   int numInputDims = input.dim();
   if (numInputDims == 4) {
+    dimc++;
     dimh++;
     dimw++;
   }
+  const auto ichannel = input.size(dimc);
   const auto iheight = input.size(dimh);
   const auto iwidth = input.size(dimw);
   const auto oheight = iheight + padT + padB;
   const auto owidth = iwidth + padL + padR;
+
+  // Validate the channel (plane) dimension in addition to the spatial dims.
+  // Without this check a grad_output whose channel count does not match the
+  // input (e.g. 0 channels) passes the width/height checks, a non-empty
+  // grad_input is allocated from the input shape, and the kernel reads
+  // grad_output out of bounds -> SIGSEGV. Mirrors the 3d backward check and
+  // the CPU/CUDA fix in pytorch/pytorch#189463.
+  TORCH_CHECK(
+      ichannel == grad_output.size(dimc),
+      "gradOutput channel unexpected. Expected: ",
+      ichannel,
+      ", Got: ",
+      grad_output.size(dimc));
 
   TORCH_CHECK(
       owidth == grad_output.size(dimw),
--- a/test/regressions/test_replication_pad_channel_mismatch.py
+++ b/test/regressions/test_replication_pad_channel_mismatch.py
@@ -29,9 +29,7 @@ class TestReplicationPadChannelMismatch(TestCase):
         grad_output = torch.ones(2, 0, 6, 8, device=xpu_device)
         inp = torch.ones(2, 2, 4, 4, device=xpu_device)
         with self.assertRaisesRegex(RuntimeError, "gradOutput channel unexpected"):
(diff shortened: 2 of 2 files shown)


laifenxiawucha:
Verified on Intel Data Center GPU Max 1100 (torch `2.14.0a0`, source build).

Reproducer — mismatched channel now raises, valid case still works:
```
mismatch raised: gradOutput channel unexpected. Expected: 2, Got: 0
valid backward shape (2, 3, 4, 4): PASS
```

Regression test `test/regressions/test_replication_pad_channel_mismatch.py`:
```
test_replication_pad2d_backward_channel_mismatch_xpu ... ok
test_replication_pad2d_backward_matches_cpu_xpu ... ok
test_replication_pad2d_backward_valid_xpu ... ok
Ran 3 tests ... OK
```

PR #4761, opened 2026-07-30
Compile SYCL kernels with `-fno-sycl-id-queries-fit-in-int`

chunhuanMeng:
Adding new compiler option prevents SYCL ID/range queries from being narrowed to int, allowing kernels to correctly handle larger ranges and offsets.


guangyey:
This change may impact performance based on the compiler team's comments. However, they don't know how much it is.

guangyey (approved):

xuhancn (changes requested):
Reviewing this against the 15 UTs affected by CMPLRLLVM-77158 (64-bit / large-tensor tests tripped by `fsycl-id-queries-range=int` on DLE 2026.2). The direction is right, but `=uint` is insufficient for 2 of the 15 tests and the flag also needs to reach the device-link step. Details in the inline comment on the flag line.

xuhancn:
Verified PR #4761 from-source on the JIRA-77158 config (PyTorch `30aabec` + torch-xpu-ops `390cceb8`, DLE 2026.2, PVC Data Center GPU Max 1550). Build + 15-UT run per config.

**PR #4761 as written: 0/15 PASS** — the flag is added to `SYCL_KERNEL_OPTIONS` only, not `SYCL_DEVICE_LINK_FLAGS`. The `torch_xpu_ops` device-link command in `build.ninja` carries no id-queries flag (`SYCL_KERNEL_OPTIONS` does not feed `SYCL_LINK_DEVICE_OBJECTS`), so the executed device image keeps `int` range and all 15 UTs fail with `... -fsycl-id-queries-range=int ... storable in an int`. Identical to the unflagged baseline.

For reference, the `fsycl-id-queries-range` flag **only takes effect when it also reaches `SYCL_DEVICE_LINK_FLAGS`**. With the flag on both compile + device-link (a change separate from this PR), the per-flag-value result on the 15 UTs is:

| `fsycl-id-queries-range` | PASS | Fail | Error |
|---|---|---|---|
| `=int` (default, no flag) | 0 / 15 | 15 | `… storable in an int` |
| `=uint` (PR #4761's value) | 8 / 15 | 7 | `… storable in an uint32_t` (range > UINT32_MAX) |
| `=size_t` | 15 / 15 | 0 | — |

So `=uint` (the value this PR proposes) covers only 8 of the 15 even when the flag correctly reaches the executed image; 7 tests launch ranges exceeding UINT32_MAX (cumsum_64bit, cumsum_outer_64bit, layer_norm, replicatepad, upsamplingnearest2d_backward, coalesce, pdist). `=size_t` covers all 15.

Separately (unrelated to this PR's flag change): the branch does not build on DLE 2026.2 without a `NullKernelClass` `if constexpr` guard in `src/comm/DeviceProperties.h` — 55× `getKernelNameHelper(...NullKernelClass)` undefined references at link. That is a 2026.2 build-blocker, not something this PR introduces or addresses.

guangyey:
Hi @chunhuanMeng 

The overall impact of enabling this flag is relatively small. Let's enable `size_t` in torch-xpu-ops. And land it after the branch cut.

guangyey (commented):
2.14 branch cut. I think we could land this to see if any unexpected happend.

guangyey (approved):

xuhancn (approved):

chunhuanMeng:
merge -f "unrelated ut failure"

chunhuanMeng:
merge -f

chunhuanMeng:
@torchxpubot merge -f

PR #4763, opened 2026-07-31
Deduplicate and template the atomics in Atomics.h

cyyever:
Two-step modernization of Atomics.h (review the two commits in order):

1. Collapse the per-address-space x per-size CAS specializations into two base
   templates (`AtomicIntegerImplBase`, `AtomicFPImplBase`) parameterized on the
   `sycl::atomic_ref` alias and constrained by C++20 concepts; drop a duplicate
   `atomic_ref` alias already provided by `comm/SYCLHelpers.h`.

2. Replace the `SYCL_ATOMIC_*` macros and the hand-written
   `atomicAdd`/`atomicAddLocal` overloads with one address-space-overloaded
   function template per op (`sycl_global_ptr` vs `sycl_local_ptr`). The body
   uses native `fetch_*` for 4/8-byte ints and float/double Add, and otherwise
   falls back to the CAS-emulated cores; `atomicCAS` becomes a local-only
   template. `Indexing` follows: `atomicAddLocal` -> `atomicAdd`, and its
   `atomicCAS` argument is wrapped for template deduction.

The macro set and the `atomicXLocal` naming are gone; overloads are visible in
source, and the pointer type selects global vs work-group-local.

### DIFF
--- a/src/ATen/native/xpu/sycl/Atomics.h
+++ b/src/ATen/native/xpu/sycl/Atomics.h
@@ -16,6 +16,7 @@
 #include <sycl/sycl.hpp>
 
 #include <concepts>
+#include <functional>
 #include <type_traits>
 
 namespace at::native::xpu {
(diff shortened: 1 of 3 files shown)


guangyey (approved):
Great!

guangyey:
build failed

PR #3910, opened 2026-06-09
[CI] Refactor E2E and PT2E - execution, summary and regression reporting

mengfei25:
## Summary

Refactor the E2E benchmark comparison flow into `run_benchmarks.summary`, add PT2E-aware reporting/export, and enable regression-based CI failure reporting.

## Motivation

The previous comparison flow was built around a monolithic script and did not handle PT2E results or regression reporting cleanly. This change makes the summary pipeline easier to maintain, improves result visibility for both regular E2E and PT2E runs, and allows CI to fail automatically on meaningful regressions.

## What changed

- Replaced the old monolithic `.github/scripts/compare-e2e.py` flow with a modular summary package:
  - `.github/scripts/run_benchmarks/summary/__main__.py`
  - `.github/scripts/run_benchmarks/summary/constants.py`
  - `.github/scripts/run_benchmarks/summary/loader.py`
  - `.github/scripts/run_benchmarks/summary/merge.py`
  - `.github/scripts/run_benchmarks/summary/report.py`
- Updated `.github/workflows/acceptance_ondemand.yml` to invoke the new summary entrypoint
- Reworked `.github/ci_expected_accuracy/check_expected.py` to compare accuracy CSVs against expected references and report regressions/new passes/new models/expected failures

## PT2E support

- Added PT2E accuracy summary using `top1` only
- Added PT2E performance summary with `fp32`, `symm`, and `asymm`
- Added PT2E console reporting:
  - pass rates
  - target-only geomean
  - target/baseline geomean ratio
- Added PT2E export support:
  - separate PT2E CSVs for csv output
  - `PT2E Accuracy` / `PT2E Performance` sheets in xlsx output

## Regression behavior

- Exit with code `1` when regressions are detected
- Report:
  - new accuracy failures
  - new performance failures
  - new performance drops
  - new PT2E accuracy failures
  - new PT2E performance failures
  - new PT2E performance drops
- Distinguish:
  - target `0` + baseline `> 0` => failure
  - target null + baseline present => not run
- Ignore:
  - target-null failure cases
  - both-target-and-baseline-failed cases

PR #4766, opened 2026-07-31
Refactor E2E and PT2E benchmark runner

mengfei25:
Split out from #3910.

## Overview

Replaces the inline-bash E2E/PT2E benchmark execution and reporting with a single, testable Python runner (`run_benchmarks`) and aligns the summary/gating scripts to its output. Behavior of the CI/Nightly accuracy gate is unchanged; only the mechanism and the result-file schema change.

## What changed

- **New runner package** `.github/scripts/run_benchmarks/` — one entry point (`python -m run_benchmarks`) drives all suites (huggingface, timm_models, torchbench, pt2e) uniformly: task generation, per‑GPU/CPU worker binding (numactl), a shared result‑CSV schema, and per‑task hang / OOM / timeout monitoring with worker replacement.
- **E2E and PT2E actions** now just invoke the runner instead of bespoke sharding/summary bash.
- **Summary pipeline migrated** to the runner’s per‑task CSVs (`e2e_summary.sh`, `check_expected.py`, `perf_comparison.py`, `calculate_best_perf.py`): suite/dtype/mode are read from CSV columns rather than parsed from filenames.
- **Device detection** emits `NUM_GPUS` (consumed by the runner).
- **PT2E launch/summary removed from CI**: dropped from the nightly/weekly suite matrix and from the summary’s result check. PT2E **execution code is retained** in the runner (and its action) for manual/on‑demand use.

## Why

- The previous per‑suite bash sharding + summary was hard to maintain, test, and extend; a single runner is unit‑testable and consistent across suites.
- A uniform CSV schema removes filename‑parsing fragility and makes the downstream summary/gating deterministic.
- Built‑in hang/OOM/timeout handling prevents a single stuck model from blocking a whole run, which better for clients testing, such as BMG
- Narrowing PT2E to execution‑only keeps this PR focused on the E2E path while preserving the capability.

## Final result schema

Per‑task file: `inductor-<suite>-<dt>-<mode>-<device>-<scenario>.csv`; scenario roll‑ups: `summary_accuracy.csv` / `summary_performance.csv`.

Common prefix: `dev, backend, suite, dtype, mode, name, scenario, batch_size, elapsed`, then:

| Result | Appended columns |
|---|---|
| accuracy | `accuracy` |
| performance | `speedup, abs_latency` (inductor latency)`, eager_latency` (= abs_latency × speedup) |
| pt2e accuracy | `top1, top5` |
| pt2e performance | `throughput, quantization` |

## Effect

- **Accuracy gating unchanged** — the summary still gates on accuracy only (`exit acc_failed`); performance/regressions remain warnings.
- **Downstream compatible** — summary/perf scripts read by column name, so the new schema and the extra `eager_latency` column are non‑breaking; also covers `_performance_comparison.yml`.
- **PT2E** no longer auto‑runs in CI/Nightly; it stays runnable manually or via a follow‑up wiring PR.
- Validated end‑to‑end locally (CI + Nightly paths) with fake results: report generation, accuracy gate, and performance comparison all pass.

Out of scope: expected‑accuracy reference/reporting content changes, model‑list overlay changes, and unrelated CI image/runner‑input updates.


mengfei25:
@torchxpubot lint

PR #3438, opened 2026-04-23
ci: run ubuntu-24.04 jobs in docker container for self-hosted runner support

chuanqi129:
## Summary

- Add `container: image: ubuntu:24.04` to all workflow jobs that use `runs-on: ubuntu-24.04` as plain GitHub-hosted runners
- This allows self-hosted runners labeled `ubuntu-24.04` to pick up these jobs when GitHub-hosted runners are congested, while the container ensures environment consistency
- Each job gets a minimal "Setup container environment" step installing only the tools it needs

## Motivation

GitHub-hosted `ubuntu-24.04` runners are often congested, causing long queue times. By adding container isolation, we can attach self-hosted runners with the `ubuntu-24.04` label to absorb overflow, and the `ubuntu:24.04` Docker container guarantees identical execution environments.

## Changes

| File | Job(s) | Tools Installed |
|------|--------|----------------|
| `pull.yml` | `preci-lint-check` | git, curl, ca-certificates, gh, jq |
| `pull.yml` | `conditions-filter-win` | git |
| `_linux_ut.yml` | `summary` | git, curl, ca-certificates, gh |
| `nightly_ondemand.yml` | `Conditions-Filter` | none (basic shell only) |
| `_windows_ut.yml` | `summary` | git, curl, ca-certificates |
| `_windows_ut.yml` | `summary_check` | git, curl, ca-certificates, gh |
| `_linux_e2e_summary.yml` | `summary` | git, curl, ca-certificates, gh, rsync |
| `_linux_op_benchmark.yml` | `op_benchmark_test_results_check` | git, curl, ca-certificates, gh, rsync |
| `_linux_transformers.yml` | `report` | git, curl, ca-certificates |
| `_performance_comparison.yml` | `Performance-Comparison` | git, curl, ca-certificates, gh |

## Notes

- `sudo` removed from existing apt-get steps (containers run as root)
- `actions/setup-python` / `actions/checkout` work inside containers
- `_linux_test_image.yml` intentionally NOT changed (needs Docker daemon on host to build images)

PR #4776, opened 2026-08-03
[CI] Use ARC Ubuntu runner for hosted workflow jobs

mengfei25:
## Summary
- Replace all explicit GitHub-hosted Ubuntu workflow runner labels (`ubuntu-24.04`, `ubuntu-latest`) with the ARC scale set runner label `xpu-ubuntu-24.04`.
- Keep caller-provided runner inputs and existing container image tags unchanged.
- Cover workflow helper/status jobs as well as CI bot, label, monitor, summary, and comparison jobs that previously ran on GitHub-hosted Ubuntu runners.

## Motivation
GitHub-hosted runners have platform and capacity limitations for this project’s XPU CI usage. Moving these jobs to the ARC-backed `xpu-ubuntu-24.04` runner label keeps CI execution on project-managed self-hosted infrastructure and avoids depending on GitHub-hosted Ubuntu availability for these workflow paths.

If ARC scale set capacity is unavailable, fallback should be handled by the runner infrastructure behind the label or by a future selector workflow. GitHub Actions does not support a true OR fallback directly in one `runs-on` value.

Authored with GitHub Copilot.

### DIFF
--- a/.claude/skills/action/runner/arc-runner/SKILL.md
+++ b/.claude/skills/action/runner/arc-runner/SKILL.md
@@ -1,0 +1,178 @@
+---
+name: arc-runner-solution
+description: Use for the ARC + K8s GitHub Actions runner solution for intel/torch-xpu-ops, including why it is used, what it provides, how to implement it, and how it works.
+---
+
+# ARC Runner Solution
+
+## 1. Why To Use
+
+Use ARC + K8s to provide stable, elastic GitHub Actions runners for `intel/torch-xpu-ops` on the local node.
+
+This solution is preferred because it:
+
+- Runs jobs in clean ephemeral runner pods.
+- Scales from zero idle runners to the required job capacity while leaving node headroom for ARC and K8s control components.
+- Avoids managing many long-running runner services by hand.
+- Uses Kubernetes scheduling, isolation, lifecycle management, and resource limits.
+- Bakes common tools into the runner image to avoid installing them during every job.
+- Reuses selected cache directories across ephemeral pods to reduce setup time.
+- Keeps `ubuntu-24.04` GitHub-hosted runners separate from the local self-hosted label.
+
+## 2. What Can Do
+
+The active setup provides:
+
+- Repository: `https://github.com/intel/torch-xpu-ops`
+- Workflow label: `xpu-ubuntu-24.04`
+- Helm release: `xpu-ubuntu-24-04`
+- Runner namespace: `arc-runners`
+- Controller namespace: `arc-systems`
+- Runner image: `arc-xpu-runner:ubuntu-24.04-tools`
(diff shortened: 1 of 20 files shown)


mengfei25:
@torchxpubot lint

mengfei25:
@torchxpubot lint

mengfei25:
@torchxpubot rebase

.claude/skills/action/runner/arc-runner/SKILL.md
chuanqi129, line 1: Why do we need this Skill? and how to use this skill in the repo & local?

.github/workflows/_linux_build.yml
```diff
@@ -244,7 +244,7 @@ jobs:
   build-status-check:
     needs: [build]
     if: ${{ always() }}
-    runs-on: ubuntu-24.04
+    runs-on: xpu-ubuntu-24.04
```
chuanqi129, line 247: I guess we can't migrate the original jobs based on github free runners to the ARC runners directly, because the official free runners have pre-installed some sw packages, refer https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md. I have tried to move those free runners based jobs to container based jobs by PR   https://github.com/intel/torch-xpu-ops/pull/3438. so that we can smoothly use any self-hosted runners whatever it's baremetal or ARC

.github/workflows/_linux_e2e_summary.yml
```diff
@@ -21,7 +21,7 @@ defaults:
 
 jobs:
   summary:
-    runs-on: ubuntu-24.04
+    runs-on: xpu-ubuntu-24.04
```
chuanqi129, line 24: On the other hand, I'm wondering that can we use same label with Github provide official free runners, such as `ubuntu-24.04` for those ARC runners? The motivation in here is, we can leverage offcial runners if have, otherwise will use our ARC self-hosted runners. I'm not sure if it's possible

mengfei25: No, not support such case, if leverage ours and github, we need to modify workflows to add logical for runners check.

.github/workflows/_linux_build.yml
mengfei25: We can also pre-install some packages in ARC image as github runners do, such as, https://github.com/intel/torch-xpu-ops/pull/4776/changes#diff-c02446077840934980d503df65936b7617a668f8c2267222d9208dedf4a4f143R32
Or install packages what we need online, but it will take more time.

mengfei25:
This change does not affect the actual test cases.

mengfei25:
Replaced by #4814, which uses the in-repository branch `intel:xpu-arc-ubuntu-runners` so CI runs from intel/torch-xpu-ops instead of the fork.

PR #4789, opened 2026-08-04
ci: fix test stage of dpclang workflow

### LINKED ISSUE #4787: ci: tests in dpclang workflow fails with RuntimeError: No XPU devices are available
Tests in [dpclang workflow](https://github.com/intel/torch-xpu-ops/blob/main/.github/workflows/pull_dpclang.yml) fail with `RuntimeError: No XPU devices are available`. When PR which added workflow was created tests were running without this error, example (6/16/2026):
* https://github.com/intel/torch-xpu-ops/actions/runs/27625400157

Then on scheduled event after the merge this issue appeared (6/28/2026):
* https://github.com/intel/torch-xpu-ops/actions/runs/28318715085

and persisted thru scheduled and PR events. For example with the most recent PR run:
* https://github.com/intel/torch-xpu-ops/actions/runs/30613713926

CC: @mengfei25 @chuanqi129

dvrogozh:
Tests in dpclang workflow were failing with the no XPU devices found error because dpclang binaries were built with the newer version of the L0 library not available on the test system. This commit fixes the dpclang build by requesting dpclang to fetch, build and carry its own L0 library.

Additionally commit rearranges the setup of dpclang which is currently sensitive to the order of steps in the workflow as
it configures `PATH`, `LD_LIBRARY_PATH`, `PKG_CONFIG_PATH`, etc. environment variables some of which are being overridden by github actions (such as `actions/setup-python`).

Finally, `op_ut` test takes too long and gets cancelled. Likely some tests in the suite get stuck. For now this commit limits
execution to `basic` test scope.

Fixes: https://github.com/intel/torch-xpu-ops/issues/4787

CC: @chuanqi129, @mengfei25

### DIFF
--- a/.github/actions/linux-testenv/action.yml
+++ b/.github/actions/linux-testenv/action.yml
@@ -52,6 +52,25 @@ runs:
       uses: actions/checkout@v4
       with:
         path: torch-xpu-ops
+    - name: Download dpclang
+      if: ${{ inputs.category == 'dpclang' }}
+      uses: actions/download-artifact@v8
+      with:
+        name: dpclang
+    # NOTE: placement of this step is important as some environment variables it sets
+    # might be overriden or reset by certain actions. For example, PKG_CONFIG_PATH is
+    # overriden by actions/setup-python.
+    - name: Setup dpclang
+      if: ${{ inputs.category == 'dpclang' }}
+      shell: bash -xe {0}
+      run: |
+        rm -rf install-dpclang && mkdir install-dpclang
+        tar xzvf install-dpclang.tgz -C install-dpclang
+        echo "CMAKE_PREFIX_PATH=$(pwd)/install-dpclang" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


dvrogozh:
@chuanqi129 , @mengfei25 , this PR fixes the issue of dpclang tests not seeing XPU devices reported in https://github.com/intel/torch-xpu-ops/issues/4787. I also have addressed another issue which was initially obscured by the lack of dpclang support in Triton - now Triton related tests are passing. Overall this PR returns status quo on the dpclang tests and even improves on top of that. Still there are tests failing which will require additional triage. I will file respective issues and follow up on them separately. Also, initially I have overlooked that `op_ut` tests are taking too long rerunning and stepping into some timeout issues. I am disabling these tests for now entirely to save test time and till I will nail down issues in the `basic` tests. Please, help approve this PR so we will get dpclang workflow completely unblock and showing actual issues so we can start working on them.

mengfei25 (approved):

chuanqi129 (approved):

Issue #4233, opened 2026-07-06
[BMG] MixOrderReductionTest.test_rms_norm_bwd_with_dynamic_shape fails: RuntimeError: tensor does not have a device (eager rms_norm backward w/ dynamic shapes)


RUIJIEZHONG66166:
## Description

Three parametrized cases of `MixOrderReductionTest.test_rms_norm_bwd_with_dynamic_shape` fail on XPU with `RuntimeError: tensor does not have a device`:

- `test_rms_norm_bwd_with_dynamic_shape_dynamic_dims0`
- `test_rms_norm_bwd_with_dynamic_shape_dynamic_dims1`
- `test_rms_norm_bwd_with_dynamic_shape_dynamic_dims2`

`test/inductor/test_mix_order_reduction.py` (class `MixOrderReductionTest`, base method `test_rms_norm_bwd_with_dynamic_shape`, decorated `@parametrize("dynamic_dims", ([0], [1], [0, 1]))`).

## Root cause

The failure occurs in the **eager reference backward pass**, before `torch.compile` runs. In the test, `ref = fwd_bwd(f)` (line 464) calls `out.backward(dy)` (line 445) on an eager `F.rms_norm` output where the input `x` was marked dynamic on dims (0, 1) via `torch._dynamo.mark_dynamic(x, (0, 1))`. The error is raised from the C++ autograd engine (`run_backward`), independent of the inductor mix-order-reduction codegen the test is meant to exercise.

This is an XPU backend / autograd bug, not a test-code issue: the test uses `device=GPU_TYPE` (device-agnostic), carries no `@skipIfXpu`, and the base method is present upstream (introduced by commit `93fef4bd1dd`, an ancestor of the tested HEAD).

## Reproduction

```
python -m pytest "test/inductor/test_mix_order_reduction.py::MixOrderReductionTest::test_rms_norm_bwd_with_dynamic_shape_dynamic_dims0" -v
```

Reproduced locally on `torch 2.13.0+xpu`:

```
FAILED [0.9454s] MixOrderReductionTest.test_rms_norm_bwd_with_dynamic_shape_dynamic_dims0
Traceback (most recent call last):
  File "test/inductor/test_mix_order_reduction.py", line 464, in test_rms_norm_bwd_with_dynamic_shape
    ref = fwd_bwd(f)
  File "test/inductor/test_mix_order_reduction.py", line 445, in fwd_bwd
    out.backward(dy)
[...]

PR #3253, opened 2026-04-01
Fix illegal tensor in dbeta of RMSNorm Backward

Wetitpig:
A fused implementation for RMS Norm has been implemented in PR #2205, which introduces the `rms_norm` flag to the function `layer_norm_backward_kernel_impl`.

`dbeta` has been deliberately set to not defined here, as `dbeta` and `dgamma` has been set to the same tensor: https://github.com/intel/torch-xpu-ops/blob/a2d516a58c64f18b76880f3a77efbc02885d65af/src/ATen/native/xpu/sycl/LayerNormKernels.cpp#L1239-L1245

However, the sum here does not take into account that `dbeta` is defined while `dbeta_blocks` has not been initialised:
https://github.com/intel/torch-xpu-ops/blob/a2d516a58c64f18b76880f3a77efbc02885d65af/src/ATen/native/xpu/sycl/LayerNormKernels.cpp#L1292

This leads to `RuntimeError: tensor does not have a device` with this traceback:
```
C++ CapturedTraceback:
#4 std::_Function_handler<std::shared_ptr<c10::LazyValue<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > const> (), c10::SetStackTraceFetcher(std::function<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > ()>)::{lambda()#1}>::_M_invoke(std::_Any_data const&) from Logging.cpp:0
#5 c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) from ??:0
#6 c10::detail::torchCheckFail(char const*, char const*, unsigned int, char const*) from ??:0
#7 at::TensorBase::options() const from LegacyBatchingRegistrations.cpp:0
#8 at::meta::resize_reduction(at::impl::MetaBase&, at::Tensor const&, c10::OptionalArrayRef<long>, bool, c10::ScalarType, bool) from ReduceOps.cpp:0
[...]

PR #4790, opened 2026-08-04
Fix RMSNorm backward crash from aliasing dgamma into the dbeta slot

songhappy:
### Bug

`aten::_fused_rms_norm_backward` on XPU fails with `RuntimeError: tensor does not have a device` whenever the weight gradient is requested (`output_mask=[True, True]`) and the row count exceeds `xe_core_count * 1024` (57344 on a Data Center GPU Max 1550 tile).

`rms_norm_backward_kernel` passes `dgamma` as **both** the `dgamma` and `dbeta` arguments of the shared `layer_norm_backward_kernel_impl`:

https://github.com/intel/torch-xpu-ops/blob/22be5ddd/src/ATen/native/xpu/sycl/LayerNormKernels.cpp#L1507-L1508

The impl selects its column-reduction arm by testing `dbeta->defined()`. Because the alias makes `dbeta` look defined, RMSNorm takes the `dgamma->defined() && dbeta->defined()` arm, which ends in

```cpp
*dbeta = dbeta_blocks.sum(0);
```

on a buffer that is allocated only under `if constexpr (!rms_norm)` -- i.e. never allocated for RMSNorm. Summing the undefined tensor raises the error.

Introduced by #2205 (Fused RMSNorm). Exposed region: `normalized_shape < 2048` **and** `rows > xe_core_count * 1024` **and** `output_mask == [True, True]`. Below the threshold the simple path guards every `dbeta` write with `if constexpr (!rms_norm)`, so the alias is inert there.

This reaches real models through a per-head QK-norm, where `normalized_shape` is `head_dim` and the row count is `batch * seq * num_heads`: a Qwen3-0.6B QK-norm at batch 2, seq 2048, 16 heads is 65536 rows and crashes, while the 8-KV-head `k_norm` at 32768 rows in the same backward succeeds.

### Fix

One hunk at the call site: pass an undefined `Tensor` instead of aliasing `dgamma` into the `dbeta` slot, so control reaches the `dgamma->defined() && !dbeta->defined()` arm that RMSNorm actually wants. `dgamma` is unchanged. Only two callers of the impl exist and the LayerNorm one instantiates `rms_norm=false`, so it is untouched.

### Relationship to existing reports

- #3253 diagnoses this same bug (thanks @Wetitpig) and fixes it by wrapping the `*dbeta = dbeta_blocks.sum(0)` in the both-defined arm with `if constexpr (!rms_norm)`. That does stop the crash. This PR instead corrects the call site, on the reasoning that the alias is the actual defect: it makes the impl's `dbeta->defined()` dispatch report something untrue, and a guard leaves that in place for future callers -- note there is a second unguarded `*dbeta = dbeta_blocks.sum(0)` in the `!dgamma->defined() && dbeta->defined()` arm, which is unreachable today only *because* the alias makes both pointers equal. Maintainers should take whichever they prefer; I am happy to close this in favour of #3253 if the guard is the preferred shape, but the regression tests below are worth landing either way.
- #4233 reports the same `tensor does not have a device` from `MixOrderReductionTest.test_rms_norm_bwd_with_dynamic_shape`, in the eager reference backward. Same signature and same op; if the shapes there cross the threshold, this fixes it.

### Tests

New `test/regressions/test_rms_norm.py` covers the crash, the numerics, and the paths that must not regress:

- weight grad above the threshold, with `dgamma` checked against an fp32 recomputation, not just a liveness check
- weight grad below the threshold, to pin the path that already worked
- `nn.RMSNorm` autograd above the threshold (how the bug reaches users)
- `output_mask=[True, False]` above the threshold, which was never broken
- `nn.LayerNorm` above the threshold, asserting `bias.grad` is exactly the row count -- this is the case that would break if the `dbeta` plumbing were disturbed

The threshold is derived from `gpu_subslice_count`, which PyTorch defines as the same `gpu_eu_count / gpu_eu_count_per_subslice` quotient the kernel calls `xe_core_count`, so the tests target the branch on any part rather than hardcoding a row count.

```
pytest test/regressions/test_rms_norm.py -v
```

### Verification

Verified on one tile of an Intel Data Center GPU Max 1550 against an unmodified `2.12.0+xpu` wheel, A/B in the same process by `LD_PRELOAD`ing only the patched translation unit (mangled symbol names confirmed identical first).

Crash, `output_mask=[True, True]`, `normalized_shape=128`:

| rows | stock | patched |
|---|---|---|
| 1024 | pass | pass |
| 32768 | pass | pass |
| 57344 | pass | pass |
| 57345 | **fail** | pass |
| 65536 | **fail** | pass |
| 131072 | **fail** | pass |

Numerics, `dgamma` vs a float64 CPU reference with fp32 inputs:

| rows | arm | rel L2 | cosine |
|---|---|---|---|
| 1024 | simple | 1.47e-07 | 1.0000000000 |
| 32768 | simple | 6.52e-07 | 1.0000000000 |
| 57344 | simple | 7.43e-07 | 1.0000000000 |
| 57345 | two-stage | 1.34e-07 | 1.0000000000 |
| 65536 | two-stage | 1.24e-07 | 1.0000000000 |
| 131072 | two-stage | 1.39e-07 | 1.0000000000 |

The two-stage arm the fix routes into is slightly *more* accurate than the simple arm, as expected for a tree reduction over more blocks. Below-threshold results agree to all printed digits before and after, and LayerNorm `bias.grad` is exact in both arms.

Not yet verified: the patch has been exercised as an interposer, not as part of a full PyTorch build, and `test/regressions/test_rms_norm.py` has not itself been run end-to-end on hardware (the probes above cover the same assertions, and the machine I have access to has no XPU device visible). CI is the real integration test here.

### DIFF
--- a/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
@@ -1398,7 +1398,9 @@ void layer_norm_backward_kernel_impl(
            static_cast<size_t>(tile_size_n < SIMD ? tile_size_n : SIMD)},
           getCurrentSYCLQueue(),
           kfn);
-      *dbeta = dbeta_blocks.sum(0);
+      if constexpr (!rms_norm) {
+        *dbeta = dbeta_blocks.sum(0);
+      }
     } else {
       return;
     }
@@ -1504,8 +1506,18 @@ void rms_norm_backward_kernel(
       "rms_norm_backward_xpu",
       [&]() {
         using accscalar_t = acc_type_device<scalar_t, kXPU>;
+        Tensor unused_dbeta;
         layer_norm_backward_kernel_impl<scalar_t, accscalar_t, scalar_t, true>(
-            dY.contiguous(), X, rstd, rstd, gamma, M, N, dX, dgamma, dgamma);
+            dY.contiguous(),
+            X,
+            rstd,
+            rstd,
+            gamma,
+            M,
+            N,
+            dX,
+            dgamma,
+            &unused_dbeta);
       });
 }
 
--- a/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
@@ -1504,8 +1504,24 @@ void rms_norm_backward_kernel(
       "rms_norm_backward_xpu",
       [&]() {
         using accscalar_t = acc_type_device<scalar_t, kXPU>;
+        // RMSNorm has no bias, so there is no dbeta. Pass an undefined tensor
+        // instead of aliasing dgamma into the dbeta slot: the shared impl
+        // branches on dbeta->defined() to pick its column-reduction arm, and
+        // aliasing makes it take the both-defined arm, which reads the
+        // dbeta_blocks buffer that the rms_norm specialization deliberately
+        // never allocates.
+        Tensor unused_dbeta;
         layer_norm_backward_kernel_impl<scalar_t, accscalar_t, scalar_t, true>(
-            dY.contiguous(), X, rstd, rstd, gamma, M, N, dX, dgamma, dgamma);
+            dY.contiguous(),
+            X,
+            rstd,
+            rstd,
+            gamma,
+            M,
+            N,
+            dX,
+            dgamma,
+            &unused_dbeta);
       });
 }
 
--- a/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
@@ -1504,24 +1504,8 @@ void rms_norm_backward_kernel(
       "rms_norm_backward_xpu",
       [&]() {
         using accscalar_t = acc_type_device<scalar_t, kXPU>;
-        // RMSNorm has no bias, so there is no dbeta. Pass an undefined tensor
-        // instead of aliasing dgamma into the dbeta slot: the shared impl
-        // branches on dbeta->defined() to pick its column-reduction arm, and
-        // aliasing makes it take the both-defined arm, which reads the
-        // dbeta_blocks buffer that the rms_norm specialization deliberately
-        // never allocates.
-        Tensor unused_dbeta;
         layer_norm_backward_kernel_impl<scalar_t, accscalar_t, scalar_t, true>(
-            dY.contiguous(),
-            X,
-            rstd,
-            rstd,
-            gamma,
-            M,
-            N,
-            dX,
-            dgamma,
-            &unused_dbeta);
+            dY.contiguous(), X, rstd, rstd, gamma, M, N, dX, dgamma, dgamma);
       });
 }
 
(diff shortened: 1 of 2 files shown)


src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -1504,8 +1504,24 @@ void rms_norm_backward_kernel(
       "rms_norm_backward_xpu",
       [&]() {
         using accscalar_t = acc_type_device<scalar_t, kXPU>;
+        // RMSNorm has no bias, so there is no dbeta. Pass an undefined tensor
+        // instead of aliasing dgamma into the dbeta slot: the shared impl
+        // branches on dbeta->defined() to pick its column-reduction arm, and
+        // aliasing makes it take the both-defined arm, which reads the
+        // dbeta_blocks buffer that the rms_norm specialization deliberately
+        // never allocates.
+        Tensor unused_dbeta;
```
guangyey:
```suggestion
```

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -1504,8 +1504,24 @@ void rms_norm_backward_kernel(
...
+            dY.contiguous(),
+            X,
+            rstd,
+            rstd,
+            gamma,
+            M,
+            N,
+            dX,
+            dgamma,
+            &unused_dbeta);
```
guangyey, line 1522:
```suggestion
            dY.contiguous(), X, rstd, rstd, gamma, M, N, dX, dgamma, dgamma);
```

guangyey (approved):
Thanks!

guangyey:
@torchxpubot please help me add rms guard for `*dbeta = dbeta_blocks.sum(0);`

guangyey:
@torchxpubot review "please help me add rms guard for *dbeta = dbeta_blocks.sum(0);"

chuanqi129:
> @torchxpubot review "please help me add rms guard for *dbeta = dbeta_blocks.sum(0);"

@guangyey currently, the bot can't accept any customized prompt by the comment to avoid potential security risk

guangyey:
@songhappy Could you please help check the failure, It seems related to this PR.

songhappy:
Thanks @guangyey — I checked again; the failures are pre-existing CI flakiness, not from
  this PR.

  Only linux-ut (op_ut) / summary (the gate step) is red. test-in-container for both op_ut and basic,
  both builds, and preci-lint-check passed. The 6 "new failures" are dynamo / sparse / multigpu cases —
  none touches RMSNorm, LayerNorm, or either file this PR changes (grep -ri rms over the full failure
  list: zero hits). Causes are environmental: ModuleNotFoundError: No module named 'functorch.compile',
  ShapeEnv not equal, worker 'gw9' crashed, a free-memory race, and aten::linalg_vector_norm missing for
  SparseXPU.

  Two checks that they aren't mine:
  1. 4 of the 6 appear verbatim on unrelated PRs #4802, #4818, #4830, #4836, #4837; the other 2 are in
  all of those PRs' full ut_failure_list.csv (the known-issues baseline is fetched live per run, so it
  drifts).
  2. The prior run on this branch (30947145321) had a disjoint set of 4 new failures — zero overlap on
  the same code.

  Meanwhile all 5 cases in test/regressions/test_rms_norm.py passed on hardware
  (basic/passed_op_regression.log). Note test/regressions/ runs under op_regression, i.e. the basic
  scope, not op_ut — so the red job doesn't cover this PR's test anyway.

Could you please help to merge it?

guangyey:
Thanks!

PR #4800, opened 2026-08-04
Fix test_profiler_correlation_id

PawelSwider2000:
Align test_profiler_correlation_id with upstream by additional check: `event.activity_type() == "cpu_op"`


test/xpu/profiler/test_profiler_xpu.py
```diff
@@ -1649,8 +1649,14 @@ def test_profiler_type(self):
...
-        We expect the correlation_id to be unique across multiple invocation of the profiler,
-        So we will reuse id_uniqueness_set.
+        We expect the correlation_id of CPU operator events to be unique across
+        multiple invocations of the profiler, so we will reuse id_uniqueness_set.
+        Only cpu_op events are checked: with CUDA available, profile() also
+        collects CUDA activities, and CUPTI's device-enumeration calls at startup
+        (cudaGetDeviceCount, etc.) are recorded as cuda_runtime events whose
+        correlation ids come from a separate counter (also starting at 1) and
+        would otherwise collide with the operator ids.
+        The same happens with other devices as XPU.
```
PatrykWilczewski, line 1659: IMO this comment is a bit too long.
Maybe:
"Only cpu_op events are checked: runtime device-enumeration calls at startup use a separate correlation-id counter and would otherwise collide with operator ids"
Or move to commit's description.

Approved anyway

PatrykWilczewski (approved):

test/xpu/profiler/test_profiler_xpu.py
PawelSwider2000: I agree, however this is a copy of PyTorch description with added line at the end:
`The same happens with other devices as XPU.`

so keeping it similar will be easier for later enabling tests of profiler for XPU in upstream PyTorch

kdrozd-dev (approved):

kdrozd-dev:
@torchxpubot merge -f "unrelated CI failures"

Issue #4048, opened 2026-06-17
[release/2.13] test_linalg_xpu.py::TestLinalgXPU::test_cond_errors_and_warnings_xpu_float64 Windows fatal exception: access violation


bjarzemb:
### 🐛 Describe the bug

test_linalg_xpu.py::TestLinalgXPU::test_cond_errors_and_warnings_xpu_float64 failed with `Windows fatal exception: access violation` on Windows with Intel XPU (Arc B580). Test failed with `AssertionError: Scalars are not equal!` for PT2.12.

## Error Message

<details>

```

============================= test session starts =============================
platform win32 -- Python 3.12.13, pytest-7.3.2, pluggy-1.6.0 -- C:\Users\gta\miniforge3\envs\pytorch_2.13\python.exe
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('C:\\Users\\gta\\repositories\\pytorch\\pytorch\\third_party\\torch-xpu-ops\\test\\xpu\\.hypothesis\\examples')
rootdir: C:\Users\gta\repositories\pytorch\pytorch
configfile: pytest.ini
plugins: hypothesis-6.56.4, cpp-2.3.0, flakefinder-1.1.0, rerunfailures-14.0, subtests-0.13.1, timeout-2.4.0, xdist-3.3.1, xdoctest-1.3.0
timeout: 600.0s
timeout method: thread
timeout func_only: False
created: 1/1 worker
1 worker [1 item]

scheduling tests via LoadScheduling

test_linalg_xpu.py::TestLinalgXPU::test_cond_errors_and_warnings_xpu_float64 Windows fatal exception: access violation

Thread 0x0000312c (most recent call first):
  File "C:\Users\gta\miniforge3\envs\pytorch_2.13\Lib\threading.py", line 359 in wait
  File "C:\Users\gta\miniforge3\envs\pytorch_2.13\Lib\threading.py", line 655 in wait
  File "C:\Users\gta\miniforge3\envs\pytorch_2.13\Lib\threading.py", line 1431 in run
  File "C:\Users\gta\miniforge3\envs\pytorch_2.13\Lib\threading.py", line 1075 in _bootstrap_inner
  File "C:\Users\gta\miniforge3\envs\pytorch_2.13\Lib\threading.py", line 1032 in _bootstrap

Thread 0x00003194 (most recent call first):
[...]

PR #4802, opened 2026-08-04
Remove what() call in rethrow exception on Windows

jmamzax:
Calling what() on rethrow_exception cause access violation on Windows. This is because of passing objects or memory allocations across dynamic-link library. Issue: https://github.com/intel/torch-xpu-ops/issues/4048

### DIFF
--- a/src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
+++ b/src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
@@ -58,16 +58,33 @@ void error_handle(
   auto ids = be.ids();
 
   for (size_t i = 0; i < errs.size(); ++i) {
+    if (!errs[i]) {
+      info_cpu[ids[i]] = -1;
+      continue;
+    }
     try {
       std::rethrow_exception(errs[i]);
     } catch (const oneapi::mkl::lapack::exception& e) {
+#ifdef _WIN32
+      // what() returns a char* into mkl_sycl_lapack.dll's CRT heap; unsafe
+      // across DLL boundaries on Windows. info() and detail() return plain
+      // integers, safe.
       TORCH_WARN(
-          "Caught lapack exception:\nWhat: ",
+          "Caught lapack exception:",
+          "\nInfo: ",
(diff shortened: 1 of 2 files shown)


src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -57,7 +57,41 @@ void error_handle(
...
+      info_cpu[ids[i]] = -1;
+      continue;
+    }
+    try {
+      std::rethrow_exception(errs[i]);
+    } catch (const oneapi::mkl::lapack::exception& e) {
+      // what() returns a char* into mkl_sycl_lapack.dll's CRT heap; unsafe
+      // across DLL boundaries on Windows. info() and detail() return plain
+      // integers, safe.
+      TORCH_WARN(
```
pbielak, line 72: Currently, there is a lot of duplication between the ifdef and else branches - maybe the ifdef-else should be restricted only to the `TORCH_WARN` call. WDYT?

pbielak (approved):

PatrykWilczewski (approved):

test/xpu/test_linalg_xpu.py
```diff
@@ -789,7 +789,7 @@ def cond_errors_and_warnings(self, device, dtype):
     a = torch.ones(3, 3, dtype=dtype, device=device)
     for p in ["wrong_norm", 5]:
         with self.assertRaisesRegex(
-            RuntimeError, f"linalg.cond got an invalid norm type: {p}"
+            ValueError, f"linalg.cond got an invalid norm type: {p}"
```
draghan, line 792: Why do we want to change the exception type?

jmamzax: It is already changed by this PR: https://github.com/intel/torch-xpu-ops/pull/4221

draghan: Ok, thanks.

draghan (approved):

jmamzax:
@torchxpubot merge -f

PR #4806, opened 2026-08-05
Order the norm global-reduce handoff and drop its unused buffers

cyyever:
The norm backward global reduce handed off its scratchpad through a semaphore
`fetch_add` with no memory order, so the last workgroup could read stale slots.
Adds an `acq_rel`-default alias plus the device-scope acquire the rest of the
workgroup needs. Also drops two buffers sized for a first moment the rms path
never reads: the scratchpad applied `sizeof(acc_t)` twice (4-8x over-allocation),
and `reduce_project` wrote `a_data` unconditionally.

Not validated on hardware.

Test: **upstream PyTorch tests** — `pytest -v test/xpu/test_nn_xpu.py -k "layer_norm or rms_norm"`

Authored with Claude Code.

### DIFF
--- a/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
@@ -1179,8 +1179,7 @@ void layer_norm_backward_kernel_impl(
           M,
           N);
       Tensor semaphores, scratchpad;
-      config.template init_global_reduce<accscalar_t>(
-          X, semaphores, scratchpad);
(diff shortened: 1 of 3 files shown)


guangyey (approved):

jianyizh:
@torchxpubot review

jianyizh (approved):

jianyizh (approved):

jianyizh:
@torchxpubot ut-check

jianyizh:

PR #4807, opened 2026-08-05
Trim oneMKL LAPACK scratchpad init, host syncs and re-allocation

cyyever:
The non-batched `orgqr`/`ungqr` paths waited on the host once per loop iteration
although all iterations share one scratchpad. That reuse is safe because XPU
streams are in-order queues, not because of the wait, so one wait after the loop
keeps the same host-visible synchronization and drops `batch_size - 1` pipeline
drains. The wait is kept rather than dropped because, unlike `getrf`/`getrs`,
these two have no `batch_error` handler.

The broadcast `lu_solve` path re-queried the scratchpad size and reallocated per
matrix with identical arguments; both hoist above the loop.

The `at::zeros` -> `at::empty` scratchpad change is dropped per review — oneMKL
reads the scratchpad before writing it, so the zero-fill is load-bearing.

Not validated on hardware.

Test: **upstream PyTorch tests** — `pytest -v test/xpu/test_linalg_xpu.py -k "orgqr or householder_product or lu_solve"`

Authored with Claude Code.

### DIFF
--- a/src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
+++ b/src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
@@ -251,22 +251,13 @@ static void apply_lu_solve_xpu_(
   std::vector<int32_t> info_vec(batch_size, 0);
   int32_t* info_data = info_vec.data();
 
+  // The scratchpad depends only on batch_size, so callers hoist it.
   auto execute_mkl_getrs = [&](const scalar_t* a,
                                scalar_t* b,
                                const int64_t* ipiv,
-                               int64_t batch_size) {
-    int64_t scratchpad_size = mkl_getrs_scratchpad<scalar_t>(
-        queue,
-        trans,
-        n,
-        nrhs,
-        lda,
-        stride_a,
-        stride_ipiv,
-        ldb,
-        stride_b,
-        batch_size);
-    Tensor scratchpad_at = at::empty({scratchpad_size}, b_.options());
+                               int64_t batch_size,
+                               scalar_t* scratchpad,
+                               int64_t scratchpad_size) {
     try {
       mkl_getrs<scalar_t>(
           queue,
@@ -282,11 +273,26 @@ static void apply_lu_solve_xpu_(
           ldb,
           stride_b,
           batch_size,
-          reinterpret_cast<scalar_t*>(scratchpad_at.data_ptr()),
+          scratchpad,
           scratchpad_size);
     } catch (const oneapi::mkl::lapack::batch_error& be) {
       error_handle(info_data, be);
     }
+  };
+
+  auto alloc_getrs_scratchpad = [&](int64_t batch_size) {
+    int64_t size = mkl_getrs_scratchpad<scalar_t>(
+        queue,
+        trans,
+        n,
+        nrhs,
+        lda,
+        stride_a,
+        stride_ipiv,
+        ldb,
+        stride_b,
+        batch_size);
+    return std::make_pair(at::empty({size}, b_.options()), size);
   };
 
   bool is_broadcast = false;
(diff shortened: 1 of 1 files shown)


src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -503,8 +527,8 @@ Tensor& ungqr_mkl(Tensor& result, const Tensor& tau) {
       const int64_t scratchpad_size =
           oneapi::mkl::lapack::ungqr_batch_scratchpad_size<T>(
               queue, m, n, k, lda, matrix_stride, batch_tau_stride, batch_size);
-      Tensor scratchpad_at = at::zeros({scratchpad_size}, result.options());
-      auto* scratchpad = reinterpret_cast<T*>(scratchpad_at.data_ptr());
+      Tensor scratchpad_at = at::empty({scratchpad_size}, result.options());
```
pbielak: We intentionally used `at::zeros` here, as there is a bug in MKL, which causes values from `at::empty` to be included into computations, resulting in garbage outputs.

The bug might be fixed in the current MKL version, but please re-run the following test to make sure the outputs are correct: `test_linalg_xpu.py::TestLinalgXPU::test_det_logdet_slogdet_xpu_float64`

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -572,8 +598,8 @@ Tensor& orgqr_mkl(Tensor& result, const Tensor& tau) {
       const int64_t scratchpad_size =
           oneapi::mkl::lapack::orgqr_batch_scratchpad_size<T>(
               queue, m, n, k, lda, matrix_stride, batch_tau_stride, batch_size);
-      Tensor scratchpad_at = at::zeros({scratchpad_size}, result.options());
-      auto* scratchpad = reinterpret_cast<T*>(scratchpad_at.data_ptr());
+      Tensor scratchpad_at = at::empty({scratchpad_size}, result.options());
```
pbielak: Ditto

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -592,14 +618,16 @@ Tensor& orgqr_mkl(Tensor& result, const Tensor& tau) {
     } else {
       const int64_t scratchpad_size =
           oneapi::mkl::lapack::orgqr_scratchpad_size<T>(queue, m, n, k, lda);
-      Tensor scratchpad_at = at::zeros({scratchpad_size}, result.options());
-      auto* scratchpad = reinterpret_cast<T*>(scratchpad_at.data_ptr());
+      Tensor scratchpad_at = at::empty({scratchpad_size}, result.options());
```
pbielak: Ditto

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -523,14 +547,16 @@ Tensor& ungqr_mkl(Tensor& result, const Tensor& tau) {
     } else {
       const int64_t scratchpad_size =
           oneapi::mkl::lapack::ungqr_scratchpad_size<T>(queue, m, n, k, lda);
-      Tensor scratchpad_at = at::zeros({scratchpad_size}, result.options());
-      auto* scratchpad = reinterpret_cast<T*>(scratchpad_at.data_ptr());
+      Tensor scratchpad_at = at::empty({scratchpad_size}, result.options());
```
pbielak: Ditto

pbielak (changes requested):

pbielak (approved):

src/ATen/native/xpu/mkl/BatchLinearAlgebra.cpp
```diff
@@ -608,9 +636,9 @@ Tensor& orgqr_mkl(Tensor& result, const Tensor& tau) {
             lda,
             tau_batch_ptr,
             scratchpad,
-            scratchpad_size)
-            .wait();
+            scratchpad_size);
       }
+      last_event.wait();
```
guangyey, line 641: I can't understand why we need `wait` here. cc @mwiktor-intel, could you please help me understand a bit.

guangyey (approved):
I am sure that these host sync (event.wait) were unnecessary. But it was introduced by the original PR not this one.
Let's land this PR and remove them in a separate PR. cc @CuiYifeng

CuiYifeng:
> I am sure that these host sync (event.wait) were unnecessary. But it was introduced by the original PR not this one. Let's land this PR and remove them in a separate PR. cc @CuiYifeng

Agree.

PR #4808, opened 2026-08-05
Reduce op count in sparse softmax get_offsets

cyyever:
`get_offsets` used a host loop over every (element, sparse dim) pair, about five
ATen ops each, to compute a strided sum — `O(nnz * ndim)` dispatches, on all four
call sites. It now folds one dim per `add_` with the host stride as alpha, so the
cost is one add per dim regardless of `nnz`, and the device strides tensor it fed
goes away.

Also drops the prefill on four staging buffers that are fully written before any
read.

Not validated on hardware.

Test: **upstream PyTorch tests** — `pytest -v test/xpu/test_foreach_xpu.py -k max`, `pytest -v test/xpu/test_sparse_xpu.py -k softmax`

Authored with Claude Code.


guangyey (approved):
LGTM.

guangyey:
@torchxpubot ut-check

Issue #2140, opened 2025-10-09
Consider how to avoid copy in FFT kernels


CuiYifeng:
## Summary
The issue proposes avoiding the use of `copy_` in the three XPU FFT kernel output functions: `_fft_c2c_mkl_out`, `_fft_c2r_mkl_out`, and `_fft_r2c_mkl_out`. This is a performance enhancement to eliminate unnecessary tensor copies in the FFT kernel pipeline.

## Type
- **Category:** performance
- **Platform:** xpu
- **Related Components:** _fft_c2c_mkl_out, _fft_c2r_mkl_out, _fft_r2c_mkl_out (FFT kernels)

## Objective
Eliminate the `copy_` call in `_fft_c2c_mkl_out`, `_fft_c2r_mkl_out`, and `_fft_r2c_mkl_out` to avoid unnecessary tensor copies during FFT operations.

## Action Items
- [x] 🔍 Issue formatted (Discovery Agent)
  <details><summary>Discovery log</summary>
  
  **[2026-05-10 22:37:04]**
  **Summary:** The issue proposes avoiding the use of `copy_` in the three XPU FFT kernel output functions: `_fft_c2c_mkl_out`, `_fft_c2r_mkl_out`, and `_fft_r2c_mkl_out`. This is a performance enhancement to eliminate unnecessary tensor copies in the FFT kernel pipeline.
  **Failed tests:** N/A
  **Dependency:** N/A
  **Commit scope:** N/A
  
  </details>

## Environment
<details><summary>collect_env</summary>

</details>

## Original Issue
<details><summary>Original issue body</summary>

@/tmp/issue_2140_original.txt

</details>

CuiYifeng:
@laifenxiawucha xpu-issue-triage

laifenxiawucha:
Review requested in [this mention](https://github.com/intel/torch-xpu-ops/issues/2140#issuecomment-4265039245).

---

## Likely enhancement

This is a performance optimization request to avoid unnecessary tensor copies in FFT MKL backend output functions.

## Next step

Investigate whether the output tensor `out` can be pre-allocated and passed directly to the underlying `_fft_c2c_mkl`, `_fft_c2r_mkl`, and `_fft_r2c_mkl` functions instead of allocating a temporary `result` tensor and then copying it with `out.copy_(result)`.

## Additional context needed

To evaluate this enhancement properly, please provide:

[...]

PR #4811, opened 2026-08-05
Write FFT out kernels directly to output

laifenxiawucha:
Related to [#2140](https://github.com/intel/torch-xpu-ops/issues/2140).

### Why

The XPU MKL FFT `c2c`, `r2c`, and `c2r` `out=` paths previously computed into a new tensor, then copied the result to the caller's `out`. That extra allocation and device copy are avoidable when the FFT can safely use `out` as its destination.

### What changed

The functional and `out=` paths now share implementations that place the final FFT pass in `out` when its dtype and contiguous strides match, the input does not alias it, and the transform preserves its layout. Multi-pass transforms use scratch storage as needed. Other cases retain the old result-then-copy behavior, including conversions, unsupported layouts, and aliases. ComplexHalf cast/normalization order is preserved. The follow-up also makes `promote_fft_input` reuse the single dtype mapping in `promote_fft_dtype`.

**Before:** FFT → allocated result → `copy_` to `out`  
**After (eligible cases):** FFT final pass → `out`, without the final copy.

### Measured results

On an Intel Data Center GPU Max 1100, implementation commit [`e68217d2`](https://github.com/intel/torch-xpu-ops/commit/e68217d25f0a20cf42fef982db4354079b6b7ff8) was compared with base [`cedc716c`](https://github.com/intel/torch-xpu-ops/commit/cedc716cf3cbdb62e7e02a683daac1255ebb3541) using the same PyTorch/toolchain and four order-balanced rounds. Across 15 representative `out=` cases, the geometric mean of paired wall-time ratios was:

| Operation | Cases | Time vs base |
| --- | ---: | ---: |
| c2c | 6 | -8.35% |
| c2r | 4 | -11.96% |
| r2c | 5 | -4.87% |

`unitrace` confirmed the removed work in representative direct-write cases (single-pair Level Zero device times):

| `out=` workload | Base → candidate µs/op | Copies/op |
| --- | ---: | ---: |
| c2c complex128, 65536×64 | 297.187 → 122.195 | 1 → 0 |
| r2c float32, 256×256×256 | 311.792 → 148.984 | 1 → 0 |
| c2r complex32, 256×256×129 | 727.914 → 521.226 | 6 → 4 |

A prime-size r2c float32 1000×1021 `out=` case measured +4.02% wall time with 12.71% within-run spread; unitrace measured +0.13% device time while removing one copy. This noisy case does not prove the absence of a regression.

Correctness tests on [`23a2479a`](https://github.com/intel/torch-xpu-ops/commit/23a2479ab89944bf3d82517a71d4e87682a1bceb): FFT `out=` regressions 14 passed; hfft2 references 8 passed; spectral suite 340 passed, 6 skipped, 1 deselected.

### DIFF
--- a/src/ATen/native/xpu/mkl/SpectralOps.cpp
+++ b/src/ATen/native/xpu/mkl/SpectralOps.cpp
@@ -22,6 +22,7 @@
 #include <ATen/ops/mul.h>
 #include <ATen/ops/real.h>
 #include <ATen/ops/zeros_like.h>
+#include <c10/util/strides.h>
 #include <comm/SYCLContext.h>
 #include <comm/TensorInfo.h>
 #include <oneapi/mkl.hpp>
@@ -306,6 +307,10 @@ static bool _exec_fft_preserves_layout(const Tensor& self, IntArrayRef dim) {
   return true;
 }
 
+static bool _has_exact_contiguous_strides(const Tensor& tensor) {
+  return tensor.strides() == c10::contiguous_strides(tensor.sizes());
+}
+
 // Result type of promote_fft_input, without materializing the promoted tensor.
 ScalarType promote_fft_dtype(ScalarType dtype) {
   if (dtype == ScalarType::Half || dtype == ScalarType::BFloat16)
@@ -329,13 +334,17 @@ static void _fft_c2c_mkl_out_impl(
   auto sorted_dims = impl::_sort_dims(self, dim);
   auto out_sizes = self.sizes();
   auto input_sizes = self.sizes();
+  const auto result_dtype = orig_self.scalar_type();
+  const bool needs_result_conversion = self.scalar_type() != result_dtype;
 
   const auto pass_count =
       (sorted_dims.size() + impl::mkl_max_ndim - 1) / impl::mkl_max_ndim;
   const bool needs_type_conversion = self.scalar_type() != out.scalar_type();
   // A multi-pass transform hands its last pass the outermost dims, so the
   // permutation is no longer the identity even when the full dim set is.
-  const bool can_write_out = !needs_type_conversion && out.is_contiguous() &&
+  const bool can_write_out = !needs_type_conversion &&
+      (!preserve_out_layout || out.scalar_type() == result_dtype) &&
+      impl::_has_exact_contiguous_strides(out) &&
       (!preserve_out_layout ||
        (pass_count == 1 &&
         impl::_exec_fft_preserves_layout(self, sorted_dims)));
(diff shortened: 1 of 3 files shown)


src/ATen/native/xpu/mkl/SpectralOps.cpp
```diff
@@ -332,58 +332,73 @@ Tensor promote_fft_input(const Tensor& input) {
...
     const auto max_dims =
         std::min(static_cast<size_t>(impl::mkl_max_ndim), sorted_dims.size());
     auto fft_dims =
         IntArrayRef(sorted_dims).slice(sorted_dims.size() - max_dims, max_dims);
+    const auto remaining_passes = pass_count - pass;
+    Tensor* pass_out = remaining_passes % 2 == 1 ? &fft_out : &scratch;
 
     impl::_exec_fft(
-        out,
+        *pass_out,
```
CuiYifeng: Can we use `Tensor&` instead of `Tensor*`?

src/ATen/native/xpu/mkl/SpectralOps.cpp
```diff
@@ -482,86 +527,133 @@ Tensor& _fft_c2r_mkl_out(
...
+  const auto pass_count = 1 + c2c_pass_count;
+  Tensor scratch;
+  if (pass_count > 1) {
+    scratch = at::empty(out_sizes, self.options().dtype(fft_dtype));
+  }
 
   auto working_tensor = self.contiguous();
 
   // First do the R2C transform on the last dimension
+  Tensor* pass_out = pass_count % 2 == 1 ? &fft_out : &scratch;
```
CuiYifeng: Ditto.

test/xpu/test_spectral_ops_xpu.py
```diff
@@ -138,11 +138,54 @@ def _compare_xpu_cpu(self, xpu_result, cpu_result, t):
     self.assertEqual(xpu_result, cpu_result, exact_dtype=False)
 
 
+def _test_fft_out_variants(self, device):
```
CuiYifeng: Please add xpu specific UT in `test/regression/`.

CuiYifeng (changes requested):
Do we have any data to ensure there's no performance regression?

laifenxiawucha:
> Do we have any data to ensure there's no performance regression?

The updated data is included in the PR description.

src/ATen/native/xpu/mkl/SpectralOps.cpp
```diff
@@ -282,60 +283,151 @@ Tensor& _exec_fft(
...
+}
+
+// Result type of promote_fft_input, without materializing the promoted tensor.
+ScalarType promote_fft_dtype(ScalarType dtype) {
+  if (dtype == ScalarType::Half || dtype == ScalarType::BFloat16)
+    return ScalarType::Float;
+  if (dtype == ScalarType::ComplexHalf)
+    return ScalarType::ComplexFloat;
+  return dtype;
+}
```
CuiYifeng, line 321: There is a function `promote_fft_input` that is semantically similar to `promote_fft_dtype`: https://github.com/intel/torch-xpu-ops/blob/8bc8bfe57874ceb750aa324fbb959be020225464/src/ATen/native/xpu/sycl/FFTKernelFunctor.cpp#L177-L184
Could you unify these two functions?

laifenxiawucha: done

PR #4820, opened 2026-08-05
Fix carry block mapping in the pstl scan

cyyever:
`_scan_kernel` scans each block, then adds the per-block carry back. Step 1 writes
`carry_ptr[group_id]` for blocks of `kssc_wgroup_size`, but step 3 re-derived its
group count from `syclMaxWorkGroupSize(kfn3)` and indexed `carry_ptr` by its own
group id. Those are separate per-kernel queries — `ScanAccumulateKernelFunctor`
uses one element of local memory where `KSScanWithCarrierKernelFunctor` uses one
per work item — so the sizes may differ. A larger accumulate size gives every
block past the first the wrong carry, silently corrupting any scan with `N` above
one work group; a smaller one reads past the `ngroups`-element carry tensor.
Step 3 now reuses step 1's size.

Latent rather than observed: it only misbehaves when the two queries disagree,
which they do not on the hardware this has run on. Affects everything routed
through `pstl::exclusive_scan`/`inclusive_scan` — sparse softmax, sparse CSR add,
embedding backward, indexing, nonzero.

Not validated on hardware.

Test: **upstream PyTorch tests** — `pytest -v test/xpu/test_sparse_xpu.py -k softmax`, `pytest -v test/xpu/test_indexing_xpu.py`, `pytest -v test/xpu/test_nn_xpu.py -k embedding`

Authored with Claude Code.


src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h
```diff
@@ -307,14 +307,11 @@ static inline OutputIt _scan_kernel(
   _scan_kernel<0>(carry_ptr, carry_ptr + ngroups, carry_ptr, (T)0);
 
   // 3. reduce among all work groups and flush data to dst
+  // Same work-group size as step 1: each item finds its carry by group id.
   ScanAccumulateKernelFunctor<OutputIt, T> kfn3(d_first, carry_ptr, N);
-
-  const auto sa_wgroup_size = syclMaxWorkGroupSize(kfn3);
-  ngroups = (N + sa_wgroup_size - 1) / sa_wgroup_size;
```
guangyey, line 313: Good catch!
Let's add a `TORCH_CHECK(kssc_wgroup_size <= sa_wgroup_size, "...")` here to ensure the `kssc_wgroup_size` is a good value to kernel `kfn3`

guangyey (approved):

guangyey:
@torchxpubot review

src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h
```diff
@@ -307,14 +307,15 @@ static inline OutputIt _scan_kernel(
   _scan_kernel<0>(carry_ptr, carry_ptr + ngroups, carry_ptr, (T)0);
 
   // 3. reduce among all work groups and flush data to dst
+  // Same work-group size as step 1: each item finds its carry by group id.
   ScanAccumulateKernelFunctor<OutputIt, T> kfn3(d_first, carry_ptr, N);
 
   const auto sa_wgroup_size = syclMaxWorkGroupSize(kfn3);
-  ngroups = (N + sa_wgroup_size - 1) / sa_wgroup_size;
+  TORCH_CHECK(kssc_wgroup_size <= sa_wgroup_size, "_scan_kernel: work group size doesn't match!")
```
guangyey:
```suggestion
  TORCH_CHECK(
      kssc_wgroup_size <= sa_wgroup_size,
      "_scan_kernel: work group size doesn't match!")
```

src/ATen/native/xpu/sycl/pstl/PSTLFunctions.h
```diff
@@ -307,14 +307,17 @@ static inline OutputIt _scan_kernel(
...
 
   // 3. reduce among all work groups and flush data to dst
+  // Same work-group size as step 1: each item finds its carry by group id.
   ScanAccumulateKernelFunctor<OutputIt, T> kfn3(d_first, carry_ptr, N);
 
   const auto sa_wgroup_size = syclMaxWorkGroupSize(kfn3);
-  ngroups = (N + sa_wgroup_size - 1) / sa_wgroup_size;
+  TORCH_CHECK(
+      kssc_wgroup_size <= sa_wgroup_size,
+      "_scan_kernel: work group size doesn't match!")
```
guangyey:
```suggestion
  TORCH_INTERNAL_ASSERT(
      kssc_wgroup_size <= sa_wgroup_size,
      "_scan_kernel: work group size doesn't match!");
```

guangyey (approved):

PR #4826, opened 2026-08-06
Replace type-punning casts with std::bit_cast

cyyever:
Reinterpreting a value through a casted pointer in the radix-sort key conversions and the scatter/gather scalar-fill paths violates strict aliasing. `std::bit_cast` (C++20) is well-defined and generates identical code.

The `__*_as_*` wrappers in `src/comm/Scalar.h` are then unused, so the header and its lone include site in `SYCLHelpers.h` are removed.

## Test Plan

Build only; codegen is identical.

---


guangyey (commented):
I think this looks good to me. Let's see if there are any other comments or if anyone has a different perspective.

jianyizh (approved):

guangyey (approved):

guangyey:
@torchxpubot merge -f

PR #4829, opened 2026-08-06
[xccl] Register SymmetricMemory allocator for XPU MemPool

Chao1Han:
## Summary

Makes `torch.distributed._symmetric_memory` usable through a `torch.xpu.MemPool` on XPU.
Three pieces:

1. Register the XPU SymmetricMemory allocator as a MemPool allocator backend.
2. Teach `rendezvous()` about non-zero offsets, which is what the MemPool path
   actually produces.
3. Tests.
## Motivation

The device-agnostic symmetric memory Python layer already lists XPU:

```python
_MEMPOOL_DEVICE_TYPES = ("cuda", "xpu")
```

but nothing on the XPU side ever called `register_mempool_allocator()`, so
`symm_mem.get_mem_pool(torch.device("xpu:0"))` failed with

```
SymmetricMemory MemPool did not find backend for device type xpu
```

That blocks every consumer of the pooled path, including:

* `symm_mem.empty()` itself — it routes through the implicit pool by default
  (`TORCH_SYMMMEM_IMPLICIT_POOL` defaults to `"1"`),
* FSDP2 (`_fsdp_collectives.py`, `ProcessGroupAllocMixin.allocate()`), whose whole
  reason for using a MemPool is *"leverage MemPool to reuse the symmetric buffer,
  avoiding allocation and rendezvous overhead"*,
* the low-contention all-gather path in `_symmetric_memory/__init__.py`,
* Inductor's `EnterCudaMemPoolContextLine` codegen.

## Notes

* `no_split=True` / `use_on_oom=False` on the pool are deliberate: one segment owns
  exactly one signal pad, so a segment must not be split between tensors, and a
  non-symmetric allocation must never be allowed to borrow from it.
* XPU lays a block out as `[data | signal_pad]` and keys `ptr_to_block_` by the
  allocation base, whereas CUDA uses `[signal_pad | data]` keyed by the data pointer.
  Both are internally consistent; this PR does not change the layout, it only makes
  the offset arithmetic explicit.
* **Behavior change:** two `rendezvous()` calls on the same tensor no longer return
  the identical handle object (`hdl is hdl2` goes `True` -> `False`). They now return
  distinct lightweight `XPUSymmetricMemory` objects sharing one cached
  `XPUPeerAllocInfo`, so `hdl.get_buffer(...).data_ptr()` still matches. This is the
  same semantics as CUDA.
  
  ## Dependency

Requires https://github.com/pytorch/pytorch/pull/192048 https://github.com/pytorch/pytorch/pull/197065


src/xccl/XPUSymmetricMemoryTypes.hpp
```diff
@@ -2,11 +2,6 @@
 
 namespace c10d::symmetric_memory {
 
-// Default signal-pad size for each rank's control area.
-// 2048 keeps parity with the CUDA-side default and has worked as a practical
-// baseline for channelized signaling. This is a default value; higher-level
-// symmetric-memory configuration can override the effective pad size.
-constexpr size_t signal_pad_size = 2048;
```
zhangxiaoli73, line 9: Remove this define, then what's the signal pad size that you use now?

Chao1Han: This is dead code, get_signal_pad_size is actually defined in `torch\csrc\distributed\c10d\symm_mem\SymmetricMemory.cpp` and the value is 9216(share with cuda). So I want to remove it.

```cpp
constexpr int max_cuda_p2p_domain_size = 72;   // NVL72
constexpr int symm_max_nblocks = 32;
constexpr size_t default_signal_pad_size =
    symm_max_nblocks * max_cuda_p2p_domain_size * sizeof(uint32_t); =9216
 ```

zhangxiaoli73 (approved):

zxd1997066 (approved):

Chao1Han:
@torchxpubot merge -f "distributed cases passed"

PR #4845, opened 2026-08-07
Detect the existence of kernel on XPU before registering fallback

CuiYifeng:
This PR allows the fallback list to safely contain ops that may have gained native XPU support in pytorch without requiring synchronized removal.
Before registering a CPU fallback for an op, query the c10::Dispatcher to check if a kernel is already registered for that op on the XPU dispatch key. If so, skip the fallback registration to avoid the warning of overriding a previously registered kernel.


guangyey (commented):
This will silently ignore the overide issue. Is it possible to add a warning.

CuiYifeng:
> This will silently ignore the overide issue. Is it possible to add a warning.

Updated.

guangyey (approved):

pbielak (approved):

CuiYifeng:
@torchxpubot ut-check

CuiYifeng:
@torchxpubot merge -f "All failed cases also exist in the other PRs, so these failures are irrelevant to this PR"

PR #4846, opened 2026-08-07
Modernize path and policy handling in SYCL CMake modules

cyyever:
Two independent changes, one per commit so they can be reverted separately.

**`cmake_path` instead of `get_filename_component`** in the per-source library loops. `FindONEMKL` deliberately keeps `REALPATH`: on Windows `SYCL_ROOT` arrives backslashed from `setvars.bat`, and `cmake_path` (which only recognizes `/`) would collapse it to `../mkl/latest` and fail the oneMKL lookup. A comment records this so it isn't "modernized" later.

**Policy pin in `run_sycl.cmake`** — the script runs under `cmake -P`, which inherits no policy version, so it set `CMP0007`/`CMP0010` by hand. Declaring the dialect covers those plus `CMP0054`, which governs the `if()` calls there and was left at OLD.

> Merging the second commit forces one full SYCL rebuild: `FindSYCL.cmake` lists `run_sycl.cmake` in `DEPENDS` for every SYCL object.

## Test Plan

Not built. Verified with `cmake -P`:

- `NAME_WLE` vs `STEM LAST_ONLY` over the real source glob — 377 sources, 0 differences
- `REALPATH` vs `NORMAL_PATH` over POSIX and native-Windows `SYCL_ROOT`, against a replica oneAPI tree with real symlinks — identical for POSIX, divergent for backslashed input

---

### DIFF
--- a/src/BuildOnLinux.cmake
+++ b/src/BuildOnLinux.cmake
@@ -34,7 +34,7 @@ endmacro()
 if(BUILD_SEPARATE_OPS)
   setup_common_libraries()
   foreach(sycl_src ${ATen_XPU_SYCL_SRCS})
-    get_filename_component(name ${sycl_src} NAME_WLE)
+    cmake_path(GET sycl_src STEM LAST_ONLY name)
     set(sycl_lib torch-xpu-ops-sycl-${name})
     sycl_add_library(
       ${sycl_lib}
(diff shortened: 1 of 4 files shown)


guangyey (approved):

src/BuildOnLinux.cmake
```diff
@@ -34,7 +34,7 @@ endmacro()
 if(BUILD_SEPARATE_OPS)
   setup_common_libraries()
   foreach(sycl_src ${ATen_XPU_SYCL_SRCS})
-    get_filename_component(name ${sycl_src} NAME_WLE)
```
guangyey, line 37: FYI @jianyizh 
If this code change makes `BUILD_SEPARATE_OPS` build fail, please tell me.

PR #4862, opened 2026-08-07
[WIN] Skip slow forward-AD composite-compliance test on Windows

### LINKED ISSUE #4835: [UT] Test case runs more than 7 hours on CI windows runner
### 🐛 Describe the bug

Please refer https://github.com/intel/torch-xpu-ops/actions/runs/31014497073/job/92334961094#step:9:4600 which runs on a MTL Windows node

```
Wed, 05 Aug 2026 14:39:06 GMT
test_ops_xpu.py::TestCompositeComplianceXPU::test_forward_ad_nn_functional_mse_loss_xpu_float32 PASSED [ 86%]
Wed, 05 Aug 2026 22:10:18 GMT
test_ops_xpu.py::TestCompositeComplianceXPU::test_forward_ad_nn_functional_multi_head_attention_forward_xpu_float32 PASSED [ 86%]
```

Failed on BMG Linux
Cases:
op_extended,third_party.torch-xpu-ops.test.xpu.extended.test_ops_xpu.TestCompositeComplianceXPU,test_forward_ad_nn_functional_multi_head_attention_forward_xpu_float32

### Versions

main

PatrykWilczewski:
Problem: TestCompositeComplianceXPU.test_forward_ad_nn_functional_multi_ head_attention_forward_xpu_float32 ran >7h on the MTL Windows CI runner, hanging the "Run OP UT" job. The thread-based 600s tiemout cannot preempt the GIL-holding (Global Interpreter Lock) native XPU call, so the test never times out.

Root cause: composite_compliance.check_forward_ad_formula wraps 2^(num_tensor_args) subclass combinations at two nesting levels (args x tangent_args). Multi_head_attention_forward's require_grad sample has 10 tensor-like args -> 2^20 = 1.048.576 forward executions plus 2 assertEqual each. This is an intrinsic upstream test-infra cost; it only becomes a CI timeout on the slow Meteor Lake integrated GPU, and doesn't hand the non-Windows runners.

Fix:
- skip_list_win.py: skip that single forward-AD variant on Windows only, so non-Windows coverage is preserved.
- run_test_with_windows_nightly.py: copy the merged skip list into a fresh list. The skip_list_common value for test_ops_xpu.py is a tuple, so the existing skip_list.extend(...) raised AttributeError (and mutated the shared dict for list values). This fix is required for the Windows skip to take effect on the nightly runner.

Fixes: https://github.com/intel/torch-xpu-ops/issues/4835

### DIFF
--- a/test/xpu/skip_list_win.py
+++ b/test/xpu/skip_list_win.py
@@ -86,4 +86,7 @@ skip_dict = {
         "test_reference_numerics_large_sin_xpu_complex64",
         "test_reference_numerics_small_acos_xpu_complex32",
     ),
+    "test_ops_xpu.py": (
+        "test_forward_ad_nn_functional_multi_head_attention_forward_xpu_float32",
+    ),
 }
--- a/test/xpu/run_test_with_windows_nighltly.py
+++ b/test/xpu/run_test_with_windows_nighltly.py
@@ -107,9 +107,7 @@ for key in skip_dict:
         print(f"\n=== Skipping test file: {key} ===")
         continue
 
-    skip_list = skip_dict.get(key)
-    if skip_list is None:
-        skip_list = []
+    skip_list = list(skip_dict.get(key) or [])
 
     if IS_WINDOWS and key in skip_dict_win:
         win_skip_list = skip_dict_win[key]


test/xpu/skip_list_win.py
```diff
@@ -86,4 +86,7 @@
         "test_reference_numerics_large_sin_xpu_complex64",
         "test_reference_numerics_small_acos_xpu_complex32",
     ),
+    "test_ops_xpu.py": (
```
tszulist-hbn, line 89: Could you double-check the skip location? I may be misreading the log. In run 31014497073 the 7h31m gap (log line 6971, 14:39:06 -> 22:10:18) looks like it falls in the "Run XPU OP Extended UT" step, not "Run OP UT" - the step's `##[group]` markers are at lines 2372 and 7816, and its output is serial (no `[gw0]` prefixes). The later 02:44:33 occurrence is in "Run OP UT" and is already SKIPPED there on main.

 If that's right, the entry added to `test/xpu/skip_list_win.py` wouldn't take effect: the extended runner is launched with cwd `test/xpu/extended/` (_windows_ut.yml:204-205), so its `from skip_list_win import skip_dict` picks up `test/xpu/extended/skip_list_win.py` instead. Should this go in `extended/skip_list_win_mtl.py`, or into the `@skipOps` at extended/test_ops_xpu.py:249? 

Related: upstream test_ops.py:2014 already skips this op for `test_forward_ad`, and locally the test report

dominikjastrzebskix:
I will add my take here:
This test takes ~4h30m on ubuntu and B580. 
We could shorten it instead of skipping it.

In `common_methods_invocations` we have a function: [sample_inputs_multi_head_attention_forward](https://github.com/pytorch/pytorch/blob/d73d1231c6de8e90c49a9a99ba70ffed5e69a44b/torch/testing/_internal/common_methods_invocations.py#L10090)

In this function we could change the logic for `requires_grad` from:
```
bias_k = make_input(emb_size)
bias_v = make_input(emb_size)
in_proj_bias = make_input(emb_size * 3)
out_proj_weight = make_input(emb_size, emb_size)
out_proj_bias = make_input(emb_size)
sample_args = (
    k, v, emb_size, num_heads, in_proj_weight,
    in_proj_bias, bias_k, bias_v, False,
    dropout_p, out_proj_weight, out_proj_bias
)
```

to:

```
if requires_grad:
# "Remove" 4 tensors from sample_args 
    bias_k = bias_v = in_proj_bias = out_proj_bias = None
else:
    bias_k = make_input(emb_size)
    bias_v = make_input(emb_size)
    in_proj_bias = make_input(emb_size * 3)
    out_proj_bias = make_input(emb_size)
sample_args = (
    k, v, emb_size, num_heads, in_proj_weight,
    in_proj_bias, bias_k, bias_v, False,
    dropout_p, out_proj_weight, out_proj_bias
)
```
Then this tests finishes in a ~minute. We could go further and change:
`mask_types = ("2d",)` -> `mask_types = (None,)`
Which effectively removes "5th tensor" and drops the time below 20s. 

Without these changes I'd argue the better way is simply removing this test. In the current state this test will never "fix itself" or start working in a future, because the core assumption here is wrong. "Skipping" it will make it a dead code forever (until someone changes it's logic or removes it).

PatrykWilczewski:
@dominikjastrzebskix done as suggested.
Commit in pytorch - https://github.com/pytorch/pytorch/pull/192989
Closing this PR.

PatrykWilczewski:
Reopening due to https://github.com/pytorch/pytorch/pull/192989#pullrequestreview-5079169630

PR #4865, opened 2026-08-07
Optimize contiguous cat with vectorized stores

lchen2331:
Optimizes the contiguous cat path by adding a vectorized copy kernel that uses wide stores in addition to the existing vectorized loads.

The existing `CatArrayBatchedCopy_alignedK_contig` path loads multiple elements at once, but still computes the output offset and stores each element separately. For layouts where each input slice and the output row pitch are multiples of the vector width, the mapping can be simplified to a direct vector-to-vector copy.

This PR adds `CatArrayBatchedCopy_vectorized`, which:

- Uses 16-byte vector loads/stores when possible, with an 8-byte fallback.
- Computes output offsets per vector instead of per element.
- Checks slice geometry and alignment before selecting the vectorized path.
- Falls back to the existing kernels for unsupported layouts.

**Performance**
KV-cache concat, bf16, [16,16,1139,128] + [16,16,1,128], dim=2
  | Latency | Achieved BW
-- | -- | --
Baseline | 1.00x | 1.00x
This PR | 0.66x | 1.52x

### DIFF
--- a/src/ATen/native/xpu/sycl/Shape.cpp
+++ b/src/ATen/native/xpu/sycl/Shape.cpp
@@ -27,6 +27,9 @@
 #include <ATen/ops/narrow.h>
 #include <ATen/ops/size_native.h>
 
+#include <algorithm>
+#include <vector>
+
 #include <ATen/native/xpu/sycl/Philox4x32.h>
 #include <ATen/native/xpu/sycl/ShapeKernels.h>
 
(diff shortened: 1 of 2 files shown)


src/ATen/native/xpu/sycl/Shape.cpp
```diff
@@ -464,6 +574,40 @@ void parallel_cat(
     if (max_elements_per_tensor == 0)
       continue;
 
+      // Independent of Dims, so dispatched ahead of the HANDLE_CASE switch.
+#define LAUNCH_VECTORIZED(VEC_BYTES)                                       \
+  {                                                                        \
+    sycl::range<2> vecGroup, vecRange;                                     \
+    std::tie(vecRange, vecGroup) = getCatRangeContig<scalar_t, VEC_BYTES>( \
+        max_elements_per_tensor, batchCounter);                            \
```
jianyizh, line 617: Take the shape in the PR description as an example.  [16,16,1139,128] + [16,16,1,128], dim=2

global range will be [2 (two tensor), min(num_eu*32, 16*16*1139*128 / vec_size(8) / wg_size(256) ) * 256)
This means some work items are idle ( just return) because tensor 1 has less elements than tensor 0

Also in getCatRangeContig, it use work group size=256 and const unsigned int num_eu = syclGpuEUCountPerSubslice();
  item_groups = std::min(num_eu * max_group_per_eu, item_groups); I don't think it's correct. You can check https://github.com/intel/torch-xpu-ops/pull/4722

jianyizh (changes requested):

src/ATen/native/xpu/sycl/Shape.cpp
lchen2331: Thanks for pointing these out.
1. For the idle work-items caused by tensors with very different sizes, I changed the host-side batching to group inputs by size and split batches when the estimated wasted work becomes too large. Since output offsets are computed in the original argument order, inputs can be reordered for batching without changing the concat result.
2. For the launch-range cap, I replaced `syclGpuEUCountPerSubslice() * max_group_per_eu` with a cap based on `syclMaxWorkItemsPerTile() / wg_size`, following the approach in #4722.

src/ATen/native/xpu/sycl/Shape.cpp
```diff
@@ -452,18 +612,50 @@ void parallel_cat(
...
       // We need max elements per tensor to compute range parameters
       max_elements_per_tensor = std::max(
           max_elements_per_tensor, catMetaData.nElements[batchCounter]);
     }
 
-    // Skip if the tensor is empty. Otherwise, the range dim is invalid
+    // Skip if every input in the batch is empty. Otherwise the range dim is
+    // invalid. Sorting puts the empty inputs last, so this ends the loop.
     if (max_elements_per_tensor == 0)
       continue;
```
jianyizh: You said Sorting puts the empty inputs last, should be break or continue?

src/ATen/native/xpu/sycl/Shape.cpp
```diff
@@ -398,36 +478,116 @@ void parallel_cat(
...
+      : fitsVector(ALIGNED_VEC_LOAD_BYTES_8)  ? ALIGNED_VEC_LOAD_BYTES_8
+                                              : 0;
+
+  // Batch inputs by size to reduce idle work items. A batch's launch range is
+  // determined by its largest input, so grouping similarly sized inputs avoids
+  // overprovisioning smaller ones. Output offsets are computed in argument
+  // order, allowing inputs to be reordered for batching without changing the
+  // result.
+  const unsigned nInputs = static_cast<unsigned>(inputs.size());
+  std::vector<int64_t> inputOffset(nInputs);
```
jianyizh: c10::SmallVector

src/ATen/native/xpu/sycl/Shape.cpp
```diff
@@ -398,36 +478,116 @@ void parallel_cat(
...
+    // exceed the useful work; otherwise, start a new batch with the remaining
+    // inputs.
+    const int64_t batchMax = inputs[order[i]].get().numel();
+    int64_t usefulElements = batchMax;
+    int64_t idleElements = 0;
+    unsigned batchEnd = i + 1;
+    while (batchEnd < nInputs &&
+           (batchEnd - i) < static_cast<unsigned>(batch_size)) {
+      const int64_t n = inputs[order[batchEnd]].get().numel();
+      if (idleElements + (batchMax - n) > usefulElements + n)
```
jianyizh, line 551: add a comment about the heuristic?

jianyizh (approved):

src/ATen/native/xpu/sycl/Shape.cpp
lchen2331: You're right. `break` is more appropriate here. I've updated it.

src/ATen/native/xpu/sycl/Shape.cpp
lchen2331: Thanks for the suggestion. Updated to use `c10::SmallVector`.

src/ATen/native/xpu/sycl/Shape.cpp
lchen2331: Added a comment explaining the heuristic and when batching stops based on idle versus useful work.

src/ATen/native/xpu/sycl/Shape.cpp
```diff
@@ -337,6 +348,72 @@ struct CatArrayBatchedCopy_alignedK_contig {
...
+    using LT = memory::aligned_vector<T, kILP>;
+
+    const IndexType group = item.get_group(0);
+
+    // Exact: the host guarantees whole vectors per slice.
+    const IndexType nVectors = inputs.nElements[group] / kILP;
+    // Empty inputs are valid and require no work. Return early to avoid
+    // dividing by inputSliceVectors, which is zero in this case.
+    if (nVectors == 0)
+      return;
```
CuiYifeng: `nVectors` can be zero in two cases: (1) `inputs.nElements[group] == 0`, or (2) `inputs.nElements[group] < kILP`.

For the vectorized path, I notice that no kernel is launched for empty tensors (the host-side `batchMax == 0` break prevents it), and all input tensors are required to be aligned to `ALIGNED_VEC_LOAD_BYTES` — which implicitly guarantees `nElements >= kILP`.
Could you explain under what conditions either (1) or (2) can actually be triggered here? If neither can occur on this path, the guard adds unnecessary overhead to device code and could be removed.

CuiYifeng (commented):
Overall, LGTM.

src/ATen/native/xpu/sycl/Shape.cpp
lchen2331: Thanks for points this out. You’re right about case (2). But case (1) was still possible. For example, with sorted input sizes `[N, 0]`, the batching condition for the empty input becomes:
```
0 + (N - 0) > N + 0
N > N  // false
```
So the empty input would be added to the batch and get a kernel group.
The device-side guard wasn’t the best way to handle this. I have changed the host batching logic to stop before adding empty inputs:
```
if (n == 0)
  break;
```
With empty inputs excluded from kernel launches, I removed the `nVectors == 0` guard.

CuiYifeng: Thanks for the elaboration!

CuiYifeng (approved):

lchen2331:
The failing UT, `test_execution_trace_env_enabled_with_pt2_xpu`, is caused by a missing file generated by TorchInductor:

`FileNotFoundError: [Errno 2] No such file or directory: '/tmp/torchinductor_root/r3/cr3z25rxexxnkmt5ftijiorxcqsimwwb56y25efpcvhpvsaf4x4m.py'`

This failure is unrelated to this PR.

lchen2331:
@torchxpubot merge -f "Failed case test_execution_trace_env_enabled_with_pt2_xpu is irrelevant to this PR"

lchen2331:
@torchxpubot merge -f "Failed case test_execution_trace_env_enabled_with_pt2_xpu is irrelevant to this PR"

Issue #4824, opened 2026-08-06
[Release/2.14] Kineto profiler: The XPU time for the launch kernel is incorrect


RUIJIEZHONG66166:
### 🐛 Describe the bug

When testing profile feature on xpu(BMG/PVC), found the XPU time for the launch kernel(urEnqueueKernelLaunchWithArgsExp) is incorrect(Previously, XPU total and XPU time avg are 0). And what's the Instrumentation?(It wasn't there before.)

### Reproduce:
python3 -m venv stock2.14
source stock2.14/bin/activate
pip install --pre torch==2.14.0.dev20260805+xpu --index-url https://download.pytorch.org/whl/nightly/xpu
python torch-xpu-ops/test/profiling/correlation_id_mixed.py

### Details:
```
/home/sdp/ruijie/stock2.14/lib/python3.12/site-packages/torch/_subclasses/functional_tensor.py:368: UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at /__w/pytorch/pytorch/torch/csrc/utils/tensor_numpy.cpp:84.)
  cpu = _conversion_method_template(device=torch.device("cpu"))
USDT:2026-08-05 19:31:21 2160542:2160542 SyncActivityProfilerHandler.cpp:39] profiler_start
USDT:2026-08-05 19:31:21 2160542:2160542 SyncActivityProfilerHandler.cpp:46] profiler_stop
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------
                                                   Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg      Self XPU    Self XPU %     XPU total  XPU time avg    # of Calls
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------
                                              aten::add        74.38%      57.301ms        99.99%      77.029ms      25.676ms      18.400us       100.00%       2.106ms     702.027us             3
                       urEnqueueKernelLaunchWithArgsExp         0.16%     124.311us         0.58%     450.242us     150.081us       0.000us         0.00%       2.088ms     695.893us             3
                                        Instrumentation         0.28%     212.556us         0.28%     212.556us       0.741us       2.088ms     11346.09%       2.088ms       7.274us           287
[...]

PR #4892, opened 2026-08-10
[CI] Turn xpu_profiling UTs into pytest cases

RUIJIEZHONG66166:
## Motivation

The cases under `test/profiling` were standalone scripts. CI ran them, and telling whether a run was correct meant opening each log and eyeballing the profiler tables. Nothing failed the job, and `ut_result_check.sh` skipped the `xpu_profiling` suite altogether:

```bash
echo "💡 Not check the test suite results: ${ut_suite}"
```

This PR turns them into pytest cases so the suite reports pass/fail like the other UT suites, while keeping the tables in the log for manual cross checking.

## What is checked

Every case asserts two common criteria:

1. the chrome trace contains at least one device kernel with a positive duration;
2. every runtime op (`ur*`, e.g. `urEnqueueKernelLaunchWithArgsExp`) has **no XPU time** and a **positive CPU time**.

`ze*` rows are only printed for now. They are all `0.000us` in every run observed so far, but the assertion is deliberately left out until a few CI rounds confirm it.

On top of that, each case asserts what its original issue was about:

| Case | Extra check |
| --- | --- |
| `test_rn50_profiling.py` | `profiling.fp32.train.pt` is generated, holds a parseable table, and that table satisfies (1) and (2) |
| `test_correlation_id_mixed` | at least one `aten::` op is credited with device time |
| `test_missing_gpu_kernel_time` | `aten::gather` and the backward ops have non-zero device time |
| `test_time_precision` | no negative CPU/device duration over the whole iteration loop |
| `test_partial_runtime_ops` | the reported `ur*` ops are exactly the expected set |
| `test_triton_xpu_ops_time` | a triton kernel shows up with a positive duration |
| `test_llama_profiling.py` | the gemm kernel call count is stable across iterations and matches the expected value |

## Implementation notes

Device time is taken from the **exported chrome trace**, not from `key_averages().self_device_time_total`: the latter can report `0` on XPU, and it mis-attributes time when ops are nested. Kernels are mapped back to the launching op through `args["External id"]`, and the enclosing ops are recovered from the CPU timeline so an op gets the same total the `XPU total` column means.

Criterion (2) is the one exception and stays on `key_averages()` — "this row has no device time" is a property of the aggregated row, not of the trace.

Each report also prints a cross check between `sum(key_averages device rows)` and `sum(trace device activity)`. The delta is `0.000us` on every case today; it is printed rather than asserted for now.

The wrapped scripts keep working standalone (`python test/profiling/correlation_id_mixed.py` etc. print exactly what they used to), the refactor only exposes a `run_profile()` entry point.

`llama_summary.py` and `llama_baseline.csv` are removed: the gemm call count they compared is now asserted directly by `test_llama_profiling.py`.

## CI

`linux-uttest` now runs the suite through pytest with `--junitxml`, and `ut_result_check.sh` gets a `check_profiling_ut` that sums up the reports and fails the job on any failure or error, listing the failing case names.

`xpu_profiling` is not part of `pull.yml`, so this only takes effect in the nightly / on-demand runs.


RUIJIEZHONG66166:
Ondemand test results with latest nightly wheel: https://github.com/intel/torch-xpu-ops/actions/runs/31453631412

RUIJIEZHONG66166:
@torchxpubot ut-check

RUIJIEZHONG66166:
@torchxpubot ut-check

.github/actions/linux-uttest/action.yml
```diff
@@ -99,45 +99,60 @@ runs:
...
-        python -u test/profiling/time_precision_in_profile.py | \
-          tee ${{ github.workspace }}/ut_log/xpu_profiling/issue_reproduce/time_precision_in_profile.log
-        python -u test/profiling/profile_partial_runtime_ops.py | \
-          tee ${{ github.workspace }}/ut_log/xpu_profiling/issue_reproduce/profile_partial_runtime_ops.log
-        python -u test/profiling/triton_xpu_ops_time.py | \
-          tee ${{ github.workspace }}/ut_log/xpu_profiling/issue_reproduce/triton_xpu_ops_time.log
-        python -m pytest -s test/profiling/test_profiler_correctness.py | \
+        # xdist (-n / --tx from detect-devices.sh) captures stdout and would drop
+        # the profiler tables the cases print for manual check
+        export PYTEST_ADDOPTS="-v --timeout 3600 --timeout_method=thread"
```
mengfei25, line 104: there is already PYTEST_ADDOPTS for parallel, we need extend it instead of overwrite, otherwise, the parallel will be lost

chuanqi129:
> Ondemand test results with latest nightly wheel: https://github.com/intel/torch-xpu-ops/actions/runs/31453631412

Hi @RUIJIEZHONG66166 , I just checked the results, there are some profiling test cases failures. Are they expected?

RUIJIEZHONG66166:
> > Ondemand test results with latest nightly wheel: https://github.com/intel/torch-xpu-ops/actions/runs/31453631412
> 
> Hi @RUIJIEZHONG66166 , I just checked the results, there are some profiling test cases failures. Are they expected?

Yes, they are expected becuase of this issue: https://github.com/intel/torch-xpu-ops/issues/4824.
Additionally, after switching to `pytest` for unit test execution, I retained the profile table output, which can be viewed in the corresponding logs. As shown in the image below, the nightly wheel at that time still exhibited the previously mentioned issue.
image

chuanqi129:
> Yes, they are expected becuase of this issue: #4824.

Thanks @RUIJIEZHONG66166 , got it. Let's retrigger a test since the issue has been fixed.

RUIJIEZHONG66166:
> > Yes, they are expected becuase of this issue: #4824.
> 
> Thanks @RUIJIEZHONG66166 , got it. Let's retrigger a test since the issue has been fixed.

Double check with Latest nightly wheel, results: https://github.com/intel/torch-xpu-ops/actions/runs/32699667607/job/97350650816. This issue(https://github.com/intel/torch-xpu-ops/issues/4824) have been fixed. Therefore, all the unit test cases I converted to the pytest format have passed.
The remaining failures are unrelated to this issue. These unit tests were originally in pytest format.

RUIJIEZHONG66166:
Ondemand test results on BMG: https://github.com/intel/torch-xpu-ops/actions/runs/34190295075. This issue(https://github.com/intel/torch-xpu-ops/issues/4824) have been fixed. Device lost because of this issue https://github.com/intel/torch-xpu-ops/issues/5080.

chuanqi129:
@RUIJIEZHONG66166  Are those failures expected? https://github.com/intel/torch-xpu-ops/actions/runs/34190295075/job/101948481307#step:10:19

chuanqi129 (approved):

RUIJIEZHONG66166:
> @RUIJIEZHONG66166 Are those failures expected? https://github.com/intel/torch-xpu-ops/actions/runs/34190295075/job/101948481307#step:10:19

For failures in test_profiler_tree, its known issue: https://github.com/intel/torch-xpu-ops/issues/5107
For failures in test_for_overlapping_kernels, the failure is `AssertionError: Scalars are not equal!`. This error appeared in both the nightly builds and my previous on-demand tests, but since I couldn't find an existing issue for it, I opened a new one(https://github.com/intel/torch-xpu-ops/issues/5341).

mengfei25 (approved):

RUIJIEZHONG66166:
@torchxpubot merge

RUIJIEZHONG66166:
@torchxpubot merge -f

RUIJIEZHONG66166:
@torchxpubot merge -f

PR #4899, opened 2026-08-10
XPU: enable lu_factor with pivot=False in oneMKL path  

### LINKED ISSUE #3951: Implement torch.linalg.lu_factor without pivoting

Currently `torch.linalg.lu_factor` is enabled via oneMKL but if `pivot==False` it throws error. Op can't fallback to CPU, as CPU doesn't have such implementation either.

Goal of this task is to add support for lu_factor without pivoting. oneMKL `getrfnp`/`getrfnp_batch` should be used

### Alternatives

_No response_

### Additional context

Gap was found in unit tests, not model

jkosnox:
torch.linalg.lu_factor(..., pivot=False) on XPU failed for two reasons:
the oneMKL LU path unconditionally rejected no-pivot mode, and the
batch_size == 1 route fell back to CPU where no-pivot LU is not
implemented. This made XPU behavior diverge from CUDA.
    
This change implements no-pivot execution in the oneMKL LU path by
threading a get_pivots flag into the LU helper, passing null pivot
storage to getrf when pivots are not requested, and returning
CUDA-style trivial pivots (1..k) for the API output. The pivoted
path remains unchanged. The kernel routing is also adjusted so
single-batch no-pivot LU stays on oneMKL instead of falling back to CPU.

Fixes: [#3951](https://github.com/intel/torch-xpu-ops/issues/3951)

### DIFF
--- a/src/ATen/native/xpu/BatchLinearAlgebra.cpp
+++ b/src/ATen/native/xpu/BatchLinearAlgebra.cpp
@@ -12,7 +12,6 @@
 #include <ATen/native/BatchLinearAlgebra.h>
 #include <ATen/native/DispatchStub.h>
 #include <ATen/native/LinearAlgebraUtils.h>
-#include <iostream>
 #if defined(USE_ONEMKL_XPU)
 #include <ATen/native/xpu/mkl/BatchLinearAlgebra.h>
 #endif // USE_ONEMKL_XPU
@@ -24,17 +23,6 @@ void lu_solve_kernel_xpu(
     const Tensor& pivots,
     const Tensor& B,
     TransposeType trans) {
-  std::cout << "[LU_FACTOR_DEBUG] lu_solve_kernel_xpu"
-            << " LU.device=" << LU.device()
-            << " LU.dtype=" << LU.scalar_type()
-            << " LU.shape=" << LU.sizes()
-            << " pivots.dtype=" << pivots.scalar_type()
-            << " pivots.shape=" << pivots.sizes()
-            << " B.device=" << B.device()
-            << " B.dtype=" << B.scalar_type()
-            << " B.shape=" << B.sizes()
-            << " trans=" << static_cast<int>(trans)
-            << std::endl;
 #if defined(USE_ONEMKL_XPU)
   native::xpu::lu_solve_mkl(LU, pivots, B, trans);
 #else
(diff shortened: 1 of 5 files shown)


jkosnox:
XPU
case,m,n,batch,dtype,pivot,status,median_ms,scratch_bytes,samples,reason
small_square_pivot,64,64,1,float32,True,OK,12.15664750088763,22528,30,
small_square_nopivot,64,64,1,float32,False,OK,11.853495499963174,38400,30,
medium_square_pivot,256,256,1,float32,True,OK,734.9880029996712,331776,30,
medium_square_nopivot,256,256,1,float32,False,OK,751.957003998541,591872,30,
batched_small_pivot,64,64,16,float32,True,OK,12.851288500314695,340992,30,
batched_small_nopivot,64,64,16,float32,False,OK,12.63951500004623,594944,30,
batched_medium_pivot,128,128,16,float32,True,OK,95.35240150034952,1337344,30,
batched_medium_nopivot,128,128,16,float32,False,OK,96.14856100051838,2368512,30,


CPU
case,m,n,batch,dtype,pivot,status,median_ms,scratch_bytes,samples,reason
small_square_pivot,64,64,1,float32,True,OK,0.012641499779419973,0,30,
small_square_nopivot,64,64,1,float32,False,ERROR,,,0,RuntimeError: linalg.lu_factor: LU without pivoting is not implemented on the CPU
medium_square_pivot,256,256,1,float32,True,OK,0.26650250038073864,0,30,
medium_square_nopivot,256,256,1,float32,False,ERROR,,,0,RuntimeError: linalg.lu_factor: LU without pivoting is not implemented on the CPU
batched_small_pivot,64,64,16,float32,True,OK,0.6632274998992216,0,30,
batched_small_nopivot,64,64,16,float32,False,ERROR,,,0,RuntimeError: linalg.lu_factor: LU without pivoting is not implemented on the CPU
batched_medium_pivot,128,128,16,float32,True,OK,0.759450000259676,0,30,
batched_medium_nopivot,128,128,16,float32,False,ERROR,,,0,RuntimeError: linalg.lu_factor: LU without pivoting is not implemented on the CPU

dominikjastrzebskix (approved):

src/ATen/native/xpu/BatchLinearAlgebra.cpp
```diff
@@ -60,13 +59,7 @@ void lu_factor_kernel_xpu(
     const Tensor& infos,
     bool compute_pivots) {
 #if defined(USE_ONEMKL_XPU)
-  int64_t batch_size = native::batchCount(input);
-  // TODO: optimize lu_factor performance on XPU when batch_size = 1
-  if (batch_size == 1) {
```
pbielak, line 65: Please compare speedup ratios for this branch (old implementation vs. new one)

test/xpu/test_linalg_xpu.py
```diff
@@ -887,8 +888,9 @@ def run_test(A, pivot, singular, fn):
                     self.assertEqual(B_, X_ @ A)
 
     sizes = ((3, 3), (5, 5), (4, 2), (3, 4), (0, 0), (0, 1), (1, 0))
-    batches = ((0,), (), (1,), (2,), (3,), (1, 0), (3, 5))
-    pivots = (True,)
+    # batches = ((0,), (), (1,), (2,), (3,), (1, 0), (3, 5))
```
pbielak: Why modify the `batches` variable? Please check the upstream implementation of this test case

test/xpu/test_linalg_xpu.py
```diff
@@ -930,9 +932,68 @@ def run_test(A, pivot, singular, fn):
                 f(torch.empty(1, 2, 2), pivot=False)
 
 
+@skipIfTorchDynamo("Runtime error with torch._C._linalg.linalg_lu_factor")
+@dtypes(torch.float)
+def linalg_lu_factor_no_pivot_regression(self, device, dtype):
```
pbielak: Shouldn't we keep regression tests in `test/regressions/`?

pbielak (changes requested):

test/xpu/test_linalg_xpu.py
jkosnox: Done

test/xpu/test_linalg_xpu.py
jkosnox: Done

src/ATen/native/xpu/BatchLinearAlgebra.cpp
jkosnox: Working on it, will comment the ratios as soon as I have results

jkosnox:
Cost of the device path relative to the CPU fallback:
lu_factor

Generally for small batch sizes this implementation is significantly slower, however the larger the batch size, the more it catches up and eventually wins with CPU

PR #4160, opened 2026-06-26
skills: refactor fix/ into generic pipeline stages with on-demand domain packs, shared by issue-handler and nightly-ci-fix

ZhaoqiongZ:
## Summary

Introduces the `fix/` skill tree — a set of generic pipeline stages for reproducing, triaging, implementing, and verifying bug fixes. Domain-specific knowledge is factored into on-demand domain packs loaded by the orchestrator after triage. Both `issue-handler` and `xpu-nightly-ci-fix` are updated to share the same pipeline.

## New files

**Pipeline stages** (all new):
- `fix/reproduce/SKILL.md` — three-stage reproduce: nightly wheel → source build → CI env
- `fix/triage/SKILL.md` — root cause analysis; emits `domain` field + Step 5.5 sanity check
- `fix/implement/SKILL.md` — apply the fix; defers skip ops to `fix/pytorch-skip`
- `fix/verify/SKILL.md` — run tests against source build; optional before/after diff and lint

**Domain packs** (all new):
- `fix/domains/xpu-kernel/SKILL.md` — XPU-specific: nightly wheel URL, scratch dir, target repo paths, upstream cross-reference, failure categories, test path resolution
- `fix/domains/cuda-porting/SKILL.md` — CUDA UT porting: locate CUDA test, sketch XPU reproducer, diff behavior, root cause table

**Standalone skill** (new):
- `fix/pytorch-skip/SKILL.md` — find/remove/add `@skipIfXpu` and all OpInfo/`instantiate_device_type_tests` skip patterns; temp-remove for reproduce; tracking issue requirement for adding

**Shared references** (all new):
- `fix/references/environment-setup.md` — generic activate + rebuild pattern
- `fix/references/failure-categories.md` — routing table: failure type → domain skill
- `fix/references/run-test.md` — generic test runner: pytest/unittest, result interpretation

**CI environment script** (new):
- `fix/reproduce/scripts/ci_env_setup.sh` — Stage 3 automation: finds the latest `pytorch/pytorch` xpu workflow run with all linux build jobs succeeded, downloads wheel artifacts from S3, derives the CI docker image tag from the `.ci/docker` tree hash, pulls the image (skips if already local), and generates a `docker run` command with `/dev/dri`, pytorch checkout, and wheels all mounted

## Modified files

- `issue-handler/SKILL.md` — adds Stage 3.5: load `fix/domains/<domain>` after triage; removes now-redundant inline fix logic (superseded by `fix/` pipeline)
[...]

PR #4918, opened 2026-08-12
skills: refactor bug-fix pipeline into fix/ leaf skills

ZhaoqiongZ:
## Summary

Refactor the bug-fix workflow under `.claude/skills/` from a monolithic
`issue-handler` skill into a `fix/`-prefixed leaf-skill pipeline
(`fix/triage` -> `fix/reproduce` -> `fix/implement` -> `fix/verify`)
with per-domain knowledge packs (`fix/domains/{xpu-kernel,inductor,upstream-pytorch}`)
and cross-cutting utilities (`fix/skip-management`). Both orchestrators
(`issue-handler`, `xpu-nightly-ci-fix`) are rewritten on top of the same
leaf skills so they share triage/reproduce/verify logic instead of
duplicating it.

Net: **20 files, +2298 / -845.**

## Review order

The PR is 40 commits. For efficient review, read the **final state** of
these files roughly in this order rather than commit-by-commit:

1. **`fix/triage/`, `fix/reproduce/`, `fix/implement/`, `fix/verify/`** -
   the new leaf skills. Each has explicit Inputs and Output blocks;
   `fix/implement` emits a machine-readable JSON block that the
   orchestrators and `fix/verify` consume by name (`changed_files`,
   `skip_added`, `tracking_issue`, `patch_proposal_mode`,
   `ready_for_verify`).
2. **`fix/domains/{xpu-kernel,inductor,upstream-pytorch}/`** - pure
   domain knowledge (paths, signatures, CUDA cross-reference). Loaded
   by orchestrators only after triage names the domain.
3. **`fix/skip-management/`** - cross-cutting: XPU skip decorator
   management (find / temp-remove for reproduction / add with tracking
   issue). Independent of any domain because skip decorators live in
   the pytorch test tree regardless of the triaged domain.
4. **`fix/reproduce/scripts/ci_env_setup.sh`** - Stage 3 CI environment
   alignment (nightly workflow -> S3 wheel artifacts -> ghcr CI image
   -> docker run). Verified against a live pytorch xpu workflow run.
5. **`issue-handler/`** - orchestrator for a single GitHub issue.
   Sequences the `fix/` leaves; owns per-issue branch isolation,
   patch-proposal mode, review subagent, and comment contract. The
   `issue-fix/`, `test-verification/` subdirs and `references/` are
   deleted; their content is inlined or moved to the leaf skills.
6. **`xpu-nightly-ci-fix/`** - orchestrator for batch nightly CI
   failures. Sequences the same `fix/` leaves per failure; owns batch
   scheduling, progress tracking (`agent_space_xpu/runs/<date>/`),
   branch strategy, and commit format.
7. **`xpu-issues-triaging/`, `xpu-ops-pr-creation/`, `xpu-build-pytorch/`,
   `skill-writer/`** - small updates. `xpu-issues-triaging` now marks
   itself as a standalone/manual entry (the automated pipeline uses
   `fix/triage`). `skill-writer` allows both basename and slash
   hierarchy in `name:` (uniqueness is the only hard requirement).

## Key contract decisions

- **`allow_skip`** is a leaf-skill input. `issue-handler` always passes
  `false`; `xpu-nightly-ci-fix` passes `true` (nightly may skip with a
  tracking issue when the fix is out of scope for a nightly run).
- **`patch_proposal_mode`** is an explicit `fix/implement` input.
  `issue-handler` sets it to `true` when `target_repo != pr_repo`
  (typically pytorch fix reached via a torch-xpu-ops issue); in that
  mode the deliverable is a diff on the issue, not a PR.
- **`ready_for_verify`** gates `fix/verify` calls. When `fix/implement`'s
  mandatory Step 3.5 skip-guard subagent rejects the diff and the
  implementer bails with `NEEDS_HUMAN`, verify is skipped entirely.
- **`fix/verify` "all skipped" handling** — verify parses pytest's
  `SKIPPED [N] <file:line>: <reason>` line: XPU-marker reasons
  (`skipIfXpu`, `xfailIfXPU`, `expectedFailureXPU`, `skipXPU`, or
  reasons containing "xpu") return `FAILED` (the fix should have
  removed the stale skip). Environmental skips return `CANNOT_VERIFY`.
  This is intentionally different from `fix/reproduce`, which
  temp-unskips via `fix/skip-management` to confirm the bug exists.
- **`fix/reproduce`** `NO_REPRODUCER` covers both a missing
  `reproducer_command` and pytest reporting `collected 0 items`
  (referenced test does not exist).

## Deduplication and cleanup

- "Forbidden workarounds" (new skip decorators, tolerance loosening
  > 1 order of magnitude, hardcoded seeds, deleted assertions, broad
  `try/except` around the failing call) is a single authoritative list
  in `fix/implement` Step 3.5. `issue-handler`'s STRICT
  patch-acceptance rules reference it.
- `Never stage third_party/xpu.txt` is a single HARD RULE in
  `fix/implement`; the three domain skills no longer repeat it.
- Tolerance-loosening threshold is `"more than an order of magnitude"`
  in both `fix/implement` and `issue-handler`.

## `ci_env_setup.sh` — validated against live CI

Fixed against a real pytorch xpu workflow run (id 31556968368):

- Job-name regex changed from `linux.*build-osdc` (matches zero real
  jobs) to `^linux.*/ build$` (real job names are like
  `linux-noble-xpu-n-py3.10 / build`).
- `BUILD_ENVS` array replaced with a `BUILD_ENV_IMAGE` associative
  array. Removed nonexistent py3.11 entries (xpu.yml only defines
  py3.10). Each build env maps to its own `docker-image-name` suffix
  (`-py3` vs `-py3-client` vs `-jammy-xpu-n-1-py3`).
- `ci-docker-hash` computed as `git rev-parse HEAD:.ci/docker` of the
  run's commit, matching `_runner-determinator.yml`'s "Compute
  .ci/docker tree hash" step exactly. Image URL:
  `ghcr.io/pytorch/ci-image:<suffix>-<hash>` (pytorch publishes to
  ghcr and ECR; ghcr is used because it does not require AWS auth).
- `unzip` no longer swallows stderr; exit code 11 (no matching files)
  triggers the `*.whl` fallback, any other non-zero exit aborts.
- `docker rm -f` on a pre-existing container only runs for the
  auto-generated default `CONTAINER_NAME` (`pytorch_xpu_ci_<timestamp>`);
  when the user passed `--container-name`, the script aborts loudly
  instead of destroying user state.
- Mount paths quoted in both the printed command and the generated
  `run_container.sh`. Script rejects paths with whitespace up front.
- `--outdir` default changed from `~/ci_env_artifacts` to
  `agent_space_xpu/ci_env`, matching the `fix/reproduce` SKILL.md
  example.

## Testing

Docs and shell only; no runtime code changed.

```bash
# Lint every changed file (all clean)
lintrunner $(git diff --cached --name-only)

# Shell syntax check
bash -n .claude/skills/fix/reproduce/scripts/ci_env_setup.sh
```

Verified `ci_env_setup.sh` assumptions against live pytorch/pytorch
xpu workflow via `gh api`:
- Workflow id 79954307 named `xpu` exists.
- Latest completed run (`31556968368`) has build jobs matching the new
  regex.
- Real docker-image name pattern extracted from build job log:
  `ci-image:pytorch-linux-noble-xpu-n-py3` +
  `ci-docker-hash: 0c3a9da87a44dbc22e9fbbe45cdb07cb76b3a062`.
- `docker manifest inspect ghcr.io/pytorch/ci-image:pytorch-linux-noble-xpu-n-py3-<hash>`
  resolves.

Authored with AI assistance (opencode).


.claude/skills/skill-writer/SKILL.md
```diff
@@ -84,10 +84,14 @@ description: Brief description of what this does and when to use it
...
+  - Lowercase letters, numbers, hyphens only; forward slashes (`/`) are
+    allowed (but not required) as hierarchy separators for nested skills
   - Max 64 characters
-  - Must match directory name
-  - Good: `pdf-processor`, `git-commit-helper`
+  - Should be unique within the skills tree; using the directory path
+    relative to `.claude/skills/` (e.g. `fix/reproduce` for a skill at
+    `.claude/skills/fix/reproduce/`) is a common convention but the
+    basename (`reproduce`) is also acceptable
+  - Good: `pdf-processor`, `git-commit-helper`, `fix/reproduce`, `reproduce`, `fix/domains/xpu-kernel`
```
CuiYifeng: `skill-writer` should be independent of `issue-handler` refactor. Can we separate the update of `skill-writer` from this PR?

.claude/skills/issue-handler/SKILL.md
```diff
@@ -1,146 +1,614 @@
...
+  follow-up PR after review. Accepted explicit values: `pytorch`,
+  `torch-xpu-ops`, or `none` (never open a PR on this run — everything is
+  patch-proposal).
 
-## Execution modes
+## Pipeline
+
+```
+issue-format → reproduce → triage → implement → verify → review → report
+```
```
CuiYifeng, line 480: The pipeline described earlier, "Sequences `fix/reproduce`, `fix/triage`, `fix/implement`, `fix/verify`, and a fresh-context **review subagent** into a single pipeline for one GitHub issue", is inconsistent with the one here.

.claude/skills/issue-handler/SKILL.md
```diff
@@ -1,146 +1,614 @@
...
+
+### Stage 1 — issue-format
+
+Classify as `bug`, `skip-list`, or `nonbug` and extract metadata.
```
CuiYifeng: Do we have any plan to refactor this part as a sub skill like `fix/*`?

.claude/skills/issue-handler/SKILL.md
```diff
@@ -1,146 +1,614 @@
...
+files in `target_repo`'s local checkout to produce and verify a concrete
+patch, but is NOT allowed to open a PR there. Deliverable is the diff
+posted to the issue for human review. A follow-up PR (if warranted) is a
+separate decision made by the reviewer, not by this skill.
+
+### Stage 3.5 — Load domain skill
+
+Read the `domain` field from the triage output. Use the skill tool to load
+`fix/domains/<domain>` (e.g. `fix/domains/xpu-kernel`). If no domain skill
+exists for the reported domain, proceed without it.
```
CuiYifeng: We need a router for domain specific skills to indicate which skills are required in future.

.claude/skills/issue-handler/issue-format/SKILL.md
```diff
@@ -40,10 +26,12 @@ Indicators: error tracebacks, failing test names, "RuntimeError", "AssertionErro
 **Non-bug** — feature requests, tasks, performance issues, questions, discussions,
 tracking issues, enhancement proposals, feature gaps.
 Indicators: "Enable", "[Task]", "Consider", "Align", "feature gap", "clarification",
-checklists of work items, "implement", `enhancement` label, `performance` label,
-no failing tests, discussion-style content.
+checklists of work items, `enhancement` label, `performance` label, no failing tests.
+
+**Labels are authoritative** — if labels say `agent_test: ut`, test_type is `ut`.
+
+## Output
```
CuiYifeng: According to “Classify as `bug`, `skip-list`, or `nonbug` and extract metadata.”, `skip-list` is not mentioned in `issue-format` but is requested by `issue-handler`.

.claude/skills/fix/triage/SKILL.md
```diff
@@ -0,0 +1,177 @@
+---
+name: fix/triage
+description: >
+  Analyze a failure and determine root cause, fix strategy, target repo,
+  domain, and verdict (IMPLEMENTING or NEEDS_HUMAN). Analysis-only — no code
+  changes. Used by both issue-handler and nightly-ci-fix orchestrators.
+---
```
CuiYifeng: Seems the functions of `fix/triage` and `issue-format` partially overlap.

.claude/skills/fix/reproduce/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+The orchestrator decides what to do with the output — this skill only reports
+the result.
+
+## Inputs
+
+- `reproducer_command` — pytest/python command or test name. If absent, return
+  `NO_REPRODUCER` immediately. Also return `NO_REPRODUCER` if the command runs
+  but pytest reports `collected 0 items` (the referenced test does not exist
+  in the tree at this base) — this is the same "no runnable reproducer"
+  condition as an absent command, just discovered a step later.
```
CuiYifeng, line 24: Can we indicate the provider for `reproducer_command`? I understand that one of the provider is issue body. Should the other provider be `fix/triage`?

.claude/skills/fix/implement/SKILL.md
```diff
@@ -0,0 +1,276 @@
...
+### Fix strategies by category
+
+See `fix/triage` Step 1 for domain routing. Common strategies:
+
+- **Tolerance:** match upstream `atol`/`rtol` values exactly.
+- **Regression:** find the guilty commit (`git log --oneline -20 -- <file>`),
+  apply a fix aligned with upstream intent; document any divergence in comments.
+- **Newly added test:** enable backend support. If `allow_skip=false` and
+  support is genuinely missing, report `NEEDS_HUMAN` — do not add a skip. If
+  `allow_skip=true`, load `fix/skip-management` and use its "Add a new skip"
```
CuiYifeng, line 88: I am confused about the positioning of the new component, it is neither a domain specific nor a stage of issue-handler.

CuiYifeng (changes requested):

.claude/skills/skill-writer/SKILL.md
ZhaoqiongZ: sure

chuanqi129:
@copilot+claude-opus-5 review with /skill-writer

.claude/skills/skill-writer/SKILL.md
ZhaoqiongZ: Done — split to #4975. This PR now reverts the skill-writer changes (commit cab6efe5) so the two PRs don't overlap.

.claude/skills/issue-handler/SKILL.md
ZhaoqiongZ: Fixed in commit 22725839: the description now lists all pipeline components and matches the Pipeline section below. Further updated in 86557e66 to mention `fix/skip-triage` as the Stage 1 dispatch target for skip-list issues.

.claude/skills/issue-handler/SKILL.md
ZhaoqiongZ: Done in commit 86557e66. Extracted the skip-list branch into `fix/skip-triage` as a peer of `fix/triage` (not a child). Contract:

- **Inputs:** `issue_body`, `pytorch_dir`, `pr_repo`.
- **Output JSON:** `verdict_table` (per-test classification) + `sub_bugs` (per `STILL_FAILING` entry: `root_cause`, `patch_diff`, `reproducer_command`, `verify_before`, `verify_after`, `status`, `target_repo`, `domain`) + `outcome` (`DONE_SKIP_TRIAGED` | `SKIP_TRIAGED_NEEDS_HUMAN`).
- `issue-handler` Stage 1 shrinks from 46 lines of inline logic to a short dispatch that posts `fix/skip-triage`'s outputs as GitHub comments and sets labels from `outcome`.

The skill is a mini-orchestrator itself (per-entry `fix/reproduce` loop + recursive full bug pipeline per sub-bug in patch-proposal mode), which is why it earned peer status with `fix/triage` rather than living under it.

.claude/skills/issue-handler/SKILL.md
ZhaoqiongZ: Done in commit 8c639a54. Added `.claude/skills/fix/domains/README.md` as the authoritative domain registry:

- Table with `domain`, `skill_path`, `target_repo`, `applies_when`, `test_locations`, `fix_locations`.
- Machine-readable JSON list of valid `domain` values.
- Explicit contracts for producer (`fix/triage`) and consumers (`issue-handler` Stage 3.5, `xpu-nightly-ci-fix` Step 4.5).
- Adding-a-new-domain checklist.

**Producer side:** `fix/triage` MUST emit a `domain` from the registry (unknown case → `NEEDS_HUMAN`, not invent a new value). Step 5.5 sanity-checks the emitted `target_repo` against the registry entry for the chosen `domain`.

**Consumer side:** Orchestrators consult the registry before loading a domain skill and fail loudly (`NEEDS_HUMAN`) on:
- `domain` value not in the JSON list
- `skill_path` directory missing
- `target_repo` mismatch between triage and registry

The previous "proceed without domain skill" silent-no-op fallback is removed — that was the exact bug PR #4160 review flagged (`upstream-pytorch` domain missing, only caught by reviewer, not by the pipeline).

.claude/skills/issue-handler/issue-format/SKILL.md
ZhaoqiongZ: Fixed in commit 22725839:

- Added a **Skip-list** classification section with indicators ("Bug Skip" template, `agent_test: skip-list` label, node-id checklist with `~~strike-through~~`, no fresh traceback).
- Updated the output enum to `"issue_type": "bug | skip-list | nonbug"`.
- Extended the label-authority note to cover `agent_test: skip-list`.

Now consistent with `issue-handler` Stage 1 (which further dispatches to `fix/skip-triage` for this class since commit 86557e66).

.claude/skills/fix/triage/SKILL.md
ZhaoqiongZ: will rename issue-format to issue-triage, fix/triage to fix/root-cause
issue-triage will give a basic analysis with issue body/label/type etc
fix/analysis will read source code and analysis for a root cause and precise scope to fix

.claude/skills/fix/reproduce/SKILL.md
ZhaoqiongZ: Documented the three real providers in commit 22725839:

1. **Issue body** — extracted by `issue-format` (via `issue-handler`).
2. **CI failure log** — the failing pytest node id from the nightly CI report (via `xpu-nightly-ci-fix`).
3. **Per-entry loop** — one test entry from a skip-list issue (via `fix/skip-triage`, formerly inlined in `issue-handler` Stage 1; extracted in commit 86557e66).

.claude/skills/fix/implement/SKILL.md
ZhaoqiongZ: `fix/skip-management` is a **cross-cutting utility** — not a pipeline stage and not a domain pack. Added a "Category" header at the top of `fix/skip-management/SKILL.md` in commit 22725839 explaining the placement:

- XPU skip decorators live in the pytorch test tree regardless of which domain triaged the failure (`xpu-kernel`, `inductor`, `upstream-pytorch` all use `@skipIfXpu`, OpInfo `DecorateInfo`, and `instantiate_device_type_tests` dict entries).
- Loaded on demand by `fix/reproduce` (temp-remove to confirm the bug) and `fix/implement` (add-with-tracking-issue when `allow_skip=true`, remove-stale when the fix lands), and no one else.
- Putting it under `fix/domains/` would force each domain to re-import it; making it a numbered stage would misrepresent it as always-run.

ZhaoqiongZ:
Confirmed 1-5 all landed on the branch (via 04a746a5 after rebase).
Two additional architectural items @CuiYifeng raised were implemented
in the same push:

- 8c639a54: fix/domains/README.md as authoritative domain registry;
  orchestrators fail loudly on unknown / missing / target_repo-mismatch
  instead of the silent-no-op fallback.
- 86557e66: fix/skip-triage extracted from issue-handler Stage 1
  (46 lines inline → short dispatch). Fix 1 (STILL_FAILING deferred
  to per-entry triage) carried over to the new skill.

Slash-hierarchy naming: agreed, follow #4975.

ZhaoqiongZ:
## Naming and responsibility cleanup (commits 555e653c, 16a8a232, 5c66cb75)

Addresses an ambiguity @CuiYifeng raised in the review — that the original `issue-format` and `fix/triage` had overlapping functions and unclear boundaries — and cleans up the "triage" naming so a single word doesn't mean three different things across the skill tree.

### The original ambiguity

Before this cleanup:

- **`issue-format`** — "format" implied a passive metadata extractor, but the skill was actually deciding classification (`bug` / `nonbug`).
- **`fix/triage`** — "triage" is a decision-making word, but this skill's real job is deep source-reading root-cause analysis, not classification.
- A reader couldn't tell which one made the "can this be fixed" call without reading both SKILL.md files.

The `issue-format` description also didn't mention some things it was actually doing — including that `issue-handler` Stage 1 depended on it emitting `skip-list`, but the skill's classification list didn't include `skip-list`.

### The new split

One clear division of labor along the shallow/deep axis:

| Skill | Layer | Access | Job |
|---|---|---|---|
| `issue-triage` | shallow | text only (issue body / labels) | Classify (`bug` / `skip-list` / `nonbug`), identify `runtime_dependencies`, estimate `scope`, emit **preliminary** `verdict`. NEEDS_HUMAN here stops the pipeline immediately. |
| `fix/root-cause` | deep | reads source | Determine **final** `root_cause`, `fix_strategy`, `target_repo`, `domain`, `verdict`. Has authority to override `issue-triage`'s preliminary `scope`/`verdict` after seeing the code. |

`issue-triage` and `fix/root-cause` are called at different pipeline stages on different inputs, so they never "both classify the same thing" the way the old pair could be read as doing.

### Concrete new behavior in `issue-triage`

Not just a rename — the shallow skill gained three real responsibilities driven by review feedback and the "what should Stage 1 actually decide" question:

1. **`runtime_dependencies`** — closed-set array (`triton`, `onednn`, `onemkl`, `driver`, `sycl`, `ipex`, `xccl`). Populated only from explicit mentions in the issue text; a traceback path alone is not enough.
2. **`scope`** — `"pytorch" | "torch-xpu-ops" | "both" | "unclear"`. Preliminary. `"unclear"` is the common case for most bug reports and is **not** `NEEDS_HUMAN`. `"both"` is passed down; `fix/root-cause` decides whether to attempt or fall back to `NEEDS_HUMAN`.
3. **`verdict`** — `"agent-fixable" | "NEEDS_HUMAN"`. `NEEDS_HUMAN` covers cases where the issue text itself lacks signal: nonbug/umbrella, no error message, no reproducer or test-name reference, hardware-only, non-public dependency.

The old `dependency: "upstream" | ""` field was removed — that signal is now carried by `scope=pytorch` / `scope=both` (one axis, one concept).

### Why the extra renames after that

Fixing `issue-format` → `issue-triage` created a new problem: "triage" now appeared in three skill names with three different meanings (`issue-triage`, `fix/triage`, `fix/skip-triage`), plus a shallow/deep convention (`fix/triage` vs `fix/analysis`) that wasn't obvious from the names. Cleaned up in commits 555e653c and 5c66cb75:

- `fix/triage` → `fix/analysis` → **`fix/root-cause`** (self-describing: says what it produces)
- `fix/skip-triage` → **`fix/skip-list`** (names the object processed, consistent with `fix/reproduce`, `fix/implement`, `fix/verify`)

**"Triage" is now used in exactly one skill name (`issue-triage`)**, so the word has a single meaning in the codebase: shallow text-only classification at the pipeline entry.

Not renamed (kept intentionally, per limited-blast-radius decision): `xpu-nightly-ci-fix` Step 4 heading "Triage each reproduced failure" (descriptive verb, not a skill name); `issue-handler` state-machine labels `TRIAGING`/`TRIAGED` and the `agent:triaged` GitHub label (existing on live issues).

### Downstream consumer updates

- `issue-handler` Stage 1 routes on `verdict` first, then `issue_type` (NEEDS_HUMAN stops the pipeline regardless of issue type). `scope` and `runtime_dependencies` flow forward into Stage 3 as hints.
- `issue-handler` Stage 3 documents `fix/root-cause`'s override authority and Step 4 explicitly handles `scope=both`.
- `fix/root-cause` `Inputs` section declares `preliminary_scope` and `runtime_dependencies` as optional hints (absent when called from `xpu-nightly-ci-fix`).

@CuiYifeng @chuanqi129 flagging both of you since this concentrates the issue-format/fix-triage overlap thread and the "concentrate 'triage' in one place" naming discussion into one comment.

ZhaoqiongZ:
@copilot+claude-opus-5  review with /skill-writer

ZhaoqiongZ:
@copilot+claude-opus-5:high review this pr in depth

ZhaoqiongZ:
Found that this pr is huge and not readable for reviewers, will seperate it to several PRs for easy review

ZhaoqiongZ:
Closing this PR — reorganizing the work into a series of focused, independent PRs, each adding one skill (or one closely-related set) to keep the review surface small and let each piece land on its own.

The refactor turned out to be too many things bundled together: `issue-triage` (shallow classification), `fix/reproduce`, `fix/root-cause`, `fix/implement`, `fix/verify`, `fix/skip-list`, `fix/skip-management`, `fix/domains/*` registry, `issue-handler` orchestrator, and `xpu-nightly-ci-fix` refresh — plus the `skill-writer` naming-rule tweak already split to #4975.

Each of these has independent value and can be reviewed against a narrower set of concerns. The current PR carries 55 commits and three rounds of deep-review fixes, which makes it hard to see the shape of any individual change.

Plan:

1. First follow-up: `issue-triage` (shallow, text-only classification of a GitHub issue; can be invoked standalone).
2. Next: `fix/reproduce` and `fix/root-cause`.
3. Then: `fix/implement`, `fix/verify`, and the skip-management pieces.
4. Then: `fix/domains/*` registry and domain packs.
5. Finally: the orchestrators (`issue-handler` refresh, `xpu-nightly-ci-fix` refresh).

Branch `skills_refactor/zzq` is preserved on origin so anything from this PR's work can be cherry-picked into the follow-up PRs. Review discussion here stays as reference.

PR #4937, opened 2026-08-13
Use const_data_ptr for read-only tensor arguments

cyyever:
`data_ptr<T>()` returns a mutable pointer however the tensor is used, so nothing stops a read-only input from being written through. This converts the call sites whose consumer *already* declares the pointer `const`, so the compiler enforces what the code assumes.

| File | Converted | Left alone |
| --- | --- | --- |
| `SparseCsrTensorAddKernels.cpp` | 6 | - |
| `DeformConv2dKernels.cpp` | 20 | `data_col`, `grad_im`, `grad_offset`, `grad_mask` |
| `RoiPoolKernels.cpp` | 5 | `output`, `argmax`, `grad_input` |
| `PsRoiPoolKernels.cpp` | 5 | `output`, `channel_mapping`, `grad_input` |
| `PsRoiAlignKernels.cpp` | 5 | `output`, `channel_mapping`, `grad_input` |
| `RoiAlignKernels.cpp` | 4 | `output`, `grad_input` |

Each conversion was checked against the consuming declaration, not inferred from the name: the sparse CSR sites bind to `const index_t*` / `const scalar_t*` locals, and the deform-conv and ROI functors declare the corresponding members `const`. Note `argmax` / `channel_mapping` differ by direction - `int*` on the forward path (written), `const int*` on the backward one (read) - so only the backward side is converted.

A wrong conversion here cannot change behavior silently; `const T*` where `T*` is expected is a compile error.

This is the first slice; roughly 316 plain `data_ptr` sites remain in `src/`, and follow-ups can take them by op family.

## Test Plan

No local XPU build is available, so compilation rests on CI.

```bash
lintrunner -a --skip CLANGTIDY $(git diff --name-only)   # ok No lint issues
git clang-format --diff HEAD    # clang-format did not modify any files
```

### DIFF
--- a/src/ATen/native/sparse/xpu/sycl/SparseCsrTensorAddKernels.cpp
+++ b/src/ATen/native/sparse/xpu/sycl/SparseCsrTensorAddKernels.cpp
@@ -202,12 +202,12 @@ void add_out_sparse_csr_kernel(
               auto a_val = A.values().to(output_values_dtype);
               auto b_val = B.values().to(output_values_dtype);
 
-              const index_t* a_crow_ptr = a_crow.data_ptr<index_t>();
-              const index_t* a_col_ptr = a_col.data_ptr<index_t>();
-              const index_t* b_crow_ptr = b_crow.data_ptr<index_t>();
-              const index_t* b_col_ptr = b_col.data_ptr<index_t>();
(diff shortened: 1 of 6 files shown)


guangyey (approved):
Very nice, thanks!

guangyey:
@torchxpubot review

PR #4971, opened 2026-08-15
Use const_data_ptr for read-only tensor arguments in sparse kernels

cyyever:
`data_ptr<T>()` returns a mutable pointer regardless of use. This converts read-only call sites in the sparse XPU kernels to `const_data_ptr`, const-qualifying the consuming functor members where they were only read. Written buffers are left alone. A wrong conversion is a compile error, not a silent behavior change.


guangyey (approved):
Thanks!

PR #4975, opened 2026-08-17
skills: allow forward slashes in skill-writer naming rule

ZhaoqiongZ:
## Summary

Split out from #4918. This is a standalone rule change to the `skill-writer` spec — independent of the `issue-handler` / `fix/` refactor.

The `skill-writer` spec said "lowercase letters, numbers, hyphens only" and "Must match directory name", which rules out slash-hierarchy names entirely. That's too strict:

- **Nested skills already exist on `main`** and use the *basename* form (`action/sycl/source-oneapi` → `name: source-oneapi`, `action/unitrace/setup` → `name: setup`, `issue-handler/issue-fix` → `name: issue-fix`). The strict rule as written would have required either renaming these or their directories.
- **#4918** proposes further nesting (`fix/reproduce`, `fix/domains/xpu-kernel`, etc.) where the *slash-path* form is a clearer fit than the basename.
- The skill tool accepts both forms as-is from the `available_skills` list.

Updated the field requirements and the pre-submit checklist to:

- Allow forward slashes as hierarchy separators (previously hyphens only)
- Accept either the directory basename or the directory path relative to `.claude/skills/` as the `name:` value
- Require uniqueness across all skills — this is the only hard rule
- Add nested examples alongside the flat ones

Author's choice for each new skill: basename or slash-path. Existing skills stay as they are.

## Testing

Docs only. `lintrunner` clean on the changed file.

Cherry-picked from #4918 (commit 850e5a9) at the request of a reviewer to keep the naming-rule change reviewable independently of the larger orchestrator refactor. #4918 will revert this commit so the two PRs don't overlap. Includes a follow-up refinement (commit cfad535, co-authored with Copilot) that relaxed the rule from "must match path" to "basename or path" after review confirmed no existing nested skill uses the strict path form.

### DIFF
--- a/.claude/skills/skill-writer/SKILL.md
+++ b/.claude/skills/skill-writer/SKILL.md
@@ -84,10 +84,12 @@ description: Brief description of what this does and when to use it
 **Field requirements**:
 
 - **name**:
-  - Lowercase letters, numbers, hyphens only
+  - Lowercase letters, numbers, hyphens only; forward slashes (`/`) are
+    allowed as hierarchy separators for nested skills
   - Max 64 characters
-  - Must match directory name
-  - Good: `pdf-processor`, `git-commit-helper`
+  - Must match the skill's directory path relative to `.claude/skills/`
+    (e.g. skill at `.claude/skills/fix/reproduce/` → `name: fix/reproduce`)
+  - Good: `pdf-processor`, `git-commit-helper`, `fix/reproduce`, `fix/domains/xpu-kernel`
   - Bad: `PDF_Processor`, `Git Commits!`
 
 - **description**:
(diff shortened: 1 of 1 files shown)


chuanqi129:
@copilot+claude-opus-5 review with /skill-writer

ZhaoqiongZ:
Confirmed all three fixes in cfad535:

1. **Basename-or-path rule** — right call. I checked the frontmatter of every skill on main; you're correct that nested skills (`action/sycl/source-oneapi`, `action/unitrace/setup`, `issue-handler/issue-fix`) all use the basename form, so the "must match path" rule as I originally wrote it would have made all four non-compliant. The relaxed rule (basename or path, uniqueness as the only hard requirement) matches actual practice.

2. **Step 8 consistency** — good catch on the file-structure block; that was a leftover I missed.

3. **ASCII arrow** — fixed.

Also updated the PR description to drop the incorrect "the repo already uses slash-separated names" claim (that was true on #4918's branch, not on main) and to reflect the "basename or path" rule the final spec landed on.

chuanqi129 (approved):

ZhaoqiongZ:
Closing along with #4918 — the reorganized follow-up PRs will use flat, hyphen-only skill names, so the "allow forward slashes as hierarchy separators" rule change this PR proposed is not required.

If a future refactor ever wants to reintroduce slash-hierarchy naming, we can revisit at that point. Branch `skill-writer/naming-rule` is preserved on origin.

PR #4979, opened 2026-08-17
Add issue-only `fix` command to the bot workflow

chuanqi129:
Adds an issue-only `fix` command to the bot, and splits the bot's command
surfaces between issues and pull requests.

- `parse-command` now also handles issue comments; routing is enforced in the parser.
- `fix` is issue-only, requires an open issue, and is gated to OWNER/MEMBER.
  `help` works on both surfaces. All other commands are PR-only.
- New always-populated `issue_number` output replaces `pr_number` in the
  comment-posting jobs.
- `help` output is split into a PR table and an issue table.
- New `fix` job runs on `xpu-agent` in the GPU container, then `claude-code-action`
  runs `/issue-handler` in pipeline mode.

### Scope of `fix`

The runner has a prebuilt nightly PyTorch-XPU wheel and no build tree, so
`libtorch_xpu.so` cannot be recompiled there. The prompt and help text state the
boundary: Python-level fixes are verified and pushed to an `agent/<slug>` branch
for a human to open a PR; C++/SYCL findings are reported as a root cause plus a
suggested patch, not as a verified fix. Giving the job a real build path needs a
measured spike and is tracked separately.

`fix` runs `/issue-handler` in pipeline mode, which rewrites the issue body with
status markers and applies stage labels. This is disclosed in the help text.

Draft until verified live:

- [ ] `@torchxpubot help` on this PR shows the PR table
- [ ] `@torchxpubot help` on an open issue shows the issue table
- [ ] `@torchxpubot fix` on this PR is rejected
- [ ] `@torchxpubot merge` on an issue is rejected
- [ ] `@torchxpubot fix` on a closed issue is rejected
- [ ] `@torchxpubot fix` from a non-maintainer is rejected
- [ ] `@torchxpubot fix` on an open issue runs end to end on `xpu-agent`


chuanqi129:
@torchxpubot review

chuanqi129:
Verified each item against the codebase. Response below; no code changes yet.

### Confirmed

- **No build path in the `fix` job.** `references/environment-setup.md:48-49` requires copying changed files into `$PYTORCH_DIR/third_party/torch-xpu-ops/` and running `ninja -C $PYTORCH_DIR/build`. The job has neither a PyTorch source tree nor a build tree, so a C++/SYCL change is never compiled and Stage 5 re-verification runs against the untouched nightly. This is the blocking issue.
- **`/xpu-ops-pr-creation` does not auto-open PRs** (`SKILL.md:116`). The prompt and the help row both promise a PR. Contradiction confirmed.
- **`xpu-agent` label:** `grep -rn xpu-agent .github/` returns zero hits outside `bot.yml`.
- **Missing `chown 1000:1000` cleanup:** 7 workflows do this (`_linux_ut`, `_linux_e2e`, `_linux_reproducer`, `_linux_transformers`, `_linux_accelerate`, `_linux_op_benchmark`, `bisect_search`).
- **Missing `DOCKER_REGISTRY_AUTH_TOKEN`:** same 7 workflows set it.
- **`HF_TOKEN`/`HUGGING_FACE_HUB_TOKEN` unused** by any step in the job.
- **`always()` on the verify step** is absent from both `review` and `ut-check`; it will misattribute setup failures and cancellations to Bedrock.
- **Pipeline mode rewrites the issue body** (`references/execution-modes.md`: status markers, labels, and a comment). Neither the help row nor the PR description discloses this.

### Corrections

- **`linux-testenv` alone does not close the build gap.** With `pytorch: nightly_wheel` it runs `uv pip install torch ... nightly/xpu` (`action.yml:78-79`) and then clones matching PyTorch source — it never builds. It supplies `$PYTORCH_DIR` and the test paths, which is a real improvement over the current hand-rolled setup, but the recompile step still has to be added on top. The recommendation is directionally right, not sufficient.
- **Checkout path convention was missed.** `linux-testenv` reads `${{ github.workspace }}/torch-xpu-ops` (`action.yml:115`), i.e. sibling jobs check out into a `torch-xpu-ops/` subdirectory. The `fix` job checks out to the workspace root, so adopting `linux-testenv` also requires changing the checkout path.
- **Security finding is a policy decision, not a defect.** The untrusted-input-to-`MERGE_TOKEN` path is real and worth deciding on, but it is a deliberate tradeoff for an autonomous fix command, not something that can be labeled incorrect. Noting the `merge -f` / `maintainerOnly` precedent as the relevant prior.

### Not covered by the review

Whether the bot's own comments re-trigger the workflow. Comments are posted with `MERGE_TOKEN` (a PAT), so unlike `GITHUB_TOKEN` they **do** raise `issue_comment` events. All 35 comment bodies in `bot.yml` were enumerated: none start with `@torchxpubot` (rejection strings start with a backtick, failure pings with `@chuanqi129`, AI results with a `> Replying to` header), so no loop today. That safety rests entirely on wording, so an author guard is worth adding:

```yaml
&& github.event.comment.user.login != 'torchxpubot'
```

Staying in draft while the environment and PR-handoff items are decided.

.github/workflows/bot.yml
```diff
@@ -968,6 +1013,166 @@ jobs:
...
+
+      - name: Setup Python
+        uses: ./.github/actions/setup-python
+        with:
+          python: '3.10'
+
+      - name: Install latest nightly PyTorch-XPU
+        run: |
+          # Reproduce/verify environment: latest nightly XPU build.
+          uv pip install --prerelease=allow torch torchvision torchaudio \
```
ZhaoqiongZ, line 1092: is uv preinstalled in the container?

.github/workflows/bot.yml
```diff
@@ -968,6 +1013,166 @@ jobs:
...
+        uses: actions/github-script@v9
+        with:
+          github-token: ${{ secrets.MERGE_TOKEN }}
+          script: |
+            const fs = require('fs');
+            const file = '${{ steps.ai-fix.outputs.execution_file }}';
+            if (!file || !fs.existsSync(file)) return;
+            const data = JSON.parse(fs.readFileSync(file, 'utf8'));
+            const result = data.filter(e => e.type === 'result').pop();
+            if (!result || result.is_error || !result.result) return;
```
ZhaoqiongZ, line 1143: fix is a long task, even failure, it's better to have a partial output noted rather than return with nothing

ZhaoqiongZ (approved):
let's land first to validate the functionality

.github/workflows/bot.yml
chuanqi129: Added partial failure reason comment

.github/workflows/bot.yml
chuanqi129: in setup python step

PR #4822, opened 2026-08-05
[SKILL] Add /label-issue skill for issue auto label

daisyden:
## Summary

Adds a Claude/OpenCode **skill** (`.claude/skills/label-issue/`) that proposes a label set for a single GitHub issue. Given an issue number or URL — and an optional `pytorch_folder` local PyTorch checkout for root-cause tracing — it reads the issue, root-causes the representative failure, and derives labels across every axis, writing a `labels.md` table to disk.

The skill is **analysis-only**: it never edits, comments on, closes, or labels the issue on GitHub, and it never creates sub-issues. Its single artifact is `labels.md`, left for a human or a downstream workflow to apply. It also does **no local test reproduce** — decisions come from the issue text plus (optionally) reading the local PyTorch source.

## How it works

1. **Extract** (`extract-issue/` subskill) — fetches the issue with `gh` and emits `extract.json` (title, body, traceback, per-case `error_message`, `test_cases`, reproduce steps, platform, and the native Type/Priority/project fields). No script, no Python.
2. **Group** the reported failures by cause (`reference/group_issue.md`). A multi-cause issue is flagged with a `need_split` recommendation row; the skill never files sub-issues itself.
3. **Root-cause** the representative case (`reference/triage_issue.md`) — traced mode when `pytorch_folder` is provided, evidence-only mode otherwise.
4. **Label** every axis — type, test, module, priority, dtype, symptom, dependency, os, hw, and the triage duplicate/wontfix/need_split labels.
5. **Output** a 3-column `axis | value | reason` table per `reference/output_format.md`, written under `agent_space/label_issue/`.

## Design principle: labels are data, not code

Every label name, keyword, and evidence rule lives once in `reference/proposed_labels.json`. The skill reads them at decision time and never hard-codes label spellings. The `reference/*.md` packs supply only the judgment the JSON cannot carry (decision-priority ordering, traceback-origin evidence, operator-to-dependency mapping, platform detection, duplicate search, output shape).

## Files

- `SKILL.md` — the orchestrating workflow (inputs, prerequisites, steps, constraints, hard stops).
- `extract-issue/SKILL.md` — the issue-extraction subskill.
- `reference/proposed_labels.json` — authoritative label taxonomy (names, keywords, evidence).
- `reference/*.md` — reasoning packs: `group_issue`, `triage_issue`, `dependency` (+ `dependency_info`), `duplicates`, `platform_specific`, `testcase_rules`, `text_rules`, `output_format`.

## Examples
<details>
[...]

PR #4983, opened 2026-08-17
skills: add issue-triage — shallow text-only issue classification

ZhaoqiongZ:
### Summary

Adds a standalone `issue-triage` skill that classifies a issue based only on its title, body, and labels — no source reading, no test runs. Emits a markdown table as a comment on the issue, applies matching labels, and returns a JSON summary to the caller.

Also wires a `@torchxpubot triage` command in `bot.yml` so the skill can be invoked directly from an issue comment, in the same style as `@torchxpubot review` / `@torchxpubot ut-check`.

This is the first in a planned series of small, focused PRs that split what #4918 tried to do all at once. Each follow-up PR adds one skill or one closely-related set. This PR **only** adds `issue-triage` (and its bot entry point); downstream pieces (reproduce / root-cause / implement / verify / orchestrator) will come in separate PRs.

### What the skill produces

For every issue, three artifacts:

**1. Markdown comment on the issue** (marked `<!-- agent:triage -->`):

```markdown
## Issue Triage

| Field                | Value                                      |
|----------------------|--------------------------------------------|
| Reproduction missing | yes / no                                   |
| Scope                | pytorch / torch-xpu-ops / both / unclear   |
| Dependencies         | triton, onednn / (none)                    |
| Handling             | agent-fixable / needs-human                |

**Reason:** <one-line, only when handling=needs-human>
```

Re-running on the same issue **updates the comment in place** (find by marker, `gh api PATCH` the body).

**2. GitHub labels:**

- `agent:reproduction-needed` when `reproduction_missing = yes`
- `agent:needs-human` when `handling = needs-human`

Scope and dependency values are metadata (JSON only), not labels.

**3. JSON returned to the caller:**

```json
{
  "issue_type": "bug | skip-list | nonbug",
  "reproduction_missing": true | false,
  "scope": "pytorch | torch-xpu-ops | both | unclear",
  "runtime_dependencies": ["triton", "onednn", "onemkl", "driver", "sycl", "xccl"],
  "handling": "agent-fixable | needs-human",
  "reason": "<one-line reason when needs-human, else empty>"
}
```

### The Handling decision

Derived from the other signals, first match wins:

```
if issue_type in ("nonbug", "umbrella task"):
    handling = needs-human   # not a fixable bug
elif reproduction_missing == yes:
    handling = needs-human   # agent has nothing to run
elif len(runtime_dependencies) > 0:
    handling = needs-human   # any external dep requires human triage
elif hardware-only or non-public dependency:
    handling = needs-human   # agent has no access
else:
    handling = agent-fixable
```

`scope=both` and `scope=unclear` do NOT force needs-human — a downstream deep-triage skill (future PR) decides the final target repo after reading source.

### Runtime dependency enumeration

Closed set of values the skill will emit:

| Value | What it means |
|---|---|
| `triton` | Inductor / torch.compile GPU codegen backend |
| `onednn` | Intel oneDNN library (matmul, conv, etc.) |
| `onemkl` | Intel oneMKL library (BLAS, LAPACK, sparse) |
| `driver` | GPU driver, level-zero, compute-runtime, `libze_intel_gpu.so` |
| `sycl` | SYCL runtime / DPC++ compiler |
| `xccl` | XCCL communication library |

Only reported when **explicitly named** in the issue. A stack-trace path through `torch._inductor` alone does not imply `triton` — the issue must say so or show a triton-side error.

### Bot integration (`@torchxpubot triage`)

Adds `triage` as an **issue-only** command wired to this skill.

Usage:

```
User comments on any open issue:  @torchxpubot triage
  → Bot job runs, checks out the repo, loads /issue-triage.
  → Skill fetches the issue, produces the table + labels + JSON.
  → Skill posts a single <!-- agent:triage --> comment on the issue.
  → Repeated invocations update the same comment in place.
```

Parser (`parse-command` job) changes:

- `triage` added to `validCommands`.
- Issue-only rejection: `@torchxpubot triage` on a PR or on a closed issue lands on the existing `invalid-command` path.
- New `issue_number` output — always the issue number (both `issue` and `pull_request` payloads carry it under `issue.number`). Existing `pr_number` semantics are unchanged.

New `triage` job:

- Runs on `ubuntu-latest`, 15-minute timeout.
- Modeled directly on the existing `review` job: same `actions/checkout@v7`, same `claude-code-action@v1` invocation, same verify tail, same cost-reporting to `COST_TRACKING_ISSUE`.
- Differences from `review`: prompts `/issue-triage` instead of `/pr-review`, targets `issue_number` instead of `pr_number`, uses a smaller turn budget (20 vs 60), and does **not** post a separate result comment — `issue-triage` writes its own `<!-- agent:triage -->` comment.

Help table is not updated in this PR — the current help output on `main` is a single PR-only table, and correctly documenting an issue-only command requires the issue/PR help split proposed in #4979. Leaving that to follow up so this diff stays minimal.

### Positioning vs existing skills

| Skill | Depth | Reads source | Overlaps with this PR? |
|---|---|---|---|
| **`issue-triage`** (this PR) | shallow, text-only | no | — |
| `xpu-issues-triaging` (existing) | deep | yes | different depth; no overlap |
| Future `fix/root-cause` (later PR) | deep | yes | different depth; no overlap |

`issue-triage` is the **entry gate** — it decides whether an issue is worth further work. Deep root-cause analysis (existing `xpu-issues-triaging` or a future `fix/root-cause`) runs only after this skill returns `handling = agent-fixable`.

### Repository label

Created the new label in the repo so `add-label` has a real target:

```
agent:reproduction-needed  (F9D0C4)
  Issue needs a reproducer or minimal repro script before agent can proceed
```

The other label the skill uses (`agent:needs-human`) already exists.

### Testing

Docs and workflow YAML only; no build or test surface.

- `lintrunner` clean on both files.
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/bot.yml'))"` passes.
- Live bot verification (`@torchxpubot triage` on an open issue) requires the workflow to land on main first, since the bot always runs from the default branch.

### Related

- Supersedes #4918 (closed). The full refactor is being resplit into focused per-skill PRs so each piece can be reviewed independently. Branch `skills_refactor/zzq` is preserved on origin as reference for the follow-up PRs that will build on this one.
- Supersedes #4975 (closed). The flat, hyphen-only `name:` this PR uses (`issue-triage`, single component, no slash) fits within the existing `skill-writer` spec unchanged.
- Rebases cleanly on top of #4979 (`bot fix` command) once one of the two lands — both add new jobs to `bot.yml` in different regions of the file and both add small parallel changes to the `parse-command` output surface (`issue_number` output). The one that merges second only needs to reconcile the two new job blocks and the two new items in `validCommands`.

### DIFF
--- a/.claude/skills/issue-triage/SKILL.md
+++ b/.claude/skills/issue-triage/SKILL.md
@@ -29,8 +29,15 @@ this skill).
 ## Inputs
 
 - A GitHub issue: URL, number, or raw title+body+labels. If given a
-  number or URL, fetch with `gh issue view --json title,body,labels`.
-- Read-only. This skill never runs tests or edits files.
+  number or URL, fetch it:
+
+  ```bash
+  gh issue view "$N" --repo "$OWNER/$REPO" --json title,body,labels
+  ```
+
+- Read-only with respect to the repository: this skill never reads
+  source, runs tests, or edits files. Its only writes are the issue
+  comment and the labels in Step 6.
 
 ## Shell helpers
 
@@ -89,9 +96,8 @@ Report `yes` when the issue lacks all of:
 
 Report `no` when at least one of the above is present.
 
-Skip-list issues are exempt — they list already-skipped test node ids
-in their body, which counts as "test node id reference" per the second
-bullet.
+Skip-list issues list already-skipped test node ids in their body, so
+they satisfy the second bullet and report `no`.
 
 ## Step 3: Estimate `scope`
 
@@ -160,11 +166,15 @@ human-readable output:
 ```markdown
 | Field | Value |
 |-------|-------|
-| Reproduction missing | <yes | no> |
-| Scope | <pytorch | torch-xpu-ops | both | unclear> |
-| Dependencies | <comma-separated list, or "(none)"> |
-| Handling | <agent-fixable | needs-human> |
+| Reproduction missing | yes / no |
+| Scope | pytorch / torch-xpu-ops / both / unclear |
+| Dependencies | comma-separated list, or (none) |
+| Handling | agent-fixable / needs-human |
 ```
+
+Each `Value` cell above lists the allowed choices; emit exactly one of
+them. Never leave a literal `|` inside a cell — it splits the cell and
+breaks the rendered table.
 
 When `Handling == needs-human`, append a `**Reason:** <one-line>` line
 below the table.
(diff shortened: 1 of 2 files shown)


.claude/skills/issue-triage/SKILL.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Step 2: Detect `reproduction_missing`
+
+Report `yes` when the issue lacks any of:
+
+- A reproducer command (pytest node id, `python -c ...`, shell command).
+- A test node id reference (e.g. `test_foo.py::TestBar::test_baz`).
+- A minimal code snippet that triggers the failure.
+
+Report `no` when at least one of the above is present.
```
CuiYifeng: `lacks any of` -> yes
`at least one of the above is present` -> no
Contradiction?

.claude/skills/issue-triage/SKILL.md
```diff
@@ -0,0 +1,294 @@
...
+`torch._inductor` does not by itself imply `triton` — the issue must
+say so or show a triton-side error).
+
+Empty array `[]` when none are named.
+
+## Step 5: Derive `Handling`
+
+Evaluate in order; first match wins:
+
+1. `issue_type` is `nonbug` or an umbrella tracking task → **needs-human**
```
CuiYifeng: What's the definition of "an umbrella tracking task"?

CuiYifeng (commented):
@chuanqi129 Please help review.

.claude/skills/issue-triage/SKILL.md
ZhaoqiongZ: the skip list issues with multi ut issue tracking, a "parent" issue

.claude/skills/issue-triage/SKILL.md
ZhaoqiongZ: Updated.

.claude/skills/issue-triage/SKILL.md
ZhaoqiongZ: Updated.

chuanqi129:
@copilot+claude-opus-5 review it with /skill-writer

chuanqi129 (commented):
Based on my understanding, this issue triage skill more like issue label skill, there is a PR https://github.com/intel/torch-xpu-ops/pull/4822 for issue label, could you please check whether we can align them as one?

.claude/skills/issue-triage/SKILL.md
```diff
@@ -0,0 +1,307 @@
+---
+name: issue-triage
+description: >
+  Shallow (text-only) triage of a GitHub issue on pytorch or torch-xpu-ops.
+  Classify as bug/skip-list/nonbug, detect whether a reproducer is present,
+  estimate scope (pytorch/torch-xpu-ops/both/unclear), identify runtime
+  dependencies (triton/onednn/onemkl/driver/sycl/xccl), and emit a handling
+  decision (agent-fixable / needs-human) based on those signals. Posts the
+  result to the issue as a comment, applies matching GitHub labels, and
```
chuanqi129: Let's simplify the description here and remove the actively issue comment

chuanqi129 (approved):

CuiYifeng (approved):

CuiYifeng:
@torchxpubot merge

Issue #4336, opened 2026-07-14
Exception: The supported dtypes for stft on device type xpu are incorrect!


PatrykWilczewski:
### 🐛 Describe the bug

1 UT fail on Windows with Exception: The supported dtypes for stft on device type xpu are incorrect!
The test pass on LINUX

**Affected Test Cases**
op_ut,third_party.torch-xpu-ops.test.xpu.test_ops_xpu.TestCommonXPU,test_dtypes_stft_xpu

**Error Message**
 FAILED test_ops_xpu.py::TestCommonXPU::test_dtypes_stft_xpu - Exception: The supported dtypes for stft on device type xpu are incorrect!

### Versions

<details>
<summary>
Click to expand
</summary>
PyTorch version: 2.14.0a0+git1c6fd72
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A
 
OS: Microsoft Windows 11 Pro Insider Preview (10.0.29618 64-bit)
GCC version: Could not collect
Clang version: Could not collect
CMake version: version 4.3.4
Libc version: N/A
 
Python version: 3.12.5 | packaged by conda-forge | (main, Aug  8 2024, 18:24:51) [MSC v.1940 64 bit (AMD64)] (64-bit runtime)
Python platform: Windows-11-10.0.29618-SP0
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
Is XPU available: True
XPU used to build PyTorch: 20260000
Intel GPU driver version:
* 32.0.101.8860 (20260625000000.******+***)
Intel GPU models onboard:
* Intel(R) Arc(TM) B570 Graphics
Intel GPU models detected:
[...]

PR #4985, opened 2026-08-17
Use computed SMEM_SIZE instead of define

jmamzax:
With commit https://github.com/intel/torch-xpu-ops/commit/b0e262470aa56eb7f2e9a174d29a5ca365cfcba3 #define SMEM_SIZE 4096 was introduced for Index_add. But value is too big for MTL integrated GPU and complex128 data type. For MTL GPU shared local memory is 64KB and for discrete GPU it is 128KB. To make it work on MTL GPU SMEM_SIZE should be set to 2048. Because for discrete GPU the 4096 value is correct the computed version of SMEM_SIZE was introduced as it should works correctly on different devices and no performance drop should be observed. Issue: https://github.com/intel/torch-xpu-ops/issues/4336


jmamzax:
Fixed by: https://github.com/intel/torch-xpu-ops/pull/4228

PR #4997, opened 2026-08-18
Add fix-reproduce skill: three-stage bug reproduction verifier

ZhaoqiongZ:
**Merge order:** issue-triage (#4983) → **fix-reproduce (#4997)** → fix-root-cause (#5011) → fix-implement/fix-verify (#5012) → issue-handler (#5014). PR 2/5.

### Summary

Adds `fix-reproduce`, a leaf skill that verifies whether a bug still reproduces before an orchestrator spends time on a fix. Three-stage approach — nightly wheel (fast) → source build at CI commit (precise) → CI-env alignment (last resort) — returning a four-state verdict. Called by orchestrators, not users; no bot command.

### Contract

**Inputs:**
- `reproducer_command` — pytest/python command or test name (provided by the caller).
- `stage` — `auto` (full three-stage fallback, default) or `nightly` (Stage 1 only; PASS returns `NOT_REPRODUCED` immediately, no fall-through).
- `ci_commit` — fallback base when `origin/main` fails to build (ignored when `stage=nightly`).
- `pytorch_dir` — local checkout; cloned if absent (ignored when `stage=nightly`).

**Outputs** — four-state verdict:
- `REPRODUCED` (stage, base, refined_command, env_diff)
- `NOT_REPRODUCED` (checked_stages, reason)
- `NO_REPRODUCER` (no_command / collected_zero)
- `CANNOT_VERIFY` (stage, blocker)

`NO_REPRODUCER` covers both an absent command and `collected 0 items` — the runtime complement to `issue-triage`'s text-only `reproduction_missing` check.

### Stages

1. **Nightly wheel** (always) — install latest XPU nightly; working-dir rules keep `import torch` off the in-tree source checkout; `@skipIfXpu` markers temp-removed before concluding.
2. **Source build at `origin/main`** (skipped when `stage=nightly`) — builds via `xpu-build-pytorch`, falls back once to `ci_commit`, records `base` for the orchestrator to branch off.
3. **CI-env alignment** (skipped when `stage=nightly`) — `ci_env_setup.sh` pulls matching CI wheels + docker image; restores `xpu.txt` + `git clean -fdx` first so a Stage 2 dev-override doesn't leak a stale `torch/_C.so`.

Carries three fixes from #4918 review: SF8 (`rm -rf $WHEELS_DIR` before extract), SF13 (Stage 2→3 pin/clean restore), SF15 (`curl --range 0-0` instead of HEAD — CloudFront 405s on HEAD).

### Not in this PR

- No bot integration.
- `fix/skip-management` slash references kept as-is (renamed in a later PR of the series — renaming now would break the reference).

### Test Plan

Docs/shell only.
```bash
bash -n .claude/skills/fix-reproduce/scripts/ci_env_setup.sh
lintrunner --paths-cmd='git diff --name-only origin/main'
```
Live behavior verifiable once an orchestrator calls this skill (#5014).

### DIFF
--- a/.claude/skills/fix-reproduce/SKILL.md
+++ b/.claude/skills/fix-reproduce/SKILL.md
@@ -1,0 +1,708 @@
+---
+name: fix-reproduce
+description: >
+  Use when asked to reproduce a bug, verify a nightly CI failure, or
+  confirm a failure still exists on latest source. Verifies whether a
+  bug still reproduces before an orchestrator commits time to a fix.
+  Runs a three-stage fallback (nightly wheel -> source build -> CI
+  environment alignment) and returns REPRODUCED / NOT_REPRODUCED /
+  NO_REPRODUCER / CANNOT_VERIFY. Called by both issue-handler and
+  xpu-nightly-ci-fix orchestrators.
+---
+
+# Reproduce — Verify the Bug Exists
+
+Runs a test and determines whether the bug reproduces. Uses a three-stage
+approach: nightly wheel first (fast), source build at CI commit second (precise),
+CI environment alignment third (last resort).
+
+The orchestrator decides what to do with the output — this skill only reports
+the result.
+
+## Inputs
+
+- `reproducer_command` — the sequence of shell commands that triggers the
+  failure. Any of the forms below is valid; Stage 1's "Reproducer forms"
+  section routes execution:
+
+  1. A pytest node id or `pytest ...` invocation:
+     `pytest -v test/xpu/test_ops.py::TestFooXPU::test_bar_xpu_float32`
+  2. A `python -c "..."` snippet or a single `python script.py` line.
+  3. A multi-line shell block (as is common in issue bodies): env setup,
+     `git clone`, `pip install`, followed by the actual failing command.
+
+  Missing or unrunnable inputs return `NO_REPRODUCER` up-front — see
+  "## Preflight" below. Non-`NO_REPRODUCER` verdicts come from the
+  Stage 1/2/3 execution flow.
+
+  Providers (set by the orchestrator, not this skill):
+  - **Issue body** — extracted by `issue-triage` from the reproducer section
+    (via `issue-handler`).
+  - **CI failure log** — the failing pytest node id from the nightly CI report
+    (via `xpu-nightly-ci-fix`).
+- `stage` — which reproduction path to run. Default `auto`.
+  - `auto` — run the full three-stage fallback chain (nightly → source_build →
+    ci_env). Used by orchestrators that need a definitive verdict.
+  - `nightly` — only run Stage 1 (nightly wheel). PASS returns
+    `NOT_REPRODUCED(checked_stages=[nightly])` immediately; do NOT fall through
+    to source build or ci_env. Cheapest option — suitable for a fast "does this
+    still reproduce on latest nightly?" answer.
+- `ci_commit` — upstream commit hash from the CI report. Only used as a
+  fallback base when `origin/main` fails to build (optional; ignored when
+  `stage=nightly`).
+- `pytorch_dir` — path to a local PyTorch checkout (optional). Used
+  whenever Prepare determines `needs_tree=yes` (any pytest form with a
+  repo-relative path) as well as by Stage 2's source build. If absent,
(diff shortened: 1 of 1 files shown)


chuanqi129:
@copilot+claude-opus-4.6 review with /skill-writer

.claude/skills/fix-reproduce/SKILL.md
```diff
@@ -0,0 +1,696 @@
+---
+name: fix-reproduce
+description: >
```
chuanqi129, line 3: Refer the bot review comment, better to add "Use when ..."

.claude/skills/fix-reproduce/SKILL.md
```diff
@@ -0,0 +1,696 @@
...
+
+#### B2. Pick the right `build_env`
+
+pytorch/pytorch's `xpu.yml` currently defines only `py3.10` linux builds:
+
+| build_env                          | Hardware | Image (reference)                                             |
+|------------------------------------|----------|---------------------------------------------------------------|
+| `linux-noble-xpu-n-py3.10`         | PVC      | `ghcr.io/pytorch/ci-image:pytorch-linux-noble-xpu-n-py3-<docker-tree-hash>`|
+| `linux-noble-xpu-n-py3.10-client`  | BMG      | `ghcr.io/pytorch/ci-image:pytorch-linux-noble-xpu-n-py3-client-<docker-tree-hash>`|
+| `linux-jammy-xpu-n-1-py3.10`       | PVC      | `ghcr.io/pytorch/ci-image:pytorch-linux-jammy-xpu-n-1-py3-<docker-tree-hash>`|
```
chuanqi129, line 540: Will this skill create container with those images by itself? The xpu-agent runner already based on docker, not sure docker-in-docker work or not

ZhaoqiongZ: will not create by those containers, the self runner host will load the container you choose and try repro the ci env with installing the specific the artifacts.
remain those context since it is already conclude and may reduce some time when investigating locally, so choose not to drop the container related info.

.claude/skills/fix-reproduce/SKILL.md
ZhaoqiongZ: updated

.claude/skills/fix-reproduce/SKILL.md
```diff
@@ -0,0 +1,708 @@
...
+| `PASSED` | Return `NOT_REPRODUCED(checked_stages=[nightly, source_build, ci_env])` — issue no longer exists; orchestrator reports to user or triage collects reason |
+
+## Output
+
+Return one of these to the orchestrator:
+
+```
+REPRODUCED
+  stage: nightly | source_build | ci_env
+  base: origin/main | <ci_commit_sha>    # base for downstream build (default origin/main; ci_commit_sha only when source_build fell back)
```
CuiYifeng: Referring to `git -C "$pytorch_dir" checkout --detach "$TORCH_COMMIT_ID"`, will base value be `$TORCH_COMMIT_ID` different from origin/main and <ci_commit_sha>?

ZhaoqiongZ: Good catch — `$TORCH_COMMIT_ID` here is the wheel's build commit (`torch.version.git_version`), and it is different from both `origin/main` and `<ci_commit_sha>`. But it is only a **temporary alignment**: Stage 3 checks out the source tree to the wheel commit so the `pytorch/test/` support modules (test base classes, helpers) match the installed wheel's binary and the test can actually run. It is not the fix base and is never returned as `base`.

Downstream fixes always branch off `origin/main` (Stage 2's rule), so `stage=ci_env` reports `base=origin/main` even though the tree is momentarily detached at `TORCH_COMMIT_ID`. `<ci_commit_sha>` only appears as base when `stage=source_build` had to fall back to the CI commit because trunk failed to build.

Clarified this in 33df6985 in three places: a note after the checkout snippet, the Stage 3 REPRODUCED decision row (now spells out `base=origin/main`), and the Output `base` comment.

CuiYifeng (approved):

ZhaoqiongZ:
@torchxpubot merge

chuanqi129 (approved):

Stonepia (approved):
LTGM

CuiYifeng:
@torchxpubot merge

Issue #4765, opened 2026-07-31
[XPU][INT4] _weight_int4pack_mm has large performance gap vs oneDNN W4A16 and incorrect results for M=1


YangKai0616:
## Description
**SGLang** and **vLLM** are preparing to integrate **PyTorch**'s native XPU operator for INT4 dense linear layers, avoiding a direct framework-level dependency on the custom oneDNN operator.

But `aten._weight_int4pack_mm` on XPU shows a significant performance gap compared with `vllm-xpu-kernels`' oneDNN W4A16 operator (`torch.ops._xpu_C.int4_gemm_w4a16`). It also produces incorrect results when `M=1`.

### Performance

| M | K | N | torch: `_weight_int4pack_mm` (us) | vllm: `int4_gemm_w4a16` (us) | Speedup |
|---:|---:|---:|---:|---:|---:|
| 8 | 4096 | 6144 | 1189.2 | 76.2 | 15.60x |
| 256 | 4096 | 6144 | 1412.1 | 205.1 | 6.89x |
| 8 | 4096 | 4096 | 571.9 | 76.4 | 7.49x |
| 256 | 4096 | 4096 | 605.2 | 154.0 | 3.93x |
| 8 | 4096 | 28672 | 4292.5 | 184.6 | 23.25x |
| 256 | 4096 | 28672 | 4463.7 | 736.1 | 6.06x |
| 8 | 14336 | 4096 | 2155.1 | 114.2 | 18.87x |
| 256 | 14336 | 4096 | 2259.2 | 402.2 | 5.62x |

The current `M > 1` implementation appears to dequantize the complete INT4 weight to BF16 before calling `matmul`, while the oneDNN implementation uses a fused W4A16 path.

### M=1 correctness issue

For every tested shape, `aten._weight_int4pack_mm` produced incorrect output when `M=1`, with approximately 97-98% of elements failing the configured tolerance. The oneDNN W4A16 output passed for the same inputs.

| M | K | N | torch op mismatch rate | vllm op mismatch rate |
|---:|---:|---:|---:|---:|
| 1 | 4096 | 6144 | 96.9727% | 0.0000% |
| 1 | 4096 | 4096 | 97.4854% | 0.0000% |
| 1 | 4096 | 28672 | 97.2726% | 0.0000% |
| 1 | 14336 | 4096 | 98.1201% | 0.0732% |

The benchmark script used to reproduce these results is attached below:
```python
#!/usr/bin/env python3
import argparse
[...]

PR #5001, opened 2026-08-18
[int4_mm] Fix `_weight_int4pack_mm_xpu`

hoshibara:
## Problem

On XPU, `_weight_int4pack_mm(A, B, qGroupSize, qScaleAndZeros)` returns wrong results when M == 1 and qGroupSize > 32: ~96-98% element mismatch vs the dequantized-weight reference (intel/torch-xpu-ops#4765). M > 1 and qGroupSize == 32 are unaffected.

## Root cause

For M == 1, `linear_int4_kernel` assigns a 32-element K-slice per subgroup (TileK = 32, constexpr in `LinearInt4KernelFunctor`). The scale/zero-point slot must be the quantization-group index of the subgroup's starting K element, i.e. `(sg_id * TileK) / qGroupSize`.

The old code computed `sg_id * (TileK / qGroupSize)`: the parenthesized division was evaluated first and truncated to 0 whenever qGroupSize > TileK, so every subgroup read quantization-group 0's scale/zero-point and the accumulated products were wrong. qGroupSize == 32 coincidentally produced the correct value, which is why failures were specific to M == 1 with group sizes 64/128/256.

## Fix

Reorder the integer division so the global K offset is divided by the group size:

    int scale_offset = (sg_id * TileK / blocksize) * ld_scale_zp;


src/ATen/native/xpu/sycl/LinearInt4.cpp
```diff
@@ -77,8 +77,8 @@ struct LinearInt4KernelFunctor : public __SYCL_KER_CONFIG_CONVENTION__ {
         for (int iu = 0; iu < Unroll; iu++) {
           const uint8_t* tmps8 =
               reinterpret_cast<const uint8_t*>(bptr + sg_id * TileK / 2);
-          int scale_offset = sg_id * (TileK / blocksize) * ld_scale_zp;
-          int zp_offset = sg_id * (TileK / blocksize) * ld_scale_zp;
+          int scale_offset = (sg_id * TileK / blocksize) * ld_scale_zp;
+          int zp_offset = (sg_id * TileK / blocksize) * ld_scale_zp;
```
CuiYifeng, line 81: Is the root cause that `TileK` may not be an integer multiple of `blocksize`? Dose CUDA have this problem?

CuiYifeng (commented):
Please update PR description to clarify what's the problem, how to fix and why, thanks.

src/ATen/native/xpu/sycl/LinearInt4.cpp
hoshibara: No. In `aten/src/ATen/native/cuda/int4mm.cu`, the group index is always derived from the global K offset first:

    int32_t groupStart = (kTileStart * kKTileSize) / QGroupSize;
    ...
    int curQGroup = (curKTile * kKTileSize) / QGroupSize;

hoshibara:
> Please update PR description to clarify what's the problem, how to fix and why, thanks.

Updated

CuiYifeng (approved):

xiaowangintel (approved):

CuiYifeng:
@torchxpubot ut-check

CuiYifeng:
@torchxpubot ut-check

CuiYifeng:
@torchxpubot merge -f

Issue #4000, opened 2026-06-15
Add XPU GPU IPC support for tensor multiprocessing


xuhancn:
## Problem

PyTorch's tensor multiprocessing (Queue-based sharing) relies on GPU IPC for cross-process mutation visibility. CUDA provides this via `cudaIpcMemHandle` — both processes map the same GPU memory allocation, so mutations in one process are immediately visible in the other.

XPU currently has no GPU IPC mechanism, which blocks proper tensor multiprocessing support.

## Impact

In pytorch/pytorch PR #186793, XPU tensor multiprocessing was implemented using CPU copy fallback. This correctly preserves values but cannot support cross-process mutation visibility — 3 device-generic tests must be skipped:

- `test_simple_sharing`: child's `fill_(4)` not visible to parent
- `test_variable_sharing`: parent's post-send mutation not visible to child
- `test_parameter_sharing`: same pattern

## What's needed

1. XPU-level IPC mechanism (equivalent to `cudaIpcMemHandle` + `cudaIpcGetMemHandle`/`cudaIpcOpenMemHandle`)
2. `XPUCachingAllocator::shareIpcHandle()` method
3. C++ IPC types for XPU (ref-counting, event sync) — see `CUDACachingAllocator.cpp` and `CudaIPCTypes.h` for reference
4. `_share_xpu_()` Python binding exposing the IPC handle

## References

- CUDA IPC path: `torch/multiprocessing/reductions.py:341-375` (`reduce_tensor` CUDA branch)
- CUDA IPC rebuild: `torch/multiprocessing/reductions.py:153-218` (`rebuild_cuda_tensor`)
- PyTorch PR #186793: https://github.com/pytorch/pytorch/pull/186793
- PyTorch issue #186585: https://github.com/pytorch/pytorch/issues/186585

PR #5008, opened 2026-08-18
Skip TestMultiprocessingDeviceType tests that fail on XPU

jmamzax:
Skip test for Windows as IPC is yet not supported. On Linux, case with empty tensor works correctly so only Windows skip is needed. This change is in place of https://github.com/pytorch/pytorch/pull/189020 according to comment. This can be removed once https://github.com/pytorch/pytorch/pull/191725 is merged. Issue: https://github.com/intel/torch-xpu-ops/issues/4068

### DIFF
--- a/test/xpu/test_multiprocessing_xpu.py
+++ b/test/xpu/test_multiprocessing_xpu.py
@@ -82,13 +82,6 @@ if __name__ == "__main__":
     self.assertRegex(stderr, "Cannot re-initialize XPU in forked subprocess.")
 
 
-@unittest.skipIf(IS_WINDOWS, "XPU IPC tensor sharing is not supported on Windows")
-@unittest.skipIf(not torch.xpu.is_available(), "XPU not available")
-def _test_empty_tensor_sharing_xpu(self):
-    self._test_empty_tensor_sharing(torch.float32, torch.device("xpu"))
-    self._test_empty_tensor_sharing(torch.int64, torch.device("xpu"))
-
-
 @unittest.skipIf(
     platform == "darwin", "file descriptor strategy is not supported on macOS"
 )
@@ -100,7 +93,6 @@ def _test_is_shared_xpu(self):
 
 TestMultiprocessing.test_cuda_bad_call = _test_cuda_bad_call
 TestMultiprocessing.test_wrong_cuda_fork = _test_wrong_cuda_fork
-TestMultiprocessing.test_empty_tensor_sharing_cuda = _test_empty_tensor_sharing_xpu
 TestMultiprocessing.test_is_shared_cuda = _test_is_shared_xpu
 
 
--- a/test/xpu/test_multiprocessing_xpu.py
+++ b/test/xpu/test_multiprocessing_xpu.py
@@ -17,7 +17,10 @@ import unittest
 
 import torch
 import torch.multiprocessing as mp
-from torch.testing._internal.common_device_type import instantiate_device_type_tests
+from torch.testing._internal.common_device_type import (
+    instantiate_device_type_tests,
+    skipXPUIf,
+)
 from torch.testing._internal.common_utils import IS_WINDOWS, run_tests, TestCase
 
 try:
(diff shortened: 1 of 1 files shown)


test/xpu/test_multiprocessing_xpu.py
BBBela: Those two test functions no longer exist in upstream.
They have been renamed and moved to `TestMultiprocessingDeviceType` test class, which is instantiated with `instantiate_device_type_tests`, but were skipped recently due to lack of IPC XPU support in [THIS COMMIT](https://github.com/pytorch/pytorch/commit/4f26a5718ee210e63d35ba59df203c263bfc547a)
So basically there is no more failing test anymore.
We should remove the `_test_empty_tensor_sharing_xpu` and `_test_is_shared_xpu` and instead import `TestMultiprocessingDeviceType` and instantiate it with `instantiate_device_type_tests(TestMultiprocessingDeviceType, globals(), allow_xpu=True)` - but it will fail until the IPC support is added.

jmamzax: Test in torch-xpu-ops repo still runs and for Windows it FAIL. But as you mention upstream was changed so I remove _test_empty_tensor_sharing_xpu test and once IPC will be done this should be reworked.

BBBela: I think some cleaning of this test file should be performed as it diverged from the upstream. Could you please do the following?:
- Remove the `_test_is_shared_xpu` as there is no more `TestMultiprocessing.test_is_shared_cuda` in the upstream.
- Import `TestMultiprocessingDeviceType` using `with XPUPatchForImport`.
- Instantiate those tests using `instantiate_device_type_tests(TestMultiprocessingDeviceType, globals(), allow_xpu=True)`
- Create new issue or update the current one with failing test cases from `TestMultiprocessingDeviceType`.
- Wait until IPC support is added and then close the issue.

I think the issue can stay open with description or comment that it waits for IPC implementation.

What do you think?

jmamzax: It can be done like that. I will update PR.

jmamzax: @BBBela Hi, I updated PR, but also skipped all tests form TestMultiprocessingDeviceType that fail. With such solution this can be merged and use then for IPC tests if needed.

BBBela (commented):
Once we started using `instantiate_device_type_tests` to instantiate `TestMultiprocessingDeviceType`, we should use `skipXPU` or `skipXPUIf` decorators instead of `unittest.skip()`.

I think there should be also created new issue with those failing test cases, not to loose this problem and enable those tests once storage sharing is implemented for XPU - unless there is already such issue.

jmamzax:
PR, updated. Skipped tests:
test_empty_tensor_sharing handled in https://github.com/intel/torch-xpu-ops/issues/4068
test_simple_sharing, test_variable_sharing, test_parameter_sharing mention here: https://github.com/intel/torch-xpu-ops/issues/4000
test_leaf_variable_sharing, test_integer_parameter_serialization: new issue https://github.com/intel/torch-xpu-ops/issues/5219

BBBela (dismissed):
Thanks!

pbielak:
@jmamzax if there are already issues present, we should link them directly in the `skipXPUIf` decorator, e.g., 

```py
TestMultiprocessingDeviceType.test_empty_tensor_sharing = skipXPUIf(
    IS_WINDOWS, "https://github.com/intel/torch-xpu-ops/issues/4068"
)(TestMultiprocessingDeviceType.test_empty_tensor_sharing)
```

Please update the PR

jmamzax:
Updated.

BBBela:
The upstream test file has been changed in my recent commit: https://github.com/pytorch/pytorch/commit/3c964308088e0c2535609b70522e13624be85de7

I added skips for failing tests directly in the upstream, so likely we don't need those additional skips introduced by this PR. Maybe we could just launch the upstream test file directly now? By removing the local copy and changing entry in `skip_list_common.py` from:
```python
    "test_multiprocessing_xpu.py": None,
```
to 
```python
    f"{PYTORCH_TEST_DIR}/test_multiprocessing.py": None,
```
But it needs to be double checked if the test coverage isn't lost.

I will work on aligning local torch-xpu-ops tests after upstream generalization, so if you want, you can reassign the issue to me, but you can also continue the work - up to you 😉

PR #5022, opened 2026-08-20
Fix safe-softmax zeroing rows that contain NaN

cyyever:
The fused kernels infer "fully masked row" from the sentinel `max_value == lowest()`. The group max reduce can drop a NaN, so a masked row still holding one (`NaN + -inf = NaN`) hit that sentinel and was zeroed, where upstream keeps it NaN. Gate the early-out on `sum_value` too: a genuinely masked row sums to 0, never NaN.

Scope: the two `inner_size == 1` kernels (`dim=-1`, the path SDPA uses). `SpatialSoftmaxForwardKernelFunctor` carries the same sentinel and is **not** fixed here -- it seeds its max from an element rather than `lowest()`, so a legitimately masked column already produces a NaN sum and this guard cannot be reused there.

Test: test/regressions/test_safe_softmax.py -- added `test_sm_nan_in_masked_row`, whose two dim sizes hit both kernels in scope.

### DIFF
--- a/src/ATen/native/xpu/sycl/GroupReduceUtils.h
+++ b/src/ATen/native/xpu/sycl/GroupReduceUtils.h
@@ -82,6 +82,8 @@ inline T& GroupReduceSumWithoutBroadcast(
   return val;
 }
 
+// Keep max_impl below: sycl::maximum under reduce_over_group can drop a NaN,
+// which torch.max/amax must not.
 template <typename T, int SIMD, int DIM>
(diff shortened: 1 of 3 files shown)


jianyizh (approved):

guangyey (approved):

guangyey:
@torchxpubot ut-check

guangyey:
@torchxpubot merge -f "unrelated failure"

PR #5026, opened 2026-08-20
Use at::empty for batch_norm backward-reduce buffers

cyyever:
`batch_norm_backward_reduce_channels_last_template` allocated `grad_weight` and `grad_bias` with `at::zeros`, but the reduce kernel assigns every channel before anything reads them, so the zero-init is a wasted memset. Upstream CUDA allocates these with `at::empty`; the counter `semaphores` stay zeroed.

With both arms allocating the same way, the `if (weight.defined())` block collapses to a size and an options pick.

Test: test/regressions/test_batch_norm.py, test/xpu/test_nn_xpu.py -- existing batch-norm coverage; no behavior change intended.


guangyey (approved):
Thanks!

pbielak (approved):

guangyey:
@torchxpubot ut-check

guangyey:
@torchxpubot merge -f "unrelated failure"

Issue #5018, opened 2026-08-19
ALIGNMENT_TRIAGE_ISSUE


laifenxiawucha:
## XPU alignment triage

The daily XPU alignment workflow posts audit summaries and independently reviewed candidates here.

Every run for which the deterministic gate produces a decision posts one Run Summary, separate from any candidate draft:

- `XPU alignment run complete`: collection and review completed without terminal warnings.
- `XPU alignment run completed with warnings`: one or more units were isolated, while the remaining reviewed results are still valid and the workflow stays green.
- `XPU alignment run completed with partial collection`: fetching stopped before every source completed; source progress and the terminal error are reported, trusted candidates may still be published or queued, and the workflow finishes red.
- `XPU alignment run failed`: a malformed artifact, coverage gap, required producer failure, or other global blocker prevents all candidate publication and the workflow finishes red.

A Run Summary is a snapshot of the completed workflow attempt; each candidate draft comment carries the current filing state.

Only terminal conditions that affect the final outcome are shown in the Run Summary. Recovered retries, transient API failures, and internal tool errors remain in workflow logs and artifacts. A scheduled retry updates the existing Summary for the same run ID; duplicate scheduled Summary markers fail closed. Dry runs may post repeated summaries. If the publisher itself or the GitHub Issues API fails, the Actions result and gate artifact are the fallback because a comment cannot be guaranteed.

Formal scheduled drafts may be approved one at a time with:

```text
@torchxpubot file <unit-id>
```

- Run the command on this issue only.
- Only repository `OWNER` or `MEMBER` accounts may approve filing.
- A formal reviewed draft from either a complete scheduled scan or a structurally valid partial scheduled scan can be filed.
- A scheduled run with exactly one reviewed payload may file it automatically, including when collection is structurally valid but partial.
- Partial runs report source progress, last cursor, rate-limit reset, and terminal errors, notify maintainers, and finish red even when trusted candidates were published or queued.
- `[DRY RUN]` drafts are test output, may repeat, never notify, and cannot be filed.
- Malformed artifacts, incomplete stage coverage, core environment failure, or a failed producer job produce only a failed Run Summary and no formal draft.
- A runner-backed blocker for one unit excludes that unit without invalidating other fully reviewed candidates.
- Historical diagnostic comments remain historical evidence and cannot be filed because they do not carry a formal unit marker.
- The bot copies the reviewed formal draft into a new `[xpu-alignment]` issue with exactly the `ai_generated` label.
[...]

PR #5003, opened 2026-08-18
[Workflow] Add the daily XPU alignment scan, review and filing pipeline

laifenxiawucha:
## What this does

Adds a daily XPU alignment pipeline with deterministic collection, credential-free XPU execution, independent review, deterministic gating, and controlled GitHub publishing.

```text
02:00 UTC collector
  -> scan-prepare agent
  -> credential-free XPU runner
  -> scan-finalize agent
  -> independent review agent
  -> deterministic gate
  -> publisher
```

Scheduled runs scan the previous complete UTC day and may publish. Manual `workflow_dispatch` is always a dry run and exposes only an optional `scan_date`, defaulting to yesterday UTC.

## Publishing policy

| Result | Scheduled run | Dry run | Workflow |
|---|---|---|---|
| complete, 0 payloads | audit summary, no mention | `[DRY RUN]` summary | green |
| complete, 1 payload | create one formal issue and notify | draft and summary only | green |
| complete, 2+ payloads | formal fileable drafts and notify | drafts and summary only | green |
| structurally valid partial | use the same 0/1/2+ policy, plus incomplete progress and notification | normal dry-run drafts plus incomplete progress, no notification | red |
| malformed evidence or global producer/coverage failure | blocker summary only and notify | dry-run blocker summary only | red |

A single runner-backed unit failure excludes only that unit. Other fully covered, independently reviewed payloads remain publishable; the scheduled summary warns maintainers, while a unit-only warning does not fail the workflow.

Scheduled formal markers are stable and idempotent. Dry-run comments may repeat. `@torchxpubot file <unit-id>` remains OWNER/MEMBER-only on #5018 and accepts formal reviewed scheduled drafts, including drafts from a structurally valid partial collection. It never accepts dry-run or historical diagnostic comments and never runs an agent after approval.

## Runner and evidence boundary

- Both XPU jobs are fixed to `xpu-agent` and `intelgpu/ubuntu-26.04-rolling:26.18`.
- The runner executes frozen scripts serially as `xpu-repro`, with no GitHub, Bedrock, cloud, or publishing credentials and no outbound network.
- The runner owns the XPU environment probe. Core PyTorch/XPU availability failures block the run; optional path or device-name failures are recorded as warnings.
- Every unit receives a private writable HOME/TMPDIR/compile-debug scratch directory while collection, prepare, and scripts remain immutable.
- Both XPU jobs restore runner ownership for the workspace and shared Actions mounts without deleting shared caches.
[...]

PR #5028, opened 2026-08-20
[Workflow] Add the daily XPU alignment scan, review and filing pipeline

laifenxiawucha:
## What this does

Runs the `xpu-alignment` skill against `pytorch/pytorch` once a day, has a second
agent review the result, and then lets a CPU-only gate decide what reaches GitHub.

```
02:00 UTC ──▶ scan (pvc) ──▶ independent review (pvc) ──▶ gate (ubuntu-latest)
                                                             │
                    ┌────────────────────────────────────────┤
                    ▼                    ▼                   ▼
              0 actionable        exactly 1 actionable   2+ actionable
              nothing to do       file it automatically  post drafts to the
                                                         triage issue for a
                                                         maintainer to approve
```

Approval happens with `@torchxpubot file <unit-id>` on the standing triage issue
(#5018).

## Why one automatic issue per day

This is a **risk budget**, not a claim that two candidates would be suspicious.
Two or more is an ordinary busy day — when this process was run by hand it
produced 4 issues on 2026-07-30 and 1 each on 07-09 and 07-29 — so the busy path
keeps the run **green** and simply asks a human to work through the list. Only a
blocked or unfinished run turns the job red.

Capping the unattended path at one per day bounds how much a bad scan can put in
the repository before a person notices, without blocking the days that only need
one issue.

## How the gate decides

The gate reads `artifacts/candidate_ledger.jsonl` — the ledger the skill already
maintains — instead of trusting a status the reviewer reports about itself.

- **Day finished?** No row with `title_status == pass` and `deep_status != reject`
  and `local_status == pending`. This is the skill's own audit rule (SKILL.md
  §Step 3.1).
- **Review complete?** Every finished actionable row carries a reviewer verdict.
  A review that runs out of time now fails hard instead of silently reporting
  fewer candidates than the scan found.

| decision | condition | result |
|---|---|---|
| `none` | nothing actionable | nothing written, green |
| `file-one` | day finished, exactly 1 actionable, auto-file on | 1 issue filed + audit line on #5018, green |
| `triage` | 2+ actionable, or day unfinished | one draft comment per candidate on #5018, green (red if unfinished) |
| `blocked` | evidence does not add up | nothing written, red |

An unfinished day still publishes its finished candidates for human triage — that
GPU time is not thrown away — but the unattended path is switched off, because it
assumes a complete picture of the day.

## No agent runs after the gate

Issue bodies are generated during the scan and review phase as
`reports/final_issue_<unit-id>.json`. Both publishing paths copy a payload that
already existed when the decision was made.

This is deliberate. In an earlier revision the gate wrote its decision to a file
and *then* an agent with `Write` access ran in the same workspace before the
filing step re-read that file. A reproduction showed the agent could change which
candidate was approved and pass every downstream check with exit code 0. Removing
the step removes the window rather than guarding it.

The same property holds for the human path: `@torchxpubot file` copies the draft
comment's bytes into the new issue, so what a maintainer reads is exactly what
gets published.

## Deliberate deviation from the skill

`.claude/skills/xpu-alignment/SKILL.md` rule #2 says:

> Never file issues automatically. Write local drafts, complete independent
> review, then ask the user before filing a review-approved draft on GitHub.

The scan agent still obeys the first sentence — it is told to stop before Step 4
and never writes to GitHub. **This workflow deliberately overrides the second
sentence on one path only**: a day with exactly one reviewed `needs-xpu-fix`
candidate is filed without asking.

Compensating controls for that path:

- a separate reviewer agent must return a verdict for it
- at most one such issue per day
- the day must be complete; an unfinished scan disables the path
- the issue is public, labelled `ai_generated`, and recorded on #5018
- two or more candidates always go to a human

Every other path still asks a person.

## Security

- Both agents run with the read-only `github.token`.
- `MERGE_TOKEN` appears only in the `env` of the two steps that write, never
  inside an agent's tool scope.
- Titles must start with `[xpu-alignment]` and labels must be exactly
  `["ai_generated"]`; a payload that fails either check is refused.
- Unit ids are restricted to `[A-Za-z0-9][A-Za-z0-9._-]{0,63}` because they reach
  shell arguments, glob fragments and comment markers.
- Duplicate detection uses comment markers, not a search query built from an
  LLM-authored title.
- The whole workflow shares one concurrency group: a scheduled run cannot know its
  scan date before a step runs, so grouping by date would not actually serialise
  same-day runs.

## Files

| file | lines | |
|---|---|---|
| `.github/workflows/xpu_alignment.yml` | 322 | scan → review → gate |
| `.github/scripts/xpu_alignment_gate.py` | 260 | ledger + coverage → decision |
| `.github/scripts/alignment_triage.py` | 150 | triage issue comment protocol |
| `.github/scripts/xpu_alignment_publish.py` | 78 | automatic path |
| `.github/scripts/bot_alignment_file.py` | 58 | human path |
| `.github/workflows/bot.yml` | +86 | `file` command |
| tests | 474 | 37 cases |

## Testing

- 37 unit tests across the gate and the comment protocol.
- Structural checks on both workflows: YAML parse, job dependency graph, `bash -n`
  on all 15 `run` blocks, embedded Python compiles, no `secrets.` interpolated
  inside a `run` block.
- End-to-end smoke run against a stub `gh` with a realistic artifact layout,
  covering: automatic filing and its idempotency on re-run, triage plus human
  approval, rejection of a duplicate/unknown/wrong-issue approval, an unfinished
  scan disabling the unattended path, a coverage gap blocking with zero writes,
  and `auto_file=false`.

## Before merging

- #5018 is currently empty. It should carry a short usage note
  (`@torchxpubot file <unit-id>`, OWNER/MEMBER only) so the first person to see a
  draft there knows what to do.
- Worth one `workflow_dispatch` run with `auto_file=false` to confirm both agents
  produce the expected files on real hardware.

## Known limits

- The ledger is written by the scan agent. Checking the reviewer against it
  catches a review that stopped early; it cannot catch a candidate the scan
  dropped before writing the ledger. There is no independent enumeration of
  upstream activity to compare against.
- Whether the pre-generated issue bodies are good enough to file unattended has
  not been observed on real data yet.

## Relationship to #5003

#5003 stays open and untouched. This branch is a rewrite from a different set of
requirements, so it starts fresh from `main` rather than continuing that history.

### DIFF
--- a/.github/scripts/alignment_triage.py
+++ b/.github/scripts/alignment_triage.py
@@ -19,7 +19,9 @@ import sys
 import tempfile
 
 UNIT_MARKER = "<!-- alignment-unit: {unit_id} -->"
-RUN_MARKER = "<!-- alignment-run: {run_id} {scan_date} -->"
+# Provenance is visible text, not an HTML comment: a triager reading a draft
+# needs the run that produced it in order to re-read the underlying evidence.
+PROVENANCE_LINE = "<sub>alignment scan `{scan_date}`, run `{run_id}`</sub>"
 FILED_MARKER = "<!-- alignment-unit-filed: #{number} -->"
 FILED_MARKER_RE = re.compile(r"<!-- alignment-unit-filed: #(\d+) -->")
 TITLE_LINE_RE = re.compile(r"^### (.+)$", re.MULTILINE)
@@ -61,7 +63,7 @@ def list_comments(repo: str, issue: int) -> list[dict]:
 def render_draft(unit_id: str, title: str, body: str, run_id: str, scan_date: str) -> str:
     return (
         f"{UNIT_MARKER.format(unit_id=unit_id)}\n"
-        f"{RUN_MARKER.format(run_id=run_id, scan_date=scan_date)}\n"
+        f"{PROVENANCE_LINE.format(run_id=run_id, scan_date=scan_date)}\n"
         f"### {title}\n\n{body}\n"
     )
 
@@ -71,8 +73,8 @@ def render_filed_note(unit_id: str, issue_url: str, run_id: str, scan_date: str)
     return (
         f"{UNIT_MARKER.format(unit_id=unit_id)}\n"
         f"{FILED_MARKER.format(number=issue_number(issue_url))}\n"
-        f"{RUN_MARKER.format(run_id=run_id, scan_date=scan_date)}\n"
-        f"`{scan_date}` automatically filed `{unit_id}` as {issue_url}\n"
+        f"`{scan_date}` automatically filed `{unit_id}` as {issue_url} "
+        f"in run `{run_id}`\n"
     )
 
 
--- a/.github/scripts/alignment_triage.py
+++ b/.github/scripts/alignment_triage.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python3
+# Copyright 2026 Intel Corporation
+# Licensed under the Apache License, Version 2.0
+
+"""The comment protocol of the standing XPU alignment triage issue.
+
+Drafts live as bot comments carrying a unit marker. Publishing a draft copies
+its bytes verbatim into a new issue, so a human reads exactly what gets filed
+and no agent runs between approval and filing.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import subprocess
+import sys
+import tempfile
+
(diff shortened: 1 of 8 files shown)


laifenxiawucha:
Superseded by #5003. The complete implementation and subsequent review fixes have been rebased onto the latest main and moved there.

PR #5033, opened 2026-08-20
Use std::min/std::max consistently in device code

cyyever:
Device code mixed `std::` and `sycl::` min/max. This converges on `std::` across the three files that still used the `sycl::` spelling.

`sycl::min`/`sycl::max` cannot become the single convention: they are constrained to SYCL generic types and reject `c10::Half`/`c10::BFloat16`, and on floating point they follow `fmin`/`fmax`, which discards NaN. `std::` already covers most of the tree and works for every type, so one spelling removes the risk of converting a float site by analogy and silently changing its NaN contract.

All 88 converted sites are `int`/`int64_t`/`index_t` index and bound arithmetic, where the two are equivalent. The one remaining `sycl::max` is in SYCL-TLA flash attention, on a float accumulator where `fmax` semantics is intended.

### DIFF
--- a/src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
+++ b/src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
@@ -39,14 +39,14 @@ struct AvgPool2dKernelFunctor {
       int hstart = ph * stride_h_ - pad_h_;
       int wstart = pw * stride_w_ - pad_w_;
       int hend =
-          sycl::min(hstart + kernel_h_, static_cast<int>(height_ + pad_h_));
+          std::min(hstart + kernel_h_, static_cast<int>(height_ + pad_h_));
       int wend =
-          sycl::min(wstart + kernel_w_, static_cast<int>(width_ + pad_w_));
+          std::min(wstart + kernel_w_, static_cast<int>(width_ + pad_w_));
       const int pool_size = (hend - hstart) * (wend - wstart);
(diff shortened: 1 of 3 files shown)


guangyey (commented):
Thanks, this PR is valuable. Compiler team commits that std::xxx and sycl::xxx will be equivalent from 2026.3.
@jianyizh what do you prefer?
1. land this PR for now
2. land this PR once upgrade to 2026.3

jianyizh:
Let's wait for 2026.3

guangyey:
Make sense.

dvrogozh:
> Let's wait for 2026.3

When I will mark PR as a Draft for now.

dvrogozh:
Generally speaking SYCL device code does not support arbitrary functions declared in C++ or C standard libraries. Overall supported are:
1. Math functions declared in `sycl::` and `sycl::native` namespaces (per spec)
2. Math functions from C or C++ standard library implemented thru the SYCL extensions.

I think at the moment there is no broad support of `std::` primitives. If so, I have a concern aligning with `std::` as it will be extremely difficult to track which `std::` primitives can and which can not be used. For math this seems to be covered by https://github.com/intel/llvm/blob/v7.1.0/sycl/doc/extensions/supported/C-CXX-StandardLibrary.rst. But what about `#include <algorithm>` primitives which we are using such as `std::min` or `std::plus`?

@gmlueck, can you, please, post recommendation here?

PR #5036, opened 2026-08-20
Merge Bitfield specializations and make the field width a template parameter

cyyever:
`Bitfield<uint64_t>` computed its mask with `(1u << len) - 1u`, a 32-bit shift copied from the `unsigned int` sibling, so `len >= 32` was undefined behavior. Writing it as `(T{1} << len) - 1` makes the two specializations identical, so one template replaces both. `len` then moves to a template parameter -- it is a compile-time constant at every call site -- so an over-wide field becomes a compile error rather than UB. `pos` stays a runtime value.

The `pos`/`len` masking is dropped with the specializations: it mirrored PTX `bfe`/`bfi` operand truncation, which is also the only reason CUDA splits by width. What XPU kept was upstream's host fallback.

The defect is unreachable today (`len` is 2 or 4), so no regression test is possible. Checked out-of-tree that the template is bit-identical to both old specializations across the reachable domain, and that the width and signedness guards reject bad instantiations.


guangyey (approved):
Thanks!

jianyizh:
@torchxpubot review

PR #5059, opened 2026-08-21
Add profiler use case smoke tests for XPU

kkarwot:
Adds use-case tests for the four XPU profiler workflows in the Profiler User Guide - eager, torch.compile, XPUGraph capture/replay, and hardware-metric collection - each profiling a small GEMM and asserting on the exported
trace.


aostrowski-hbn (approved):

test/xpu/skip_list_common.py
```diff
@@ -343,6 +343,7 @@
     "profiler/test_cpp_thread_xpu.py": None,
     "profiler/test_execution_trace_xpu.py": None,
     "profiler/test_profiler_xpu.py": None,
+    "profiler/test_xpu_profiler_use_cases.py": None,
```
aostrowski-hbn: This key does not match the file added in this PR: the test file is `profiler/test_profiler_use_cases.py`, but the skip-list entry says `profiler/test_xpu_profiler_use_cases.py`.

`run_test_with_skip.py` iterates the keys of `skip_dict` and calls `launch_test(key)` -> `pytest <key>`, so a key pointing at a non-existent path makes pytest exit 4 and the `op_ut` job fail.

```suggestion
    "profiler/test_profiler_use_cases.py": None,
```

Side note: `xpu_profiling` is not part of PreCI (the matrix is `op_ut, basic`), so the line added to `linux-uttest/action.yml` only runs in nightly/on-demand. With the key fixed, `op_ut` would cover these tests in PreCI too.

aostrowski-hbn (changes requested):
One blocker before this lands: the skip-list key does not match the new test file, which will fail the `op_ut` job. Details inline.

test/xpu/skip_list_common.py
kkarwot: Good catch! Thanks, updated

aostrowski-hbn (approved):

.github/actions/linux-uttest/action.yml
```diff
@@ -118,6 +118,8 @@ runs:
         # Overlapping kernels test (depends on rn50.log.json from RN50 run above)
         python -m pytest -s test/profiling/test_for_overlapping_kernels.py | \
           tee ${{ github.workspace }}/ut_log/xpu_profiling/issue_reproduce/test_for_overlapping_kernels.log
+        python -m pytest -s test/xpu/profiler/test_profiler_use_cases.py | \
```
RUIJIEZHONG66166: Please add xml generated which can refer to the changes in https://github.com/intel/torch-xpu-ops/pull/4892/changes#diff-6c0469c6e8fb3c4eff76c5ff0a72b84174118d1ef86ec284d6771ccc0943556d. Subsequently, the profile UT allows for the easy confirmation of whether the UT has failed through the downstream processing and recognition of the XML.

kkarwot: Updated

RUIJIEZHONG66166 (approved):

kdrozd-dev:
@torchxpubot merge -f

PR #2994, opened 2026-03-05
Remove -fsycl-host-compiler, use pure icpx compilation

chunhuanMeng:
This pull request makes several improvements to the SYCL build system and device code error checking. The main themes are simplifying the SYCL host/device compilation process and improving assertion handling in device code. Notably, it removes the use of a separate host compiler for SYCL, updates related CMake logic, and standardizes kernel assertions for better device compatibility.

**Build system and SYCL compilation improvements:**

* Removed the use of a separate host compiler for SYCL, simplifying the build process by having `icpx` handle both host and device code. This change is reflected in the removal of `SYCL_HOST_COMPILER` and related flags from CMake files, and updates to how compilation flags and defines are handled. [[1]](diffhunk://#diff-1ef557c3e9c9fc2f266b2aed39984c8a488734e654043b5e79e99851d6dc9c60L103-R112) [[2]](diffhunk://#diff-d82bb39ac5cb9040c26a2047779a96ea826156519c3231126c13d65e904085feL62-L65) [[3]](diffhunk://#diff-5403684696b143181d44560ec2f35e37ffd9330ae5f3895361585cb1ca3cc18aL32) [[4]](diffhunk://#diff-5403684696b143181d44560ec2f35e37ffd9330ae5f3895361585cb1ca3cc18aL42-L52) [[5]](diffhunk://#diff-5403684696b143181d44560ec2f35e37ffd9330ae5f3895361585cb1ca3cc18aL65-R79) [[6]](diffhunk://#diff-5403684696b143181d44560ec2f35e37ffd9330ae5f3895361585cb1ca3cc18aL141-L142)
* Updated the way SYCL compilation flags are assembled and passed, ensuring that necessary preprocessor defines (e.g., `-D` flags) are correctly propagated to the device compiler, and added platform-specific handling for Intel library linking.
* Combined `SYCL_KERNEL_OPTIONS` and `SYCL_HOST_FLAGS` into `SYCL_COMPILE_FLAGS` for a more unified and correct flag application during compilation.
* Removed the `-Wno-interference-size` warning suppression for GNU compilers, as the host compiler is icpx.

**Device code and error handling improvements:**

[...]

PR #5062, opened 2026-08-21
Refactor SYCL flag handling for host and device compilation

chunhuanMeng:
## Motivation
PR [#2994](https://github.com/intel/torch-xpu-ops/pull/2994) removed `-fsycl-host-compiler` and enabled pure `icpx` compilation for SYCL kernels. Host flags were initially merged into the SYCL compiler flags because `icpx` compiles both host and device code.

However, host and device compilation may require different options. For example, a Debug build may use `-O0 -g` for host code while keeping device code optimized without debug information. Passing all flags through one list prevents such configurations and may incorrectly apply host-only options to the device frontend.

This PR separates host flags from common/device flags. Host options are forwarded individually with `-Xarch_host`, while `SYCL_COMPILE_FLAGS` contains only common and device compilation options.

## Additional warning fixes

Forwarding the complete host flag set to the `icpx` host frontend exposed three additional categories of warnings.

### GCC-only warning options

Some warning options inherited from PyTorch are supported by GCC but not by the Intel SYCL/Clang host frontend. Passing them through `-Xarch_host` produced an `unknown-warning-option` diagnostic for every SYCL translation unit.

These options are now excluded only when constructing the `icpx` command. They remain available to regular sources compiled by GCC.

### SYCL-TLA third-party warnings

The Clang host frontend reports additional warnings in the third-party SYCL-TLA/CUTLASS headers. With `WERROR=1`, those warnings prevented the SYCL-TLA target from compiling.

The affected diagnostics are demoted with `-Wno-error=*` only for the `torch-xpu-ops-sycltla` target. They remain visible as warnings, and other SYCL targets are unaffected.

### Host-pass source diagnostics

The Clang host pass also detected warnings in torch-xpu-ops sources that were not reported by the previous GCC host compiler.

These were fixed locally instead of disabling their warning classes globally:

- annotate helpers that are used only by the SYCL device pass;
- remove an unused lambda capture;
- remove unused functor fields, constructor state, and constants;
- preserve existing interfaces where message support has not yet been implemented.


cmake/Modules/FindSYCL/run_sycl.cmake
```diff
@@ -59,18 +65,14 @@ endforeach()
 # Choose host flags in FindSYCL.cmake
 @SYCL_host_flags@
 
+# These GCC-only warning controls are inherited from PyTorch's global host
+# flags. They are not accepted by the Intel SYCL/Clang frontend, even when
+# forwarded with -Xarch_host, and produce one warning block for every kernel.
+list(FILTER CMAKE_HOST_FLAGS EXCLUDE REGEX "^-Wno-stringop-overflow$|^-Wno-dangling-reference$|^-Werror=dangling-reference$|^-Wno-error=dangling-reference$")
```
dvrogozh: The `cmake/Modules/*` scripts are to define high level functions, macros and helpers. I do not think that this is the right place to define or modify flags. I suggest to:
1. Limit `cmake/Modules/*` to cmdline composition and other high level helpers
2. Deal with flags entirely in `cmake/BuildFlags.cmake`

cmake/BuildFlags.cmake
```diff
@@ -238,7 +238,7 @@ macro(set_build_flags)
     message(STATUS "Compile Intel GPU AOT Targets for ${AOT_TARGETS}")
   endif()
 
-  list(APPEND SYCL_COMPILE_FLAGS ${SYCL_KERNEL_OPTIONS} ${SYCL_HOST_FLAGS})
+  list(APPEND SYCL_COMPILE_FLAGS ${SYCL_KERNEL_OPTIONS})
```
dvrogozh: This means that `SYCL_KERNEL_OPTIONS` is now redundant. Can we just populate `SYCL_COMPILE_FLAGS` from the very beginning?

dvrogozh (changes requested):

cmake/BuildFlags.cmake
chunhuanMeng: `SYCL_KERNEL_OPTIONS` and `SYCL_COMPILE_FLAGS` intentionally have different meanings. `SYCL_KERNEL_OPTIONS` contains kernel/device-specific options defined by this file, while `SYCL_COMPILE_FLAGS` also contains externally supplied common SYCL flags from `SYCL_FLAGS`. We keep the two variables separate to preserve this distinction and combine them only when assembling the final compile command.

cmake/Modules/FindSYCL/run_sycl.cmake
chunhuanMeng: Agreed. These warning options are valid for standalone GCC host compilation, but are not accepted by the Intel SYCL/Clang frontend when forwarded with -Xarch_host. I defined the SYCL-specific exclusion policy in `BuildFlags.cmake` and pass it to `run_sycl.cmake`. The latter will only compose the icpx command line by omitting those entries, without modifying the original host flags used by GCC.

cmake/BuildFlags.cmake
```diff
@@ -73,6 +75,18 @@ macro(set_build_flags)
   elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     list(APPEND SYCL_HOST_FLAGS -fPIC)
     list(APPEND SYCL_HOST_FLAGS -std=${CPP_STD})
+    # These GCC warning options remain available to regular host compilation,
+    # but are not supported by the Intel SYCL/Clang host frontend.
+    list(APPEND SYCL_HOST_FLAGS_EXCLUDED_FROM_SYCL
+      -Wno-stringop-overflow
+      -Wno-dangling-reference
+      -Werror=dangling-reference
+      -Wno-error=dangling-reference)
```
dvrogozh, line 82: Are you still debugging? these dangling-reference options contradict each other. You either have `-Wdangling-reference -Werror=dangling-reference` or `-Wno-dangling-reference -Wno-error=dangling-reference`.

dvrogozh (changes requested):

cmake/BuildFlags.cmake
dvrogozh: I do not see that you use `SYCL_COMPILE_FLAGS` anywhere else. Same for `SYCL_KERNEL_OPTIONS`. This means that `SYCL_COMPILE_FLAGS = SYCL_KERNEL_OPTIONS` and one of them is redundant. I believe you can just populate `s/SYCL_KERNEL_OPTIONS/SYCL_COMPILE_FLAGS/`.

```
cmake/BuildFlags.cmake:42:  set(SYCL_KERNEL_OPTIONS)
cmake/BuildFlags.cmake:90:  # -- Kernel flags (SYCL_KERNEL_OPTIONS)
cmake/BuildFlags.cmake:104:  # icpx -fsycl -fsycl-target=${SYCL_TARGETS_OPTION} ${SYCL_KERNEL_OPTIONS} kernel.cpp -o kernel.o
cmake/BuildFlags.cmake:111:  list(APPEND SYCL_KERNEL_OPTIONS -fno-sycl-unnamed-lambda)
cmake/BuildFlags.cmake:112:  list(APPEND SYCL_KERNEL_OPTIONS -fno-sycl-id-queries-fit-in-int)
cmake/BuildFlags.cmake:113:  list(APPEND SYCL_KERNEL_OPTIONS -sycl-std=2020)
cmake/BuildFlags.cmake:114:  list(APPEND SYCL_KERNEL_OPTIONS -foffload-fp32-prec-div)
cmake/BuildFlags.cmake:115:  list(APPEND SYCL_KERNEL_OPTIONS -foffload-fp32-prec-sqrt)
cmake/BuildFlags.cmake:131:  list(APPEND SYCL_KERNEL_OPTIONS ${_sycl_strict_fp_flag})
cmake/BuildFlags.cmake:136:    list(APPEND SYCL_KERNEL_OPTIONS -Qstd=${CPP_STD})
cmake/BuildFlags.cmake:137:    list(APPEND SYCL_KERNEL_OPTIONS /Qfma)
cmake/BuildFlags.cmake:138:    list(APPEND SYCL_KERNEL_OPTIONS /Qftz-)
cmake/BuildFlags.cmake:140:    list(APPEND SYCL_KERNEL_OPTIONS -Wno-ignored-attributes)
cmake/BuildFlags.cmake:142:    list(APPEND SYCL_KERNEL_OPTIONS -std=${CPP_STD})
cmake/BuildFlags.cmake:143:    list(APPEND SYCL_KERNEL_OPTIONS -Wno-absolute-value)
cmake/BuildFlags.cmake:147:    list(APPEND SYCL_KERNEL_OPTIONS -ffp-contract=fast)
cmake/BuildFlags.cmake:152:      list(APPEND SYCL_KERNEL_OPTIONS -no-ftz)
cmake/BuildFlags.cmake:164:    list(APPEND SYCL_KERNEL_OPTIONS ${_sycl_intel_lib_libirc_flag})
cmake/BuildFlags.cmake:174:    list(APPEND SYCL_KERNEL_OPTIONS -g -O0 -Rno-debug-disables-optimization)
cmake/BuildFlags.cmake:176:    list(APPEND SYCL_KERNEL_OPTIONS
cmake/BuildFlags.cmake:215:    list(APPEND SYCL_KERNEL_OPTIONS ${SYCL_TARGETS_OPTION})
cmake/BuildFlags.cmake:239:          list(APPEND SYCL_KERNEL_OPTIONS -fsycl-fp64-conv-emu)
cmake/BuildFlags.cmake:244:      list(APPEND SYCL_KERNEL_OPTIONS ${SYCL_TARGETS_OPTION})
cmake/BuildFlags.cmake:252:  list(APPEND SYCL_COMPILE_FLAGS ${SYCL_KERNEL_OPTIONS} ${SYCL_HOST_FLAGS})
dvrogozh@DUT7308PVC:~/git/pytorch/pytorch2/third_party/torch-xpu-ops$
dvrogozh@DUT7308PVC:~/git/pytorch/pytorch2/third_party/torch-xpu-ops$
dvrogozh@DUT7308PVC:~/git/pytorch/pytorch2/third_party/torch-xpu-ops$
dvrogozh@DUT7308PVC:~/git/pytorch/pytorch2/third_party/torch-xpu-ops$ fgrep -rsn SYCL_COMPILE_FLAGS
cmake/BuildFlags.cmake:43:  set(SYCL_COMPILE_FLAGS ${SYCL_FLAGS})
cmake/BuildFlags.cmake:252:  list(APPEND SYCL_COMPILE_FLAGS ${SYCL_KERNEL_OPTIONS} ${SYCL_HOST_FLAGS})
cmake/Modules/FindSYCL/run_sycl.cmake:36:set(SYCL_compile_flags @SYCL_COMPILE_FLAGS@) # list
cmake/Modules/FindSYCL.cmake:21:#  SYCL_COMPILE_FLAGS
```

cmake/BuildFlags.cmake
chunhuanMeng: Good catch. The inherited host flags contain -Wno-dangling-reference and -Wno-error=dangling-reference, but not -Werror=dangling-reference. Intel Clang reports the unsupported -Wno-error=dangling-reference option in the diagnostic as -Werror=dangling-reference, which led me to add the latter defensively. I verified the flags individually and confirmed that the additional -Werror=dangling-reference was unnecessary, so I removed it.

src/ATen/native/xpu/sycl/LinearInt4.cpp
```diff
@@ -14,15 +14,15 @@
 #include <comm/SYCLContext.h>
 
 namespace at::native::xpu {
-static inline int padto_le(int src, int padding) {
+[[maybe_unused]] static inline int padto_le(int src, int padding) {
```
dvrogozh: Why do we need `[[maybe_unused]]` in all these `padto_le` functions? The function is kind of used - see link below. Maybe that's just one of the variants which is not used so we can drop it?

https://github.com/intel/torch-xpu-ops/blob/bd4ed4fba10455a2697b18628d58f65b0e83373d/src/ATen/native/xpu/sycl/LinearInt4.cpp#L108

src/ATen/native/xpu/sycl/DilatedMaxPool2d.cpp
```diff
@@ -34,7 +34,7 @@ DISABLE_RETURN_TYPE_WARNING_BEGIN
 
 namespace at::native::xpu {
 
-static inline int p_start(
+[[maybe_unused]] static inline int p_start(
```
dvrogozh: Why do we need `[[maybe_unused]]` in all the `p_start` functions? The function is kind of used - see link below.

https://github.com/intel/torch-xpu-ops/blob/bd4ed4fba10455a2697b18628d58f65b0e83373d/src/ATen/native/xpu/sycl/DilatedMaxPool2d.cpp#L574

src/ATen/native/xpu/sycl/DilatedMaxPool2d.cpp
```diff
@@ -45,7 +45,11 @@ static inline int p_start(
       : (size + pad - ((kernel - 1) * dilation + 1)) / stride + 1;
 }
 
-static inline int p_end(int size, int pad, int pooled_size, int stride) {
+[[maybe_unused]] static inline int p_end(
```
dvrogozh: Same as above. Usage is here:
https://github.com/intel/torch-xpu-ops/blob/bd4ed4fba10455a2697b18628d58f65b0e83373d/src/ATen/native/xpu/sycl/DilatedMaxPool2d.cpp#L575

src/ATen/native/xpu/sycl/TensorCompareKernels.cpp
```diff
@@ -192,36 +192,32 @@ struct AssertAsyncKernelFunctor1 {
   void operator()(sycl::nd_item<1> item) const {
     SYCL_KERNEL_ASSERT(input_[0] != 0);
   }
-  AssertAsyncKernelFunctor1(const scalar_t* input, Msg msg)
-      : input_(input), msg_(msg) {}
+  AssertAsyncKernelFunctor1(const scalar_t* input, Msg) : input_(input) {}
```
dvrogozh, line 197: Actually you are making only a partial fix because you address the warning but does not eliminate root cause. Which is that this kernel accepts the argument which is not actually used. If I grepped everything correctly, then the entrypoint to this behavior is here:
https://github.com/intel/torch-xpu-ops/blob/408612134e5686edf0ccbd03cd97d76f77e3f31e/src/ATen/native/xpu/TensorCompare.cpp#L89

Note second `""` empty argument - that's our message :). I believe that tensor compare kernels are the only once which are using this `AssertAsyncKernelFunctor1` kernel. Which is suspicious. Overall:
1. Is this `AssertAsyncKernelFunctor1` needed at all? why no other kernel is using it?
2. If it's still needed, then  we at least must remove the message argument from all the functions in the chain.

src/ATen/native/xpu/sycl/LossNLL2dKernels.cpp
```diff
@@ -265,7 +265,6 @@ void nll_loss2d_forward_kernel(
        &target_,
        &output,
        &total_weight,
-       &input,
```
dvrogozh, line 268: Why this one was removed? (you don't remove the class member associate with that, so this looks suspicios)

dvrogozh (changes requested):

src/ATen/native/xpu/sycl/LossNLL2dKernels.cpp
chunhuanMeng: `input` is no longer used; `input_` is used instead.

src/ATen/native/xpu/sycl/LinearInt4.cpp
chunhuanMeng: In the current build process, icpx compiles the source file in both host and device compilation passes. These functions are used by the device code, but have no references during the host compilation pass, so the host pass reports them as unused.
Therefore, they cannot be dropped as unused variants; they are still required by the device pass. `[[maybe_unused]]` only suppresses the unused-function warning in the host pass.

src/ATen/native/xpu/sycl/DilatedMaxPool2d.cpp
chunhuanMeng: Same as above

src/ATen/native/xpu/sycl/DilatedMaxPool2d.cpp
chunhuanMeng: Same as above

src/ATen/native/xpu/sycl/TensorCompareKernels.cpp
chunhuanMeng: Thanks for pointing this out. The limited scope of my change was intentional. I noticed the existing SYCL_KERNEL_ASSERT_MSG is not ready comment and interpreted the current Msg plumbing as scaffolding for enabling assertion-message support once the corresponding SYCL functionality becomes available. Therefore, I removed only the unused stored member that triggers the new compiler warning, while preserving the existing interface and call-chain structure.
In particular, `_assert_async_msg_xpu()`must retain the message argument because its signature implements the PyTorch operator schema. In addition, `_assert_async_msg_kernel()` is declared with `TORCH_XPU_API`, so removing its argument would change an exported function signature rather than merely cleaning up a private functor.
My goal in this PR was to address the warning exposed by the compiler change without introducing a broader API and implementation cleanup.
Given the existing comment and exported interface, I would prefer to keep the current message path and limit this PR to the warning fix.

src/ATen/native/xpu/sycl/LinearInt4.cpp
dvrogozh: In this case functions should be declared under `#ifdef __SYCL_DEVICE_ONLY__` and `[[maybe_unused]]` attributes dropped.

src/ATen/native/xpu/sycl/TensorCompareKernels.cpp
dvrogozh: Ok, that's fine then if these functions are plumbed into pytorch API. However, can you add a source comment with `TODO` before this constructor declaration explaining why the `Msg` argument is there, but not used? If we are waiting for SYCL functionality to be available, then we need an issue at https://github.com/intel/llvm/issues and add it to this TODO.

src/ATen/native/xpu/sycl/LossNLL2dKernels.cpp
dvrogozh: Ah, ok, that's a lambda capture.

src/ATen/native/xpu/sycl/LinearInt4.cpp
chunhuanMeng: This works for `DilatedMaxPool2d.cpp` because the calls are dependent on the template parameter `index_t`, so lookup can be deferred to the device compilation pass.

For these cases, the calls are non-dependent because their arguments have concrete int types. The functions therefore need to remain visible during the host pass, where `[[maybe_unused]]` is required to suppress `-Wunneeded-internal-declaration`.

dvrogozh: @tahonermann, can you, please, comment on that? Does SYCL has a way to mark functions as belonging to device or host paths beyond `#ifdef __SYCL_DEVICE_ONLY__`? The usage of `[[maybe_unused]]` does not seem quite right to me in this case. Maybe there are some `[[sycl_device]]` and `[[sycl_host]]` to mark functions or they need to be introduced?

src/ATen/native/xpu/sycl/TensorCompareKernels.cpp
dvrogozh: @chunhuanMeng , please, help to address this.

src/ATen/native/xpu/sycl/LinearInt4.cpp
tahonermann: I think this usage of `[[maybe_unused]]` is ok, but it has the unfortunate affect of also suppressing a warning during device compilation if a change should make the function unused there as well. Perhaps this would be a preferable approach:
```
#ifdef __SYCL_DEVICE_ONLY__
#define MAYBE_UNUSED_ON_HOST
#else
#define MAYBE_UNUSED_ON_HOST [[maybe_unused]]
#endif

MAYBE_UNUSED_ON_HOST static inline int padto_le(int src, int padding) {
```

> Maybe there are some `[[sycl_device]]` and `[[sycl_host]]` to mark functions or they need to be introduced?

DPC++ does support an `__attribute__((sycl_device_only))` attribute, but it is ignored during host compilation and has semantic affects that are not likely what you want here (it impacts overloading), so probably won't help with this.

src/ATen/native/xpu/sycl/TensorCompareKernels.cpp
chunhuanMeng: Thanks for the suggestion. I will open a torch-xpu-ops issue to track enabling SYCL_KERNEL_ASSERT_MSG.

For this PR, I prefer to keep the current minimal warning fix and preserve the existing message interface. The scope of this PR is SYCL flag refactoring, so the full assertion-message implementation will be handled separately.
issue link: #5165

chunhuanMeng:
@dvrogozh, could you please re-review this PR? If there are no remaining concerns, an approval would be appreciated. Thanks!

src/ATen/native/xpu/sycl/TensorCompareKernels.cpp
dvrogozh: > the full assertion-message implementation will be handled separately.

That's fine. However, can you, please, add a TODO in the source comment?


```suggestion
  // TODO: add actual handling for the message argument
  // See: https://github.com/intel/torch-xpu-ops/issues/5165
  AssertAsyncKernelFunctor1(const scalar_t* input, Msg) : input_(input) {}
```

src/ATen/native/xpu/sycl/LinearInt4.cpp
dvrogozh: @chunhuanMeng , the suggestion from Tom makes sense to me. This way we will make sure that function is actually used for the device. Can you consider to implement that, please?

src/ATen/native/xpu/sycl/TensorCompareKernels.cpp
```diff
@@ -192,36 +192,34 @@ struct AssertAsyncKernelFunctor1 {
...
-  Msg msg_;
 };
 
 struct AssertAsyncKernelFunctor2 {
   void operator()(sycl::nd_item<1> item) const {
     SYCL_KERNEL_ASSERT(input_[0] != c10::complex<float>(0, 0));
   }
-  AssertAsyncKernelFunctor2(const c10::complex<float>* input, Msg msg)
-      : input_(input), msg_(msg) {}
+  AssertAsyncKernelFunctor2(const c10::complex<float>* input, Msg)
```
dvrogozh, line 209:
```suggestion
  // TODO: add actual handling for the message argument
  // See: https://github.com/intel/torch-xpu-ops/issues/5165
  AssertAsyncKernelFunctor2(const c10::complex<float>* input, Msg)
```

src/ATen/native/xpu/sycl/TensorCompareKernels.cpp
```diff
@@ -192,36 +192,34 @@ struct AssertAsyncKernelFunctor1 {
...
-  Msg msg_;
 };
 
 struct AssertAsyncKernelFunctor3 {
   void operator()(sycl::nd_item<1> item) const {
     SYCL_KERNEL_ASSERT(input_[0] != c10::complex<double>(0, 0));
   }
-  AssertAsyncKernelFunctor3(const c10::complex<double>* input, Msg msg)
-      : input_(input), msg_(msg) {}
+  AssertAsyncKernelFunctor3(const c10::complex<double>* input, Msg)
```
dvrogozh, line 222:
```suggestion
  // TODO: add actual handling for the message argument
  // See: https://github.com/intel/torch-xpu-ops/issues/5165
  AssertAsyncKernelFunctor3(const c10::complex<double>* input, Msg)
```

dvrogozh (commented):
@chunhuanMeng , other than 2 more nit comments, I am good with the PR and ready to approve. The problem is that CI (build) is failing. Seems some warnings treated as error coming from SYCL TLA.

dvrogozh:
```
/home/jenkins/actions-runner/_work/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/src/ATen/native/transformers/xpu/flash_attn/sycltla/mha_bwd_launch.h:577:17: error: unused variable 'kNSGs' [-Werror,-Wunused-variable]
  577 |   constexpr int kNSGs = Trait::kNSGs;
      |                 ^~~~~
```

And other like that fail CI build. Not sure if that's related to this PR.

dvrogozh (approved):
LGTM

cmake/Modules/FindSYCL/run_sycl.cmake
```diff
@@ -61,17 +69,13 @@ endforeach()
...
-  # Extract -D (GCC/Clang) or /D (MSVC) defines from CMAKE_HOST_FLAGS and pass
-  # them directly to icpx, since host compiler is removed. This is needed for
-  # macros like HAVE_AVX512_CPU_DEFINITION that control signatures, to avoid
-  # symbol mismatch with libtorch_cpu.so.
-  if(flag MATCHES "^-D")
-    list(APPEND SYCL_compile_flags "${flag}")
-  elseif(flag MATCHES "^/D")
-    string(REGEX REPLACE "^/D" "-D" converted_flag "${flag}")
-    list(APPEND SYCL_compile_flags "${converted_flag}")
+  if(NOT flag IN_LIST SYCL_host_flags_excluded_from_sycl)
```
CuiYifeng: I wonder if `-Wno-stringop-overflow ` (one more space at the end) can bypass this check. If so, do we need to match with Regex?

CuiYifeng (commented):
The remaining part of this PR LGTM.

cmake/Modules/FindSYCL/run_sycl.cmake
chunhuanMeng: Good point. IN_LIST performs exact matching, so trailing whitespace could bypass the exclusion. I normalized the flag with string(STRIP) before checking it, instead of using a broad regex.

CuiYifeng (approved):

chunhuanMeng:
@torchxpubot merge -f “unrelated failure”

PR #5064, opened 2026-08-21
Add XPU dispatch for linalg.polar

Silv3S:
This patch removes `linalg.polar.out` from hardcoded CPU-fallback list, so it can be registered in upstream PyTorch `native_functions.yaml`. This op decomposes into other ops e.g. `svd` that will fallback to CPU anyway, but it will be easier to add native implementation in the future

Upstream PR: https://github.com/pytorch/pytorch/pull/188253


BBBela (approved):

pbielak (approved):

guangyey:
@torchxpubot merge -f "https://github.com/pytorch/pytorch/pull/188253 was merged!"

PR #5116, opened 2026-08-26
bot: let `@torchxpubot fix` source-build and verify C++/SYCL fixes

ZhaoqiongZ:
Lets `@torchxpubot fix` source-build and verify C++/SYCL fixes, not just Python ones.

Previously the `fix` job had only a nightly wheel and no oneAPI toolchain, so C++/SYCL changes under `src/` couldn't be compiled and were handed off unverified.

- **Container** → `intel/omix:0.3.0-devel-ubuntu26.04` (ships oneAPI); a fast `icpx --version` gate fails the job in seconds if the toolchain is missing.
- **One tree, no divergence**: for a C++/SYCL fix the agent works in the pytorch clone's `third_party/torch-xpu-ops` submodule (cloned by `/fix-reproduce` into `agent_space_xpu/pytorch`), commits it on `agent/fix-issue-<N>`, points that tree's `third_party/xpu.txt` at the branch HEAD so CMake builds the fix, and builds/verifies that same tree — so the verified tree is the kept tree. Python fixes run against the nightly wheel in the workspace checkout. `fix-verify` accepts the fix staged or committed.
- **Not pushed — exported as a patch**: the branch is never pushed (a human decides that, and the pytorch-side tree can't be pushed as a branch anyway). The workflow runs `git format-patch` and uploads the fix as a downloadable **patch artifact**, alongside the transcript and `fix_result` JSON. `/issue-handler` owns the batch fan-out.
- **Build fits the runtime**: job timeout 180 → 300 min; only `BASH_MAX_TIMEOUT_MS` is raised to just under it (the agent passes an explicit long timeout on the build call) while `BASH_DEFAULT_TIMEOUT_MS` stays at the stock 2min so an unrelated hung command fails fast. `TORCH_XPU_ARCH_LIST=bmg,pvc` pins AOT to the runner GPUs instead of every supported arch.
- **Secrets**: redaction runs before anything reads the agent output — the result/cost comments (agent free text posted to a public issue) and the transcript / patch / JSON artifacts are all scrubbed first, and every publish is gated on redaction succeeding (a missing `perl` is a hard failure). Redaction covers the three job secrets, `GITHUB_TOKEN`, and the base64 `x-access-token:<token>` form `actions/checkout` writes into `.git/config`.
- **Cleanup**: `AGENT_SPACE` scratch dir is reclaimed at job end; the workspace `chown 1000:1000 /__w /github ./ -R` (matching the repo's other container jobs) hands ownership back so the next run's checkout isn't denied.

Rebased onto latest `main`; `references/environment-setup.md` was dropped along with main (its content now lives in `/xpu-build-pytorch`).

**Known limitation:** the "before" is a nightly-wheel failure and the "after" is a source build with the fix, so a PASSED verdict does not by itself prove the fix caused the pass (a bug already fixed upstream after the nightly was cut would also pass). `fix-verify` documents this; a human reviewing the patch confirms the change is responsible. Verification is otherwise self-reported (no separate gate / reviewer).

### DIFF
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -1368,27 +1368,12 @@ jobs:
               });
             }
 
-      # A build-in-the-loop run leaves a full pytorch clone + build tree (tens
-      # of GB, hundreds of thousands of files) under $AGENT_SPACE on this shared
-      # self-hosted runner. The fix was already pushed to its branch and
-      # reported in the issue comment, so reclaim it now — this also keeps the
-      # chown below from having to walk that tree.
+      # Reclaim the tens-of-GB build tree before the chown below has to walk it.
       - name: Reclaim agent scratch space
         if: ${{ always() }}
         run: rm -rf "$AGENT_SPACE" || true
 
-      # This privileged container runs as root, so every file it writes into the
-      # runner's bind mounts is root-owned (claude-code-action's `bun install`
-      # under _actions, tool installs under _tool, step files under _temp, the
-      # checked-out repo under _work/<repo>). _actions/_tool/_temp are SEPARATE
-      # bind mounts under /__w, so `chown -R /__w` does not reliably recurse into
-      # them; name each explicitly. If left root-owned, the NEXT job on this
-      # shared self-hosted runner dies in "Prepare workflow directory" (Access
-      # to .../_actions/... denied) before any step runs. The runner admins will
-      # not add a pre-job hook, so each job must hand these back itself. Derive
-      # the runner uid:gid from /__e (the runner's externals, a read-only mount
-      # it owns and never writes to) instead of hardcoding 1000 — the service
-      # user here is not uid 1000.
+      # Hand back root-owned files (uid from /__e, not hardcoded) so the next job on this shared runner isn't denied access to its bind mounts.
       - name: Cleanup workspace
         if: ${{ always() }}
         run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/bot.yml
```diff
@@ -1143,7 +1153,8 @@ jobs:
 
       - name: Configure git identity
         run: |
-          git config --global --add safe.directory "${{ github.workspace }}"
+          # "*": a source-build fix makes nested root-owned git repos under agent_space_xpu/ that git would otherwise reject.
+          git config --global --add safe.directory "*"
```
Stonepia: Is this safe? I kind of think it is AI trying to do some lazy change so avoid the actual error.

Stonepia: @copilot+claude-opus-5  verify if my suspection is correct or not. What is the suggested change? DO NOT modify the code yourself

Stonepia (approved):

chuanqi129:
@torchxpubot review

ZhaoqiongZ:
@torchxpubot review

ZhaoqiongZ:
@torchxpubot merge

ZhaoqiongZ:
@torchxpubot review

chuanqi129 (approved):
Please check the bot review comments before land it. Especially for the security part

ZhaoqiongZ:
@torchxpubot review

ZhaoqiongZ:
@torchxpubot review

ZhaoqiongZ:
@torchxpubot review

ZhaoqiongZ:
@torchxpubot review

ZhaoqiongZ:
@torchxpubot merge

PR #5140, opened 2026-08-28
Refine gamma/beta load for LayerNorm

CuiYifeng:
This PR is to refine the vectorization for gamma/beta load in `VectorizedLayerNormKernelFunctor` to generate `load.ugm.d32x2.a64` instead of two `load.ugm.d32.a64`. The main change is the introduction of local variables to hold `gamma_vec[i]` and `beta_vec[i]` before entering the loop of gamma/beta processing.
Due to the usually low proportion of gamma/beta among all inputs, this issue has no significant impact on performance, but it should still be fixed.

### DIFF
--- a/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
+++ b/src/ATen/native/xpu/sycl/LayerNormKernels.cpp
@@ -498,34 +498,38 @@ struct VectorizedLayerNormKernelFunctor
       // Computation is performed in T_ACC, X is cast to T_ACC and result is
       // implicitly cast to T
       if (gamma_vec != nullptr && beta_vec != nullptr) {
+        vec_t gamma_data = gamma_vec[i];
+        vec_t beta_data = beta_vec[i];
 #pragma unroll
         for (int ii = 0; ii < vec_size; ii++) {
           if constexpr (!rms_norm) {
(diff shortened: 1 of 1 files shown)


LuFinch (approved):
LGTM.  Do you see any performance improvement?

CuiYifeng:
> LGTM. Do you see any performance improvement?

Far less than the fluctuation in measurement.

src/ATen/native/xpu/sycl/LayerNormKernels.cpp
```diff
@@ -498,34 +498,38 @@ struct VectorizedLayerNormKernelFunctor
       // Computation is performed in T_ACC, X is cast to T_ACC and result is
       // implicitly cast to T
       if (gamma_vec != nullptr && beta_vec != nullptr) {
+        vec_t gamma_data = gamma_vec[i];
+        vec_t beta_data = beta_vec[i];
```
guangyey: When `rms_morm` is `true`, this will become unused code.

CuiYifeng: Good catch! Updated.

guangyey (approved):

CuiYifeng:
@torchxpubot merge

PR #5144, opened 2026-08-29
Use FunctionSchema::name() instead of OperatorName::name field

cyyever:
```
OperatorName::name is expected to become a private field behind a
name() accessor in a future PyTorch release, so
op.schema().operator_name().name will stop compiling once that lands.

FunctionSchema already exposes its own name() (a thin wrapper over the
same underlying string) as a stable public API independent of that
internal change, so op.schema().name() works identically before and
after the change - it's also one hop shorter.
```

Drafted with AI assistance (Claude Code).


guangyey (approved):

guangyey:
Thanks!

PR #5147, opened 2026-08-31
[CI] Add bmg distributed test in CI and nightly wheel weekly

zxd1997066:
Add BMG runner coverage for distributed tests, plus a weekly-only BMG suite using a temporary validation branch.

Changes:
- pull.yml: linux-distributed now runs xpu_distributed on both `distributed` and `bmg-test` runners.
- nightly_ondemand.yml: new `xpu_distributed_bmg` ut scope, only auto-added on the Saturday nightly-wheel schedule (30 16 * * 5); also usable via on-demand dispatch.
- linux-uttest/action.yml: new `xpu_distributed_bmg` step that clones the `daisyden/distributed_2.13` branch inside the step and runs `run_distributed_local.py`.
- ut_result_check.sh: route `xpu_distributed_bmg` through the same result-parsing path as `xpu_distributed`.
- run_distributed.py: fix ZE_AFFINITY_MASK on BMG. BMG topology has no XL-linked GPU groups (all SYS), so the old logic collapsed to a single GPU index; now falls back to the first WORLD_SIZE GPUs when no group is found.


.github/actions/linux-uttest/action.yml
```diff
@@ -156,6 +156,36 @@ runs:
           tee ${{ github.workspace }}/ut_log/xpu_distributed/xpu_distributed_test.log
         find ../ -type f -name "*.xml" -exec cp {} ${{ github.workspace }}/ut_log/ \;
 
+    - name: xpu_distributed_bmg
```
mengfei25: I think we can add it into `xpu_distributed` directly instead of a new, and next we will move our all tests on BMG.

.github/workflows/nightly_ondemand.yml
```diff
@@ -29,7 +29,7 @@ on:
       ut:
         type: string
         default: ''
-        description: UT scope. 'op_regression,op_extended,op_ut,skipped_ut,xpu_profiling,torch_xpu,op_regression_dev1,xpu_distributed,microbench,windows'. Delimiter is comma
+        description: UT scope. 'op_regression,op_extended,op_ut,skipped_ut,xpu_profiling,torch_xpu,op_regression_dev1,xpu_distributed,xpu_distributed_bmg,microbench,windows'. Delimiter is comma
```
chuanqi129, line 32: Why we need to add a new `xpu_distributed_bmg` scope?

.github/actions/linux-uttest/action.yml
chuanqi129, line 159: Can we merge this step and original one together?

.github/actions/linux-uttest/action.yml
```diff
@@ -156,6 +156,36 @@ runs:
...
+      shell: timeout 36000 bash -xeu -o pipefail {0}
+      if: ${{ inputs.ut_name == 'xpu_distributed_bmg' }}
+      run: |
+        xpu-smi topology -m
+        mkdir -p ut_log/xpu_distributed_bmg
+
+        rm -rf pytorch/third_party/torch-xpu-ops
+        git clone --depth 1 --branch daisyden/distributed_2.13 \
```
chuanqi129, line 167: Why we still need depends on this branch?

.github/actions/linux-uttest/action.yml
zxd1997066: The original one is a small scope for CI, this is for full weekly test

.github/actions/linux-uttest/action.yml
zxd1997066: Some test files are not upstream yet, put them on daisyden/distributed_2.13 temporarily

.github/workflows/nightly_ondemand.yml
zxd1997066: it is for full distributed weekly test, will rename it to `xpu_distributed_weekly`

.github/actions/linux-uttest/action.yml
chuanqi129: Can we move them into torch-xpu-ops directly?

.github/actions/linux-uttest/action.yml
```diff
@@ -156,6 +156,36 @@ runs:
           tee ${{ github.workspace }}/ut_log/xpu_distributed/xpu_distributed_test.log
         find ../ -type f -name "*.xml" -exec cp {} ${{ github.workspace }}/ut_log/ \;
 
+    - name: xpu_distributed_weekly
```
chuanqi129, line 159: I think we still can merge them in one step and use a flag to detect whether in CI or weekly. Seems we already have CI flag, @mengfei25 is it right? or check by event type, weekly has specific timer trigger.

.github/actions/linux-uttest/action.yml
zxd1997066: yes, they are in torch-xpu-ops currently https://github.com/intel/torch-xpu-ops/tree/daisyden/distributed_2.13/test/xpu/distributed, or do you mean move ~50 files to main branch?

.github/actions/linux-uttest/action.yml
mengfei25: Yes, trigger event and timer can distinguish them. And we can also setup a input or env.

PR #5151, opened 2026-08-31
[Workflow] Unify XPU alignment run summaries

laifenxiawucha:
## Summary

The alignment workflow had three related problems: review results did not clearly distinguish upstream-owned work from XPU work that needs a tracker, run comments varied by outcome, and publishing still required manual filing while preparation occupied an XPU runner.

This PR:

- judges upstream landing against the collected default-branch snapshot and separates tracker ownership from the repository where the fix belongs;
- publishes one consistent Run Summary for complete, warning, partial, and failed runs, with links back to the workflow;
- automatically files 1–3 reviewed candidates and leaves all candidates as drafts for batches of 4 or more;
- removes `@torchxpubot file` and adds the scan date to issue titles;
- moves `scan-prepare` to `ubuntu-latest`, leaving only credential-free reproducer execution on the XPU runner;
- records per-role usage and keeps GitHub publishing under the `torchxpubot` identity.

Before: verdict ownership was ambiguous, run status was inconsistent, filing needed a separate command, and preparation consumed XPU capacity.

After: reviewed XPU work is reported and published through one traceable flow, while XPU machines are reserved for reproducer execution.

## Validation

Validated with Python compilation, skill validation, YAML parsing, temporary gate/publisher and summary fixtures, `lintrunner`, and `git diff --check`.


ZhaoqiongZ:
@torchxpubot review

laifenxiawucha:
> > Replying to [this comment](https://github.com/intel/torch-xpu-ops/pull/5151#issuecomment-5475908117) by @ZhaoqiongZ
> 
> ## PR Review: #5151
> ### Summary
> Replaces the branchy XPU alignment run note with a single deterministic summary classified into four states (`complete`, `complete-with-warnings`, `partial`, `failed`), and makes the scheduled summary update in place on retry. The state classifier and the workflow's error/warning split line up correctly, but the summary can misreport a blocked run as successful, loses the identifying detail from global blockers, and the PR deliberately ships without the tests the author already wrote.
> 
> ### Correctness
> **`run_state` is a second source of truth for the run outcome, with no consistency check** (`xpu_alignment_publish.py:47-57`). The gate already computes the authoritative outcome as `decision` / `would_decision`, but `run_state` re-derives it from the raw `global_blockers` / `collection_status` / `unit_blockers` fields. `main()` cross-checks `mode` against `verdict` ("Dry-run mode and decision do not agree") but performs no equivalent check here, so the two can disagree. Rendered directly:
> 
> ```
> decision = {'decision': 'blocked', 'mode': 'schedule', 'collection_status': 'complete', 'payloads': []}
> -> XPU alignment run complete
>    - Run status: success
>    notify False
> ```
> 
> A `blocked` artifact whose `global_blockers` list is empty (or renamed by a future gate change) is published as a clean, un-notified success. Add `if (verdict == "blocked") != bool(decision.get("global_blockers")): fail(...)` alongside the existing dry-run agreement check, or key `run_state` off `would_decision`.
> 
> **`_normalize_global_blocker` destroys the identifying part of every blocker, and truncation happens after normalization** (`xpu_alignment_publish.py:95-98`, `168-175`). Rendered with six real gate blocker codes:
> 
> ```
> - Global blockers: 6
>   - scan invalid unit
>   - scan invalid unit
>   - runner log digest mismatch
>   - execution digest mismatch
>   - decision invalid id
>   - 1 additional blocker(s) omitted
> ```
> 
> Two lines are identical, no unit id survives, and the one blocker that carried genuinely distinct information (`collection-invalid: HTTP 403 rate limited`) is the one the `[:5]` slice dropped. Only `{label}-unreadable:{error}` and `collection-invalid:{error}` embed exception text; every other gate code embeds a unit id or a field name (`scan-invalid-unit:{id}`, `execution-digest-mismatch:{id}`, `runner-environment-invalid:{field}`). Preserve the suffix when it matches `UNIT_ID_RE`, and de-duplicate before slicing to 5.
> 
> **`_partial_progress` interpolates the raw collector exception string** (`xpu_alignment_publish.py:114-120`). `collection_progress[].error.message` is `str(error)` from `xpu_alignment_collect.py:493,507` — free-form and possibly multi-line. Rendered:
> 
> ```
>   - `b`: 1 page(s), 5 item(s), last cursor `none`, rate reset `...` — HTTP 403: rate limit
> retry later
> ```
> 
> The second line escapes the bullet and breaks the list. This also contradicts the policy `_normalize_global_blocker` states one function above ("Keep stable error codes in comments, not raw exception details") — use the same treatment, or at minimum collapse newlines and wrap in a code span. Separately, the line dropped the per-source `status` field the old output printed, while the `if not affected` fallback at `xpu_alignment_publish.py:110-111` lists _all_ sources including complete ones — so the reader cannot tell which source was actually partial.
> 
> **"Drafts awaiting filing" counts drafts that are already filed** (`xpu_alignment_publish.py:141`, `152-159`). In `triage` mode a payload whose draft carries `FILED_MARKER` takes the `else` branch at line 258 ("Skipping ... already present"), never enters `filed`, and therefore appears under `fileable_drafts` with the instruction "Approve drafts one at a time with `@torchxpubot file <unit-id>`". A triager who follows that instruction gets `parse_draft`'s hard failure: "`<unit-id>` was already filed as #N". `existing` is in scope at the call site — filter the awaiting list with `FILED_MARKER_RE`.
> 
> **Dry-run still creates the duplicate-marker state that `find_run_note` now treats as fatal** (`xpu_alignment_publish.py:242`, `alignment_triage.py:104-109`). `existing_run_note` is forced to `None` for dry-run, so re-running a dry-run workflow run posts a second comment carrying the same `RUN_NOTE_MARKER`. Since the marker is already run-id scoped, distinct runs get distinct comments under the update path anyway; the special case buys nothing and only manufactures the ambiguity the new duplicate check calls out.
> 
> **`RUN_NOTE_MARKER`'s comment no longer describes the mechanism** (`alignment_triage.py:30`): "One notification per run, so a re-run does not ping anyone twice." That was structurally guaranteed by not posting twice. It now depends on GitHub not sending notifications for `@mentions` added by editing an existing comment — load-bearing, non-obvious, and exactly the kind of global context the repo's comment policy asks to record.
> 
> **`Review outcomes: none` is ambiguous** (`xpu_alignment_publish.py:60-70`). It reads the same whether no unit was actionable, or review returned nothing for actionable units. For a comment that is meant to be the deterministic record of the run, include the inventory/triage counts — `mandatory_units` is already in the artifact.
> 
> ### Testing
> The PR body states the branches were exercised locally "before removing the temporary test module from the PR." That is the wrong direction: `.github/workflows/bot.yml:1800-1822` exists specifically for this, and its own rationale comment says "Pure python, no GPU, seconds -- so a regression in it is caught on the PR that introduces it instead of by losing a fix in production." `run_state`, `run_note`, `_excluded_units`, `_review_outcomes`, and `_normalize_global_blocker` are pure functions over a dict, and every finding above was reproducible in a few lines. Commit the test module. Note the job's glob is `test_bot_*.py`, so the file must either match that pattern or the glob must be widened.
> 
> ### Recommendation
> **Request Changes**
> 
> The blocked-renders-as-success classification gap and the deleted test module are blocking. The global-blocker normalization and the raw exception interpolation both make the new summary less debuggable than the output it replaces, which is the opposite of the PR's goal. The state-to-workflow mapping (error + `exit 1` for `failed`/`partial`, `::warning::` for isolated unit blockers and verification gaps) is consistent with the gate's `needs_attention` computation and needs no change.
> 
> Custom skills applied: pr-review.
> 
> _Generated by [`review` job](https://github.com/intel/torch-xpu-ops/actions/runs/33373982197/job/99431202467)._

done

ZhaoqiongZ:
@torchxpubot review

laifenxiawucha:
@torchxpubot review

ZhaoqiongZ:
@torchxpubot review

laifenxiawucha:
@ZhaoqiongZ done

.github/scripts/xpu_alignment_publish.py
```diff
@@ -156,94 +214,111 @@ def main() -> int:
...
+                print(f"Skipping {unit_id}: already filed as {filed_issue_url}.")
+                continue
+            if verdict == "triage":
+                print(f"Queued {unit_id} for manual triage on #{args.triage_issue}")
+                continue
+
+            issue_url = _file_candidate(args.repo, payload, draft, comment_id)
+            filed.append((unit_id, issue_url))
+            print(f"Filed {unit_id} as {issue_url}")
+        except (SystemExit, ValueError, KeyError, TypeError, OSError):
```
ZhaoqiongZ: use except (Exception, SystemExit) as error and then print both unit_id & error

ZhaoqiongZ (approved):

.claude/skills/xpu-alignment/references/automation-contract.md
Stonepia: I think this is not needed. You don't have to explain to the Agent about these things.

.github/scripts/alignment_triage.py
```diff
@@ -96,43 +108,42 @@ def post_comment(repo: str, issue: int, body: str) -> int:
     return int(created["id"])
 
 
-def has_unit(comments: list[dict], unit_id: str) -> bool:
+def find_unit_comments(comments: list[dict], unit_id: str) -> list[dict]:
     marker = UNIT_MARKER.format(unit_id=unit_id)
-    return any(marker in (comment.get("body") or "") for comment in comments)
+    matches = [comment for comment in comments if marker in (comment.get("body") or "")]
```
Stonepia: This line only checks if `marker` in the body. However, if the body contains the copy of previous comments, this will gets wrong.

For example, a reviewer may quote and then reply. This "marker check" would treat this as a match, then something may fail

.claude/skills/xpu-alignment/references/automation-contract.md
laifenxiawucha: done

.github/scripts/alignment_triage.py
laifenxiawucha: done.It now verifies the identity of the comment author.

Stonepia (approved):

ZhaoqiongZ:
@torchxpubot merge

Issue #5079, opened 2026-08-24
[BMG] test_combo_kernels test_combo_kernel_split_large_reductions fails on BMG (pass on PVC) in torch-2.14


daisyden:
## Summary

BMG-specific Inductor combo kernel failures observed with torch-2.14. These cases **pass on PVC** but fail on BMG, so they are hardware (BMG) specific.

## Affected cases

- `inductor/test_combo_kernels.py::ComboKernelTests::test_combo_kernel_split_large_reductions`
- `inductor/test_combo_kernels.py::ComboKernelTestsPerSubkernelBlocks::test_combo_kernel_split_large_reductions`

## Environment

- Torch: 2.14
- Hardware: Intel BMG (fails)
- Reference: Intel PVC (passes)

## Reproduction

```bash
python test/inductor/test_combo_kernels.py ComboKernelTests.test_combo_kernel_split_large_reductions
python test/inductor/test_combo_kernels.py ComboKernelTestsPerSubkernelBlocks.test_combo_kernel_split_large_reductions
```

## Notes

- BMG-only regression: both cases pass on PVC with the same torch-2.14 build.

ZhaoqiongZ:
@torchxpubot fix

ZhaoqiongZ:
@torchxpubot fix

torchxpubot:
Automated fix session for #5079 (BMG-specific `test_combo_kernel_split_large_reductions` failure).



## Issue Triage

| Field | Value |
|-------|-------|
| Issue type | single-bug |
| Reproduction missing | no |
| Scope | pytorch |
| Dependencies | (none named) |
| Handling | agent-fixable |

Reproducer is an explicit `test/inductor/test_combo_kernels.py` invocation, and this runner has the BMG hardware the issue names, so the hardware-specific failure is reachable here.



## Reproduce

Both node ids fail on this runner (Intel Arc Pro B60 / BMG, `torch 2.15.0.dev20260830+xpu`) with `AssertionError: Scalars are not equal! Expected 5 but got 4`, including under `TORCHINDUCTOR_FORCE_DISABLE_CACHES=1`, so this is not stale-cache pollution. Verdict **REPRODUCED**.



## Root-cause Analysis

- **Issue type:** inductor bug (test portability)
- **Fix repo:** pytorch
- **Analyzed at:** pytorch@69c9b15d
- **Verdict:** IMPLEMENTING

[...]

PR #5161, opened 2026-09-01
bot: let the fix pipeline own its issue comment end to end

ZhaoqiongZ:
### Problem

`@torchxpubot fix` had two publishers writing two different comments, and which ones showed up was a coin flip:

- the workflow's `Post result comment` step posted the agent's closing summary (`result.result`);
- the `issue-handler` skill was separately told to maintain an accumulating **session comment** of per-stage blocks via `gh issue comment --edit-last`.

The fix container ships without `gh`, so the skill's instruction could not be followed as written. Both runs hit `gh: command not found`; what happened next differed:

| | #3426 | #5079 |
|---|---|---|
| stage reports | posted (agent improvised a `urllib` REST fallback) | not posted |
| workflow summary | posted | posted |
| comments left | 2, disagreeing | 1, summary only |

Same workflow commit, same image, same prompt. The only difference was the agent's choice after `gh` turned out to be missing.

### Change

Install `gh`, and hand the pipeline the comment the workflow already creates:

- `Install gh CLI` — pinned v2.98.0 upstream tarball. Not apt: `gh` is a single static binary and the distro package lags far enough to miss subcommands the skills use (2.4.0 on a current Ubuntu).
- `Post starting message` passes its comment id to the agent as `SESSION_COMMENT_ID`. Triage overwrites the placeholder text, stages 3-5 append their blocks, Stage 6 appends the closing summary.
- `Post result comment` is **deleted**. `result.result` now only reaches the transcript artifact.

One comment per run, and it fills in live rather than appearing whole at the end (a fix job runs 30min-5h):

```
<!-- agent:session -->
<!-- agent:triage -->       Stage 1, replaces "Starting automated fix..."
<!-- agent:root-cause -->   Stage 3
<!-- agent:implement -->    Stage 4 (diff)
<!-- agent:verify -->       Stage 5
<!-- agent:summary -->      Stage 6 (verdict / branch / caveats)
_Generated by [fix job](...)_
```

### Reviewer note on redaction

The redact step runs after the agent exits, so it no longer gates anything public — the comment is written while the job runs. Post-hoc redaction would not help anyway: GitHub keeps comment edit history visible. The prompt therefore constrains the comment to stage reports only, never raw command output or environment. This is not a new exposure — #3426 already posted from inside the agent unredacted — it is now deliberate rather than accidental. Artifact redaction (transcript, patches, `fix_result*.json`) is unchanged.

### Test plan

```bash
python3 -c "
import yaml
steps = yaml.safe_load(open('.github/workflows/bot.yml'))['jobs']['fix']['steps']
print([s.get('name') for s in steps])
assert not any(s.get('name') == 'Post result comment' for s in steps)
"

curl -fsSL https://github.com/cli/cli/releases/download/v2.98.0/gh_2.98.0_linux_amd64.tar.gz | tar -xz -C /tmp
/tmp/gh_2.98.0_linux_amd64/bin/gh --version    # gh version 2.98.0 (2026-08-20)
/tmp/gh_2.98.0_linux_amd64/bin/gh api --help   # -f/--raw-field ... @<path> reads value from file

lintrunner -m origin/main                                  # ok, no lint issues
python3 -m pytest .github/scripts/test_bot_export_fix_patch.py -q   # 11 passed
```

End-to-end coverage is `@torchxpubot fix` on a live issue; the comment shape cannot be exercised from a PR.

Authored with the assistance of an AI coding agent (opencode).


Stonepia (approved):

CuiYifeng (approved):

ZhaoqiongZ:
@torchxpubot merge

Issue #5126, opened 2026-08-27
[Dynamo][Reductions] test_tensor_compare_ops_empty fails on CPU and XPU with FakeTensorObservedException


gplutop7:
### 🐛 Describe the bug

Under Dynamo-wrapped execution, the following tests fail:
- `TestReductionsDeviceCPU.test_tensor_compare_ops_empty_cpu`
- `TestReductionsDeviceXPU.test_tensor_compare_ops_empty_xpu`

Both fail during fake-tensor/meta evaluation of `aten.amax.default` with:
`RuntimeError: reducing over zero-size dimension for reduction operation without identity`

instead of matching the test expectation (`IndexError` with message containing `Expected reduction dim`).

Repro:
`PYTORCH_TEST_WITH_DYNAMO=1 python test_reductions.py \
  TestReductionsDeviceCPU.test_tensor_compare_ops_empty_cpu \
  TestReductionsDeviceXPU.test_tensor_compare_ops_empty_xpu -v`

Branch:
gplutop7:gplutopx/reductions-math-tests

### Versions

PyTorch version: 2.15.0a0+gitd43095a
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: Ubuntu 24.04.3 LTS (x86_64)
GCC version: (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version: Could not collect
CMake version: version 4.2.1
Libc version: glibc-2.39

Python version: 3.12.3 (main, Jan 22 2026, 20:57:42) [GCC 13.3.0] (64-bit runtime)
Python platform: Linux-6.14.0-1012-intel-x86_64-with-glibc2.39
Is CUDA available: False
CUDA runtime version: No CUDA

daisyden:
<details>
<summary>label-issue: intel/torch-xpu-ops#5126</summary>

Root cause: under `PYTORCH_TEST_WITH_DYNAMO=1`, fake-tensor/meta evaluation of `aten.amax.default` on a zero-size dim raises `RuntimeError: reducing over zero-size dimension ... without identity` instead of the expected `IndexError (Expected reduction dim)`; fails identically on CPU and XPU (upstream meta defect).

Trace mode: evidence-only from issue error message (self-sufficient: names `aten.amax` meta and both error types).

| axis | value | reason |
|---|---|---|
[...]

PR #5162, opened 2026-09-01
bot: make the fix pipeline's session comment reliable

ZhaoqiongZ:
Two follow-ups to #5161, which made the `fix` pipeline the sole owner of its issue comment.

### 1. Wrong `gh api` flag (the comment body became a literal `@filename`)

`-f/--raw-field` does **not** interpret a leading `@` as a filename -- only `-F/--field` does. So the PATCH posted the string `@session_comment.md` as the comment body. First run on the new path shows it: https://github.com/intel/torch-xpu-ops/issues/5126#issuecomment-5489603018

Verified with the real binary, using the markdown API as an echo:

```console
$ printf 'hello **world**' > t.md
$ gh api -X POST /markdown -f text=@t.md
<p>@t.md</p>
$ gh api -X POST /markdown -F text=@t.md
<p>hello <strong>world</strong></p>
```

The read half of the recipe was already correct (`--jq .body` prints the raw string).

### 2. Nothing enforced that the comment gets written at all

The prompt asks the pipeline to update the comment; nothing checks that it did. #5079 posted no stage reports at all, and the flag bug above wrote garbage -- both leave the issue showing only `Starting automated fix on this issue`, with the run's outcome reachable only from the job log.

New `Backfill session comment` step: after the agent exits, if the comment carries no `<!-- agent:session -->` marker, overwrite it with the agent's closing summary (`result.result` -- the text the deleted `Post result comment` step used to post).

- The marker is the first thing the pipeline writes, so a comment it filled even partially (triage only, then the job timed out) keeps its stage reports and is left alone.
- Still one comment per run: it edits the id the pipeline was given, never creates a second.
- Runs after `Redact secrets` for the same reason the old step did: `result.result` is agent free text going to a public issue.

Worst case now degrades to the pre-#5161 behaviour (one summary) instead of a bare placeholder.

### Test plan

```bash
# gate + body assembly + size clamp
node -e '
function backfill(bodyNow, resultText){
  if ((bodyNow || "").includes("<!-- agent:session -->")) return null;
  const footer = "\n\n---\n_footer_";
  let body = "<!-- agent:session -->\n\n" + resultText + footer;
  if (body.length > 65000) body = body.slice(0, 64800 - footer.length) + "\n\n_(truncated)_" + footer;
  return body;
}
console.log(backfill("Starting automated fix on this issue.", "S") !== null);  // true
console.log(backfill("@session_comment.md", "S") !== null);                    // true
console.log(backfill("<!-- agent:session -->\n<!-- agent:triage -->", "S"));   // null
console.log(backfill("placeholder", "S".repeat(200000)).length <= 65536);      // true
'

# step order
python3 -c "
import yaml
names = [s.get('name') for s in yaml.safe_load(open('.github/workflows/bot.yml'))['jobs']['fix']['steps']]
assert names.index('Backfill session comment') > names.index('Redact secrets from agent artifacts')
"

lintrunner -m origin/main
```

The `@torchxpubot fix` run in flight on #5126 keeps the old prompt, so its comment stays broken; re-run it after this lands.

Authored with the assistance of an AI coding agent (opencode).

### DIFF
--- a/.claude/skills/issue-handler/references/execution-modes.md
+++ b/.claude/skills/issue-handler/references/execution-modes.md
@@ -105,8 +105,11 @@ gh api "/repos/$OWNER/$REPO/issues/comments/$SESSION_COMMENT_ID" \
   --jq .body > session_comment.md
 cat next_block.md >> session_comment.md
 gh api --method PATCH "/repos/$OWNER/$REPO/issues/comments/$SESSION_COMMENT_ID" \
-  -f body=@session_comment.md
+  -F body=@session_comment.md
 ```
+
+`-F`, not `-f`: only `--field` reads a value from `@file`; `--raw-field`
+would post the literal string `@session_comment.md` as the comment.
 
 Keep the accumulating body under 65000 chars (the GitHub limit is 65536,
 and a PATCH over it fails outright): if the diff in the implement block
--- a/.claude/skills/issue-handler/references/execution-modes.md
+++ b/.claude/skills/issue-handler/references/execution-modes.md
@@ -108,8 +108,8 @@ gh api --method PATCH "/repos/$OWNER/$REPO/issues/comments/$SESSION_COMMENT_ID" 
   -F body=@session_comment.md
 ```
 
-`-F`, not `-f`: only `--field` reads a value from `@file`; `--raw-field`
-would post the literal string `@session_comment.md` as the comment.
+`--raw-field` would post the literal string `@session_comment.md` as the
+comment; only `--field` reads the value from a file.
 
 Keep the accumulating body under 65000 chars (the GitHub limit is 65536,
 and a PATCH over it fails outright): if the diff in the implement block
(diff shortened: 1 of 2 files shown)


.claude/skills/issue-handler/references/execution-modes.md
```diff
@@ -105,9 +105,12 @@ gh api "/repos/$OWNER/$REPO/issues/comments/$SESSION_COMMENT_ID" \
   --jq .body > session_comment.md
 cat next_block.md >> session_comment.md
 gh api --method PATCH "/repos/$OWNER/$REPO/issues/comments/$SESSION_COMMENT_ID" \
-  -f body=@session_comment.md
+  -F body=@session_comment.md
 ```
 
+`-F`, not `-f`: only `--field` reads a value from `@file`; `--raw-field`
```
Stonepia: I think those comments are not needed to explicitly write out, the Agent surely will have the information. I would like to keep every file simpler, it is not for human reading anyway

Stonepia (approved):

.claude/skills/issue-handler/references/execution-modes.md
ZhaoqiongZ: remove the explanation and statement, only keep the intention to use raw comment rather than file name

CuiYifeng (approved):

ZhaoqiongZ:
@torchxpubot merge

PR #5167, opened 2026-09-01
[Workflow] Refine alignment review for unmerged upstream PRs

laifenxiawucha:
## Summary

- judge upstream landing against the collected default-branch snapshot
- distinguish unresolved XPU work from an existing upstream-owned fix
- create ops tracker payloads for confirmed XPU work even when the code change belongs in PyTorch or another GitHub repository
- run generated Python reproducers unbuffered so timeout and signal diagnostics reach the runner log

`implementation_repository` identifies where code should change; publishable `needs-xpu-fix` payloads still create trackers in `intel/torch-xpu-ops`. Observation-only `track-upstream` results do not create payloads or notifications. The runner records the exact `python -I -u` invocation in its artifact.

## Test plan

```text
python -m py_compile .github/scripts/xpu_alignment_gate.py .github/scripts/xpu_alignment_runner.py
python /home/jyz/.codex/skills/.system/skill-creator/scripts/quick_validate.py .claude/skills/xpu-alignment
python <temporary gate and unbuffered-runner fixtures>
python <temporary Markdown reference and YAML check>
lintrunner -a -m origin/main --skip CLANGTIDY,CLANGFORMAT
git diff --check
```

No test file or manual workflow dispatch is included.


ZhaoqiongZ:
@torchxpubot review

laifenxiawucha:
@ZhaoqiongZ done

laifenxiawucha:
@torchxpubot review

laifenxiawucha:
@torchxpubot review

ZhaoqiongZ:
@torchxpubot review

ZhaoqiongZ (approved):

ZhaoqiongZ (approved):
fix the blocking issue

Stonepia (approved):

.github/scripts/xpu_alignment_gate.py
```diff
@@ -20,6 +20,7 @@
 SCHEMA_VERSION = 1
 UNIT_ID_RE = re.compile(r"[A-Za-z0-9][A-Za-z0-9._-]{0,63}")
 SHA256_RE = re.compile(r"[0-9a-f]{64}")
+REPOSITORY_RE = re.compile(r"[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+")
```
Stonepia, line 23: By taking this change, this seems that we only have two logic:

```
if "needs-xpu-fix" : fix in torch-xpu-ops
else: fix in that PR
```

Is this desired?

.github/scripts/xpu_alignment_gate.py
```diff
@@ -477,7 +477,16 @@ def _validate_review(
...
         verdicts[unit_id] = str(verdict)
-        if entry.get("implementation_repository") not in IMPLEMENTATION_REPOSITORIES:
+        repository = entry.get("implementation_repository")
+        if verdict == "needs-xpu-fix":
+            repository_valid = repository == "intel/torch-xpu-ops"
+        elif verdict == "track-upstream":
+            repository_valid = isinstance(repository, str) and bool(
+                REPOSITORY_RE.fullmatch(repository)
+            )
+        else:
```
Stonepia: This else seems never been triggered

Stonepia (commented):
Overall LGTM, just have some concern about the triggering logic

.github/scripts/xpu_alignment_gate.py
Stonepia: @copilot debate with my questions. Is this concern valid?

chuanqi129 (commented):
@laifenxiawucha  @ZhaoqiongZ Some late comments for the 1st PR https://github.com/intel/torch-xpu-ops/pull/5003

1. What's the `@torchxpubot file` in `bot.yml`? Why do we need it?
2. For the alignment workflow, we should track the total cost for each launch in the issue https://github.com/intel/torch-xpu-ops/issues/4251 also.
3. Why the job `scan-prepare` need run on `xpu-agent` runner?
4. What's the different between pr-review skill and review role in xpu-alignment skill?
5. Suggest moving the `ALIGNMENT_TRIAGE_NOTIFY` from workflow into the issue https://github.com/intel/torch-xpu-ops/issues/5018 body, to read those ID from the issue or convert it as a workflow secret. After that, those IDs update don't need a PR for it.
6.  Suggest using same `torchxpubot` account to comment on issue https://github.com/intel/torch-xpu-ops/issues/5018. And add job run hyperlink in the comment directly rather than a workflow run id.
7. Is there any issue created by the xpu-alignment workflow directly?

ZhaoqiongZ:
What's the @torchxpubot file in bot.yml? Why do we need it?
if xpu alignment issue >1 in a day, we would like to review it before create the issue directly, so add a file to create a issue after review & confirmed

For the alignment workflow, we should track the total cost for each launch in the issue https://github.com/intel/torch-xpu-ops/issues/4251 also.
Hi @laifenxiawucha  please do so and add model routing feature to save costs

Why the job scan-prepare need run on xpu-agent runner?

there is reproduce step need to run on a xpu machine

What's the different between pr-review skill and review role in xpu-alignment skill?
Hi @laifenxiawucha , please help clarify this

Suggest moving the ALIGNMENT_TRIAGE_NOTIFY from workflow into the issue https://github.com/intel/torch-xpu-ops/issues/5018 body, to read those ID from the issue or convert it as a workflow secret. After that, those IDs update don't need a PR for it.
Suggest using same torchxpubot account to comment on issue https://github.com/intel/torch-xpu-ops/issues/5018. And add job run hyperlink in the comment directly rather than a workflow run id.

Is there any issue created by the xpu-alignment workflow directly?
not yet for a single issue, during the scan in past days,  the issue filtere > 1  so only draft summary issue been created, Hi @laifenxiawucha please paste the summary / issues here

laifenxiawucha:
I’ll take care of items 2, 5, and 6:
Track the total cost for each alignment workflow run in #4251 and add model routing to reduce the cost.
Move ALIGNMENT_TRIAGE_NOTIFY out of the workflow, so updating those IDs will not require a PR.
Use the same torchxpubot account to comment on #5018, and include the job run hyperlink directly instead of only the workflow run ID.

AND `pr-review` reviews the code diff of a specific PR.
The `review` role in the xpu-alignment skill reviews the evidence from the scan/runner stage to determine XPU ownership, check whether a tracker already exists, and decide whether an issue payload can be generated.
@ZhaoqiongZ @chuanqi129

chuanqi129:
Thanks @ZhaoqiongZ @laifenxiawucha 

> Why the job scan-prepare need run on xpu-agent runner?
> there is reproduce step need to run on a xpu machine

Does it mean that besides the dedicated job `run-reproducers` we still need reproduce in `scan-prepare`?

laifenxiawucha:
> Thanks @ZhaoqiongZ @laifenxiawucha
> 
> > Why the job scan-prepare need run on xpu-agent runner?
> > there is reproduce step need to run on a xpu machine
> 
> Does it mean that besides the dedicated job `run-reproducers` we still need reproduce in `scan-prepare`?

According to the current rules, the `prepare` step does not require XPU. If resource usage is a concern, I can try moving it to a non-XPU runner.@chuanqi129 @ZhaoqiongZ

.github/scripts/xpu_alignment_gate.py
laifenxiawucha: DONE

.github/scripts/xpu_alignment_gate.py
laifenxiawucha: done

.claude/skills/xpu-alignment/references/automation-contract.md
```diff
@@ -289,7 +290,7 @@ artifacts. It does not execute code or sample rejected inventory. It covers ever
   "units": [{
     "id": "issue-123",
     "verdict": "needs-xpu-fix",
-    "implementation_repository": "intel/torch-xpu-ops",
+    "implementation_repository": "pytorch/pytorch",
```
Stonepia: Why have this change? Even this is just a sample, I would prefer to write it to torch-xpu-ops.

Stonepia (approved):

chuanqi129:
> According to the current rules, the `prepare` step does not require XPU. If resource usage is a concern, I can try moving it to a non-XPU runner.@chuanqi129 @ZhaoqiongZ

@laifenxiawucha let's move it first until we have real needs on xpu runner at this step

.claude/skills/xpu-alignment/references/automation-contract.md
laifenxiawucha: done

chuanqi129:
> if xpu alignment issue >1 in a day, we would like to review it before create the issue directly, so add a file to create a issue after review & confirmed

From my perspective, the `file` cmd is not necessary and also not reasonable to add one bot cmd which just can use it this specific issue. My suggestion is free the alignment job to open issue by itself, we can review the issue directly, if the issue quality is not good, we can close with comment. With that, the alignment skill can evolution with those bad cases. cc: @ZhaoqiongZ @laifenxiawucha

ZhaoqiongZ:
> > if xpu alignment issue >1 in a day, we would like to review it before create the issue directly, so add a file to create a issue after review & confirmed
> 
> From my perspective, the `file` cmd is not necessary and also not reasonable to add one bot cmd which just can use it this specific issue. My suggestion is free the alignment job to open issue by itself, we can review the issue directly, if the issue quality is not good, we can close with comment. With that, the alignment skill can evolution with those bad cases. cc: @ZhaoqiongZ @laifenxiawucha

 Hi @laifenxiawucha ， could you update with chuanqi's suggestion, but still need to block if issues > 3, which is obvious not normal, and need human review, at that time, we can create issue with local agent rather than add a bot file command

laifenxiawucha:
Okay, I’ll remove this **file** instruction later and set it so that up to three issues can be submitted automatically. If there are more than three, they will require manual review. 
@ZhaoqiongZ @chuanqi129

laifenxiawucha:
@torchxpubot merge

PR #5175, opened 2026-09-01
build: expose USE_SYCLTLA as an option

dvrogozh:
For flexibility, add USE_SYCLTLA option which allows to disable the SYCL-TLA based kernels for experiments and somewhat speedup the build.

For the work to enable upstream SYCL compilers this option becomes essential as it allows to reduce the code volume which needs to be taken into consideration on the initial enabling effort.

CC: @guangyey, @chunhuanMeng

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,13 +65,13 @@ if(USE_XCCL)
   endif()
 endif()
 
-set(USE_SYCLTLA ON)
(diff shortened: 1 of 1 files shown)


guangyey (approved):

LuFinch (approved):

chunhuanMeng (approved):

dvrogozh:
@torchxpubot merge -f

PR #5184, opened 2026-09-02
Add vectorized channels-last avg_pool2d forward kernel (VEC/WTILE/HWTILE) on XPU

yucai-pro:
## Summary

The current channels-last `avg_pool2d` forward kernel on XPU processes one
output element per work-item with a naive scalar gather loop, achieving only
~14-45% of peak HBM bandwidth across common shapes. This PR adds a
vectorized dispatch path with three kernel variants:

- **VEC**: each work-item loads `vec_size` (8/4/2) channels at once via
  `memory::aligned_vector`, with unrolled fast paths for k=3s1 and k=2s2
  interior pixels.
- **WTILE**: for k=3,s=1 shapes, each work-item computes 2 adjacent output
  columns at once, sharing the 2 overlapping input columns between them
  (33% fewer loads than two independent VEC work-items).
- **HWTILE**: variant of WTILE used when `vec_size==2`, computing a 2x2
  block of adjacent outputs per work-item (4 loads/output vs 9 in the
  scalar path).

`vec_size` is chosen at launch time from pointer alignment and an
occupancy floor (>=8192 vectorized work-items), with a bias toward `vec4`
over `vec8` when doing so lets WTILE activate. WTILE/HWTILE activate only
for k=3,s=1 shapes where the halved work-item count still clears a 65536
occupancy threshold; otherwise the kernel falls back to plain VEC, and to
the original scalar `AvgPool2dChannelsLastKernelFunctor` when no vec_size
divides the channel count.

Additionally, the scalar channels-last fallback kernel's index
decomposition (`% pooled_width` / `% channels` chains) is replaced with
`at::detail::IntDivider` magic-number division, avoiding per-work-item
hardware DIV/MOD instructions.

NCHW (contiguous) forward, and both NCHW/channels-last backward kernels
are unchanged.

## Benchmark

100 shapes covering k=1/2/3, stride=1/2, various N/C/H/W, fp16,
channels_last, XPU event timing with an L3-cache flush before each timed
iteration (20 warmup + 200 timed runs),
`BW% = (input_bytes + output_bytes) / (time * 456 GB/s peak)`.

| metric | value |
|---|---|
| geomean speedup | 2.86x |
| mean speedup | 3.06x |
| median speedup | 2.62x |
| min / max speedup | 1.01x / 5.63x |
| regressions (speedup < 1.0) | 0 / 100 |

| N | C | H | W | k | s | baseline (us, BW%) | optimized (us, BW%) | speedup |
|---|---|---|---|---|---|---|---|---|
| 1024 | 256 | 35 | 35 | 3 | 1 | 20295.24, 13.9% | 3603.47, 78.2% | 5.63x |
| 1024 | 26 | 56 | 56 | 3 | 1 | 4777.08, 15.3% | 1346.72, 54.4% | 3.55x |
| 128 | 112 | 14 | 14 | 3 | 2 | 53.57, 28.8% | 52.85, 29.1% | 1.01x |

Full 100-shape before/after table available on request.

## Testing

Custom correctness script against CPU fp32 reference across shapes
exercising every dispatch branch (vec8/vec4/vec2, WTILE(8/4/1), HWTILE(2),
scalar fallback for non-divisible channel counts): max abs error 0.0017,
consistent with fp16 rounding.

pytorch/pytorch's own avg_pool2d channels_last forward op tests
(`test/nn/test_pooling.py`), run against this branch:
```
python test/nn/test_pooling.py -v \
  TestPoolingNNDeviceXPU.test_avg_pool2d_nhwc_xpu_float16 \
  TestPoolingNNDeviceXPU.test_avg_pool2d_nhwc_xpu_float32 \
  TestPoolingNNDeviceXPU.test_avg_pool2d_nhwc_xpu_float64 \
  TestAvgPoolDeviceXPU.test_avg_pool2d_ceil_mode_xpu
```
Result: 4 passed.

---
*This PR was prepared with AI assistance (GitHub Copilot).*

### DIFF
--- a/src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
+++ b/src/ATen/native/xpu/sycl/AveragePool2dKernels.cpp
@@ -19,7 +19,9 @@
 #include <ATen/native/Pool.h>
 
 #include <ATen/native/xpu/sycl/AveragePool2dKernels.h>
+#include <ATen/native/xpu/sycl/IntegerDivider.h>
 #include <ATen/native/xpu/sycl/KernelUtils.h>
+#include <ATen/native/xpu/sycl/MemoryAccess.h>
 #include <comm/Runtime.h>
 #include <comm/SYCLContext.h>
 #include <comm/SYCLHelpers.h>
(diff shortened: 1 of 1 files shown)


yucai-pro:
Closing to recreate under a better-named branch (this is a perf change, not a fix, and only touches the forward kernel).

PR #5186, opened 2026-09-02
Use upstream `test_autocast.py` file instead of local copy.

BBBela:
Upstream `test/test_autocast.py` test file has been enabled for xpu in https://github.com/pytorch/pytorch/pull/193189, so we can start using the upstream version directly instead of local copy.

No test coverage is lost!


dominikjastrzebskix (approved):

kdrozd-dev (approved):

BBBela:
@torchxpubot merge

PR #5195, opened 2026-09-03
Corrected fallback for linalg cholesky

mwiktor-intel:
In order to create MKL call for torch xpu ops, the correct list of operators to fallback are needed. This is a temporary step, needed to register op dispatch in upstream.


mwiktor-intel:
Apart from XPUFallback.cpp entries,  an explicit temporary fallback for linalg_cholesky was added. It is necessary to avoid runtime errors when the XPU backend is added to native_functions.yaml (PR https://github.com/pytorch/pytorch/pull/195820). Now, it works correctly both with existing and future entries in native_function.yaml

src/ATen/native/xpu/BatchLinearAlgebra.cpp
```diff
@@ -179,4 +179,22 @@ Tensor& orgqr_kernel_xpu(Tensor& result, const Tensor& tau) {
 
 REGISTER_XPU_DISPATCH(orgqr_stub, &orgqr_kernel_xpu);
 
+void cholesky_kernel_fallback(const Tensor& input, const Tensor& info, bool upper) {
```
Silv3S: To pass the linter check
```suggestion
void cholesky_kernel_fallback(
    const Tensor& input,
    const Tensor& info,
    bool upper) {
```

PawelSwider2000 (approved):

kdrozd-dev (approved):

kdrozd-dev:
@torchxpubot merge -f

PR #5199, opened 2026-09-03
Enable Clang as the XPU host compiler

cyyever:
Changes:
* Widen compiler checks in `cmake/BuildFlags.cmake`, the `USE_SYCLTLA` gate in `CMakeLists.txt`, and `test/sycl/CMakeLists.txt` to accept `Clang` alongside `GNU` (SYCL-TLA requires Clang >= 19.0, per upstream [sycl-tla](https://github.com/intel/sycl-tla))
* Fix Clang 22 `-Werror` diagnostics and PyTorch API compatibility issues surfaced by an end-to-end build with Clang as host compiler (`MemoryAccess.h`, `TensorTransformationsKernels.{cpp,h}`, `SparseCsrTensorMath.cpp`, `NestedTensorTransformerFunctions.cpp`, `XPUSymmetricMemory.hpp`)

Verified with:
* Clang 22.1.8 (host compiler) + Intel oneAPI DPC++/C++ 2026.1.0 (SYCL compiler) on Arch Linux, kernel 7.2.2-arch1-1
* On https://github.com/cyyever/torch-xpu-ops/commit/b125031e605266809660eb51202d15a901c29543

```
CC=clang CXX=clang++ pip install -e . -v --no-build-isolation
```

Assisted-by: GitHub Copilot


dvrogozh:
Thank you for your contribution. Couple question:
1. Have you verified that the Pytorch XPU can be fully built with the Clang as a host compiler? 
2. Where requirement to build Pytorch XPU with Clang on Linux is coming from? Is there any actual project which is going to exercise such a build? This will help to prioritize the request and CI coverage for such build type.

cyyever:
I am building XPU with clang without any issues. The intention is to find more bugs by untested compilers.

guangyey (commented):
Currently, we don't have clang CI for this code change. @dvrogozh could you please help check if the code change makes sense.

dvrogozh (changes requested):
1. This PR is missing to extend the following to Clang:
https://github.com/intel/torch-xpu-ops/blob/5cdd4e46bd058506f1620944b27b87056e415cc9/test/sycl/CMakeLists.txt#L36

Without this change build fails with:
```
  /usr/bin/ld: test_sycl/CMakeFiles/test_sycl_build_archive.dir/main.cpp.o: in function `test_simple_kernel()':
  main.cpp:(.text+0x11): undefined reference to `itoa(float*, int)'
```

2. I also have a concern that SYCL-TLA is not included in the support. I believe it need to. Please, modify the respective condition accordingly:
https://github.com/intel/torch-xpu-ops/blob/5cdd4e46bd058506f1620944b27b87056e415cc9/CMakeLists.txt#L69

3. In respect to the PR description. It will be used as a commit message. Thus, please, clean it up to be more focused:
   1. Drop link to claude.ai session. Not everyone can access it.
   2. Replace "Generated with [Claude Code](https://claude.com/claude-code)" by `Assisted-by: Claude Code`
   3. Replace summary with the shorter message. Just `Widen compiler checks to support both GNU and Clang` should be enough
   4. Provide details on what you've actually tried and what were prerequisites.

Overall, for the PR description I expect to see something like this:
```
Changes:
* Widen compiler checks to support both `GNU` and `Clang`

Verified with:
* Clang 20 on Ubuntu 24.04
* On https://github.com/pytorch/pytorch/commit/45c1f8424f91ccef1ba32649d70f723b1cf3f0d5
* On https://github.com/intel/torch-xpu-ops/commit/5cdd4e46bd058506f1620944b27b87056e415cc9

apt-get install clang-20 libomp-20-dev
CC=clang-20 CXX=clang++-20 python -m pip install --no-build-isolation -v -e .

Assisted-by: Claude Code
```

cmake/BuildFlags.cmake
```diff
@@ -70,12 +70,20 @@ macro(set_build_flags)
...
-  elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
     list(APPEND SYCL_HOST_FLAGS -fPIC)
     list(APPEND SYCL_HOST_FLAGS -std=${CPP_STD})
     # Excluding warnings which flood the compilation output
     # TODO: fix warnings in the source code and then reenable them in compilation
     list(APPEND SYCL_HOST_FLAGS -Wno-sign-compare)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+      # PyTorch enables pedantic warnings as errors for XPU targets.
+      list(APPEND SYCL_HOST_FLAGS
```
dvrogozh: You are fixing at least some of these warning in this PR. Are these `-Wno-*` still needed? Can you do one of 2 things, please:
1. Either really fix all warnings and drop these `-Wno-*` arguments entirely for Clang
2. Or leave these arguments, revert C++ source file changes and defer warning fixes to the separate PR

src/xccl/XPUSymmetricMemory.hpp
```diff
@@ -73,8 +73,8 @@ class XPUSymmetricMemory : public SymmetricMemory {
   std::vector<c10::intrusive_ptr<AllocationRef>> alloc_refs_;
   std::vector<void*> buffers_;
   std::vector<void*> signal_pads_;
-  HandleType mc_handle_;
-  void* mc_addr_;
+  [[maybe_unused]] HandleType mc_handle_;
```
dvrogozh: These class variables seem actually unused except in constructor when they are initialized. Please:
1. Either fix constructors and constructor invocations by dropping unused variables
2. Or add in-source comment explaining why these variables are still needed. For example, if constructor specification is externally managed, then it might mandate having these variables passed to the constructor

cyyever: fix them

cmake/BuildFlags.cmake
cyyever: fix all warnings and drop these -Wno-* arguments

dvrogozh (commented):
Please, rebase as the PR got conflict with the recently merged https://github.com/intel/torch-xpu-ops/commit/6254016f01c9cb434b1106eb7758adda8ebb24f1. Note that one more compiler check was added which you might need to generalize:
https://github.com/intel/torch-xpu-ops/blob/6254016f01c9cb434b1106eb7758adda8ebb24f1/cmake/BuildFlags.cmake#L222-L224

src/ATen/native/xpu/sycl/TensorTransformationsKernels.cpp
```diff
@@ -135,10 +135,7 @@ void flip_kernel_impl(TensorIterator& iter) {
   launch_kernel(iter.numel(), loop);
 }
 
-void flip_kernel(TensorIterator& iter, bool quantized) {
-  if (quantized) {
-    TORCH_CHECK(false, "XPU current does not flip for quantized tensor");
-  }
+void flip_kernel(TensorIterator& iter) {
```
dvrogozh, line 138: You've removed the  argument of this function, but I don't see any modifications to its invocation. Can you, please, check why?

src/ATen/native/xpu/sycl/MemoryAccess.h
```diff
@@ -340,14 +340,9 @@ inline int can_vectorize_up_to(array_t pointers) {
 
 template <typename T>
 struct alignas(sizeof(T) * 4) m_vec4 {
-  union {
-    T val[4];
-    struct {
-      T x, y, z, w;
-    };
-  };
+  T val[4];
```
dvrogozh, line 343: The dropped union/struct construction is typical. What was the warning here? can it be fixed in other way without dropping the whole construct? My worry is that even if we don't use `x, y,, z, w` in the release code, we might be using these in debug code or debug print macro definitions. If so this change might regress our debug path. We at least need to try grep or ask AI to check the usage.

src/ATen/native/xpu/sycl/TensorTransformationsKernels.cpp
cyyever: These changes are wrong because it was compiled for a customized pytorch without quantized tensors.

src/ATen/native/xpu/sycl/MemoryAccess.h
cyyever: The warnings were
```
error: anonymous structs are a GNU extension [-Werror,-Wgnu-anonymous-struct]
warning: anonymous types declared in an anonymous union are an extension [-Wnested-anon-types]
```
I had encountered no error in release building for referencing ``x,y,z,w``.  I have verified the debug build.

dvrogozh (changes requested):
Build with clang-20 fails with:
```
icx: error: invalid Xarch argument: '-Xarch_host -Qunused-arguments', not all driver options can be forwared via Xarch argument
```

I think `-Qunused-arguments` option is coming from Pytorch top level [CMakeLists.txt#L1336](https://github.com/pytorch/pytorch/blob/8be2400cf5cd92040286a9605608759a89009c99/CMakeLists.txt#L1336). It is conditionally added if supported and it's not supported by GCC, but supported by CLANG. After https://github.com/intel/torch-xpu-ops/commit/6254016f01c9cb434b1106eb7758adda8ebb24f1 it is passed via `-Xarch_host` which is inappropriate way to pass the option per error above.

@chunhuanMeng , what is our reason to use `'-Xarch_host` at the first place? Why we can't just pass arguments directly to the ICPX compiler?

chunhuanMeng:
> Build with clang-20 fails with:
> 
> ```
> icx: error: invalid Xarch argument: '-Xarch_host -Qunused-arguments', not all driver options can be forwared via Xarch argument
> ```
> 
> I think `-Qunused-arguments` option is coming from Pytorch top level [CMakeLists.txt#L1336](https://github.com/pytorch/pytorch/blob/8be2400cf5cd92040286a9605608759a89009c99/CMakeLists.txt#L1336). It is conditionally added if supported and it's not supported by GCC, but supported by CLANG. After [6254016](https://github.com/intel/torch-xpu-ops/commit/6254016f01c9cb434b1106eb7758adda8ebb24f1) it is passed via `-Xarch_host` which is inappropriate way to pass the option per error above.
> 
> @chunhuanMeng , what is our reason to use `'-Xarch_host` at the first place? Why we can't just pass arguments directly to the ICPX compiler?

 `-Xarch_host` was introduced to separate host-only flags from device/common SYCL flags. A pure `icpx -fsycl` invocation drives both host and device compilation, so forwarding all `CMAKE_CXX_FLAGS` directly to `icpx` would apply host-specific options to the device compilation path as well. This is particularly problematic for flags such as host Debug `-O0 -g`, where we intentionally want different host and device configurations.

 In this respect, `-Xarch_host` serves a purpose similar to `-Xcompiler`: both route options to the host compilation path instead of applying them to device compilation. It is not intended to wrap every flag inherited from PyTorch unconditionally.

 `-Qunused-arguments` is the exception here. Although `icx` is Clang-based, `-Qunused-arguments` is a Clang/**icx driver** option, consumed by the outer `icx` driver before it dispatches host and device compilation. It is not a valid option to forward through `-Xarch_host` (or `-Xcompiler`) to the internal host compiler invocation. 
 
 We may need to handle this flag specially when Clang is enabled, since it is a driver-level option rather than a host compiler option.

cyyever:
@dvrogozh fixed

dvrogozh (approved):
LGTM

@guangyey, @chunhuanMeng, I don't have any more comments from my side. Can you, please, review as well?

chunhuanMeng (approved):

guangyey (approved):

dvrogozh:
@torchxpubot merge -f "CI failures are unrelated"

dvrogozh:
> Failed to merge PR: Server Error

Actually that's not true. PR got merged, see https://github.com/intel/torch-xpu-ops/commit/de6c911047543ae40bb17d266cc0a32162fda787. The PR was not closed however, I will do that manually. @mengfei25, @chuanqi129 : fyi.

PR #5217, opened 2026-09-04
Fix H2D metadata race in nested _transform_bias_rescale_qkv on XPU

jmamzax:
Issue: https://github.com/intel/torch-xpu-ops/issues/4138
The nested-tensor fast path of _transform_bias_rescale_qkv builds a small host tensor `offsets` holding per-batch buffer offsets and flattened input sizes, then copies it to the device and hands the raw device pointers to the kernel. The copy was issued with non_blocking=true:

    auto metadata = offsets.to(at::Device(kXPU), at::kInt, true, true);

`offsets` is a pageable host-local that is destroyed when the function returns. With a non-blocking copy the host does not wait, so the source buffer can be freed (and its memory reused) before the asynchronous H2D transfer actually runs. The kernel then reads corrupted offsets/sizes and routes work-groups to the wrong batch's data, producing numerically wrong results for valid tokens.

This pattern was inherited from the CUDA implementation (aten/src/ATen/native/transformers/cuda/attention.cu), where it is benign because the CUDA runtime performs a non-blocking H2D copy from pageable memory synchronously. On XPU the copy can genuinely run asynchronously, so the race is real. It is timing dependent: on a discrete Arc B580 it happened to always win the race, but on the Lunar Lake integrated GPU (Arc 140V, shared system memory) it reliably lost it.

The fix makes the metadata copy blocking (non_blocking=false), so the host buffer stays alive until the transfer completes.

Performance:

The metadata is tiny (2 * B + 1 ints), and the kernel that consumes it runs immediately after on the same queue, so the copy was already effectively serialized with the kernel and the non-blocking form bought no overlap. Measured on the Arc 140V, the blocking-vs-non-blocking delta for this copy is within noise (about +/-0.5 us), against a transform op of ~255-481 us and a full TransformerEncoder fast path of ~2.8-8.3 ms. Net impact is negligible.

Test Plan:

Repro/isolation and the two originally-failing tests were run on the Arc 140V (LNL) machine. Before the fix the isolated transform produced wrong, non-deterministic results (e.g. 48/50 in-process runs wrong for batch=4); after the fix all pass.

```
python test_transformers_xpu.py \
  TestTransformersXPU.test_transformerencoder_fastpath_use_torchscript_False_enable_nested_tensor_True_use_autocast_False_d_model_12_xpu \
  TestTransformersXPU.test_transformerencoder_fastpath_use_torchscript_False_enable_nested_tensor_True_use_autocast_False_d_model_256_xpu
# Ran 2 tests OK
```

This change was authored with the assistance of an AI coding assistant.


src/ATen/native/transformers/sycl/AttentionKernels.cpp
```diff
@@ -438,7 +438,9 @@ void _transform_bias_rescale_qkv_kernel(
 
           native::narrow_symint(offsets, 0, sizes.numel() + 1, sizes.numel())
               .copy_(sizes.reshape({-1}));
-          auto metadata = offsets.to(at::Device(kXPU), at::kInt, true, true);
+          // Blocking H2D: `offsets` is a pageable host-local that frees on
+          // return, so a non-blocking copy can read freed memory on XPU.
+          auto metadata = offsets.to(at::Device(kXPU), at::kInt, false, true);
```
pbielak: WDYT about adding inline argument name comments, so the `false`, `true` values are more self-explanatory?

e.g., `.to(/*device=*/at::Device(kXPU), /*dtype=*/at::kInt, /*arg_name=*/false, /*arg_name2=*/true);`

In the current version, I'm not able to tell what the purpose of the `false/true` values is

jmamzax: Updated.

pbielak (approved):

BBBela (approved):
Looks good to me, thanks 😉

PR #5226, opened 2026-09-07
[xccl] Mark isInitialized as override

Chao1Han:
pytorch/pytorch made Backend::isInitialized virtual so that split() callers can check whether a lazily initialized backend already has its parent communicator.


zxd1997066 (approved):

mengfei25 (approved):

Chao1Han:
@torchxpubot merge -f "build passed"

Issue #5221, opened 2026-09-07
[Inductor] test_var_mean_uses_two_step_for_non_split_reductions fails on XPU: two-step variance heuristic is gated on device.type == "cuda"


daisyden:
### 🐛 Describe the bug

`test/inductor/test_cooperative_reductions.py::TestVarianceReductionHeuristic::test_var_mean_uses_two_step_for_non_split_reductions`
fails on XPU because Inductor's two-step variance heuristic is gated on
`device.type == "cuda"`, so XPU always falls through to Welford.

#### Root cause

`torch/_inductor/lowering.py::use_two_step_variance` selects the Triton
two-step (sum-of-squares) path only for CUDA:

```python
    if device and device.type == "cpu":
        threshold = config.cpp.use_two_step_variance_threshold
    elif device and device.type == "cuda" and is_triton(x) and is_cuda_two_step_dtype:
        min_numel = config.triton.use_two_step_variance_min_numel
        threshold = config.triton.use_two_step_variance_threshold
        check_for_split = True
    else:
        threshold = config.unroll_reductions_threshold
```

On XPU the `else` branch is taken, so `threshold` becomes
`config.unroll_reductions_threshold` (8). A 4096-element reduction is far above
that, `use_two_step_variance` returns `False`, and Inductor emits
`triton_helpers.welford_reduce` instead of `tl.sum`. The heuristic was added for
CUDA only in pytorch/pytorch#184383 ("[inductor] Use two-step variance for small
CUDA reductions"); XPU uses the same Triton codegen path and the same
`config.triton.*` knobs, so the gate is too narrow rather than intentionally
CUDA-specific.

Note the three sibling tests in the same class (`..._keeps_welford_for_float32_reductions`,
`..._keeps_welford_for_small_reductions`, `..._keeps_welford_for_split_reductions`)
currently pass on XPU only vacuously: they assert Welford *is* used, which is
always true on XPU since the two-step path is unreachable.

#### Reproducer

```bash
python test/inductor/test_cooperative_reductions.py \
  TestVarianceReductionHeuristic.test_var_mean_uses_two_step_for_non_split_reductions
```

(The test currently skips unless `GPU_TYPE` is `cuda`; the XPU run above is with
the skip relaxed to `GPU_TYPE not in ["cuda", "xpu"]`.)

Standalone:

```python
import torch
from torch._inductor.utils import run_and_get_code

def fn(x):
    return torch.var_mean(x, dim=-1, correction=0)

[...]

Issue #4732, opened 2026-07-29
[XPU] torchrun --numa-binding fails on XPU: _XpuDeviceProperties missing pci_domain_id/pci_bus_id/pci_device_id


astachowiczhabana:
### 🐛 Describe the bug

`torch/numa/binding.py` was made device-generic in #185266 (`torch.cuda` ->
`torch.accelerator`), but XPU doesn't satisfy the implicit backend contract, so
`--numa-binding` is unusable on Intel GPUs.

`_get_numa_node_index_for_device_index` (`torch/numa/binding.py:595-604`) reads
three PCI attributes to find the device's NUMA node via sysfs:

```python
domain = device_properties.pci_domain_id
bus    = device_properties.pci_bus_id
device = device_properties.pci_device_id
pci_addr = f"{domain:04x}:{bus:02x}:{device:02x}.0"
with open(f"/sys/bus/pci/devices/{pci_addr}/numa_node") as f: ...
```

CUDA binds them at `torch/csrc/cuda/Module.cpp:1201-1203`. XPU exposes no `pci_*`
attribute at all (full list: `torch/csrc/xpu/Module.cpp:326-360`), and
`grep -rn "pci" torch/csrc/xpu/ torch/xpu/ c10/xpu/` returns zero hits. The data is
missing one layer down too — `AT_FORALL_XPU_EXT_DEVICE_PROPERTIES`
(`c10/xpu/XPUDeviceProp.h:116-139`) never queries a PCI address — so *
need a change**, not just the pybind11 binding.

### Impact

`torchrun --numa-binding=<mode>` **fails to launch** with `AttributeError`:
`torch/distributed/run.py:957-960` doesn't set
`should_fall_back_if_binding_fails`, which defaults to `False`
(`binding.py:95`), so `_handle_exception` re-raises (`binding.py:266`)
fallback enabled it instead runs unbound and loses the locality win. The upstream
default (`numa_options=None`) is unaffected.

### Secondary issue: no test enforces the contract

`test/test_numa_binding.py` passes on XPU but gives no real XPU covera
`torch.get_device_module` to a `MagicMock` returning a hand-rolled
`MockDeviceProperties` that declares the three `pci_*` fields (lines 2
It would pass with no accelerator present. That mocking is right for its purpose
(fabricating 4-socket / 8-NUMA-node topologies no CI runner has), so t
missing test at a different layer. The contract exists only in #185266's
description:

[...]

Issue #3530, opened 2026-04-29
[Bug Skip] XPU index_add_ numerical accuracy: 3 test_index_add_correctness_2d_* cases (bf16, fp32+alpha)


daisyden:
## Bug Description

`torch.Tensor.index_add_` (and the out-of-place equivalent) on XPU produces
numerically inaccurate results when used with `bfloat16` and with `float32 +
non-default alpha`, exceeding the tolerances upstream's `test_indexing.py`
applies. Three cases under `TestIndexingXPU` fail:

## Affected Tests

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_indexing_xpu.TestIndexingXPU,test_index_add_correctness_2d_alpha_xpu
op_ut,third_party.torch-xpu-ops.test.xpu.test_indexing_xpu.TestIndexingXPU,test_index_add_correctness_2d_large_xpu_bfloat16
op_ut,third_party.torch-xpu-ops.test.xpu.test_indexing_xpu.TestIndexingXPU,test_index_add_correctness_2d_out_of_place_xpu_bfloat16

### `test_index_add_correctness_2d_alpha_xpu` (float32 + alpha)

```
AssertionError: Tensor-likes are not close!
Mismatched elements: 37 / 3200 (1.2%)
Greatest absolute difference: 9.012222290039062e-05 at index (58, 13)
Greatest relative difference: 0.0005161699373275042 at index (58, 13)
File ".../test/test_indexing.py:2218"
    self.assertEqual(dst.cpu(), dst_ref, atol=1e-5, rtol=1e-5)
```

### `test_index_add_correctness_2d_large_xpu_bfloat16`

```
AssertionError: 0.016779609101716046 not less than 0.01 :
    index_add_ torch.bfloat16 L2 rel err too high: 0.016779609101716046
File ".../test/test_indexing.py:2158"
```

### `test_index_add_correctness_2d_out_of_place_xpu_bfloat16`

```
AssertionError: Tensor-likes are not close!
Mismatched elements: 52 / 6400 (0.8%)
Greatest absolute difference: 0.28363037109375 at index (38, 10)
Greatest relative difference: 0.4216286838054657 at index (177, 27)
File ".../test/test_indexing.py:2198"
    self.assertEqual(out_func.cpu().float(), ref, atol=atol, rtol=rtol)
```

## Reproduction

```bash
cd $HOME/daisy_pytorch/third_party/torch-xpu-ops/test/xpu
[...]

PR #5227, opened 2026-09-07
[bot] Keep the fix pipeline from stopping mid-run, and fail the job when it does

ZhaoqiongZ:
## Problem

Three of 19 `@torchxpubot fix` runs on 2026-09-07 finished all 26 steps green without completing the pipeline:

| Issue | Where it stopped | Cost |
|---|---|---|
| #3530 | ended the turn when the nested `issue-triage` skill returned, stage 1 of 6 | $0.50 |
| #5221 | started a `run_in_background` build, reported progress, ended the turn | $5.96 |
| #4732 | same, fix committed and left at `verdict=PENDING_VERIFY` | $12.99 |

All three exit with `stop_reason=end_turn`, `is_error=false` and a non-zero cost, which is what the verify gate accepted:

```js
outcome === 'success' && !result.is_error && result.total_cost_usd > 0
```

## Fix

Two changes in `bot.yml`.

**A `Stop` hook on the fix job.** The Stop event carries `last_assistant_message` and `background_tasks`, so the hook can refuse the stop until the run reports a final outcome with nothing left in flight. Blocking resumes the same session, so #4732's build does not start over. Claude Code caps consecutive blocks at 8, so a stuck run still ends.

**The verify step requires the same.** It matches the closing message for an outcome token. `fix_result*.json` cannot serve as the signal: a legitimate `SKIPPED` run writes none, while an abandoned one writes it at `PENDING_VERIFY`.

## Test

Pulled the transcript artifacts for all 20 runs in that batch, plus #3530's original $0.50 drop-out, and matched the pattern against each run's actual `result.result`:

- 18/18 completed runs match
- 3/3 drop-outs do not, and none of them contains any token from the list

No structured field separates the two groups. `stop_reason`, `terminal_reason` and `subtype` are identical across all 21 runs, and `num_turns` does not help either (#4732 dropped out after 138 turns). The issue-body `<!-- agent:status -->` marker is absent on all 20 issues, which are human-filed rather than created from the agent template.

Treating every intermediate assistant message as a hypothetical stopping point, 17 of 277 (6.1%) would still be let through. So the verify step is a backstop at roughly 94%. The Stop hook is the primary defense, and it reads `background_tasks` rather than vocabulary.

The hook itself only proves out on a live Stop event. #5221 and #4732 are the natural first targets, being known to stop at the build.

### DIFF
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -1263,7 +1263,7 @@ jobs:
                 "Stop": [{
                   "hooks": [{
                     "type": "prompt",
-                    "prompt": "A headless /issue-handler pipeline is trying to end its turn: $ARGUMENTS. Nothing can resume it, so a turn that ends mid-run loses the whole build. ok=false unless BOTH hold: last_assistant_message reports a terminal outcome (IMPLEMENTING(fix_verified) / NEEDS_HUMAN(reason) / SKIPPED(reason), or an equivalent per-sub-item batch report), AND background_tasks is empty. Otherwise reason must tell it to continue in this session from the stage it stopped at -- waiting on a background task means polling that task in the foreground, and a sub-skill's output is never the final answer."
+                    "prompt": "$ARGUMENTS\n\nAllow the stop only if both are true:\n1. last_assistant_message contains an explicit exit status for the whole run -- a stated final outcome, not a progress update or a partial result\n2. background_tasks is empty\n\nIf both: {\"ok\": true}\nOtherwise: {\"ok\": false, \"reason\": \"Not finished. Keep working in this session. If a background task is running, poll it in the foreground.\"}"
                   }]
                 }]
               }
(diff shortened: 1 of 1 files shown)


.github/workflows/bot.yml
```diff
@@ -1256,7 +1256,18 @@ jobs:
...
           github_token: ${{ secrets.MERGE_TOKEN }}
-          settings: '{"alwaysThinkingEnabled": true}'
+          settings: |
+            {
+              "alwaysThinkingEnabled": true,
+              "hooks": {
+                "Stop": [{
+                  "hooks": [{
+                    "type": "prompt",
+                    "prompt": "A headless /issue-handler pipeline is trying to end its turn: $ARGUMENTS. Nothing can resume it, so a turn that ends mid-run loses the whole build. ok=false unless BOTH hold: last_assistant_message reports a terminal outcome (IMPLEMENTING(fix_verified) / NEEDS_HUMAN(reason) / SKIPPED(reason), or an equivalent per-sub-item batch report), AND background_tasks is empty. Otherwise reason must tell it to continue in this session from the stage it stopped at -- waiting on a background task means polling that task in the foreground, and a sub-skill's output is never the final answer."
```
Stonepia: If we really need to inject the prompts, I would suggest that we write it simple and with clear messages and very simple binary output (Like, just success or failure). Otherwise, maybe in some case, the stop loop will also do another round of Agent loop, then making the stop loop still failure.

.github/workflows/bot.yml
```diff
@@ -1424,7 +1435,10 @@ jobs:
                 is_error: result.is_error, total_cost_usd: result.total_cost_usd
               }));
             }
-            if (outcome === 'success' && result && !result.is_error && result.total_cost_usd > 0) {
+            const reportedOutcome = /(?:^|\n)\s*(?:\*\*)?Done\b|\b(IMPLEMENTING\(fix_verified\)|NEEDS_HUMAN|SKIPPED|NOT_REPRODUCED|NO_REPRODUCER|CANNOT_VERIFY|ALREADY_FIXED|STALE_SKIP|FIXED|PASSED|FAILED)\b/
```
Stonepia: Have a little concern on what if there are other uncatched outcomes

CuiYifeng: I also wonder if the capitalization of the output will strictly adhere to assumptions. For example, do we need capitalization insensitive?

.github/workflows/bot.yml
```diff
@@ -1424,7 +1435,10 @@ jobs:
...
+              .test((result && result.result) || '');
```
CuiYifeng: Can we refine the naming of `result.result`?

.github/workflows/bot.yml
ZhaoqiongZ: Fair — cut it to two checks and a fixed `{"ok": true/false}`.

.github/workflows/bot.yml
ZhaoqiongZ: Tested it on that batch's actual output — caught all 3 drop-outs, didn't trip on any of the 18 good runs. It's just a backstop anyway, the `Stop` hook is what really does the work.

ZhaoqiongZ: Added `/i`. Didn't change any verdict, but no point betting on capitalization.

.github/workflows/bot.yml
ZhaoqiongZ: Renamed to `resultEvent`, same as the `triage` job at L1665. Inner `.result` is Claude Code's field, has to stay.

CuiYifeng (approved):

Stonepia (approved):

ZhaoqiongZ:
@torchxpubot merge

ZhaoqiongZ:
@torchxpubot merge

Issue #5027, opened 2026-08-20
test_comprehensive_combinations_xpu_float16 fails with Tensor-likes not close (numeric mismatch)


daisyden:
This issue occur in pytorch CI, but passed in my local PVC. 

### 🐛 Describe the bug

```
`test_comprehensive_combinations_xpu_float16` in `test/inductor/test_torchinductor_opinfo.py`
fails with a numeric mismatch between eager and compiled (inductor) outputs for `torch.combinations`
on XPU with float16 inputs.

Failure log:

FAILED [12.0015s] inductor/test_torchinductor_opinfo.py::TestInductorOpInfoXPU::test_comprehensive_combinations_xpu_float16 - Exception: Tensor-likes are not close!

Mismatched elements: 2 / 4 (50.0%)
Greatest absolute difference: 0.00048828125 at index (2,) (up to 1e-05 allowed)
Greatest relative difference: 0.0012941360473632812 at index (2,) (up to 0.001 allowed)

Caused by sample input at index 17: SampleInput(input=Tensor[size=(4,), device="xpu:0", dtype=torch.float16], args=(), kwargs={'r': '4', 'with_replacement': 'True'}, broadcasts_input=False, name='')
```

### Reproducer

```
PYTORCH_OPINFO_SAMPLE_INPUT_INDEX=17 python test/inductor/test_torchinductor_opinfo.py TestInductorOpInfoXPU.test_comprehensive_combinations_xpu_float16
```

### Versions

Branch: daisyden/inductor (please fill in full `collect_env.py` output / CI job link if needed)

Local passed env: 
pytorch-triton-xpu        3.6.0+git225cdbde
torch                     2.15.0.dev20260819+xpu

dpkg -l |grep intel
ii  intel-fw-gpu                            2025.13.2-398~24.04                              all          Firmware package for Intel integrated and discrete GPUs
ii  intel-gsc                               0.9.5-4~u24.04                                   amd64        Intel(R) Graphics System Controller Firmware Update Library (IGSC FUL)
ii  intel-i915-dkms                         1.25.2.44.250224.52+i62-1                        all          Out of tree i915 driver.
ii  intel-igc-cm                            1.0.225.54093-1146~24.04                         amd64        Frontend libraries for Intel(R) C for Metal Compiler
ii  intel-igc-compute                       0.1.6883                                         amd64        Intel(R) graphics compiler for OpenCL -- core libs, version 2.0 .
[...]

Issue #5020, opened 2026-08-19
test_dispatch_symbolic_meta_outplace_nn_functional_multilabel_soft_margin_loss_xpu_float32 fails: log_sigmoid_forward buffer meta/real shape mismatch


daisyden:
### 🐛 Describe the bug

Case:
`op_ut,third_party.torch-xpu-ops.test.xpu.test_meta_xpu.TestMetaXPU,test_dispatch_symbolic_meta_outplace_nn_functional_multilabel_soft_margin_loss_xpu_float32`

Repro:
```bash
pytest -v test/test_meta.py -k test_dispatch_symbolic_meta_outplace_nn_functional_multilabel_soft_margin_loss_xpu_float32
```

### Root cause

`multilabel_soft_margin_loss` calls `log_sigmoid`, dispatching to `aten.log_sigmoid_forward`, which returns `(output, buffer)`. The real XPU kernel (`log_sigmoid_forward_xpu`) returns an empty `buffer` of shape `(0,)` (same as CUDA). However the decomposition in `torch/_decomp/decompositions.py` picks the buffer shape based on `self.is_cuda or self.is_xpu`:
```python
if self.is_cuda or self.is_xpu:
    buffer = self.new_zeros((0,))
else:
    buffer = z
```
During the meta crossref, `self` is a `meta` tensor, so both `is_cuda` and `is_xpu` are `False`. The code takes the CPU branch and returns a buffer with the input shape `(5,)`, disagreeing with the real XPU impl shape `(0,)`.

### Error log
```text
RuntimeError: output 1: meta disagrees with real impl:
aten.log_sigmoid_forward.default(
  tensor(..., device='meta', size=(5,)) stride=(1,),
) = (
  (tensor(..., device='meta', size=(5,)) stride=(1,), tensor(..., device='meta', size=(5,)) stride=(1,))
)
for element 1, was torch.Size([5]) but real shape was torch.Size([0])

Caused by sample input at index 0: SampleInput(input=Tensor[size=(5,), device="xpu:0", dtype=torch.float32], args=TensorList[Tensor[size=(5,), device="xpu:0", dtype=torch.float32]], kwargs={}, broadcasts_input=False, name='')
```

### Versions
PyTorch: `2.14.0.dev20260809+xpu` (pytorch commit `e66cba9192bef866c83d6bd60fe2142ba6fc1aa7`)

---
_This issue text was drafted with the assistance of an AI agent and reviewed by the submitter._

[...]

Issue #4834, opened 2026-08-06
UT failure: test_warp_softmax_64bit_indexing_xpu_float32 - AssertionError: False is not true


daisyden:
Split from #4825.

Cases:
op_ut,third_party.torch-xpu-ops.test.xpu.test_nn_xpu.TestNNDeviceTypeXPU,test_warp_softmax_64bit_indexing_xpu_float32

## ErrorLog
```
______ TestNNDeviceTypeXPU.test_warp_softmax_64bit_indexing_xpu_float32 _______
[gw7] linux -- Python 3.10.20 /tmp/xpu-tool/Python/3.10.20/x64/bin/python
Traceback (most recent call last):
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/test_nn_xpu.py", line 14132, in test_warp_softmax_64bit_indexing
    run_test(
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/test_nn_xpu.py", line 14130, in run_test
    self.assertTrue(torch.allclose(x.grad.cpu(), xx.grad, rtol=rtol, atol=1e-3))
  File "/tmp/xpu-tool/Python/3.10.20/x64/lib/python3.10/unittest/case.py", line 687, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true
```

## Pytorch Version
71b66af50dd18422e9ed558d83e3582d14aa1639

daisyden:
[agent_triage_result]

auto-label-issues: intel/torch-xpu-ops#4834

Root cause: XPU softmax backward dispatches a dedicated 64-bit `IndexType`
path for large tensors (`src/ATen/native/xpu/sycl/SoftMaxKernels.cpp:1773-1873`); the exact numerical divergence source in that path vs. the CPU reference is unconfirmed without deeper kernel-level investigation.

| label | reason |
|---|---|
| `issue_type: Bug` | `AssertionError: False is not true` in traceback; runtime assertion failure, not a feature/task. |
| `test_module: ut` | Traceback is a `pytest` frame in `test/xpu/test_nn_xpu.py`, a UT case. |
| `module: torch-ops-reduction` | Softmax backward reduces along `dim=-1`; failure is in the reduction kernel path, not gemm/eltwise. |
| `P2` | 1 UT case, `AssertionError` (functional accuracy mismatch) without a crash. |
[...]

PR #5239, opened 2026-09-08
[bot] Fix the agent lifecycle label mapping and surface the SKIPPED reason

ZhaoqiongZ:
Four label problems, found going through the 2026-09-07 fix batch.

**SKIPPED had the wrong label.** Table said `DONE, SKIPPED -> agent:done`, but `agent:skipped` exists and is what the agent applies. Split them: one means a patch is waiting, the other means the issue can be closed.

**Runs could end up with no label.** Nothing said to swap the old one out, or that there should always be one. #4455 dropped `agent:active` and had nothing to add, since `agent:done` didn't exist. Seven runs are unlabelled right now, and #5020 finished DONE but sits on `agent:needs-human`.

**The SKIPPED reason was buried.** It can mean not a bug, no reproducer, or no longer reproduces. The outcome line landed on line 4 in #4834 but line 48 in #5027, so Stage 6 now puts it at the top with its reason.

**`agent:reproduction-needed` was redundant** with `SKIPPED(reproduction_missing)`. Dropped it and deleted the label, nothing was using it.

Also created `agent:done` and `agent:triaged`, and rewrote the three terminal label descriptions.

### DIFF
--- a/.claude/skills/issue-handler/references/execution-modes.md
+++ b/.claude/skills/issue-handler/references/execution-modes.md
@@ -69,8 +69,12 @@ Apply the matching GitHub label when advancing the marker:
 |----------|-------|
 | DISCOVERED, TRIAGING, REPRODUCING, IMPLEMENTING, VERIFYING | `agent:active` |
 | TRIAGED | `agent:triaged` |
-| DONE, SKIPPED | `agent:done` |
+| DONE | `agent:done` |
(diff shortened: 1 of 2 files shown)


Stonepia (approved):

CuiYifeng (approved):

ZhaoqiongZ:
@torchxpubot merge

PR #5240, opened 2026-09-08
[xccl] Support Float4_e2m1fn_x2 and Float8_e8m0fnu for non-reduction ops

Chao1Han:
These are the two halves of an MX tensor. torchao keeps them as separate tensors (MXTensor holds `qdata` plus `scale`), so a collective on an MX tensor is really two collectives.
Both are one byte per element -- fp4 packs two values into a byte and e8m0 is a bare 8-bit exponent -- so for ops that only move bytes they can map to onecclUint8 the same way the fp8 types already do.


zhangxiaoli73 (approved):

zxd1997066 (approved):

Chao1Han:
@torchxpubot merge -f "distributed tests passed"

PR #5241, opened 2026-09-08
Check noise shape and contiguity in rrelu_with_noise_kernel

cyyever:
_rrelu_with_noise_xpu_train writes self.numel() distinct values into noise directly. Nothing checked that noise's shape matches self's, or that noise is contiguous; a non-contiguous noise (e.g. an expanded view) got a `noise_.contiguous()` temporary written into instead and discarded, so the caller's noise tensor silently kept its original values with no error.

Adds the same shape check the CPU kernel already has, and requires noise be contiguous so the kernel can write into it directly -- matching the pattern PyTorch's own `rrelu_with_noise_out_cuda` already uses for the identical CUDA-side bug (pytorch/pytorch#196087, pytorch/pytorch#196088).

Test Plan: no XPU device available where this was written; offered for CI. `clang-format` reports no changes needed beyond what's shown in the diff.

This PR was authored with the assistance of an AI coding assistant.

### DIFF
--- a/src/ATen/native/xpu/sycl/RreluWithNoiseKernels.cpp
+++ b/src/ATen/native/xpu/sycl/RreluWithNoiseKernels.cpp
@@ -9,6 +9,7 @@
  */
 
 #include <ATen/Dispatch.h>
+#include <ATen/TensorUtils.h>
 #include <ATen/native/Math.h>
 #include <ATen/native/Resize.h>
 #include <ATen/native/TensorIterator.h>
@@ -101,7 +102,6 @@ inline void _rrelu_with_noise_xpu_train(
     const Scalar& upper_,
     const std::optional<Generator>& generator) {
   auto input = input_.contiguous();
-  auto noise = noise_.contiguous();
   Tensor tmp_output = output.contiguous();
 
   int64_t numel = input.numel();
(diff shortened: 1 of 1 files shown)


guangyey (commented):
Overall LGTM. Could you please add a test to `test/regressions/`

guangyey (approved):

test/regressions/test_rrelu_with_noise.py
```diff
@@ -0,0 +1,66 @@
...
+            RuntimeError, "refers to a single memory location"
+        ):
+            torch.rrelu_with_noise(x, noise, training=True)
+
+    def test_noise_written(self):
+        # Positive path: the kernel must write into the caller's noise.
+        x = torch.randn(64, 64, device=xpu_device)
+        noise = torch.zeros_like(x)
+        out = torch.rrelu_with_noise(x, noise, 0.1, 0.9, training=True)
+        self.assertTrue(torch.all(noise[x > 0] == 1))
```
guangyey: This seems incorrect.

guangyey:
please ensure the ut pass.

jianyizh (approved):

test/regressions/test_rrelu_with_noise.py
cyyever: fixed

PR #5268, opened 2026-09-09
bot: add a human review template to the fix session comment

ZhaoqiongZ:
A fix run records whether its own patch verified, not whether a human took it.

When the `fix` job ends it now appends this to the session comment, so reviews come back in the same shape every time:

    
    
    
    
    notes:

A reviewer copies it into a new comment, keeps one line, says why.

The block is constant text, so the workflow appends it and the skill is told not to write it.


.github/workflows/bot.yml
```diff
@@ -1386,6 +1388,33 @@ jobs:
...
 
+      # The template a human fills in to review the fix. Constant text, so the
+      # workflow appends it rather than the model.
+      - name: Append review request
+        if: always() && steps.redact.outcome == 'success'
+        continue-on-error: true
+        env:
+          ID: ${{ steps.start-comment.outputs.comment_id }}
+        run: |
+          C="/repos/${{ github.repository }}/issues/comments/$ID"
```
CuiYifeng, line 1400: Do we need to check if `steps.start-comment.outputs.comment_id` is empty? Such as `[ -n "$ID" ]`.

CuiYifeng (approved):
Overall, LGTM. Please consider more levels like `strong/weak accepted` and `strong/weak rejected`.

Stonepia (approved):

ZhaoqiongZ:
@torchxpubot merge

ZhaoqiongZ:
@torchxpubot merge

PR #5275, opened 2026-09-09
skills: one orchestrator for both kinds of batch

ZhaoqiongZ:
`issue-handler`'s Stage 1u and the nightly CI fixer were the same loop:
the wheel preflight, the shared reset recipe, the same four reproduce
verdicts, never abort on one entry, the 3-attempt cap, a `fix_result`
per sub-item, a fan-out report. execution-modes.md already kept a shared
copy of the reset recipe "so the two copies cannot drift".

Added to Stage 1u what it was missing:

* `stage` follows the issue. CI break gets `stage=auto`, everything else
  keeps `stage=nightly`. Many CI failures only reproduce on a source
  build, and `stage=nightly` returns NOT_REPRODUCED as soon as the
  nightly wheel passes, which records unfixed work as already fixed.
* `allow_skip` follows the issue. CI break may skip against a tracking
  issue; elsewhere a skip leaves the bug, so it is not allowed.
* Infra failures (runner crash, docker timeout) are split out before the
  loop instead of running the pipeline to reach CANNOT_VERIFY.
* Sub-items may arrive as bare node ids, not only as an issue checklist.
* The fan-out report gained COVERED and SKIPPED rows.
* One fix may cover several sub-items. Entries in a batch are often the
  same bug in the same class; fixing each separately runs root-cause and
  a source build per entry for the same patch. Fix one, re-run the
  related entries against the staged fix, and list what passes in
  `covers` instead of fixing it again.

`fix-implement` gains the `covers` field. `xpu-nightly-ci-fix` is
deleted and the leaf skills now name one orchestrator.


Stonepia (approved):

CuiYifeng (approved):

ZhaoqiongZ:
@torchxpubot merge

PR #5276, opened 2026-09-09
[bot] Watch #364 and hand each new disable to @torchxpubot fix

ZhaoqiongZ:
DISABLED XPU tests are announced in chuanqi129/pytorch-xpu-ci#364 
add a workflow to watch this issue and copy Disabled issues link in this repo and then use @  torchxpubot fix to fix it automatically

### DIFF
--- a/.github/workflows/ci_disabled_watch.yml
+++ b/.github/workflows/ci_disabled_watch.yml
@@ -7,8 +7,10 @@ name: ci-disabled-watch
 
 on:
   schedule:
-    # After the nightly xpu.yml run and #364's auto-skip pass have settled.
-    - cron: '0 4 * * *'
+    # Twice a day: #364 announces 1.0 new disabled issue per day on average but
+    # bursts to 4, and one run copies one announcement, so a daily poll never
+    # drains a burst. 19:00 and 04:00 Beijing, when the GPU runners are idle.
+    - cron: '0 11,20 * * *'
   workflow_dispatch:
     inputs:
       limit:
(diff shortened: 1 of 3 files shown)


Stonepia (approved):

Stonepia:
@torchxpubot review

.github/workflows/ci_disabled_watch.yml
```diff
@@ -0,0 +1,41 @@
...
+
+on:
+  schedule:
+    # After the nightly xpu.yml run and #364's auto-skip pass have settled.
+    - cron: '0 4 * * *'
+  workflow_dispatch:
+    inputs:
+      limit:
+        description: 'Max announcements to copy in one run'
+        default: '1'
```
CuiYifeng, line 18: Is `limit` to copy comments one by one?
Will `@torchxpubot fix` be added in the first line of each comment once it is copied?
If so, will the fixing tasks start one by one?

CuiYifeng (commented):
My concern is about task scheduling. In my opinion, comments from chuanqi's issue 364 can be copied in batches, but fixing tasks should ideally be executed sequentially. Please check the comments and help confirming my understanding.

.github/workflows/ci_disabled_watch.yml
ZhaoqiongZ: I suppose there should only be 1 comment copied by cron every day, 1 comment will trigger fix once but it will have multi Disabled Issues in this comment.
If there appears multi Disabled Issue comment in the tracking issue and missed some of them by cron, we can also trigger the workflow manually and it will also add just 1 Disabled Issues comment (with several Disabled issues)
the fixing task will try reproduce all of the Disabled issues and group them to several bug and fix one by one

PR #5285, opened 2026-09-09
[Workflow] Keep closed trackers from suppressing findings

laifenxiawucha:
## Summary

A confirmed XPU finding with only a closed tracker could be suppressed by review, the gate, or publisher deduplication. This PR lets it produce a new issue citing the closed tracker; an open `intel/torch-xpu-ops` tracker still prevents duplicate filing. It also raises the separate XPU environment-probe timeout from 30 to 120 seconds and prevents duplicate dates in published alignment issue titles.

Static source verification was split into #5393 and is outside this PR.

## Why and how

- Review records `canonical_tracker_state` alongside the tracker URL. `evidence.md` defines tracker, verdict, and citation decisions; the skill and workflow prompt point to it.
- The gate requires `duplicate` for an open local tracker and rejects `needs-xpu-fix` for that tracker. A new payload must cite any recorded tracker. Older review artifacts lacking tracker state must be reviewed again.
- The publisher checks issue state in both deduplication paths. It reuses open issues, files anew after closed ones, and reports lookup failures instead of silently reusing an unknown issue.
- Publication removes leading `YY-MM-DD` or `YYYY-MM-DD` date tags from reviewed titles before adding the authoritative UTC scan date.

## Before and after

| Case | Before | After |
|---|---|---|
| Closed local tracker | Could prevent a new issue | New issue can be filed and cites the old tracker |
| Open local tracker with `needs-xpu-fix` and no payload | Counted as a fix without filing | Gate rejects the inconsistent verdict; review must use `duplicate` |
| Environment probe takes 31–120 seconds | Timed out before the reproducer | Can complete within the new probe budget |
| Reviewed title already contains a date | Publisher added a second date | Draft and issue use one UTC scan date |

## Validation

16 focused local tests and 11 repository tests passed. The skill validator, workflow YAML parse, Python compilation, lintrunner, and `git diff --check` passed. The focused test harness remains local and was not pushed. The date-only follow-up passed four inline title cases, 11 existing script tests, Python compilation, and lintrunner; no test file was added.

Updated with assistance from OpenAI Codex.


laifenxiawucha:
@torchxpubot review

laifenxiawucha:
Thanks. I confirmed the reported failures and addressed the production and contract issues in `1fe19505`:

- malformed `upstream_source` / `xpu_source` values are normalized before scan validation, so the gate produces blockers instead of raising `AttributeError`;
- `scan-prepare` now explicitly owns `evidence/`, and the contract includes the static JSON shape;
- static sources are compared by SHA-256 only after both objects validate;
- the XPU snapshot must be a nonempty exact excerpt of `target_path` in the gate checkout;
- retained trackers must be cited by an anchored URL match whenever a payload is emitted;
- runner and scan artifacts require an explicit `environment` field, including `null` when there are no runtime entries;
- the docs now cover the all-rejected case and clarify that the job still schedules on `xpu-agent` even though it skips the nightly runtime and XPU probe;
- static-source blocker codes now keep the unit id at the end.

I kept the PR scope without a committed test file. The existing 11 tests, py_compile, workflow parsing, and lintrunner pass. I also exercised the malformed-source crash, fabricated/identical snapshots, tracker prefix collision, and missing/null environment cases with standalone assertions.

laifenxiawucha:
@torchxpubot review

laifenxiawucha:
Thanks. I adopted the review findings that exposed the mismatch between the claimed deterministic verification and the actual model trust boundary.

Updated in `f279a4c0`:

- removed the self-referential workspace excerpt check;
- restored `repository` and `commit` to both static source objects;
- bound the upstream coordinates to the collector repository/frozen head and the XPU coordinates to `intel/torch-xpu-ops` plus the explicit workflow SHA;
- defined snapshots as immutable audit data rather than deterministically authenticated source;
- required the independent reviewer to re-fetch both exact files, confirm each snapshot and the oracle, and emit `verification-gap` without a payload on uncertainty;
- explicitly prohibited `script`, `script_sha256`, and `timeout_seconds` for static entries;
- limited static verification to functional/API source differences, excluding performance claims;
- restored both runtime and static execution examples and reflowed the edited documentation.

The scarce-runner optimization is intentionally deferred; no-runtime plans still use the existing `xpu-agent` job but skip nightly provisioning and probing.

Per the chosen MVP scope, no test file is added. I removed the unrelated pytest result from the PR validation section rather than presenting it as feature coverage. Syntax, workflow parsing, lintrunner, and standalone gate boundary assertions pass.

laifenxiawucha:
@torchxpubot review

laifenxiawucha:
A strong motivating example for `verification: static` is Run #12: [XPU Alignment run 34302439189](https://github.com/intel/torch-xpu-ops/actions/runs/34302439189), which validated upstream [PyTorch PR #196230](https://github.com/pytorch/pytorch/pull/196230), commit [`78fa18d`](https://github.com/pytorch/pytorch/commit/78fa18d91b0d762f510d64053d01d265562a3cd4).

That upstream commit removed `adaptive_pool_empty_output_check` from `AdaptivePooling.h`, while the XPU implementation still referenced it at `src/ATen/native/xpu/AdaptiveAveragePooling2d.cpp:32`. This is a source compatibility problem.

Even if the nightly is current and contains the upstream commit, the combined PyTorch/XPU source would fail during the XPU extension build before any runtime reproducer could start. The two appropriate ways to detect this are therefore:

- a compile/integration check using the upstream and XPU revisions together; or
- a static check that reads the frozen source coordinates and confirms that one revision deleted the helper while the other still references it.

A normal runtime reproducer can only help after the combined sources compile. It may be useful for the separate pooling shape-check or error-message behavior, but it cannot establish this deleted-symbol compatibility failure.

laifenxiawucha:
@torchxpubot review

laifenxiawucha:
@torchxpubot review

ZhaoqiongZ (changes requested):
too much tracker and tracker state in the logic, is all of them really matters?
if the state get properly in time before use them? if the state is well used to prevent or keep issues?

PR #5293, opened 2026-09-10
[CI] Distinguish hang vs timeout and skip reruns for timed-out UT cases

mengfei25:
## What

Improve how timed-out UT cases are handled under `pytest` + `pytest-xdist` + `pytest-timeout`, and distinguish **hang** from **timeout**.

## Why

With `--timeout_method=thread`, a pytest timeout hard-kills the xdist worker, so xdist reports it as `crashed while running` — identical to a genuine crash. The `--only-rerun 'crashed while running'` filter therefore reran timed-out cases, which just time out again (wasteful), and the JUnit XML only showed a generic worker-crash message instead of the real reason.

## Changes

`.github/scripts/conftest.py`:
- Records a per-case **reason file** just before pytest-timeout aborts the worker. A companion timer fires within the lead window and is cancelled if the case returns control, so **only genuinely aborted cases keep a reason file** (precise detection, shared controller→workers via `pytest_configure_node`).
- **Classifies the abort** by sampling the test thread's Python stack across a short window:
  - dead thread or no stack progress ⇒ **Hang** (stuck)
  - stack still advancing ⇒ **Timeout** (slow but progressing)
  
  The configured timeout duration and the hung stack are captured in the reason.
- In `pytest_handlecrashitem`, replaces xdist's generic `crashed while running` text with the captured reason. This drops the phrase the rerun filter matches (so timeouts/hangs are **not rerun**) and surfaces the **real reason in the JUnit XML** report.
- All added hooks use `optionalhook=True`, so the file stays safe when `xdist`/`pytest-timeout` are absent.

`.github/scripts/check-ut.py`:
- Recognize `Timeout` and `Hang` in `error_types` so the UT summary shows the real reason instead of a generic label.

Also includes a prior commit bumping several UT step shell timeouts from 3600s to 7200s in `.github/actions/linux-uttest/action.yml`.

## Notes

Classification is based on Python-level progress: a case blocked in a long C call (e.g. a device sync) shows no Python movement and is labeled **Hang**, even though C code may still be active — an inherent limit of observing from another thread.


chuanqi129:
@torchxpubot review

chuanqi129 (approved):

chuanqi129:
Please address the block issue review comment from bot

PR #5300, opened 2026-09-10
[SYCL free func] Migrate NestedTensorBinaryOpsKernels to sycl free function

YzyParry:
Migrate kernels in `src/ATen/native/nested/xpu/sycl/NestedTensorBinaryOpsKernels.cpp`
to sycl free function.
Including:

- OpDenseVectorizedFunctor

`AddFunctor` and `MulFunctor` are not kernels. They are the elementwise operation
passed to the kernel as the `func_t` template parameter, so they stay as they are.


tye1:
@torchxpubot  check-ut

PR #5310, opened 2026-09-10
[xccl] Pin the sub-group size of the kernels using the LSC signal asm

Chao1Han:
The inline asm in Signal.hpp asks for execution size 32, but none of the kernels calling it declared a sub-group size, so IGC was free to pick one. The asm operands are ordinary variables of the surrounding code and are compiled at whatever width IGC chose, so any width below 32 makes the asm read and write past the end of both operands. It happened to work on pvc and bmg only because IGC picked SIMD32 there; nothing in the source asked for it. On dg2 it picks SIMD8 and IGC aborts while compiling (GSD-13398), and dg2 is in the default AOT target list.

Bind the exec size in the asm and the sub-group size of the three kernels to one constant. The value is 16 rather than 32: dg2 still fails to compile at 32 even with the sub-group size pinned, and pvc and bmg reject 8, so 16 is the only width all three accept.

### DIFF
--- a/src/xccl/Signal.cpp
+++ b/src/xccl/Signal.cpp
@@ -6,6 +6,7 @@
 namespace c10d::symmetric_memory {
 
 struct barrierKernel {
+  SYCL_REQD_SUB_GROUP_SIZE(XCCL_SIGNAL_SUB_GROUP_SIZE)
   void operator()(sycl::nd_item<1> item) const {
     auto thread_id = item.get_local_id(0);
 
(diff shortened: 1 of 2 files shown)


chunhuanMeng (approved):

zhangxiaoli73 (approved):

Chao1Han:
@torchxpubot merge -f "distributed tests passed and build successful with dg2 aot at local"

PR #5314, opened 2026-09-10
Reject a mismatched out= dtype in histc instead of casting silently

### LINKED ISSUE #5237: [xpu-alignment] histc(out=) on XPU silently casts instead of rejecting a mismatched out dtype
## Independent local verification

XPU rerun 2026-09-07 (Max 1100, torch 2.15.0.dev20260902+xpu): float64 input with int64 out was accepted and silently cast. This supersedes the skipped behavior in intel/torch-xpu-ops#2234; that issue is linked for deduplication and intentionally unchanged.

The upstream result is still tracked as upstream-dependent; this issue records the concrete XPU defect and does not assert that upstream has finalized cross-backend semantics.

<sub>alignment scan `2026-09-05`, run `34005971680`</sub>
## Upstream source

- Upstream object: https://github.com/pytorch/pytorch/pull/196100 (`Fix histc(out=) dtype validation mismatch between CPU and CUDA`) - **OPEN / unmerged** at review time, `mergedAt: null`, base `main`.
- Upstream report it fixes: https://github.com/pytorch/pytorch/issues/196008 (open).
- Scan window: `[2026-09-05T00:00:00Z, 2026-09-06T00:00:00Z)` over `pytorch/pytorch`, frozen default-branch head `071dd4d98ee0ca692fbe0cb3e9f3b95955d73329`.

The PR body states explicitly that the XPU entry points at `_histc_out_xpu`, "a same-shaped
wrapper that lives out of tree in the xpu fork; it likely shares this gap and should get the
same fix there." This review confirms on hardware that it does.

## Observed XPU behavior

`torch.histc` with an `out=` whose dtype differs from the input is accepted on XPU and the
floating-point histogram is silently truncated into the `out` dtype. CPU raises for the same
call today, so this is a current CPU-versus-XPU argument-validation divergence and does not
depend on PR 196100 landing.

CPU path: `histc.out` dispatches to `histogram_histc_out` ->
`histogramdd_prepare_out` (`aten/src/ATen/native/Histogram.cpp:132`), which raises
`torch.histogram: input tensor and hist tensor should have the same dtype, ...`.

## Reproducer

```python
import torch

torch.manual_seed(0)
x = torch.randn(8, dtype=torch.float64, device="xpu")
out = torch.empty(4, dtype=torch.int64, device="xpu")

h = torch.histc(x, bins=4, min=-2.0, max=2.0, out=out)   # expected: RuntimeError
torch.xpu.synchronize()
print("returned without raising:", h.dtype, h.cpu().tolist())

matching = torch.empty(4, dtype=torch.float64, device="xpu")
h2 = torch.histc(x, bins=4, min=-2.0, max=2.0, out=matching)
print("matching out dtype:", h2.dtype, h2.cpu().tolist())
```

Observed output (`runner/logs/pr-196100.log`):

```
input dtype: torch.float64, mismatched out dtype: torch.int64
x: device=xpu:0, dtype=torch.float64
out: device=xpu:0, dtype=torch.int64
CALL: torch.histc(x, bins=4, min=-2.0, max=2.0, out=out)
returned without raising: dtype=torch.int64, values=[0, 3, 4, 0]
matching out dtype: returned dtype=torch.float64, device=xpu:0, values=[0.0, 3.0, 4.0, 0.0]
```

The same call on CPU raises
`RuntimeError: torch.histogram: input tensor and hist tensor should have the same dtype, but got input double and hist long int`.

## Target-path evidence

- `native_functions.yaml` (frozen head, line 9792) dispatches `histc.out` to
  `XPU: _histc_out_xpu`, implemented in `src/ATen/native/xpu/SummaryOps.cpp`.
- `_histc_out_xpu` computes `auto ret = _histc_xpu(...)` in the *input* dtype, then does
  `at::native::resize_output(result, ret.sizes()); result.copy_(ret);`. `copy_` casts
  silently, so any `out` dtype is accepted. There is no `TORCH_CHECK` on
  `self.dtype() == result.dtype()`.
- Runtime proof the XPU path ran: both operands are `xpu:0` tensors, the histogram was
  actually computed (`[0, 3, 4, 0]`, matching the `float64` result `[0.0, 3.0, 4.0, 0.0]`
  from the matching-dtype call), and the return dtype is the mismatched `int64`.

## Suggested fix

Add the CPU/CUDA check to `_histc_out_xpu` in `src/ATen/native/xpu/SummaryOps.cpp`, with the
same message text upstream uses:

```cpp
TORCH_CHECK(
    self.dtype() == result.dtype(),
    "torch.histogram: input tensor and hist tensor should",
    " have the same dtype, but got input ", self.dtype(),
    " and hist ", result.dtype());
```

The fix should also remove `"test_out_histc_xpu_float32"` from
`test/xpu/skip_list_common.py:183`, which is currently skipped for exactly this gap.

## Prior XPU history (not a live tracker)

https://github.com/intel/torch-xpu-ops/issues/2234 reported the same defect through
`TestCommonXPU::test_out_histc_xpu_float32` ("Expected RuntimeError when doing an unsafe cast
from a result of dtype torch.float32 into an out= with dtype torch.long"). It was closed as
completed by https://github.com/intel/torch-xpu-ops/pull/3119, which only added the test to
`skip_list_common.py` on the stated grounds that "Both tests are expected to fail, same as on
CUDA." PR 196100 removes that premise by adding the check to CUDA, and the upstream test it
adds (`test/test_reductions.py::test_histc_out_dtype`) is device-generic apart from
`@skipIfMPS`, so XPU will start failing it once the PR lands. Issue 2234 is therefore not a
canonical tracker for this work: it is closed and its resolution was a skip, not a fix.

## Ownership

The code change belongs in `intel/torch-xpu-ops`: `_histc_out_xpu` is implemented here and
PR 196100 only touches `aten/src/ATen/native/cuda/SummaryOps.cu`. No `pytorch/pytorch` issue
or PR claims the XPU side, and no open `intel/torch-xpu-ops` tracker covers it (searches for
`histc`, `histc out dtype`, and `_histc_out_xpu` returned only the closed items above and
unrelated `histc` work).

## Environment

```
torch: 2.15.0.dev20260904+xpu (git_version 3780801c1d295481c4ddeb2d43049cd00f4c5697)
torch file: /opt/conda/lib/python3.13/site-packages/torch/__init__.py
python: 3.13.13 (/opt/conda/bin/python3.13)
xpu available: True, device_count: 8
xpu device 0: Intel(R) Arc(TM) Pro B60 Graphics
```

Evidence artifacts: `scripts/repro_pr-196100.py`
(sha256 `c75c448132fd81362606622e032d5a6f077992270d996aae3493dbf99d9bdd69`),
`runner/logs/pr-196100.log`
(sha256 `3c3d468a291cc469ab31c216b0e2e4681d19ae87575d22aff2c3bc0d6be88024`).

laifenxiawucha:
`_histc_out_xpu` computes into a fresh tensor of the input dtype and then does `result.copy_(ret)`. `copy_` converts, so a float64 input written into an int64 `out=` yields a silently truncated histogram and no error. CPU raises here; its check lives in `histogramdd_prepare_out`, which the XPU out variant does not go through. The functional form allocates its own output and is unaffected.

The check goes before `resize_output` so a rejected call does not resize the caller's tensor on its way to raising, and reuses the CPU message verbatim.

`test_out_histc_xpu_float32` comes off the skip list, and needs one more change to pass: `align_db_decorators` matched `device_type == "cuda"` as a bare string, but this test's upstream xfail is scoped to the tuple `("cuda", "xpu")`, so it stayed live on XPU and the fixed test would report an unexpected success. The new branch drops only `"xpu"` from the scope and stays gated on `_cuda_xfail_xpu_pass`, which already lists `("histc", "test_out")`.

`test/regressions/test_histc.py` adds a direct case plus a matching-dtype positive control. Against the current released binary (`torch 2.12.1+xpu`, Max 1100) the rejection case fails as expected -- `RuntimeError not raised` -- and the control passes.

pytorch/pytorch#196100 adds the same check to CUDA. This neither depends on nor conflicts with it; once it lands, the tuple xfail should be dropped upstream rather than de-scoped here.

Fixes #5237.

## Post-fix XPU result

The Linux XPU `op_regression` artifact for PR head `b6601b2545e32f6e4773091820a5dcf2b8b24900` in [CI run 34858234020](https://github.com/intel/torch-xpu-ops/actions/runs/34858234020) records:

```text
test_histc_out_rejects_mismatched_dtype       PASSED
test_histc_out_matching_dtype_still_works     PASSED
2 passed, 0 failed
```

The fixed path rejects a mismatched `out=` dtype and preserves the matching-dtype positive control.

### DIFF
--- a/src/ATen/native/xpu/SummaryOps.cpp
+++ b/src/ATen/native/xpu/SummaryOps.cpp
@@ -55,6 +55,14 @@ Tensor& _histc_out_xpu(
     const Scalar& min,
     const Scalar& max,
     Tensor& result) {
+  // Checked before resize_output so a rejected call leaves result untouched.
+  TORCH_CHECK(
+      self.dtype() == result.dtype(),
+      "torch.histogram: input tensor and hist tensor should",
+      " have the same dtype, but got input ",
+      self.dtype(),
+      " and hist ",
+      result.dtype());
   auto ret = _histc_xpu(self, bins, min, max);
   at::native::resize_output(result, ret.sizes());
   result.copy_(ret);
--- a/test/regressions/test_histc.py
+++ b/test/regressions/test_histc.py
@@ -1,0 +1,36 @@
+# Copyright 2020-2026 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Owner(s): ["module: intel"]
+import torch
+from torch.testing._internal.common_utils import TestCase
+
+xpu_device = torch.device("xpu")
+
+
+class TestHistc(TestCase):
+    def test_histc_out_rejects_mismatched_dtype(self):
+        """out= must not be cast into silently.
+
+        _histc_out_xpu computed into a tensor of the input dtype and finished
+        with result.copy_(ret), so a float histogram written into an integral
+        out= came back truncated with no error. CPU rejects the same call from
+        histogramdd_prepare_out, which the XPU out variant does not go through.
+        """
+        x = torch.linspace(1, 8, 8, device=xpu_device, dtype=torch.float32)
+        out = torch.empty(4, device=xpu_device, dtype=torch.int64)
+
+        with self.assertRaisesRegex(RuntimeError, "should have the same dtype"):
+            torch.histc(x, bins=4, min=0, max=8, out=out)
+
+    def test_histc_out_matching_dtype_still_works(self):
+        x = torch.linspace(1, 8, 8, device=xpu_device, dtype=torch.float32)
+        out = torch.empty(0, device=xpu_device, dtype=torch.float32)
+
+        torch.histc(x, bins=4, min=0, max=8, out=out)
+        self.assertEqual(out, torch.histc(x, bins=4, min=0, max=8))
--- a/test/xpu/skip_list_common.py
+++ b/test/xpu/skip_list_common.py
@@ -180,7 +180,6 @@ skip_dict = {
         # Exception: The supported dtypes for linalg.multi_dot on device type xpu are incorrect!
         "test_dtypes_linalg_multi_dot_xpu",
         # For CUDA it's skipped explicitly in common_methods_invocations.py in upstream. We can skip it here
-        "test_out_histc_xpu_float32",
         "test_out_mean_xpu_float32",
         # FakeTensor mismatch in outputs_alias_inputs for aten.view.default
         # Known upstream issue: https://github.com/pytorch/pytorch/issues/159150
--- a/test/xpu/xpu_test_utils.py
+++ b/test/xpu/xpu_test_utils.py
@@ -1071,6 +1071,23 @@ class XPUPatchForImport(XPUImportCtx):
                             wrapper.device_type = "xpu"
                             replaced = True
                     elif (
+                        isinstance(wrapper.device_type, (list, tuple))
+                        and "xpu" in wrapper.device_type
+                        and unittest.expectedFailure in wrapper.decorators
+                        and (op_name, wrapper.test_name) in _cuda_xfail_xpu_pass
+                    ):
+                        # Upstream may scope one xfail to several devices at
+                        # once (device_type=("cuda", "xpu")). The "cuda" branch
+                        # above only matches the plain string, so drop XPU from
+                        # the scope here.
+                        replaced = True
+                        new_wrapper = copy.copy(wrapper)
+                        new_wrapper.device_type = tuple(
+                            d for d in wrapper.device_type if d != "xpu"
+                        )
+                        wrapper_xpu.append(new_wrapper)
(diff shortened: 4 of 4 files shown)


N0AHZACH:
I am already working on this and I am done with it............would be down to collab?

laifenxiawucha:
> I am already working on this and I am done with it............would be down to collab?

Hi @N0AHZACH , Thanks for your contribution! Our XPU alignment agent found your CPU/CUDA PR and detected the same issue on XPU, so we opened this PR. Since you're already working on it, we'll close this one and waiting for your PR. Thanks a lot for contributing and happy to collab!

N0AHZACH:
> > I am already working on this and I am done with it............would be down to collab?
> 
> 
> 
> Hi @N0AHZACH , Thanks for your contribution! Our XPU alignment agent found your CPU/CUDA PR and detected the same issue on XPU, so we opened this PR. Since you're already working on it, we'll close this one and waiting for your PR. Thanks a lot for contributing and happy to collab!

That's awesome I'll push the PR in an hour or so kinda fixing something

PR #5315, opened 2026-09-10
Add distributed xpu profiler test

kkarwot:


test/xpu/distributed/test_c10d_xccl.py
```diff
@@ -1802,5 +1804,71 @@ class SetDeviceMethod(Enum):
...
+            torch.xpu.synchronize()
+
+            with profile(
+                activities=[ProfilerActivity.CPU, ProfilerActivity.XPU],
+            ) as prof:
+                output = x @ weight
+                dist.all_reduce(output)
+                prof.step()
+                # Sync before export so async XPU kernels are recorded.
+                torch.xpu.synchronize()
```
aostrowski-hbn, line 1852: There's no schedule (wait/warmup/active step counts) so we can get rid of the `step` call.

```suggestion
                # Sync before export so async XPU kernels are recorded.
                torch.xpu.synchronize()
```

If a schedule would be used the `synchronize` call should be placed before moving to the next `step`. Otherwise part of an iteration could be cut out from the trace or some warmup artifacts could be included when they shouldn't.

test/xpu/distributed/test_c10d_xccl.py
```diff
@@ -1802,5 +1804,71 @@ class SetDeviceMethod(Enum):
...
+                    e for e in data.get("traceEvents", []) if e.get("cat") == "kernel"
+                ]
+
+            gemm_kernels = [k for k in kernels if "gemm" in k.get("name", "").lower()]
+            self.assertGreater(
+                len(gemm_kernels),
+                0,
+                f"[Rank {self.rank}] No GEMM kernel in trace; saw kernels: "
+                f"{[k.get('name') for k in kernels]}",
+            )
```
aostrowski-hbn, line 1868: As this is a distributed test I think we should check at least for distributed kernels or XCCL events. NCCL tests already do it: https://github.com/pytorch/pytorch/blob/main/test/distributed/test_c10d_nccl.py#L5218-L5238. 

I think in this case there should be an event called `xccl:all_reduce`.

Can we add similar checks here?

test/xpu/distributed/test_c10d_xccl.py
```diff
@@ -1802,5 +1804,71 @@ class SetDeviceMethod(Enum):
...
+        finally:
```
aostrowski-hbn, line 1869: TIL that there's a PT function for validating JSON traces called `validate_trace`: https://github.com/pytorch/pytorch/blob/main/torch/profiler/_trace_validator.py#L318

Can we call it here to ensure we got a proper trace? Looks like something we could use in much more tests.

test/xpu/distributed/test_c10d_xccl.py
```diff
@@ -1802,5 +1804,71 @@ class SetDeviceMethod(Enum):
...
+    def tearDown(self):
+        super().tearDown()
+        try:
+            os.remove(self.file_name)
+        except OSError:
+            pass
+
+    @requires_xccl()
+    @skip_if_lt_x_gpu(2)
+    def test_profiler_xpu_distributed(self):
```
aostrowski-hbn, line 1825: To get proper distributed traces we need to set the env vars:
- `INTEL_LIBITTNOTIFY64` should point at the location of `libpti_view.so`,
- `TORCH_PROFILER_ENABLE_COLLECTIVE_PROFILING=1`

test/xpu/distributed/test_c10d_xccl.py
```diff
@@ -1802,5 +1804,71 @@ class SetDeviceMethod(Enum):
...
+            torch.xpu.set_device(self.rank)
+
+            M = N = K = 4
+            x = torch.randn(M, K, device="xpu")
+            weight = torch.randn(K, N, device="xpu")
+
+            # Warm up so first-iteration setup costs stay out of the profiled region.
+            for _ in range(2):
+                _ = x @ weight
+            torch.xpu.synchronize()
```
aostrowski-hbn, line 1843: Could we warm up the collective too?

```suggestion
            # Warm up so first-iteration setup costs stay out of the profiled region.
            for _ in range(2):
                output = x @ weight
                dist.all_reduce(output)
            torch.xpu.synchronize()
```

aostrowski-hbn (changes requested):

PR #5316, opened 2026-09-10
[Windows] Prepare Windows build configuration for SYCL free-function rewrite

draghan:
This PR is preparing the Windows build configuration for proper compilation and linking, once more kernels will join the SYCL free-function rewrite, to avoid certain Windows limitations.

1. Main limitation, the max size of a Windows static library:

    On Windows the SYCL free-function kernel style produces heavier obj files.
  Aggregated into one archive they exceed the 4GB hard limit of a Windows static library, failing at the archive stage with LNK1248.

    Distribute the kernel sources over six static libraries so no single archive approaches the limit. The kernels are grouped in a way, that the heaviest kernel families (binary elementwise, unary/special functions, NN layers) are separated from each other.
    The common kernel library is narrowed to hold only the shared comm sources which carry symbols shared between kernel translation units.

2. MAX_PATH limit:

    A descriptive target names consume big chunk of the 260 characters of the Windows MAX_PATH limit.
    The SYCL compiler is long-path aware and still correctly writes the object, but lib.exe is not and fails to read it back, resulting in LNK1181.

    The CMake target names are changed to be deliberately short (xpuk0 ... xpuk5) and the descriptive name is applied to the produced .lib through OUTPUT_NAME.

3) command-line length limit of `cmd.exe`:

    The device object list for a group library is long enough to exceed the 8191-character cmd.exe command-line limit, past which parameters can be mis-parsed depending on quote and operator placement.
    This results in linker failures that vary with object count, order and build-directory depth, and so appear non-deterministic.

    To fix this, pass the device-link object files to `-fsycl-link` through a response file on Windows.

Linux build and link topology are unaffected.


draghan:
@torchxpubot check-ut

tye1:
@marpioch  could you pls. review and approve this PR if no more comments from your side?

marpioch (approved):

tye1:
@torchxpubot  check-ut

PR #5348, opened 2026-09-14
Add `@torchxpubot label` for issues and fold AI comment detail

chuanqi129:
Adds `@torchxpubot label` for issues, and folds the long AI output on `review` and
`ut-check` behind a summary so the verdict is readable without scrolling.

### `label`

Issue-only, joining `fix` and `triage`. PRs already get `disable_*` labels from
`auto_label.yml`. `@torchxpubot label --dry-run` reports without writing; anything else
after the command is rejected rather than ignored, so a mistyped `--dryrun` cannot fall
through to a live write.

The `label-issue` skill documents what a workflow around it must do, and the job is built
to that:

- **Analysis and application are separate steps.** The agent only writes `labels.md` --
  no write token, no `WebFetch` (the skill forbids following URLs out of issue text), no
  `Task` (it forbids subagent fan-out). A plain `run:` step applies it.
- **That split is what makes the exit code work.** `apply_labels.py` exits 2 when setting
  Priority cleared the issue's other native fields, and states a wrapping workflow must
  fail loudly there. Inside the agent's `Bash` that status is a tool result, so the data
  loss would have gone green. A failed apply also posts its own comment, because the
  comment step is skipped once a prior step fails and the verify step returns early on a
  healthy agent.
- **The apply target is pinned**, by a new `--expect <repo>#<id>` in `apply_labels.py`.
  It reads the issue to write from a `label-issue: <repo>#<id>` header inside `labels.md`,
  which the agent produced after reading text anyone can author. The vocabulary is
  allowlisted; the target was not. The check compares the value the parse resolved, not
  whether a matching header exists somewhere -- the parse binds to the first header, so a
  file carrying an injected one above the real one would walk past a whole-file check.
- **The agent holds no write scope.** The job is `issues: read`; the checkout takes no
  PAT, for the same reason the `fix` job scrubs one -- checkout persists the token into
  `.git/config` and the agent has `Bash`. Every write goes through `MERGE_TOKEN` on a
  later step.
- **The comment posts on `always()` and says what was written.** A refused or failed apply
  is the run where the proposed labels matter most, and the comment carries
  `apply_labels.py`'s output so allowlist drops and the label cap are visible rather than
  buried in the job log.
- **`labels.md` is uploaded as an artifact**, so the reasoning behind each applied label
  outlives the run instead of surviving only as the model's paraphrase.

The runner has no PyTorch checkout, so `label-issue` runs evidence-only, which is what the
skill asks for in automation.

### Folded comments

`review` and `ut-check` now lead with a `VERDICT:` line and fold the rest. `label` is
deliberately excluded: `label-issue` already wraps its artifact in a `<details>`, so
folding it again would bury the table two clicks deep.

The line is specified in **each skill's own Output Format section**, not only in the
workflow prompt. Both skills prescribe their first line (`## PR Review: #<n>`,
`## UT Result Check: PR #<n>`), so a prompt alone would lose to the more specific
instruction.

`foldDetails` searches the first few lines rather than only the first: models routinely
open with a sentence of preamble -- the bot's own review on this PR did -- which would
otherwise have left every comment unfolded. It closes only the gap the hoisted line
leaves, since a global collapse would rewrite the blank runs inside log excerpts. It lives in `.github/scripts/fold_details.js`
so its three call sites share one copy and it can be tested. Each call site tolerates the
module being absent and posts unfolded instead, because losing a whole comment to a
missing file would be worse than not folding.

Test: `.github/scripts/test_fold_details.js`, run by `script-tests`

```bash
node .github/scripts/test_fold_details.js
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/bot.yml'))"
node --check   # every embedded github-script block, via a YAML walker
bash -n        # the apply step's script
```

Both helper files are in the workflow's path filter, so the tests run on changes to
themselves. The require fallback was exercised against a missing module. The parse step's
JavaScript was run standalone: `label` is accepted on an open issue, refused on a PR and
on a closed issue, and `--dryrun`, `-n` and stray text are each rejected with the usage.
The target-pinning grep was checked against the `#77`-matching-`#770` case it exists to
prevent. Exit-code propagation was confirmed under `bash -e`, the shell Actions uses.

### Not verifiable on this PR

The fold itself. The posting jobs check out with no ref, so the `require` resolves against
`main`, where `fold_details.js` does not exist until this merges. Until then those jobs
take the unfolded fallback.

### Draft until verified live

- [ ] `@torchxpubot label --dry-run` on an issue reports without writing
- [ ] `@torchxpubot label` applies labels/type/priority and uploads `labels.md`
- [ ] A `labels.md` naming a different issue is refused by `--expect`
- [ ] A forced `apply_labels.py` exit 2 fails the job and comments
- [ ] `--max-turns 40` is enough for an issue with failures in several groups
- [ ] The cost row lands on the tracking issue

Still open, and better judged from the first live `--dry-run` than guessed at now:
`duplicate`/`wontfix`/`need_split` are applied by `apply_labels.py` rather than
human-gated, and `--no-comment` sends the Type/Priority mismatch note to stderr only.


chuanqi129:
@torchxpubot review

chuanqi129:
@copilot+claude-opus-5 /pr-review

chuanqi129:
@torchxpubot review

.github/workflows/bot.yml
```diff
@@ -995,11 +1037,28 @@ jobs:
             const trigger = context.payload.comment;
             const header = `> Replying to [this comment](${trigger.html_url}) by @${trigger.user.login}\n\n`;
             const footer = `\n\n---\n_Generated by [\`ut-check\` job](${jobUrl})._`;
+            // Hoist the model's VERDICT line and fold everything else. The line is
+            // prescribed by each skill's Output Format, but models routinely open
+            // with a sentence of preamble, so search the first few lines rather
+            // than only the first. A model that omits it entirely gets posted
+            // unfolded, and the warning makes that visible in the log.
+            const foldDetails = (text) => {
```
daisyden, line 1045: Can we use a shared foldDetails definition with line 900?

.github/workflows/bot.yml
```diff
@@ -1588,6 +1630,216 @@ jobs:
...
+        uses: actions/checkout@v7
+
+      - name: Run AI labeler
+        id: ai-label
+        uses: anthropics/claude-code-action@v1
+        with:
+          use_bedrock: "true"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          settings: '{"alwaysThinkingEnabled": true}'
+          claude_args: "--model ${{ env.BEDROCK_MODEL }} --max-turns 40 --allowedTools Bash,Read,Glob,Grep,Write,Skill"
```
daisyden, line 1656: please ensure max-turns 40 works for the issues with lot of test failures in multiple groups.

.github/workflows/bot.yml
```diff
@@ -1588,6 +1630,216 @@ jobs:
...
+
+          args=(--no-comment)
+          if [ "$DRY_RUN" != "true" ]; then
+            args+=(--apply)
+          fi
+          # Run outside the agent so a non-zero exit fails the job. apply_labels.py
+          # exits 2 when setting Priority cleared the issue's other native fields,
+          # and documents that a wrapping workflow must fail loudly rather than
+          # bury that in a log.
+          python .github/scripts/apply_labels.py "$labels_md" "${args[@]}"
```
daisyden, line 1705: According to the comments the exit code can be handled when apply_labels.py is called in worflow, but seems we did not handle it.

.github/workflows/bot.yml
```diff
@@ -1588,6 +1630,216 @@ jobs:
...
+            const jobUrl = job?.html_url || `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+            const trigger = context.payload.comment;
+            const header = `> Replying to [this comment](${trigger.html_url}) by @${trigger.user.login}\n\n`;
+            const footer = `\n\n---\n_Generated by [\`label\` job](${jobUrl})._`;
+            // Hoist the model's VERDICT line and fold everything else. The line is
+            // prescribed by each skill's Output Format, but models routinely open
+            // with a sentence of preamble, so search the first few lines rather
+            // than only the first. A model that omits it entirely gets posted
+            // unfolded, and the warning makes that visible in the log.
+            const foldDetails = (text) => {
```
daisyden, line 1739: use a shared foldDetails?

.github/workflows/bot.yml
```diff
@@ -1588,6 +1630,216 @@ jobs:
...
+        uses: actions/github-script@v9
+        env:
+          AI_LABEL_OUTCOME: ${{ steps.ai-label.outcome }}
+          BOT_ESCALATION_USERS: ${{ secrets.BOT_ESCALATION_USERS }}
+        with:
+          github-token: ${{ secrets.MERGE_TOKEN }}
+          script: |
+            const fs = require('fs');
+            const outcome = process.env.AI_LABEL_OUTCOME || 'skipped';
+            const file = '${{ steps.ai-label.outputs.execution_file }}';
```
daisyden, line 1769: May I know where the execution_file  is defined?

daisyden (commented):
please check my comments.

.github/workflows/bot.yml
chuanqi129: Done -- it is now `.github/scripts/fold_details.js`, required at all three sites.

It started out that way, then got inlined when the 12 lines of string handling looked like
poor value against a module plus a test file. Two reviewers asking independently settled
it, and pulling it back out paid for itself immediately: the shared copy now searches the
first few lines for the VERDICT rather than only the first, because the bot's own review
on this PR opened with a sentence of preamble and would have left every comment unfolded.
That fix landing in one place rather than three is the argument you were making.

One consequence worth knowing: the posting jobs check out with no ref, so the `require`
resolves against `main`. The helper is only live once this merges, and until then those
jobs take the unfolded fallback. Each call site wraps the `require` so a missing module
costs the fold, not the whole comment.

.github/workflows/bot.yml
chuanqi129: Same as above -- shared now, in `.github/scripts/fold_details.js`.

.github/workflows/bot.yml
chuanqi129: The exit code does propagate, but you were right that something was being dropped -- just
not that.

The `python` call is the last command in the `run:` block and Actions runs `run:` under
`bash -e`, so a non-zero exit becomes the step's status. Checked rather than assumed:

```
$ cat /tmp/step.sh
labels_md=/tmp/t.md
args=(--no-comment --apply)
python3 /tmp/fake_apply.py "$labels_md" "${args[@]}"   # exits 2

$ bash -e /tmp/step.sh; echo "exit=$?"
ERROR: destructive replace
exit=2
```

What was actually broken is what the requester sees. On a failed apply, `Post label
comment` is skipped -- a step with no `if:` does not run once a prior step has failed --
and `Verify: AI labeling succeeded` returns early because the agent itself succeeded. So
the job went red with no comment anywhere, which is the same silent-failure shape the
exit-2 handling exists to prevent.

Added a `Report an apply failure` step gated on `failure() && steps.apply.outcome ==
'failure'` that comments with the job link. Thanks for pulling on this.

.github/workflows/bot.yml
chuanqi129: It is an output of `anthropics/claude-code-action@v1`, reached through the step's `id`:

- `id: ai-label` on the "Run AI labeler" step
- `steps.ai-label.outputs.execution_file` here

The action writes a JSON log of the run and exposes the path under that name. The four
AI-backed jobs all read it the same way (`steps.ai-review.*` in `review`,
`steps.ai-triage.*` in `triage`, and so on) to pull out the final `result` entry for the
comment body, the success check, and the cost figure.

.github/workflows/bot.yml
chuanqi129: Agreed, and I can't settle it locally -- it needs a real multi-group issue, since the
turn count scales with the number of groups the skill root-causes separately.

It is on the PR's verification checklist. Plan is to run `@torchxpubot label --dry-run`
against the worst issue we can find first: dry run still performs the full analysis and
only skips the write, so it measures the turn budget without touching anything. If 40 is
short the run hits the cap and the job fails visibly rather than applying a partial label
set.

chuanqi129:
@torchxpubot review

chuanqi129:
All eight verified against the source; every one held up. Fixed in 710eee83.

### The pin did not pin

You are right, and this was the worst kind of wrong: a control I documented as a
security boundary that does not hold. Reproduced exactly as you described --

```
$ printf 'label-issue: intel/torch-xpu-ops#1\nx\nlabel-issue: intel/torch-xpu-ops#4752\n' > t.md
$ grep -qE 'label-issue:[[:space:]]*intel/torch-xpu-ops#4752([^0-9]|$)' t.md && echo PASS
PASS
$ python3 -c "import re;print(re.search(r'label-issue:\s*([^\s#]+)#(\d+)', open('t.md').read()).group(0))"
label-issue: intel/torch-xpu-ops#1
```

The grep asked "does a matching header exist", the parse asks "what is the first
header". I even tested the `#77`/`#770` prefix case and still missed that the two halves
were answering different questions.

Taken your better option: `apply_labels.py` now has `--expect <repo>#<id>` and compares
the value the parse actually resolved, exiting 2 on mismatch. That also moves the control
out of untested YAML shell and under `test_bot_apply_labels.py`, which covers the
injected-header case, a foreign repo, and the neighbouring numbers a substring comparison
would accept.

### `issues: write` on the analysis agent

Also right, and self-inflicted: I added that scope last round to answer a "permissions not
declared" finding, and in doing so broke the split the job exists to implement. Now
`issues: read`. Every write in the job runs through `MERGE_TOKEN` on a later step, which
does not consult job permissions, so nothing needed it.

### `label` and the fold

Confirmed: `label-issue` mentions `VERDICT` nowhere, and `output_format.md` already wraps
the whole artifact in a `<details>`. So the fold never fired for `label`, and if it had it
would have double-wrapped. Dropped `label` from the fold rather than editing a third
skill -- its output is already collapsed by design. The PR description no longer claims
all three fold.

### The rest

- **Global `\n{3,}`**: scoped to the join point. A test now asserts a stack trace's blank
  runs survive.
- **Analysis lost on apply failure**: the comment is `always()` now, so a refused or
  failed apply still delivers the proposed labels -- including the pin refusal.
- **Dry run vs. live**: the footer says which, and the comment carries
  `apply_labels.py`'s own output, so the allowlist drops and the cap are visible instead
  of living in the job log while the model's unfiltered proposal is what gets posted.

`--max-turns 40` is still unvalidated. Per your recommendation the first live run will be
`--dry-run` on a real multi-group issue, now that the pin is real.

.github/scripts/apply_labels.py
```diff
@@ -444,10 +444,28 @@ def main():
...
     repo, issue_id, rows, full_text = parse_labels_md(args.labels_md)
+
+    # labels.md is written by an agent that read attacker-authored issue text, so
+    # the target it names is untrusted. A caller that already knows which issue
+    # the run is for passes --expect and the mismatch stops here, before any
+    # write. The comparison is against the value parse_labels_md actually
+    # resolved -- checking that a matching header exists somewhere in the file
+    # would pass for a labels.md carrying an injected header above the real one,
+    # since the parse binds to the first match.
+    if args.expect:
```
daisyden, line 461: The labels are added by script, do we really need this expect augument? Shall we just check the issue_id in label.md before writing the label?

.github/workflows/bot.yml
```diff
@@ -995,11 +1032,21 @@ jobs:
             const trigger = context.payload.comment;
             const header = `> Replying to [this comment](${trigger.html_url}) by @${trigger.user.login}\n\n`;
             const footer = `\n\n---\n_Generated by [\`ut-check\` job](${jobUrl})._`;
+            // The checkout carries no ref, so this resolves against the default
```
daisyden, line 1035: Cannot understand the comments. Please help to explain what is the relationship of folding with checkout.

.github/workflows/bot.yml
daisyden, line 1022: What is the steps.ai-analysis.outputs.execution_file? It is a json file?

PR #5355, opened 2026-09-14
[XPU] Reject sparse CSR out for strided add

### LINKED ISSUE #5343: [xpu-alignment] [26-09-13] torch.add(dense, dense, out=sparse_csr) aborts the process on XPU instead of raising
## Summary

On XPU, `torch.add(a, b, out=out)` with strided `a` and `b` and a valid sparse CSR `out` terminates the process with SIGSEGV and no Python-level exception. Upstream CPU raises a catchable (if unpolished) `RuntimeError: ... INTERNAL ASSERT FAILED`, so an XPU user gets strictly less than upstream: no diagnosable error at all, and no chance to handle it.

The layout combination itself need not be supported. The defect is that `add_out_sparse_compressed_xpu` re-dispatches instead of rejecting it, so the call never terminates.

## Upstream source

- pytorch/pytorch#196906, "torch.add with dense inputs and valid sparse CSR out triggers INTERNAL ASSERT FAILED", created 2026-09-13T15:45:32Z, reported on CPU/CUDA build 2.13.0+cu126. Still OPEN with no linked fix PR or cross-referenced commit as of 2026-09-14.
- Found by an xpu-alignment scan of the half-open window `[2026-09-13T00:00:00Z, 2026-09-14T00:00:00Z)` against the `pytorch/pytorch` default-branch snapshot `ad9448bc18658ff31d7512c2e5368f0718cf1ea0`.

## Observed XPU behavior

Runner result: `returncode -11` (SIGSEGV), `timed_out: false`, 8.049 s, no `RESULT` token, no traceback.

```text
torch_version: 2.15.0.dev20260912+xpu
xpu_available: True
xpu_device_name: Intel(R) Arc(TM) Pro B60 Graphics
control_layout: torch.sparse_csr
control_device: xpu:0
control_dense: [[3.0, 0.0], [0.0, 3.0]]
a_layout_device: torch.strided xpu:0
b_layout_device: torch.strided xpu:0
out_layout_device: torch.sparse_csr xpu:0
marker: calling torch.add(dense, dense, out=sparse_csr)
```

## Target-path evidence

- `aten/src/ATen/native/native_functions.yaml` at the snapshot dispatches `add.out` for `SparseCsrXPU` to `add_out_sparse_compressed_xpu`, which this repository implements.
- The control call (sparse CSR + sparse CSR) ran first and returned `torch.sparse_csr` on `xpu:0` densifying to `[[3.0, 0.0], [0.0, 3.0]]`, so the `SparseCsrXPU` add path is reachable and functional on this device. No XPU-to-CPU fallback warning was emitted for `add` (the one fallback warning in the log comes from sparse CSR construction).
- The operand layouts and devices were printed immediately before the call, and the reproducer runs under `python -u`, so the pre-call marker being the last line places the abort inside the target call, not in setup and not after it.

## Root cause

`src/ATen/native/sparse/xpu/SparseCsrTensorMath.cpp:400`:

```cpp
Tensor& add_out_sparse_compressed_xpu(
    const Tensor& self,
    const SparseCsrTensor& other,
    const Scalar& alpha,
    SparseCsrTensor& out) {
  if (self.layout() == kStrided) {
    at::add_out(out, self, other.to_dense(), alpha);
    return out;
  } else if (other.layout() == kStrided) {
    at::add_out(out, other, self.to_dense(), alpha);
    return out;
  }
```

`Tensor::to_dense()` on a strided tensor returns that tensor unchanged, so when both inputs are strided and `out` is sparse CSR the re-dispatch of `at::add_out` selects `SparseCsrXPU` again (the key comes from `out`) with identical layouts. That recurses without bound and exhausts the native stack, which Python cannot convert into `RecursionError`; the reproducer catches `RecursionError` explicitly and it never fired.

The log contains no native backtrace, so the recursion mechanism is inferred from the source above; what the run directly establishes is an uncatchable process abort inside the target call.

The second branch has the same trap: `add(sparse_csr, dense, out=sparse_csr)` enters `else if (other.layout() == kStrided)`, re-dispatches with two strided inputs, and then takes the first branch. Both branches need the guard.

Suggested fix: `TORCH_CHECK` the supported `out` layout (strided when either input is strided) before re-dispatching, so the unsupported combination produces a user-facing error instead of unbounded re-entry.

## Ownership

Upstream CPU takes a different branch for the same inputs: `add_out_sparse_compressed_cpu` calls `add_out_dense_sparse_compressed_cpu(out, self, other, alpha)`, whose `TORCH_INTERNAL_ASSERT(src.layout() == kSparseCsr || src.layout() == kSparseCsc)` is the assertion pytorch/pytorch#196906 reports. Fixing that CPU entry point, or improving its error message, does not touch `add_out_sparse_compressed_xpu`, which is separate code in this repository. No `pytorch/pytorch` issue or PR owns the XPU function, and a search of `intel/torch-xpu-ops` (sparse csr add, add_out_sparse_compressed, SparseCsrTensorMath, xpu-alignment) found no existing tracker.

## Reproducer

```python
import torch

def make_csr(values):
    return torch.sparse_csr_tensor(
        torch.tensor([0, 1, 2], device="xpu"),
        torch.tensor([0, 1], device="xpu"),
        values,
        size=(2, 2),
        check_invariants=True,
        device="xpu",
    )

# Control: the SparseCsrXPU add path works.
control = torch.add(
    make_csr(torch.ones(2, device="xpu")),
    make_csr(torch.full((2,), 2.0, device="xpu")),
)
print(control.to_dense().cpu().tolist())

a = torch.ones((2, 2), device="xpu")
b = torch.ones((2, 2), device="xpu")
out = make_csr(torch.zeros(2, device="xpu"))
print("calling torch.add(dense, dense, out=sparse_csr)", flush=True)
torch.add(a, b, out=out)  # SIGSEGV on XPU; RuntimeError (INTERNAL ASSERT) on CPU
```

## Environment

- `torch 2.15.0.dev20260912+xpu` at `/opt/conda/lib/python3.13/site-packages/torch/__init__.py`
- Python 3.13.13 (`/opt/conda/bin/python3.13`), run as `python3.13 -I -u`
- `xpu_available: true`, device `Intel(R) Arc(TM) Pro B60 Graphics`, no environment warnings
- Run artifacts: `scripts/repro_issue-196906.py` (sha256 `b730db66dd4aba7eec8b593baedbffb1b8935563f11b140dbd43dcd3b80b57b9`), `runner/logs/issue-196906.log` (sha256 `2460b822d987aa98c2387221e0a28da94a33c1de72376f16758e355217f27528`)

laifenxiawucha:
## Background

When either input to XPU sparse-CSR `add.out` is strided, the implementation densifies the sparse operand and redispatches. If `out` is still sparse CSR, dispatch selects the same XPU function again and recurses until the process terminates with SIGSEGV.

## Upstream

- [pytorch/pytorch#196906](https://github.com/pytorch/pytorch/issues/196906) reports the unsupported dense-input/sparse-output combination on CPU, where it reaches an internal assertion. XPU has a separate redispatch path whose failure is unbounded recursion, so the guard belongs here.

## Fix

- require a strided `out` whenever either input is strided
- retain the supported sparse-input/strided-output and sparse-input/sparse-output paths
- add coverage for rejected and supported layout combinations

## Before and after

| Case | Before | After |
|---|---|---|
| Strided input with sparse CSR `out` | Native-stack recursion and SIGSEGV | Checked `RuntimeError` |
| Supported output layouts | Work | Still work |

## Testing

Linux XPU `op_regression` in [CI run 34826679782](https://github.com/intel/torch-xpu-ops/actions/runs/34826679782): 3 passed, 0 failed.

Fixes #5343


laifenxiawucha:
Implemented and opened from latest `main`. The guard prevents recursive `SparseCsrXPU` redispatch for strided inputs with sparse CSR `out`; tests cover all three rejected layouts, both dense-output operand orders, and sparse+sparse output. Local XPU reproduced the pre-fix SIGSEGV (return code 139); source-build CI validation is pending.

Issue #5341, opened 2026-09-14
[BMG] [UT] test_for_overlapping_kernels.py::TestOverlappingKernels::test_overlapping_kernels AssertionError: Scalars are not equal!


RUIJIEZHONG66166:
### 🐛 Describe the bug

During nightly test, this profile UT case failed with below log:

```
============================= test session starts ==============================
platform linux -- Python 3.10.21, pytest-7.3.2, pluggy-1.6.0 -- /__w/torch-xpu-ops/torch-xpu-ops/.venv/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/.hypothesis/examples')
rootdir: /__w/torch-xpu-ops/torch-xpu-ops/pytorch
configfile: pytest.ini
plugins: xdist-3.3.1, subtests-0.13.1, rerunfailures-14.0, cpp-2.3.0, hypothesis-6.56.4, timeout-2.4.0, xdoctest-1.3.0, flakefinder-1.1.0
timeout: 3600.0s
timeout method: thread
timeout func_only: False
collecting ... collected 1 item

test/profiling/test_for_overlapping_kernels.py::TestOverlappingKernels::test_overlapping_kernels Found 951 overlapping kernel pairs!
Top 10 longest overlaps:

Overlap #1:
  Kernel 1: ProfilerStep#19 (gpu_user_annotation) - [5877365641832.521 -> 5877367057470.267]
  Kernel 2: at::native::xpu::MaxPool2dBackwardDeterministicKernelFunctor<float, float, false, int> (kernel) - [5877367005647.266 -> 5877367019314.869]
  Overlap duration: 13667.603515625 units

Overlap #2:
  Kernel 1: ProfilerStep#19 (gpu_user_annotation) - [5877365641832.521 -> 5877367057470.267]
  Kernel 2: at::native::xpu::BatchNormBackwardKernelFunctor<16, float, float, float, int> (kernel) - [5877366937863.364 -> 5877366950277.53]
  Overlap duration: 12414.166015625 units

Overlap #3:
  Kernel 1: ProfilerStep#19 (gpu_user_annotation) - [5877365641832.521 -> 5877367057470.267]
[...]

PR #5363, opened 2026-09-14
Fix device activity filtering in overlapping kernels test

PawelSwider2000:
Use exact Kineto categories when selecting device activities in the overlapping kernels test.

This prevents categories such as `gpu_user_annotation` from being incorrectly treated as kernels due to substring matching. The test now recognizes only `kernel`, `gpu_memcpy`, and `gpu_memset`.

The fix for: #5341


PawelSwider2000:
CI failures are unrelated

test/profiling/test_for_overlapping_kernels.py
```diff
@@ -28,7 +28,7 @@ def extract_kernels(trace_data):
     kernels = []
 
     # GPU-related categories to include
-    gpu_categories = {"fwdbwd", "ac2g", "kernel", "gpu", "cuda", "hip"}
```
BBBela, line 31: Why do we remove the `"fwdbwd"` and `"ac2g"`? Were those categories unnecessarily added in the first place or something changed in the meantime? Just want to clarify.

BBBela (commented):
Generally the change looks good to me - I just have one comment below for clarification 😉

test/profiling/test_for_overlapping_kernels.py
PawelSwider2000: Yes, they was added unnecessarily

BBBela (approved):
Thanks!

pbielak (approved):

PawelSwider2000:
@torchxpubot merge -f "unrelated CI failures"

PR #5365, opened 2026-09-14
Give the execution-trace pt2 tests a private inductor cache dir

### LINKED ISSUE #5121: [Bug Skip]: [Regression] FileNotFoundError: [Errno 2] No such file or directory: '/tmp/torchinductor_root/r3/cr3z25rxexxnkmt5ftijiorxcqsimwwb56y2 in test_execution_trace_xpu.py
### 🐛 Describe the bug with skip template

Cases:
~~op_ut,third_party.torch-xpu-ops.test.xpu.profiler.test_execution_trace_xpu.TestExecutionTraceXPU,test_execution_trace_env_enabled_with_pt2_xpu~~

### FileNotFoundError: [Errno 2] No such file or directory: '/tmp/torchinductor_root/r3/cr3z25rxexxnkmt5ftijiorxcqsimwwb56y25efpcvhpvsaf4x4m.py'

```
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/profiler/test_execution_trace_xpu.py", line 508, in test_execution_trace_env_enabled_with_pt2
    fn(*inputs)
  File "/__w/torch-xpu-ops/torch-xpu-ops/.venv/lib/python3.10/site-packages/torch/_dynamo/eval_frame.py", line 1278, in compile_wrapper
    raise e.remove_dynamo_frames() from None  # see TORCHDYNAMO_VERBOSE=1
  File "/__w/torch-xpu-ops/torch-xpu-ops/.venv/lib/python3.10/site-packages/torch/_inductor/compile_fx.py", line 1311, in _compile_fx_inner
    raise InductorError(e, currentframe()).with_traceback(
  File "/__w/torch-xpu-ops/torch-xpu-ops/.venv/lib/python3.10/site-packages/torch/_inductor/compile_fx.py", line 1303, in _compile_fx_inner
    mb_compiled_graph = fx_codegen_and_compile(
  File "/__w/torch-xpu-ops/torch-xpu-ops/.venv/lib/python3.10/site-packages/torch/_inductor/compile_fx.py", line 2158, in fx_codegen_and_compile
    return scheme.codegen_and_compile(gm, example_inputs, inputs_to_check, graph_kwargs)
  File "/__w/torch-xpu-ops/torch-xpu-ops/.venv/lib/python3.10/site-packages/torch/_inductor/compile_fx.py", line 2073, in codegen_and_compile
    return CompiledFxGraph(
  File "/__w/torch-xpu-ops/torch-xpu-ops/.venv/lib/python3.10/site-packages/torch/_inductor/output_code.py", line 604, in __init__
    with open(graph.cache_path) as f:
torch._inductor.exc.InductorError: FileNotFoundError: [Errno 2] No such file or directory: '/tmp/torchinductor_root/r3/cr3z25rxexxnkmt5ftijiorxcqsimwwb56y25efpcvhpvsaf4x4m.py'

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"


To execute this test, run the following from the base repo dir:
    PYTORCH_TEST_WITH_SLOW=1 python test/xpu/profiler/test_execution_trace_xpu.py TestExecutionTraceXPU.test_execution_trace_env_enabled_with_pt2_xpu

This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0
```

## Pytorch Version

Detected in : [nightly #32733042537 / op_ut](https://github.com/intel/torch-xpu-ops/actions/runs/32733042537/job/97468718823) (2026-08-24)  
latest good : unknown  
current     : unknown

**Regression evidence**

These cases passed in the previous healthy nightly for their category and fail now.

| Category | | Run | Date | torch | torch-xpu-ops |
|---|---|---|---|---|---|
| op_ut | Last good | [#32642064357 (op_ut)](https://github.com/intel/torch-xpu-ops/actions/runs/32642064357/job/97205902557) | 2026-08-23 | unknown | unknown |
| op_ut | First seen bad | [#32733042537 (op_ut)](https://github.com/intel/torch-xpu-ops/actions/runs/32733042537/job/97468718823) | 2026-08-24 | unknown | unknown |

This is a bisect *range*, not a culprit commit.

### Versions

<details><summary>Detail</summary>

collect_env output was not captured in this run's artifact.

</details>

astachowiczhabana:
Fixes #5121

## Root cause

`test_execution_trace_env_enabled_with_pt2` and `test_triton_fx_graph_with_et`
both began with `PyCodeCache.cache_clear(purge=True)`, to force the compile that
follows to be a cache miss so a triton kernel is really generated for the trace to
capture.

That purges the **per-user** inductor cache root (`/tmp/torchinductor_<user>/`),
which every pytest-xdist worker in a CI run shares, and whose entries are **content
addressed**: `codecache.py`'s `write()` hashes the source and reuses the existing
file when the hash matches, so two workers generating byte-identical wrapper code
hold the same path. `cache_clear(purge=True)` `os.remove()`s every module the
calling process has loaded, with no cross-process guard, and CI runs with
`--dist worksteal`. A worker part-way through a compile then loses the file under
it: `CompiledFxGraph.__init__` re-reads `graph.cache_path` off disk right after
`graph.py` wrote and loaded it, and that read raises the `FileNotFoundError`
reported in #5121, killing an otherwise-successful compile.

## The fix

Wrap each test's compile and profiling in `torch._inductor.utils.fresh_cache()`,
already the idiom throughout `test/xpu/dynamo/test_aot_autograd_cache_xpu.py`. It
points `TORCHINDUCTOR_CACHE_DIR` at a private tempdir for the duration, so the
test's artifacts are no longer in a directory another worker can purge, and the
test no longer purges a directory another worker is using -- `fresh_cache()`'s own
entry/exit cleanup is `clear_caches()`, i.e. `cache_clear()` with `purge=False`,
and the `os.remove()` in `codecache.py` sits under `if purge:`, so nothing on disk
is touched.

It also serves the original intent better than the purge did: an empty private dir
is a *guaranteed* miss, whereas a purged shared dir can be repopulated by another
worker between the purge and the compile.

Ignoring the re-indentation, the change is one import, two deleted purges and two
`with fresh_cache():` lines.

## Test plan

Two real processes, no monkeypatching, using only public `PyCodeCache` API in the
order `graph.py` and `output_code.py` use it, with a rendezvous between the write
and the re-read:

```bash
xpu gpu python /work/experiments/issue-181/race_two_process.py
```

```
=== BEFORE fix: reader uses the shared default cache dir ===
  same file on disk (content-addressed sharing): True
  READER OUTCOME  : FileNotFoundError: [Errno 2] No such file or directory: '/tmp/.../cri6cthbn7h6....py'

=== AFTER  fix: reader wraps its work in fresh_cache() ===
  same file on disk (content-addressed sharing): False
  READER OUTCOME  : ok
```

The changed file, including under the parallel configuration CI uses:

```bash
pytest third_party/torch-xpu-ops/test/xpu/profiler/test_execution_trace_xpu.py::TestExecutionTraceXPU::test_execution_trace_env_enabled_with_pt2_xpu -v --timeout=600
pytest third_party/torch-xpu-ops/test/xpu/profiler/test_execution_trace_xpu.py::TestExecutionTraceXPU::test_triton_fx_graph_with_et_xpu -v --timeout=600
pytest third_party/torch-xpu-ops/test/xpu/profiler/test_execution_trace_xpu.py -v --timeout=600
pytest third_party/torch-xpu-ops/test/xpu/profiler/test_execution_trace_xpu.py --timeout=600 -n 2 --dist worksteal
```

13 passed serially, 13 passed under two workers. Both changed tests still find a
`triton_` node with a non-empty `kernel_file` attribute, so the compile is still a
real miss, and `test_triton_fx_graph_with_et` still matches its expected FX graph
text from the collected resource files. `lintrunner` on the changed file: no
issues. Device: Intel(R) Arc(TM) Pro B65 Graphics, torch `2.15.0a0+git28789c4`.

> Authored with AI assistance on Intel GPU hardware. The diff and the test output above are that run's; the
> submitter is responsible for reading the change before this is opened.


BBBela (commented):
From what I know the pytest-xdist problem is only torch-xpu-ops specific as PyTorch CI is launched different way. So the problem is that torch-xpu-ops launch CI tests in a different way than PyTorch does - not some regression or bug in XPU implementation.

Also, the mirroring test file from upstream is marked as DONE in terms of generalization of tests. If it is indeed already generalized, then I suggest we remove the local torch-xpu-ops copy of this test file and use upstream test file directly. Then this PR can be created directly in PyTorch as improvement.

If this PR lands, we won't be able to launch this test file directly from upstream as there will be the divergence introduced by this PR.

Issue #5291, opened 2026-09-10
test_foreach_xpu.py fail to collect on BMG


kaileiyx:
### 🐛 Describe the bug

Cases:
op_ut,,third_party.torch-xpu-ops.test.xpu.test_foreach_xpu

## ErrorLog
```bash
==================================== ERRORS ====================================
___ ERROR collecting third_party/torch-xpu-ops/test/xpu/test_foreach_xpu.py ____
ImportError while importing test module '/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/test_foreach_xpu.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/github/home/.local/share/uv/python/cpython-3.10-linux-x86_64-gnu/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_foreach_xpu.py:51: in <module>
    from test_foreach import TestForeach
E   ImportError: cannot import name 'TestForeach' from 'test_foreach' (/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/../../../../test/test_foreach.py)
- generated xml file: /__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/op_ut_with_skip.test_foreach_xpu.xml -
=========================== short test summary info ============================
ERROR test_foreach_xpu.py
```

## Pytorch
latest good : [ca66d98](https://github.com/pytorch/pytorch/commit/ca66d9844119d868477faa88c782bd46495e1783)
current :  [81deebd](https://github.com/pytorch/pytorch/commit/81deebd0497f96495b563af4155b99b372d8022c)
## Torch-xpu-ops
latest good : https://github.com/intel/torch-xpu-ops/commit/d6fa1aaa5d0c5100624017d22ae7932f0bd9389f
current :  5cdd4e46bd058506f1620944b27b87056e415cc9

### Versions

<details>
<summary>Detail</summary>
 Collecting environment information...
  PyTorch version: 2.15.0a0+git81deebd
  Is debug build: False
  CUDA used to build PyTorch: None
  ROCm SDK used to build PyTorch: N/A
  HIP used to build PyTorch: N/A
  
  [pip3] intel-opencl-rt==2026.1.0
  [pip3] intel-openmp==2026.1.0
[...]

PR #5366, opened 2026-09-14
Switch to upstream tests

pbielak:
Several upstream tests have been refactored, so this PR uses the updated upstream test files instead of the local overrides:

- `test_indexing_xpu.py` (fix for: https://github.com/intel/torch-xpu-ops/issues/5340 )
- `nn/test_convolution_xpu.py` (fix for: https://github.com/intel/torch-xpu-ops/issues/5259 )
- `test_foreach_xpu.py` (fix for: https://github.com/intel/torch-xpu-ops/issues/5291 )

In all cases the test coverage is at least as high as before - executing the upstream files adds some extra CPU tests, but the running time is not substantially affected.


test/xpu/skip_list_win_lnl.py
```diff
@@ -6,6 +6,8 @@
 #
 # http://www.apache.org/licenses/LICENSE-2.0
 
+PYTORCH_TEST_DIR = "..\\..\\..\\..\test"
```
BBBela: Are backward slashes necessary here? I thought python generally handles forward slashes on Windows well when used with `os` and `pathlib` libraries. Just asking to make sure if it is necessary here.

test/xpu/run_test_with_windows_nighltly.py
```diff
@@ -8,6 +8,8 @@
 res = 0
 IS_WINDOWS = sys.platform == "win32"
 
+PYTORCH_TEST_DIR = "../../../../test"
+
```
BBBela: The `PYTORCH_TEST_DIR` is repeatedly defined several times throughout the "`skip_list_<...>`" files.
I am wondering if we could define it once and import in other places - not to repeat the definition. What do you think?

BBBela (commented):
Beside those two small comments below, the rest looks good to me 😉

test/xpu/skip_list_win_lnl.py
pbielak: Thanks, updated

test/xpu/run_test_with_windows_nighltly.py
pbielak: Done

BBBela (approved):
Thanks!

AKloniecki (approved):
LGTM

pbielak:
@torchxpubot merge -f "CI failures are unrelated"

PR #5373, opened 2026-09-15
Avoid backlinking PyTorch PR in cost tracker

CuiYifeng:
Write the '#' as the HTML entity `&#35;` instead to avoid backlinking.


chuanqi129 (approved):

CuiYifeng:
@torchxpubot rebase

PR #5393, opened 2026-09-16
[Workflow] Add static alignment verification

laifenxiawucha:
## Purpose

Add a deterministic static verification path for functional and API divergences that can be established from source at fixed commits. These findings no longer depend on an XPU runtime, while runtime-based findings keep the existing reproducer flow.

This change was split from #5285. That PR retains closed-tracker handling and the 120-second XPU environment-probe timeout.

## Why this change is needed

The alignment pipeline previously required every validated unit to contain a Python reproducer and runner result. That is appropriate for behavioral and performance claims, but it is the wrong evidence model for differences such as:

- an upstream helper that moved or changed signature;
- an error string or validation rule that changed upstream;
- a check that XPU still keeps as a private source copy.

Run `34302439189` exposed the failure mode: `commit-78fa18d` identified a source divergence, but the nightly runtime predated the upstream commit. The unit became `blocked-env`, and a source-only finding was dropped even though no device execution was needed to decide it.

### Preparation contract

- Add an optional `verification` field with `runtime` as the backward-compatible default and `static` as the new mode.
- A static entry omits `script`, `script_sha256`, and `timeout_seconds`.
- It records immutable `upstream_source` and `xpu_source` objects containing repository, exact 40-character commit, original path, snapshot path, and SHA-256.
- Static verification is limited to functional and API source differences. Performance claims still require runtime evidence.

### Runner behavior

- Filter static entries out of the executable plan while preserving execution coverage checks.
- Mixed plans execute only their runtime entries.
- Static-only plans skip user switching, torch/XPU probing, and nightly XPU installation; they produce `environment: null` and an empty runtime result list.

### Deterministic gate

- Validate repository, commit, path, snapshot location, digest, and the frozen upstream/XPU commit coordinates.
- Reject identical upstream and XPU snapshots, unexpected scripts on static entries, unknown verification modes, and mismatched evidence paths.
- Allow only `confirmed` or `not-reproduced` for static units. `related-failure` and `blocked-*` remain runtime-only outcomes.
- Require the scan artifact to reference the exact two immutable snapshots recorded during preparation.

### Workflow and independent review

- Persist the `evidence/` directory with the preparation artifact.
- Detect whether a plan contains runtime entries and provision the XPU runtime only when needed.
- Pass the checked-out XPU commit into the gate so source coordinates cannot drift.
- Require the independent reviewer to re-fetch both files at their recorded repository, commit, and path. Insufficient or inconsistent evidence becomes `verification-gap` and produces no issue payload.
- Static payloads contain source coordinates, snapshots, oracle, and the observed difference; runtime payloads continue to contain the reproducer, log, and environment.

## Before and after

| Scenario | Before | After |
|---|---|---|
| Source-only functional/API divergence | Required a runtime reproducer | Uses immutable upstream and XPU source evidence |
| Upstream commit is newer than the installed nightly | Could become `blocked-env` and lose the finding | Can be decided from the frozen source snapshots |
| Static-only plan | Installed nightly XPU packages and probed a device | Skips runtime provisioning and records `environment: null` |
| Mixed static/runtime plan | Every unit was executed | Only runtime entries execute; static entries retain source evidence |
| Static evidence disproves the claim | No source-specific terminal result | Records `not-reproduced` |
| Static evidence is incomplete | Could be forced into a runtime blocker | Records `verification-gap` with no payload |
| Entry without `verification` | Runtime execution | Remains runtime execution for compatibility |

## Validation

```bash
python3 -m pytest .github/scripts/ -q
PYTHONPATH=.github/scripts python3 -m py_compile .github/scripts/xpu_alignment_gate.py .github/scripts/xpu_alignment_runner.py
python3 -c "import pathlib,yaml; yaml.safe_load(pathlib.Path('.github/workflows/xpu_alignment.yml').read_text())"
lintrunner -a -m origin/main --skip CLANGTIDY
git diff --check
```

Result: 11 repository tests passed, both scripts compiled, the workflow YAML parsed, and lintrunner reported no issues. A static-only execution-plan check also confirmed that the runner skips environment probing and writes `environment: null` with no runtime results.


laifenxiawucha:
A strong motivating example for `verification: static` is Run #12: [XPU Alignment run 34302439189](https://github.com/intel/torch-xpu-ops/actions/runs/34302439189), which validated upstream [PyTorch PR #196230](https://github.com/pytorch/pytorch/pull/196230), commit [`78fa18d`](https://github.com/pytorch/pytorch/commit/78fa18d91b0d762f510d64053d01d265562a3cd4).

That upstream commit removed `adaptive_pool_empty_output_check` from `AdaptivePooling.h`, while the XPU implementation still referenced it at `src/ATen/native/xpu/AdaptiveAveragePooling2d.cpp:32`. This is a source compatibility problem.

Even if the nightly is current and contains the upstream commit, the combined PyTorch/XPU source would fail during the XPU extension build before any runtime reproducer could start. The two appropriate ways to detect this are therefore:

- a compile/integration check using the upstream and XPU revisions together; or
- a static check that reads the frozen source coordinates and confirms that one revision deleted the helper while the other still references it.

A normal runtime reproducer can only help after the combined sources compile. It may be useful for the separate pooling shape-check or error-message behavior, but it cannot establish this deleted-symbol compatibility failure.

PR #5413, opened 2026-09-17
Disable flight recorder by default

zhangxiaoli73:
In vLLM end-to-end model tests, disabling Flight Recorder (TORCH_FR_BUFFER_SIZE=0) improved output throughput by 4.1%-8.6% across four models in FP8 TP2 1K/512 eager-mode workloads on vLLM 0.29.0.

Therefore, disable Flight Recorder by default so users benefit from this performance improvement without additional configuration.


src/xccl/ProcessGroupXCCL.cpp
frost-intel, line 407: Would recommend a log here to indicate that FR is off by default, and the env var to turn it on:

frost-intel (approved):
I don't have any issue with this, seems reasonable to disable this in every inference case. UTs set this var already so that shouldn't cause any issues.

zhangxiaoli73:
@torchxpubot  merge

Chao1Han (approved):

Chao1Han:
@torchxpubot merge -f "distributed cases passed"

PR #5416, opened 2026-09-17
Fix mha_bwd empty-batch abort

CuiYifeng:
This PR is to solve https://github.com/pytorch/pytorch/issues/196748.

The empty-batch early return in mha_bwd is after the dout contiguity check `dout.stride(-1) == 1`. The backward over an empty attention output feeds mha_bwd a stride-0 `dout` whose stride(-1) is 0, which triggers the abort `mha_bwd on xpu: dout tensor must have contiguous last dimension`.

Solution: move empty-batch early return ahead of the contiguity checks.


chunhuanMeng (approved):

guangyey (approved):

guangyey:
@torchxpubot merge

Issue #5073, opened 2026-08-24
AttributeError: type object 'TestUnaryUfuncs' has no attribute 'test_tan_complex_cuda_matches_numpy' in test_unary_ufuncs_xpu.py


kaileiyx:
### 🐛 Describe the bug

Cases:
op_ut,,third_party.torch-xpu-ops.test.xpu.test_unary_ufuncs_xpu

## ErrorLog
```bash
==================================== ERRORS ====================================
_ ERROR collecting third_party/torch-xpu-ops/test/xpu/test_unary_ufuncs_xpu.py _
Traceback (most recent call last):
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/test_unary_ufuncs_xpu.py", line 67, in <module>
    TestUnaryUfuncs.test_tan_complex_cuda_matches_numpy
AttributeError: type object 'TestUnaryUfuncs' has no attribute 'test_tan_complex_cuda_matches_numpy'. Did you mean: 'test_tan_complex_matches_numpy'?
_ ERROR collecting third_party/torch-xpu-ops/test/xpu/test_unary_ufuncs_xpu.py _
Traceback (most recent call last):
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/test_unary_ufuncs_xpu.py", line 67, in <module>
    TestUnaryUfuncs.test_tan_complex_cuda_matches_numpy
AttributeError: type object 'TestUnaryUfuncs' has no attribute 'test_tan_complex_cuda_matches_numpy'. Did you mean: 'test_tan_complex_matches_numpy'?
_ ERROR collecting third_party/torch-xpu-ops/test/xpu/test_unary_ufuncs_xpu.py _
Traceback (most recent call last):
  File "/__w/torch-xpu-ops/torch-xpu-ops/pytorch/third_party/torch-xpu-ops/test/xpu/test_unary_ufuncs_xpu.py", line 67, in <module>
    TestUnaryUfuncs.test_tan_complex_cuda_matches_numpy
AttributeError: type object 'TestUnaryUfuncs' has no attribute 'test_tan_complex_cuda_matches_numpy'. Did you mean: 'test_tan_complex_matches_numpy'?
_ ERROR collecting third_party/torch-xpu-ops/test/xpu/test_unary_ufuncs_xpu.py _
Traceback (most recent call last):
[...]

PR #5439, opened 2026-09-18
Update overrides in `test_unary_ufuncs_xpu.py`

pbielak:
Upstream generalized several test methods in `test_unary_ufuncs.py` along with method name changes - see [1]. This commit updates the local overrides and removes the obsolete ones.

Provides a fix for: https://github.com/intel/torch-xpu-ops/issues/5073

[1] https://github.com/pytorch/pytorch/pull/188656


dominikjastrzebskix (approved):

BBBela (approved):
Shouldn't the mirroring upstream test file be enabled for XPU first? Then we could launch the upstream test file directly without our local wrapper, right?

I see it is marked as done in the generalization Excel, while it is not enabled for XPU in upstream.

I mean, if this PR is to temporarily simplify this local copy of test file, before enabling XPU in the upstream, then it looks good to me, but I believe it is not the target solution.

Approving as the change itself is okay, but leaving the above comment for consideration.

pbielak:
There is a pending PR in upstream: https://github.com/pytorch/pytorch/pull/197782

but I will merge this PR to unblock `test_unary_ufuncs_xpu.py`, so it gets properly tested again.

pbielak:

PR #5445, opened 2026-09-18
Re-enable test_grad_scaling_state_dict_xpu

### LINKED ISSUE #5375: [upstream_ut] test_grad_scaling_state_dict_xpu: _scale is not torch.FloatTensor on XPU
## Description
test_grad_scaling_state_dict_xpu in test/test_torch.py fails. The test asserts
that GradScaler._scale is a torch.FloatTensor, but on XPU the scale tensor is
not an instance of torch.FloatTensor (CPU float storage type).

## Error log
```
  File "test/test_torch.py", line 5649, in test_grad_scaling_state_dict
    self.assertTrue(isinstance(s1._scale, torch.FloatTensor))
AssertionError: False is not true
```

## Reproduce
```
python test/test_torch.py TestTorchDeviceTypeXPU.test_grad_scaling_state_dict_xpu
```

pponikox:
#### Summary:
This patch fixes intel/torch-xpu-ops#5375 by removing the stale skip so the test runs in CI again.
The XPU override of `test_grad_scaling_state_dict` in test_torch_xpu.py already handles the XPU scale-tensor check, and the test passes on XPU. Its entry in skip_list_common.py is now stale.

#### Validation (Intel BMG):
```
python test_torch_xpu.py TestTorchDeviceTypeXPU.test_grad_scaling_state_dict_xpu -v
```
* before: OK (run directly; skip entry present)
* after:  still OK (skip entry removed)


pponikox:
This solution is not working as expected (root cause is still not calling XPU test).
The problem must be resolved on the PyTorch side. Closing this PR.

PR #5450, opened 2026-09-20
Decouple attention dropout mask reconstruction from XPU generator

majing921201:
Attention backward currently reconstructs its dropout mask through a temporary XPU generator, directly configuring XPUGeneratorState. We will refactor the XPUGeneratorState in stock pytorch, current code will fail in the upcoming change.

To decouple XPUGenerator, We add an attention-specific `mem_eff_attention_dropout_mask_kernel` that accepts `PhiloxXpuState`, prepares tensor metadata and execution configuration, and invokes the existing dropout launcher. The existing `dropout` and `fused_dropout_kernel` implementations remain unchanged.


Chao1Han:
@torchxpubot check-ut

Chao1Han:
@torchxpubot check-ut

Chao1Han (approved):

majing921201:
@LuFinch Please help review this PR

CuiYifeng (commented):
From the design, `PhiloxXpuState(seed, offset)` should be equivalent to the previous temporary-generator path. Could we add or point to a test that verifies this equivalence (for the same hardware)?

majing921201:
> From the design, `PhiloxXpuState(seed, offset)` should be equivalent to the previous temporary-generator path. Could we add or point to a test that verifies this equivalence (for the same hardware)?

Yes, the dropout interface currently only accepts an XPUGenerator. Previously, we used the seed and offset to construct a generator, only for dropout to extract those same values again. This change passes the seed and offset directly to the kernel, removing the dependency on XPUGenerator.
The existing test for [test_transformers_xpu.py](https://github.com/intel/torch-xpu-ops/blob/6a2d048e453e0d63324c34a98dfeaf6921f433eb/test/xpu/test_transformers_xpu.py#L5070) will cover it.

CuiYifeng (approved):

PR #5457, opened 2026-09-21
[CI] Let dynamic skip function for xpu_profiling UT

RUIJIEZHONG66166:
### Problem

`check_profiling_ut` in `.github/scripts/ut_result_check.sh` decided the verdict purely from the pytest junit xml counters:

```bash
if [[ "$total_failures" -gt 0 ]] || [[ "$total_errors" -gt 0 ]] || [[ "$total_tests" -le 0 ]]; then
```

So even when every failure was already tracked as a known issue, the summary job still reported ❌ (e.g. [run 35513152275](https://github.com/intel/torch-xpu-ops/actions/runs/35513152275/job/106193312557)). The known-issue filtering it already ran was printed but never used, unlike `op_regression` / `op_regression_dev1` / `op_extended` / `op_ut` / `test_xpu`, which pass once all failures are known.

### Change

Align `xpu_profiling` with `run_main_tests`:

- List the filtered known issues (`grep -noFf Known_issue.log failures_xpu_profiling.log`), same output style as the other suites.
- Use `check_new_failed` output `failures_xpu_profiling_filtered.log` as the failure count for the final verdict. This is also the file the summary job's "Collect new failures with error messages" step reads.
- Keep the raw xml counters as a fallback when `failures_xpu_profiling.log` or `Known_issue.log` is missing (i.e. `check-ut.py` produced nothing to filter), so real breakage is still caught.
- `total_tests <= 0` still fails the suite, so a truncated / empty run cannot silently pass.


RUIJIEZHONG66166:
Ondemand test: https://github.com/intel/torch-xpu-ops/actions/runs/35557047453/job/106204012942. Skip function passed.

chuanqi129:
Can we add profiling ut test in ci scope also?

RUIJIEZHONG66166:
> Can we add profiling ut test in ci scope also?

Add now.

.github/scripts/ut_result_check.sh
mengfei25: After profiling to pytest (results in xml files), looks like we can remove this func and use run_main_tests directly

PR #5468, opened 2026-09-21
[xccl] Keep record_param_comms open at the launch

aostrowski-hbn:
A profiler attributes a GPU kernel to whichever CPU op is open when the kernel
is launched. `RECORD_PARAM_COMMS_DATA` is built for that: it declares a
`DebugInfoGuard` and the `RecordFunction` guard whose destructors end the event,
so it has to be invoked in the scope that should bound the event — which is how
`ProcessGroupNCCL` invokes it, bare, with `return collective(...)` running
inside the record.

`RECORD_PARAM_COMMS_DATA_WITH_LOG` wrapped it in `do { } while (0)`, closing
that scope before the collective was launched. The event ended empty and the
communication kernel was attributed to the enclosing `c10d::<op>_` instead, so
nothing that reads collective information off the kernel event could find it.

**Why CUDA works and XPU did not:** the same upstream macro, a different
invocation. NCCL leaves it bare at function scope; the XCCL wrapper added a
scope around it.

`all_to_all_single` needs a second change: its two records sit inside the split
branches while the collective is launched after them, so the guard died early
there even with the macro fixed. NCCL avoids this by returning
`collective(...)` from inside each branch; duplicating that whole call here
would be worse, so the branch-specific values are computed first and the record
is taken once at function scope. Each branch keeps its validation on the same
side of the record as before, so a failing check still emits what it did
previously. That is the only deliberate divergence from the CUDA structure, and
it produces the same record.

A DDP step with no `barrier()` in the captured region goes from 0 of 6
collective kernels carrying `Collective name` to 6 of 6.

Test: test/xpu/distributed/test_c10d_xccl.py

```
ZE_AFFINITY_MASK=0,1 pytest -v --timeout 600 \
    test/xpu/distributed/test_c10d_xccl.py \
    -k test_record_param_comms_scope_covers_launch
```

As a single line, for the reproducer job to pick up:

```
pytest test/xpu/distributed/test_c10d_xccl.py -k test_record_param_comms_scope_covers_launch -v
```

The new test fails without the fix, reporting that no `record_param_comms`
scope was open when `xccl:all_reduce` was launched. `test_c10d_ops_xccl.py` is
untouched and still passes.


tsocha (approved):
There is a little difference between versions but it's probably safe.
Before: pass new empty vector to logging macro.
Now: pass reference of empty vector to logging macro.

kdrozd-dev (approved):

kdrozd-dev:
@torchxpubot merge

aostrowski-hbn:
The two commits from #5469 have been moved onto this branch, so #5469 closed as merged into it. The tree here is exactly what #5469's CI had already validated — only the branch it sits on changed — so this is safe to land without waiting for the full rerun, and I am force-merging on that basis.

Why the move: `@torchxpubot merge` failed here with *"Merging stacked PRs via this endpoint is not supported. Use the asynchronous merge endpoint instead."* The bot uses the synchronous merge endpoint, which GitHub refuses for a PR that has a stacked child, and retargeting #5469 was rejected as well — *"Cannot change the base branch because the pull request is part of a stack"*. A follow-up PR will switch the bot to the async endpoint so the next stack does not need this workaround.

aostrowski-hbn:
@torchxpubot merge -f

aostrowski-hbn:
Superseded by #5486, which carries the same five commits on a single branch off `main` — the stack could neither be merged (GitHub refuses the synchronous merge endpoint for a stacked PR) nor un-stacked (its base cannot be retargeted). Closing this one.

PR #5479, opened 2026-09-22
Widen allowed_bots for UT auto-issue, and cover xpu_profiling

RUIJIEZHONG66166:
### `allowed_bots` widened
To fix "Action failed with error" when using auto-issue skill, issue log: https://github.com/intel/torch-xpu-ops/actions/runs/35605281528/job/106589102085
On a scheduled run `github.actor` is whoever last pushed the workflow, which is
rarely a person. `claude-code-action` rejects any actor whose account type is not
`User` unless it is listed, and it failed on `github-merge-queue[bot]`: the
drafting step died and the night produced no issues. That step is
`continue-on-error`, so nothing surfaced but a step failure - the failure mode is
silent, which is why the list needs to be wider than the one actor we hit.

`github-actions`, `Copilot` and `copilot-swe-agent` are the other non-`User`
actors that realistically land a change to `nightly_ondemand.yml`. The action
normalises the `[bot]` suffix away, so it is not spelled out.

### Also here

`xpu_profiling` added to `UT_JOB_CATEGORIES`.


RUIJIEZHONG66166:
Verified ondemand mode for auto-issue: https://github.com/intel/torch-xpu-ops/actions/runs/35808233769/job/107071865654. It works.

.github/scripts/ut_collect_evidence.py
```diff
@@ -87,6 +87,7 @@
 UT_JOB_CATEGORIES = {
     "basic": ["op_regression", "op_regression_dev1", "op_extended"],
     "op_ut": ["op_ut"],
+    "xpu_profiling": ["xpu_profiling"],
```
chuanqi129, line 91: @mengfei25 check with @RUIJIEZHONG66166 to add upstream ut here also

.github/workflows/_ut_auto_issue.yml
```diff
@@ -95,6 +95,11 @@ jobs:
         with:
           use_bedrock: "true"
           github_token: ${{ secrets.MERGE_TOKEN }}
+          # On a scheduled nightly github.actor is whoever pushed the last
+          # change to nightly_ondemand.yml, which is rarely a person: a merge
+          # queue landing, a workflow-authored commit or a Copilot PR all leave
+          # a bot there, and the action rejects any not listed here.
+          allowed_bots: github-merge-queue,github-actions,Copilot,copilot-swe-agent
```
chuanqi129: where those bots come? We use Claude code directly, why still need copilot here?

RUIJIEZHONG66166: Narrow allowed_bots to github-merge-queue.
Because github.actor on a scheduled run is whoever pushed to main, and here that is always the merge queue. The last 100 scheduled nightly runs all have actor=github-merge-queue[bot], including after Copilot-authored PRs #5437 and #5197 landed; only triggering_actor changed. No workflow pushes to main or dispatches nightly_ondemand.yml, so github-actions[bot] has no path either. Narrowed to github-merge-queue.

.github/scripts/ut_collect_evidence.py
RUIJIEZHONG66166: Add now

chuanqi129 (approved):

PR #5485, opened 2026-09-22
[sycl-free-function] Refactor Indexing by free function

NeoZhangJianyu:
Refactor Indexing by free function
UT is passed locally.
rename all kernel functions' name to be with suffix "_kernel_impl".


src/ATen/native/xpu/sycl/Indexing.h
```diff
@@ -698,7 +707,6 @@ void index_kernel_impl(
       index_ptrs,
       sizes,
       strides);
-  sycl_kernel_submit(sycl::range<1>(numel), queue, kfn);
 }
```
mkucha:
```suggestion
  constexpr auto kernel_func =
      index_functor_kernel_impl<func_t, index_buf_type, offset_calc_type>;
  int64_t local_range = syclMaxWorkGroupSize<kernel_func>();
  int64_t global_range =
      ((numel + local_range - 1) / local_range) * local_range;

  sycl_kernel_submit<kernel_func, 1>(
      sycl::range<1>(global_range),
      sycl::range<1>(local_range),
      queue,
      0,
      f,
      offset_calc,
      indice_size_bytes,
      out_data,
      in_data,
      num_indices,
      index_ptrs,
      sizes,
      strides,
      numel);
}
```

src/ATen/native/xpu/sycl/Indexing.h
```diff
@@ -589,77 +614,54 @@ void small_index_kernel(
...
+    OffsetCalculatorType offset_calc,
+    int64_t indice_size_bytes,
+    char* out_data,
+    char* in_data,
+    size_t num_indices,
+    at::detail::Array<index_buf_type, XPU_MAX_TENSORINFO_DIMS> index_ptrs,
+    at::detail::Array<int64_t, XPU_MAX_TENSORINFO_DIMS> sizes,
+    at::detail::Array<int64_t, XPU_MAX_TENSORINFO_DIMS> strides) {
+  auto item_id = syclext::this_work_item::get_nd_item<1>();
+  auto linear_idx = item_id.get_global_linear_id();
```
mkucha:
```suggestion
    at::detail::Array<int64_t, XPU_MAX_TENSORINFO_DIMS> strides,
    int64_t numel) {
  auto item_id = syclext::this_work_item::get_nd_item<1>();
  auto linear_idx = item_id.get_global_linear_id();
  if (linear_idx >= static_cast<size_t>(numel)) {
    return;
  }
```

src/ATen/native/xpu/sycl/Indexing.h
mkucha: The local range is hard-coded to 1, so every work-group runs a single work-item and only one SIMD lane per sub-group does useful work.

mkucha:
I've added two suggested changes inline to fix a performance bug in `index_functor_kernel_impl`.

**Problem:** in `index_kernel_impl` the kernel is submitted with a hard-coded local range of `sycl::range<1>(1)`, so every work-group contains a single work-item. Unit tests pass because the output is correct, but the kernel is an order of magnitude slower than the functor-based version it replaces.

**Suggested changes:**
1. `index_functor_kernel_impl` takes `numel` and returns early for `linear_idx >= numel`.
2. `index_kernel_impl` queries the work-group size with `at::xpu::getKernelMaxWorkGroupSize` (same as `small_index_kernel_impl` in this file) and rounds the global range up to a multiple of it.

This is the same pattern as the col2im fix in #5496.

**Measurements** (Arc B580, Windows 11). Two torch builds identical except for this change:
- Isolated `aten::index` with the `beit_base_patch16_224` shapes (src `(732, 12)` bf16, index `(38809,)` int64): **368 µs → 19 µs** per call.
- `beit_base_patch16_224` inference, `--backend=inductor --cold-start-latency --batch-size=2`, eager latency: **7.79 → 3.81 ms** (bf16), **7.79 → 3.93 ms** (fp16).

Note: the measured builds are based on an earlier version of this code, where the kernel is still called `index_functor_kernel` and the work-group size comes from `syclMaxWorkGroupSize`. For free-function kernels that helper performs the same `kernel_device_specific::work_group_size` query as `at::xpu::getKernelMaxWorkGroupSize`, so the suggestion here only aligns with the helper this PR already uses.

src/ATen/native/xpu/sycl/Indexing.h
mkucha: The local range is hard-coded to 1, so every work-group runs a single work-item and only one SIMD lane per sub-group does useful work.

```suggestion
  int64_t wgroup_size = at::xpu::getKernelMaxWorkGroupSize<
      index_functor_kernel_impl<func_t, index_buf_type, offset_calc_type>>();
  int64_t global_size =
      ((numel + wgroup_size - 1) / wgroup_size) * wgroup_size;
  sycl_kernel_submit<
      index_functor_kernel_impl<func_t, index_buf_type, offset_calc_type>,
      1>(
      sycl::range<1>(global_size),
      sycl::range<1>(wgroup_size),
      queue,
      0,
      f,
      offset_calc,
      indice_size_bytes,
      out_data,
      in_data,
      num_indices,
      index_ptrs,
      sizes,
      strides,
      numel);
}
```

PR #5486, opened 2026-09-22
[xccl] Keep record_param_comms open at the launch

aostrowski-hbn:
Supersedes #5468 and #5469, which were opened as a stack. GitHub refuses the
synchronous merge endpoint for a stacked pull request and will not let a stacked
PR's base be retargeted, so neither could be landed; this PR carries the same
five commits on a single branch off `main`. The change-set is byte-identical to
the one both of those PRs had green CI for, rebased onto today's `main` — the
only upstream delta in the touched file is #5413, in the constructor, away from
every hunk here.

---

A profiler attributes a GPU kernel to whichever CPU op is open when the kernel
is launched. `RECORD_PARAM_COMMS_DATA` is built for that: it declares a
`DebugInfoGuard` and the `RecordFunction` guard whose destructors end the event,
so it has to be invoked in the scope that should bound the event — which is how
`ProcessGroupNCCL` invokes it, bare, with `return collective(...)` running
inside the record.

`RECORD_PARAM_COMMS_DATA_WITH_LOG` wrapped it in `do { } while (0)`, closing
that scope before the collective was launched. The event ended empty and the
communication kernel was attributed to the enclosing `c10d::<op>_` instead, so
nothing that reads collective information off the kernel event could find it.

**Why CUDA works and XPU did not:** the same upstream macro, a different
invocation. NCCL leaves it bare at function scope; the XCCL wrapper added a
scope around it.

`all_to_all_single` needs a second change: its two records sit inside the split
branches while the collective is launched after them, so the guard died early
there even with the macro fixed. NCCL avoids this by returning
`collective(...)` from inside each branch; duplicating that whole call here
would be worse, so the branch-specific values are computed first and the record
is taken once at function scope. Each branch keeps its validation on the same
side of the record as before, so a failing check still emits what it did
previously. That is the only deliberate divergence from the CUDA structure, and
it produces the same record.

A DDP step with no `barrier()` in the captured region goes from 0 of 6
collective kernels carrying `Collective name` to 6 of 6.

### Sequence number

`Seq` and `Comms Id` were missing even with the record scope fixed, because they
are gated separately: `saveNcclMeta` emits them only when `ParamCommsDebugInfo`'s
sequence number is non-negative, and it stays at its `-1` default unless
`maybeSetSequenceInfo`'s tuple overload runs. The call sites passed a bare
`static_cast<int>(...)`, which selects the no-op overload, so the sequence
identity never reached the trace. `ProcessGroupNCCL` passes
`std::make_tuple(seq, isP2P)`; this follows it, and `WorkXCCL::wait()` in this
file already recorded the tuple that way. Send and recv additionally reported the
collective counter, so they now use `seqP2P_` with the offset the op itself will
apply, marked as p2p so the `Comms Id` hash matches the other backends. The two
predictions are named `predictNextCollectiveSeqId` / `predictNextP2PSeqId` rather
than spelled out at each call site.

Test: test/xpu/distributed/test_c10d_xccl.py

```
ZE_AFFINITY_MASK=0,1 pytest -v --timeout 600 \
    test/xpu/distributed/test_c10d_xccl.py \
    -k test_record_param_comms_scope_covers_launch
```

As a single line, for the reproducer job to pick up:

```
pytest test/xpu/distributed/test_c10d_xccl.py -k test_record_param_comms_scope_covers_launch -v
```

The test fails on each half separately: without the scope fix it reports that no
`record_param_comms` scope was open when `xccl:all_reduce` was launched, and with
the scope fixed but the sequence still a bare int it reports `'Seq' not found`.


kdrozd-dev (approved):

tsocha (approved):
re-approved.

aostrowski-hbn:
@torchxpubot merge -f

aostrowski-hbn:
Those same changes were already tested in https://github.com/intel/torch-xpu-ops/pull/5468 and https://github.com/intel/torch-xpu-ops/pull/5469, so I force merged.

PR #5487, opened 2026-09-22
[CI] Merge pull requests asynchronously

aostrowski-hbn:
`@torchxpubot merge` calls the synchronous merge endpoint, which GitHub refuses
for a pull request that is part of a stack:

> Merging stacked PRs via this endpoint is not supported. Use the asynchronous
> merge endpoint instead.

So the bot cannot land a stacked PR at all, and there is no way around it from the
PR side either — GitHub rejects retargeting a stacked PR's base with *"Cannot
change the base branch because the pull request is part of a stack"*. We hit both
on #5468/#5469 and had to reopen the work as a single branch (#5486).

The asynchronous endpoint is the documented path for stacks and handles ordinary
pull requests too, so this uses it unconditionally instead of branching on whether
a stack is involved. It enqueues and reports back by uuid, and branch protection
is only evaluated once the request runs, so the outcome is read from the poll
rather than from the call: a pending request is followed for five minutes,
`enqueued` is reported as such because a merge queue decides when it lands, and a
409 (a request already enqueued for that PR) is followed by its uuid rather than
starting a second one. The head sha is passed so a push racing the command cancels
the merge instead of landing something the command never saw.

Test: none (CI-only change to the bot; verified by hand as below)

Both routes exercised against a live stacked pull request that the synchronous
endpoint had already refused:

```
gh api --method PUT /repos/intel/torch-xpu-ops/pulls/<n>/merge-async \
    -H 'X-GitHub-Api-Version: 2026-03-10' \
    -f sha=<head> -f merge_method=squash -f merge_action=default
gh api /repos/intel/torch-xpu-ops/pulls/<n>/merge-async/<uuid> \
    -H 'X-GitHub-Api-Version: 2026-03-10'
```

The enqueue returned `status: pending` with a uuid and the poll then returned
`status: failed` with the branch-protection message for a token without push
access — the stack itself no longer being an obstacle, which is what this change
relies on. With the bot's own token that last step is what succeeds. The workflow
still parses and the step's script passes `node --check`.


chuanqi129:
How many stack PRs we have?

